Issue #133, opened 2021-01-22
🚀 - Use providers to inject default configurations


fynnfeldpausch:
# 🚀 Feature request

### Is your feature request related to a problem?

I would like to adjust the default values for several inputs on different components. In our project for instance, we would like to use the checkbox component in size `l`. Given the current implementation, we need to add the size input to every checkbox. It would be nice to globally define the default values via a configuration provider. Similar functionality exists on Angular Material and other frameworks.

### Describe the solution you'd like

```
import {TUI_CHECKBOX_DEFAULT_OPTIONS} from '@taiga-ui/kit';
...
{
  imports: [TuiRootModule, ...],
  providers: [{
    provide: TUI_CHECKBOX_DEFAULT_OPTIONS,
    useValue: {
      size: 'l'
    }
  }],
  ...
}
export class AppModule {}
```

### Describe alternatives you've considered

The only alternative currently is to set the size input on every checkbox.

### Additional context

Ideally, this would be available for every component and not only for checkboxes. We can help provide PRs if you like the idea.

MarsiBarsi:
That's a cool idea! I like it

Moreover, I think it could be combined with a concept of controllers. Take a look at textfield controller, for example:

It is a set of directives:
- Docs: https://taiga-ui.dev/tui-textfield-controller
- Source code: https://github.com/TinkoffCreditSystems/taiga-ui/tree/main/projects/core/directives/textfield-controller

And we inject it in DI and use https://github.com/TinkoffCreditSystems/taiga-ui/blob/main/projects/core/components/primitive-textfield/primitive-textfield.component.ts

It allows users to set parameters for textfield itself, for its parent (e.g. `tui-input`) or set something like `tuiTextfieldSize="l"` for all inputs in a form

fynnfeldpausch:
How would that work in detail? Would the default options then not only be injected / used in the respective inputs but also in the TextfieldController?

fynnfeldpausch:
I have added an example PR to have a better base for discussions. This is the solution I had in mind. Obviously, this would then have to be introduced to all the other components too. It would also be a good place to extract the hard coded icon names, that are used in a bunch of components (e.g. the arrow component or the InputPassword). Those hard coded icon names make it hard to switch to another icon set at the moment.

waterplea:
@vladimirpotekhin could you add configuration for `InputTime` next, please?

vladimirpotekhin:
Ok, I'll add

[...]

PR #139, opened 2021-01-22
feat(checkbox): allow configuration of default checkbox options

fynnfeldpausch:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [x] Feature

Partially implements #133

## What is the new behavior?

Default checkbox options can be set via a provider. This is just an example to have a discussion base for #133 

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/kit/components/checkbox/checkbox-options.ts
+++ b/projects/kit/components/checkbox/checkbox-options.ts
@@ -1,0 +1,5 @@
+import {TuiSizeL} from '@taiga-ui/core';
+
+export interface CheckboxOptions {
+    size?: TuiSizeL;
+}
--- a/projects/kit/components/checkbox/checkbox.component.ts
+++ b/projects/kit/components/checkbox/checkbox.component.ts
@@ -6,6 +6,7 @@ import {
     forwardRef,
     HostBinding,
     Inject,
+    InjectionToken,
     Input,
     Optional,
     Self,
@@ -20,6 +21,12 @@ import {
     TuiFocusableElementAccessor,
 } from '@taiga-ui/cdk';
 import {TuiSizeL} from '@taiga-ui/core';
+import {CheckboxOptions} from './checkbox-options';
+
+/** Injection token that can be used to specify default checkbox options. */
+export const TUI_CHECKBOX_DEFAULT_OPTIONS = new InjectionToken<CheckboxOptions>(
+    'tui-checkbox-default-options',
+);
 
 @Component({
     selector: 'tui-checkbox',
@@ -39,7 +46,7 @@ export class TuiCheckboxComponent
     @Input()
     @HostBinding('attr.data-tui-host-size')
     @tuiDefaultProp()
-    size: TuiSizeL = 'm';
+    size: TuiSizeL;
 
     @ViewChild('focusableElement')
     private readonly focusableElement?: ElementRef<HTMLInputElement>;
@@ -49,9 +56,13 @@ export class TuiCheckboxComponent
         @Self()
         @Inject(NgControl)
         control: NgControl | null,
+        @Optional()
+        @Inject(TUI_CHECKBOX_DEFAULT_OPTIONS)
+        options: CheckboxOptions | null,
         @Inject(ChangeDetectorRef) changeDetectorRef: ChangeDetectorRef,
     ) {
         super(control, changeDetectorRef);
+        this.size = options?.size ?? 'm';
     }
 
     get nativeFocusableElement(): HTMLInputElement | null {


waterplea:
Hi @fynnfeldpausch 

So we discussed it and decided to go with this idea. A few comments on general approach:
1) Default options should be made into exported constant
2) Token would have a factory that returns that constant, meaning @Optional is non necessary
3) Interface should be readonly with all keys required. People who want to override it would import default constant, destruct it and override keys they need
4) Component itself should implement the interface too

waterplea (changes requested):
Please update the PR according to my comment, keeping constant and token in the same file so it's easy to see default values. We can practice the approach on this PR and checkbox component and once we all happy with how it works we can go over the rest of the components.

fynnfeldpausch:
> Please update the PR according to my comment, keeping constant and token in the same file so it's easy to see default values. We can practice the approach on this PR and checkbox component and once we all happy with how it works we can go over the rest of the components.

Let me know what you think. I also added a section to the docs. Not sure if you want to have the different members of the type described in more detail (e.g. in a table or something).

fynnfeldpausch:
@waterplea Any update on this one. Would be really cool to have a template that we can use for the other components as well.

projects/kit/components/checkbox/checkbox.component.ts
```diff
@@ -56,13 +50,12 @@ export class TuiCheckboxComponent
         @Self()
         @Inject(NgControl)
         control: NgControl | null,
-        @Optional()
-        @Inject(TUI_CHECKBOX_DEFAULT_OPTIONS)
-        options: CheckboxOptions | null,
+        @Inject(TUI_CHECKBOX_OPTIONS)
+        options: CheckboxOptions,
```
waterplea, line 54: let's just make it `private readonly` and write `size: TuiSizeL = this.options.size`

projects/kit/components/checkbox/checkbox-options.ts
```diff
@@ -1,5 +1,19 @@
+import {InjectionToken} from '@angular/core';
 import {TuiSizeL} from '@taiga-ui/core';
 
 export interface CheckboxOptions {
-    size?: TuiSizeL;
+    readonly size: TuiSizeL;
```
waterplea, line 5: I wonder why doesn't this go into conflict with size being non-readonly in checkbox component 🤔

projects/kit/components/checkbox/checkbox-options.ts
```diff
@@ -1,5 +1,19 @@
...
 }
+
+/** Default values for the checkbox options. */
+export const TUI_CHECKBOX_DEFAULT_OPTIONS: CheckboxOptions = {
+    size: 'm',
+};
+
+/** Injection token that can be used to specify checkbox options. */
+export const TUI_CHECKBOX_OPTIONS = new InjectionToken<CheckboxOptions>(
```
waterplea, line 14: @fynnfeldpausch @MarsiBarsi so what do you think if we will use this same token for icons config? If we will, then we need to move that to `PrimitiveCheckbox`. Sounds like a reasonable thing to do, I wouldn't want to create too many tokens.

waterplea (changes requested):
Added a few comments to think about and change request in constructor.

projects/kit/components/checkbox/checkbox-options.ts
fynnfeldpausch: I was wondering the exact same thing TBO. I'd find it more intuitive to remove the readonly modifier from the field as I don't think it adds much value.

projects/kit/components/checkbox/checkbox-options.ts
fynnfeldpausch: Good question. Don't know how to achieve a good separation then, as the checkbox might define other inputs as the primitive checkbox. One could either use different tokens for every component type (checkbox, checkbox-block, primitive checkbox,...) or go for a unified one with all fields and use it in every component. Don't know where to place it then. More interestingly, how would one make a distinction between the default size of the primitive checkbox and the normal checkbox component. Or use one size option and use it in every component (so basically not making them configurable individually)?...

projects/kit/components/checkbox/checkbox-options.ts
waterplea: Let's remove `readonly` here and add utility type `Readonly<CheckboOptions>` to the token then.

projects/kit/components/checkbox/checkbox-options.ts
waterplea: I was thinking one token per component. In this particular case, looks like Checkbox will just inject the same token.

projects/kit/components/checkbox/checkbox-options.ts
```diff
@@ -0,0 +1,19 @@
+import {InjectionToken} from '@angular/core';
+import {TuiSizeL} from '@taiga-ui/core';
+
+export interface CheckboxOptions {
+    size: TuiSizeL;
+}
+
+/** Default values for the checkbox options. */
+export const TUI_CHECKBOX_DEFAULT_OPTIONS: CheckboxOptions = {
```
waterplea, line 9: This one is also `Readonly<CheckboxOptions>` so nobody decides to mutate the constant instead of providing :) We love `readonly` here :)

waterplea (approved):
Other than that one little comment this looks good to me. @MarsiBarsi ?

MarsiBarsi (approved):

PR #140, opened 2021-01-23
feat(checkbox): make checkboxes react to new height variables

fynnfeldpausch:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [x] Feature

Checkboxes currently do not react to the newly introduced CSS height properties.

## What is the new behavior?

Checkboxes now react to CSS height properties. They are exactly half in size of the respective variable value. A new size has been added to keep a 16px size variance. One problem is, that the L variance is now 22px heigh (and not 24px) and does not fit into the 4px base grid. This is due to the respective height variable being 44px instead of 48px. I think it would make sense to increase this value, but I can't oversee the impact, that such a change would have. Maybe this PR is a good base for discussions.

-   [x] Yes
-   [ ] No

Size L is now 22px instead of 24px.

## Other information

A similar approach could be taken for toggles and radio buttons.

### DIFF
--- a/projects/core/components/primitive-checkbox/primitive-checkbox.component.ts
+++ b/projects/core/components/primitive-checkbox/primitive-checkbox.component.ts
@@ -65,10 +65,10 @@ export class TuiPrimitiveCheckboxComponent {
     }
 
     get iconCheck(): string {
-        return this.size === 'm' ? 'tuiIconCheck' : 'tuiIconCheckLarge';
+        return this.size === 'l' ? 'tuiIconCheckLarge' : 'tuiIconCheck';
     }
 
     get iconIndeterminate(): string {
-        return this.size === 'm' ? 'tuiIconMinus' : 'tuiIconMinusLarge';
+        return this.size === 'l' ? 'tuiIconMinusLarge' : 'tuiIconMinus';
     }
 }
--- a/projects/core/components/primitive-checkbox/primitive-checkbox.style.less
+++ b/projects/core/components/primitive-checkbox/primitive-checkbox.style.less
@@ -1,22 +1,25 @@
 @import '../../styles/taiga-ui-local.less';
 
-@sizeM: 16px;
-@sizeL: 24px;
-
 :host {
     display: block;
     border-radius: 8px;
     font-size: 0;
 
+    &[data-tui-host-size='s'] {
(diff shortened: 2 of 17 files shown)


projects/core/components/primitive-checkbox/primitive-checkbox.component.ts
```diff
@@ -65,10 +65,10 @@ export class TuiPrimitiveCheckboxComponent {
     }
 
     get iconCheck(): string {
-        return this.size === 'm' ? 'tuiIconCheck' : 'tuiIconCheckLarge';
+        return this.size === 'l' ? 'tuiIconCheckLarge' : 'tuiIconCheck';
```
waterplea, line 68: Just curious, why the switch? :)

waterplea: Oh, I see, you added small size to checkboxes

projects/core/components/primitive-checkbox/primitive-checkbox.style.less
```diff
@@ -1,22 +1,25 @@
...
-        height: @sizeM;
+        width: calc(var(--tui-height-m) / 2);
+        height: calc(var(--tui-height-m) / 2);
         border-radius: 4px;
     }
 
     &[data-tui-host-size='l'] {
-        width: @sizeL;
-        height: @sizeL;
+        width: calc(var(--tui-height-l) / 2);
```
waterplea, line 21: I noticed inconsistency in size variables and removed 40px in last commit, moving 44 to m and 56 to l, dropping 40px and xl naming. But even before that these sizes were wrong. Checkboxes have to be 16 and 24px in size. Do you feel like we could use a third size for checkboxes?

fynnfeldpausch: This is of cause a very personal decision. We are looking for a size of 20px in our design system, but I guess we could also try and adjust. I found it a nice idea to have checkboxes react to the new height variables as well and just set them to half of these values. This obviously clashes with the recent changes of m=44 and l=56.

waterplea:
@fynnfeldpausch do we decline this? You mentioned on discord that you will override height in-place for your project.

fynnfeldpausch:
Yep, feel free to decline it 👍

PR #157, opened 2021-01-27
feat(demo): add guide on using different icon set

waterplea:


projects/demo/src/modules/customization/icon-set/icon-set.template.html
```diff
@@ -0,0 +1,60 @@
+<tui-doc-page i18n-header header="Icons set">
+    <p i18n>
+        Taiga UI uses
+        <a tuiLink routerLink="/tui-svg"><code>tui-svg</code></a> component to
+        display SVG icons. It is capable of showing icons by name, link or
```
MarsiBarsi, line 5: you can say also "it can show icons by ..."

MarsiBarsi (approved):

PR #185, opened 2021-01-30
feat(kit): use lazy loading strategy in avatar component

### LINKED ISSUE #62: [FEATURE] a lazy loading image inside the Avatar component
# 🚀 Feature request

 I'm always frustrated when an image is not loading lazily. Like an image inside the Avatar component.

 All images inside the Avatar component loaded lazily OR loaded using the 'loading="lazy"' property

 Nope

vadimk7:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [x] Feature

Closes #62

## What is the new behavior?
All avatar images are loaded lazily

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/kit/components/avatar/avatar.module.ts
+++ b/projects/kit/components/avatar/avatar.module.ts
@@ -1,9 +1,10 @@
 import {CommonModule} from '@angular/common';
 import {NgModule} from '@angular/core';
+import {TuiLazyLoadingModule} from '@taiga-ui/kit/directives/lazy-loading';
 import {TuiAvatarComponent} from './avatar.component';
 
 @NgModule({
-    imports: [CommonModule],
+    imports: [CommonModule, TuiLazyLoadingModule],
     declarations: [TuiAvatarComponent],
     exports: [TuiAvatarComponent],
 })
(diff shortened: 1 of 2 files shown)


projects/kit/components/avatar/avatar.module.ts
```diff
@@ -1,9 +1,10 @@
 import {CommonModule} from '@angular/common';
 import {NgModule} from '@angular/core';
+import {TuiLazyLoadingModule} from '@taiga-ui/kit/directives/lazy-loading';
```
waterplea, line 3: @MarsiBarsi should we import from directives entry point? Maybe we should document it somehow?

MarsiBarsi: We can but we should not. It is not very important for me, but I see a small benefit here: it may be easier for IDE to add the next import from directives. Well, we can describe such "import policy" in our CONTRIBUTING.md and I am going to fix it here with web editor of Github

MarsiBarsi (approved):

MarsiBarsi (approved):

MarsiBarsi:
Spasibo, Vadim, for your contribution! :)

vadimk7:
@waterplea, @MarsiBarsi, thanks for submitting my pr 🙂 
About direct import from entrypoint, I thought this kind of import allows to include only used directives in the avatar module and optimize the resulting bundle, isn't it?

waterplea:
> @waterplea, @MarsiBarsi, thanks for submitting my pr 🙂 
> About direct import from entrypoint, I thought this kind of import allows to include only used directives in the avatar module and optimize the resulting bundle, isn't it? 

Yes, but in our library every secondary entry point becomes primary entry point for the higher level imports. This way it is already optimized, no matter how deep you import. So we try to use the highest possible point so if you, say, import another directive — it would be placed in the same import line.

PR #201, opened 2021-02-03
feat(common): added CSS custom properties for fonts

fynnfeldpausch:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [x] Feature

Font styles can not be changed easily.

## What is the new behavior?

Font styles can be adjusted via CSS custom properties. Also lots of the typography rules have been cleaned up. Docs have been updated accordingly.

-   [x] Yes
-   [ ] No

- Mixin `.text-body-xl-bold()` has been removed. Use `.text-body-xl()` and set the font weight manually.
- Mixin `.text-body-l-bold()` has been removed. Use `.text-body-l()` and set the font weight manually.
- Mixin `.text-body-m-bold()` has been removed. Use `.text-body-m()` and set the font weight manually.
- Mixin `.text-body-s-bold()` has been removed. Use `.text-body-s()` and set the font weight manually.
- Mixin `.text-body-xs-bold()` has been removed. Use `.text-body-xs()` and set the font weight manually.

- Mixin `.text-body-l(2)` has been removed. Use `.text-body-l()` and set the line height manually.
- Mixin `.text-body-l(3)` has been removed. Use `.text-body-l()` and set the line height manually.
- Mixin `.text-body-m(2)` has been removed. Use `.text-body-m()` and set the line height manually.
- Mixin `.text-body-s(2)` has been removed. Use `.text-body-s()` and set the line height manually.

- Mixin `.font_heading` has been removed.
- Mixin `.font` has been deprecated.

- Deprecated CSS property  `--tui-heading-font` has been removed.
- Deprecated CSS property  `--tui-text-font` has been removed.
- Classes, mixins and styles related to captions have been removed.
- letter spacing for `xs`-sized texts has been removed as it is very font-specific and might break customization.


projects/core/styles/theme/variables.less
```diff
@@ -1,12 +1,23 @@
...
+    --tui-font-heading-4: bold 28px/32px var(--tui-font-heading);
+    --tui-font-heading-5: bold 24px/28px var(--tui-font-heading);
+    --tui-font-heading-6: bold 20px/24px var(--tui-font-heading);
+    --tui-font-text: 'Manrope', @font-fallback;
+    --tui-font-text-xl: normal 19px/28px var(--tui-font-text);
+    --tui-font-text-l: normal 17px/24px var(--tui-font-text);
+    --tui-font-text-m: normal 15px/24px var(--tui-font-text);
+    --tui-font-text-s: normal 13px/20px var(--tui-font-text);
+    --tui-font-text-xs: normal 11px/16px var(--tui-font-text);
+    --tui-font-caption: 'Manrope', @font-fallback;
```
waterplea, line 18: I don't think we need captions, let's just drop them.

projects/core/styles/mixins/text.scss
```diff
@@ -3,128 +3,69 @@
     color: var(--tui-text-01);
 }
 
-//headings
+// headings
+
 @mixin text-h1() {
-    @include font_heading(50px, 56px);
```
waterplea, line 8: Seems like text mixins are pretty useless now. @MarsiBarsi @fynnfeldpausch they could probably be used by people already, so we cannot drop them, right? How about we mark them all as deprecated and drop SCSS ones since they are a new addition?

projects/core/styles/theme/variables.less
fynnfeldpausch: Do you mean captions in general and set the style inline or just drop the font face definition of captions and use the normal text font?

projects/core/styles/mixins/text.scss
fynnfeldpausch: Yeah, that's why I did not remove them because I figured that they might be in use. The same applies to the `.font` mixin, which is also unused now and could be deprecated.

projects/core/styles/theme/variables.less
waterplea: I don't think it is used anywhere in public code, other than on Typography page itself.

projects/core/styles/mixins/text.scss
waterplea: I don't know any way other than comment. We wanted to release Angular stuff so much we didn't stop to look that a lot of global styles and mixins are pointless :) People would still use some layout library because built in Taiga UI classes are not powerful enough. I think we should remove all of this in 3.0 to not confuse people.

projects/core/styles/theme/variables.less
fynnfeldpausch: They are in use in islands as category headlines. Should I remove the category styles and replace them with inline styles?

waterplea: Right. Yes, please do so.

fynnfeldpausch: Done

projects/core/styles/mixins/text.scss
fynnfeldpausch: Added deprecation comments to all mixins and also removed them from the docs page as deprecated code should not be actively included in the docs IMO.

waterplea (changes requested):
Let's also replace all mixins usages with CSS vars and return removed mixins and deprecate them instead. @MarsiBarsi @WynieCronje how about we remove SCSS mixins for fonts? They were only present for one version and I think we will release a new version soon so @fynnfeldpausch will have config for checkboxes, CSS vars and new translations are up too..

fynnfeldpausch:
@waterplea I think the SCSS and LESS mixins should stay in sync. I would remove them together. Thats why I deprecated it instead of removing them. Will take care of the other comments.

fynnfeldpausch:
@waterplea all mixins are back (but deprecated) and their usage has been replaced.

WynieCronje:
Thanks for updating the scss 😄

projects/addon-charts/components/ring-chart/ring-chart.style.less
```diff
@@ -32,7 +32,8 @@
 
 .content {
     .fullsize();
-    .text-body-m(2);
+    font: var(--tui-font-text-m);
+    line-height: 20px;
```
waterplea, line 36: This one makes no difference, let's remove it

projects/core/styles/theme/variables.less
```diff
@@ -1,12 +1,20 @@
 @font-fallback: -apple-system, 'BlinkMacSystemFont', system-ui, 'Roboto', 'Segoe UI', 'Helvetica Neue', sans-serif;
 
 :root {
-    // Deprecated
-    --tui-heading-font: 'Manrope', @font-fallback;
-    --tui-text-font: 'Manrope', @font-fallback;
```
waterplea, line 6: These two were in the initial release, let's keep them as deprecated to not introduce a breaking change

projects/kit/components/avatar/avatar.style.less
```diff
@@ -22,39 +22,41 @@
     overflow: hidden;
 
     &[data-tui-host-size='xs'] {
-        .text-body-s(2);
+        font: var(--tui-font-text-s);
+        line-height: 16px;
```
waterplea, line 26: Let's remove this one two

projects/kit/components/avatar/avatar.style.less
```diff
@@ -22,39 +22,41 @@
...
         width: @size-xs;
         height: @size-xs;
     }
 
     &[data-tui-host-size='s'] {
-        .text-body-m(2);
+        font: var(--tui-font-text-m);
+        line-height: 20px;
```
waterplea, line 33: And this one doesn't make a difference

projects/kit/components/badge/badge.style.less
```diff
@@ -3,7 +3,8 @@
 @padding: 12px;
 
 :host {
-    .text-body-s(2);
+    font: var(--tui-font-text-s);
+    line-height: 16px;
```
waterplea, line 7: This one has no effect as well

projects/kit/components/breadcrumbs/breadcrumbs.style.less
```diff
@@ -6,11 +6,13 @@
     height: 40px;
 
     &[data-tui-host-size='m'] {
-        .text-body-s(2);
+        font: var(--tui-font-text-s);
+        line-height: 16px;
```
waterplea, line 10: No effect either

projects/kit/components/breadcrumbs/breadcrumbs.style.less
```diff
@@ -6,11 +6,13 @@
...
     }
 
     &[data-tui-host-size='l'] {
-        .text-body-m(2);
+        font: var(--tui-font-text-m);
+        line-height: 20px;
```
waterplea, line 15: And here

projects/kit/components/input-range/input-range.style.less
```diff
@@ -4,7 +4,8 @@
 @padding2: @padding * 2;
 
 :host {
-    .text-body-m(2);
+    font: var(--tui-font-text-m);
+    line-height: 20px;
```
waterplea, line 8: This one also should be removed

projects/kit/components/input-slider/input-slider.style.less
```diff
@@ -4,7 +4,8 @@
 @padding2: @padding * 2;
 
 :host {
-    .text-body-m(2);
+    font: var(--tui-font-text-m);
+    line-height: 20px;
```
waterplea, line 8: Please remove this one as well

projects/kit/components/tag/tag.style.less
```diff
@@ -5,7 +5,8 @@
 
 :host {
     .transition(background);
-    .text-body-s(2);
+    font: var(--tui-font-text-s);
+    line-height: 16px;
```
waterplea, line 9: This one also does not matter

projects/kit/components/tag/tag.style.less
```diff
@@ -225,7 +226,8 @@
 }
 
 .edit {
-    .text-body-s(2);
+    font: var(--tui-font-text-s);
+    line-height: 16px;
```
waterplea, line 230: And this one too

waterplea (changes requested):
@fynnfeldpausch thank you very much for this PR. I decided to go over all those `line-height` overrides and marked those that make no effect with comments. Please remove them so we do not overcomplicate styles and we can merge.

fynnfeldpausch:
> @fynnfeldpausch thank you very much for this PR. I decided to go over all those `line-height` overrides and marked those that make no effect with comments. Please remove them so we do not overcomplicate styles and we can merge.

✅

waterplea (approved):

MarsiBarsi (approved):

PR #259, opened 2021-03-02
feat(error): strings in error component

### LINKED ISSUE #255: [FEATURE] Strings in error component
# 🚀 Feature request

Right now Error component only supports special class TuiValidationError, it needs to support plain string too.

podkovyrovda:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [x] Feature

Closes #255 

-   [ ] Yes
-   [x] No


projects/core/components/error/error.component.ts
```diff
@@ -15,13 +15,23 @@ import {Observable} from 'rxjs';
     animations: [tuiHeightCollapse, tuiFadeIn],
 })
 export class TuiErrorComponent {
-    @Input()
+    @Input('error')
     @tuiDefaultProp()
+    set errorSetter(error: TuiValidationError | string | null) {
```
MarsiBarsi, line 20: @waterplea should we maybe support the whole polymorphous content here?

MarsiBarsi (commented):
Hey 👋 Thank you for your contribution. We'll discuss the issue here and come back soon ;)

projects/core/components/error/error.component.ts
waterplea: `TuiValidationError` is already a wrapper around polymorpheus + context so I think it's not necessary. What is necessary though is to handle strings as validation errors for controls, not just in `Error` component. Say we have a validator that returns this error: `{ error: 'Oops!' }`.

@podkovyrovda check out `FieldError` component around line 185. At the end of that function we try to get error content from the DI. In case we do not get it, the content is undefined and context is actual string error. We need to handle this case, but also be able to show default error message. I think if we have `content === undefined` and `typeof context === 'string'` there at the end we can use `context` as `content`. This way it would show `'Oops!'` for the validator above and default error message for `{ required: true }` for example. 

@MarsiBarsi what do you think?

P.S. there's also a TODO there, you can remove it and the line it mentions, it is outdated.

waterplea (changes requested):
Please check my comment

projects/core/components/error/error.component.ts
```diff
@@ -15,13 +15,23 @@ import {Observable} from 'rxjs';
...
+        if (typeof error === 'string') {
+            this.error = new TuiValidationError(error);
+
+            return;
+        }
+
+        this.error = error;
```
waterplea, line 27: Please use a ternary operator for brevity.

projects/core/components/error/error.component.ts
```diff
@@ -15,13 +15,23 @@ import {Observable} from 'rxjs';
...
+    }
+
     error: TuiValidationError | null = null;
 
     constructor(
         @Inject(TUI_MODE) readonly mode$: Observable<TuiBrightness | null>,
         @Inject(TUI_DEFAULT_ERROR_MESSAGE)
-        readonly defauleErrorMessage$: Observable<string>,
+        readonly defaultErrorMessage$: Observable<string>,
```
waterplea, line 35: Nice catch :) thanks!

projects/core/components/error/error.component.ts
waterplea: @MarsiBarsi @podkovyrovda so what do you think?

MarsiBarsi: Well, let's do it, a bit mindblowing but solves the case

waterplea (approved):

PR #268, opened 2021-03-04
feat(kit): update tabs scroll position on activeIndex change

makkaronis4e:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [x] Feature

## What is the new behavior?

Now tabs will update scroll position on activeIndex change. Previously it worked only after click on tab.

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/kit/components/tabs/tabs/tabs.component.ts
+++ b/projects/kit/components/tabs/tabs/tabs.component.ts
@@ -138,7 +138,7 @@ export class TuiTabsComponent implements AfterViewChecked {
         moveFocus(tabs.indexOf(current), tabs, step);
     }
 
-    updateScrollPosition(element: HTMLElement) {
+    updateScrollPosition(element?: HTMLElement) {
         if (!element) {
             return;
         }
--- a/projects/kit/components/tabs/tabs/tabs.component.ts
+++ b/projects/kit/components/tabs/tabs/tabs.component.ts
@@ -126,7 +126,7 @@ export class TuiTabsComponent implements AfterViewChecked {
             return;
         }
 
-        this.activeItemIndex = index;
+        this.activeItemIndexSetter = index;
         this.activeItemIndexChange.emit(index);
     }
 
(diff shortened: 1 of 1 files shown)


projects/kit/components/tabs/tabs/tabs.component.ts
```diff
@@ -124,6 +128,20 @@ export class TuiTabsComponent implements AfterViewChecked {
 
         this.activeItemIndex = index;
```
waterplea, line 129: Please use the setter here

projects/kit/components/tabs/tabs/tabs.component.ts
```diff
@@ -124,6 +128,20 @@ export class TuiTabsComponent implements AfterViewChecked {
...
+    @HostListener('keydown.arrowRight.prevent', ['$event.target', '1'])
+    @HostListener('keydown.arrowLeft.prevent', ['$event.target', '-1'])
+    onKeyDownArrow(current: HTMLElement, step: number) {
+        const {tabs} = this;
+
+        moveFocus(tabs.indexOf(current), tabs, step);
+    }
+
+    updateScrollPosition(element: HTMLElement) {
+        if (!element) {
```
waterplea, line 142: This check doesn't make sense in terms of type, please add '?' to element argument.

waterplea (changes requested):
A few comments and we're good to merge

projects/kit/components/tabs/tabs/tabs.component.ts
makkaronis4e: why we need to use setter here? it will trigger 1 additional recalculation method. I think this line can be removed

waterplea: I don't get it. Why would it? It is programmatic setting after user. You mean it would bubble up and back with two way binding? What if user doesn't have binding at all and just uses router links?

makkaronis4e: yeah, havent even thought about it :)

waterplea (approved):

PR #369, opened 2021-05-03
feat(input-password): allow configuration of default input password options

KarimovDev:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [x] Feature

Partially implements #133 

## What is the new behavior?

Default input password options can be set via a provider. This is a partial implementation of #133

-   [ ] Yes
-   [x] No


projects/demo/src/modules/components/input-password/examples/2/index.ts
```diff
@@ -0,0 +1,41 @@
...
+    changeDetection,
+    encapsulation,
+    providers: [
+        {
+            provide: TUI_INPUT_PASSWORD_OPTIONS,
+            useValue: {
+                ...TUI_INPUT_PASSWORD_DEFAULT_OPTIONS,
+                hintMode: 'onDark',
+                icons: {
+                    hide: hideIcon,
```
waterplea, line 30: This can be just plain `string`. It should be `PolymorpheusContent` which is basically a union of `string` | `(context: C) => string` | `TemplateRef<C>` | `PolymorpheusComponent<C>`. See my comment in token definition,

projects/kit/components/input-password/input-password-options.ts
```diff
@@ -0,0 +1,29 @@
+import {InjectionToken} from '@angular/core';
+import {TuiHintMode} from '@taiga-ui/core/enums';
+import {TuiSizeL, TuiSizeS} from '@taiga-ui/core/types';
+
+export interface InputPasswordOptions {
+    readonly hintMode: Readonly<TuiHintMode | null>;
+    readonly icons: Readonly<{
+        hide: (size: TuiSizeS | TuiSizeL) => string;
```
waterplea, line 8: Please use `PolymorpheusContent<TuiContextWithImplicit<TuiSizeS | TuiSizeL>>`

projects/kit/components/input-password/input-password.template.html
```diff
@@ -25,6 +25,7 @@
         class="icon"
         tuiHintId="not_required"
         [tuiHint]="hintContent"
+        [tuiHintMode]="options.hintMode"
```
waterplea, line 28: Are you sure we need this here? Maybe it should just rely on `TuiMode` directive, like regular input field hint?

waterplea (changes requested):
Please take a look at comments. Maybe using PolymorpheusContent is excessive here but I would like to have uniformity of such API. Technically this would allow better customization like styling of this icon or animating it.

waterplea:
Thank you for this PR! This is something we would want to do for all our components so thanks for helping us :)

projects/kit/components/input-password/input-password.template.html
KarimovDev: I think you are right. I've changed it to a hint controller

projects/kit/components/input-password/input-password-options.ts
KarimovDev: OK, I've changed it

projects/demo/src/modules/components/input-password/examples/2/index.ts
KarimovDev: Yep, I agree

projects/core/directives/hint-controller/hint-controller.directive.ts
```diff
@@ -7,7 +7,7 @@ import {TUI_HINT_CONTROLLER} from './hint-controller.token';
 
 // TODO: v2.0 use in Charts
 @Directive({
-    selector: '[tuiHintContent]',
+    selector: '[tuiHintContent],[tuiHintDirection],[tuiHintMode]',
```
waterplea, line 10: @MarsiBarsi please check this too.

projects/demo/src/modules/components/input-password/examples/2/index.ts
```diff
@@ -0,0 +1,32 @@
+import {Component} from '@angular/core';
+import {FormControl, FormGroup, Validators} from '@angular/forms';
+import {
+    TUI_INPUT_PASSWORD_DEFAULT_OPTIONS,
+    TUI_INPUT_PASSWORD_OPTIONS,
+} from '@taiga-ui/kit/components/input-password/input-password-options';
```
waterplea, line 6: It should be imported from the main entry point

projects/demo/src/modules/components/input-password/examples/2/index.html
```diff
@@ -0,0 +1,9 @@
+<form class="b-form" [formGroup]="testForm">
+    <tui-input-password
+        formControlName="testValue"
+        tuiHintDirection="right"
+        tuiHintMode="onDark"
```
waterplea, line 5: Default hint also follows tuiMode directive so if the input is on dark background it automatically turns white. Maybe this extra hint should also do that?

projects/demo/src/modules/components/input-password/examples/import/define-options.txt
```diff
@@ -0,0 +1,16 @@
+import {TUI_INPUT_PASSWORD_OPTIONS, TUI_INPUT_PASSWORD_DEFAULT_OPTIONS} from '@taiga-ui/core';
```
waterplea, line 1: /kit

projects/kit/components/input-password/input-password-options.ts
```diff
@@ -0,0 +1,28 @@
+import {InjectionToken} from '@angular/core';
+import {TuiContextWithImplicit} from '@taiga-ui/cdk';
+import {TuiSizeL, TuiSizeS} from '@taiga-ui/core/types';
```
waterplea, line 3: Cross-package imports should be from the main entry point

projects/kit/components/input-password/input-password.component.ts
```diff
@@ -68,12 +82,23 @@ export class TuiInputPasswordComponent
         return !!this.textfield && this.textfield.focused;
     }
 
-    get icon(): string {
-        if (this.textfieldSize.size === 's') {
-            return this.isPasswordHidden ? 'tuiIconEyeClosed' : 'tuiIconEyeOpen';
+    get icon(): PolymorpheusContent<TuiContextWithImplicit<TuiSizeS | TuiSizeL>> {
+        if (this.isPasswordHidden) {
```
waterplea, line 86: Please use ternary operator here

projects/kit/components/input-password/input-password.component.ts
```diff
@@ -68,12 +82,23 @@ export class TuiInputPasswordComponent
...
 
-        return this.isPasswordHidden ? 'tuiIconHideLarge' : 'tuiIconShowLarge';
+        return this.options.icons.show;
+    }
+
+    get context(): TuiContextWithImplicit<TuiSizeS | TuiSizeL> {
+        return this.getContext(this.textfieldSize.size);
+    }
+
+    @tuiPure
```
waterplea, line 97: 👍

waterplea (changes requested):
A few small comments

projects/core/directives/hint-controller/hint-controller.directive.ts
MarsiBarsi: Hey 👋 It does not seem correct in the current implementation. hintController does not accumulate all the directive from DI above (textfieldController does it, for example). So, if we set `tuiHintDirection` without `tuiHintContent` we customize direction of hint that cannot exist

projects/kit/components/input-password/input-password.component.ts
KarimovDev: Done

projects/demo/src/modules/components/input-password/examples/import/define-options.txt
KarimovDev: Done

projects/demo/src/modules/components/input-password/examples/2/index.html
KarimovDev: Implemented this logic

projects/demo/src/modules/components/input-password/examples/2/index.ts
KarimovDev: Done

KarimovDev:
Also, I've changed the tests back. They are working now through component method as before instead of DOM attribute

projects/kit/components/input-password/input-password-options.ts
KarimovDev: Done. Can you please explain why?

projects/core/directives/hint-controller/hint-controller.directive.ts
KarimovDev: You are right. 
However this implementation doesn't throw any error if someone set hint direction without hint content.
@waterplea do you have any ideas how to provide tuiHintDirection and tuiHintMode through hint controller without this selectors? Should I put them back into input password options token?

projects/demo/src/modules/components/input-password/input-password.module.ts
```diff
@@ -25,6 +26,7 @@ import {ExampleTuiInputPasswordComponent} from './input-password.component';
         InheritedDocumentationModule,
         TuiAddonDocModule,
         TuiButtonModule,
+        TuiModeModule,
```
waterplea, line 29: There's no need for this module since you are not using the directive here.

projects/kit/components/input-password/input-password.component.ts
```diff
@@ -44,6 +50,8 @@ import {InputPasswordOptions, TUI_INPUT_PASSWORD_OPTIONS} from './input-password
             useExisting: forwardRef(() => TuiInputPasswordComponent),
         },
         HINT_CONTROLLER_PROVIDER,
+        TuiDestroyService,
```
waterplea, line 53: There's no need for explicit destroy service, MODE_PROVIDER already has it baked in

projects/kit/components/input-password/input-password-options.ts
waterplea: Since all our lower-level entry points are re-exported again on the higher level you get the same treeshakability if you import from the top level. When in the same library we write exact secondary entry points so that there are no circular dependencies and each entry point can clearly resolve its dependencies.

projects/kit/components/input-password/input-password.component.ts
KarimovDev: Did you mean HINT_CONTROLLER_PROVIDER? I can't find TuiDestroyService inside MODE_PROVIDER :(

projects/kit/components/input-password/input-password.component.ts
```diff
@@ -68,8 +78,18 @@ export class TuiInputPasswordComponent
...
     ) {
         super(control, changeDetectorRef);
+
+        mode$.pipe(takeUntil(destroy$)).subscribe(mode => {
+            this.globalMode = mode;
+        });
+    }
+
+    get computedMode(): TuiHintMode | TuiBrightness | null {
+        return this.hintController.mode || (this.globalMode ? this.globalMode : null);
```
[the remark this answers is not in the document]
KarimovDev: Yep, I've copied this from tooltip component. Think it needs refactoring there too

projects/demo/src/modules/components/input-password/input-password.module.ts
KarimovDev: My fault, I forgot to remove it

projects/kit/components/input-password/input-password-options.ts
KarimovDev: Thank you for the explanation

projects/kit/components/input-password/input-password.component.ts
waterplea: Right, I meant HINT_CONTROLLER_PROVIDER, sorry.

projects/kit/components/input-password/input-password.component.ts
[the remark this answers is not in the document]
waterplea: True, thanks for taking care of it!

projects/core/directives/hint-controller/hint-controller.directive.ts
waterplea: @KarimovDev is it reasonable for those to be different between tooltip and show/hide hint? How about we still wire those to the controller directive and add an example where you would use empty tuiHintContent="" so that tooltip is not shown, but the directive is matched to pass direction and mode?

KarimovDev: OK, I've done it in the way above

projects/kit/components/input-password/input-password.component.ts
```diff
@@ -54,8 +73,22 @@ export class TuiInputPasswordComponent
...
+        @Inject(TUI_INPUT_PASSWORD_OPTIONS)
+        public readonly options: InputPasswordOptions,
+        @Inject(TUI_HINT_WATCHED_CONTROLLER)
+        readonly hintController: TuiHintControllerDirective,
+        @Inject(TuiDestroyService) destroy$: Observable<unknown>,
+        @Inject(TUI_MODE) mode$: Observable<TuiBrightness | null>,
     ) {
         super(control, changeDetectorRef);
+
+        mode$.pipe(takeUntil(destroy$)).subscribe(mode => {
```
waterplea, line 85: Just came to my mind that we can get rid of the subscription here in favor of `async` pipe:

```ts
readonly computedMode$ = combineLatest(
  this.mode$,
  this.hintController.change$.pipe(map(() => this.hintController.mode)),
).pipe(
  map(([mode, controller]) => controller || mode),
  startWith(null)
)
```
Just need to check that both emit initially.

waterplea (changes requested):
@KarimovDev just came up with a little idea on how to improve this. Sorry for lots of comments, hopefully this is the last one :) Thanks for your patience!

projects/kit/components/input-password/input-password.component.ts
KarimovDev: hintController emits changes before component constructor executing. So combineLatest needs startWith operator inside hintController observable. I've done it

waterplea (approved):

waterplea:
Awesome! Thank you for this contribution :)

KarimovDev:
> Awesome! Thank you for this contribution :)

Thank you for your help :)

PR #374, opened 2021-05-06
feat(core): `TUI_ANIMATIONS_DURATION` add a new token

### LINKED ISSUE #235: [FEATURE] Speed up animationso of allow to configure animation speed
# 🚀 Feature request

I feel all animations are a bit too slow. E.g. when clicking on a tuiDropdown the contents is opened quite slowly. It feels sluggish instead of snappy.

The animations could be a little faster by default. If possible it would be very convenient if this could be configured somehow (via directive itself and/or a global config).

I have searched a bit but I don't see how I could change this myself without changing the source code (I think the releated code is here: https://github.com/TinkoffCreditSystems/taiga-ui/search?q=TUI_ANIMATION_DURATION ).

### Additional context
If you know of a quick workaround I would be happy to use it

Thank!

waterplea:
Closes #235

### DIFF
--- a/projects/addon-tablebars/components/table-bars-host/table-bars-host.component.ts
+++ b/projects/addon-tablebars/components/table-bars-host/table-bars-host.component.ts
@@ -19,7 +19,7 @@ export class TuiTableBarsHostComponent {
     readonly animation = {
         value: '',
         ...this.options,
-    };
+    } as const;
 
     constructor(
         @Inject(TuiTableBarsService) readonly service: TuiTableBarsService,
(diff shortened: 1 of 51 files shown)


projects/addon-tablebars/components/table-bars-host/table-bars-host.component.ts
```diff
@@ -15,9 +16,15 @@ import {TuiTableBarsService} from '../../services/table-bars.service';
     animations: [tuiSlideInTop],
 })
 export class TuiTableBarsHostComponent {
+    readonly animation = {
+        value: '',
+        ...this.options,
+    };
```
MarsiBarsi, line 22: nit: you can use `as const` here

MarsiBarsi (approved):

PR #398, opened 2021-05-14
Refactor/core exports

sean-perkins:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [ ] Feature
-   [x] Refactoring (no functional changes, no api changes)

All contents of a file are exported up through the barrel files.

## What is the new behavior?

Exports are named, only exporting out the public methods/variables from those classes/etc.

-   [ ] Yes
-   [x] No

## Other information

`TuiScrollbarWrapperDirective` wasn't previously exported to the public API. I didn't notice anything calling it explicitly internal, so I added that with the named exports.


sean-perkins:
@waterplea I don't actually know how to build TaigaUI locally out to peek at the bundle (outside of using the Angular CLI manually). Is there a specific way you guys are building these out locally or through Github actions? Want to make sure how I'm verifying the output aligns with how the packages are published to NPM.

I looked at the `publish` script, but it's referencing a `build:cdk` command that does not exist.

MarsiBarsi:
@sean-perkins thanks for this PR! You can build core library typing `npm run build cdk && npm run build core`

waterplea:
@sean-perkins Thank for taking this on! Please add explicit typing in `core/components/dialog/dialog.service.ts` on line 22:
`protected readonly defaultOptions: TuiDialogOptions<any> = DEFAULT_OPTIONS;`
looks like TypeScript is going crazy with imports like that in the compiled files:
`import("../../../../dist/core/types/dialog-options").TuiDialogBaseOptions`
@MarsiBarsi had a check for dist folder after build but somehow it didn't work this time, we need to check it next week.

sean-perkins:
@waterplea Can do - although I don't believe this PR solves the output issues around Angular 12 compatibility. 

With the tsconfig for prod builds having `enableIvy` as false, the generated output is still problematic for both the `main` branch and this PR. 

Setting `enableIvy` to true however, resolves the output issues for both main and this PR. 

I don't know if it's expected to start distributing libraries with Ivy enabled for 12+ now. If it is, that's definitely a major breaking change for TaigaUI. If that does resolve the issues, the tedious nature of named exports doesn't seem to add value to the library. 

I'd propose I close out this PR. Sounds like you've addressed the short-term issues with Angular 12. If there continues to be build failures with Angular 12, it may make sense to just release an Ivy build for Taiga.

Thoughts?

waterplea:
We cannot release ivy build because that would require bumping angular version to 12, I don't think that would be backwards compatible with Angular 9 projects which we must support. Unfortunately my fix didn't solve all issues, when it's installed from npm, although it seemed to fix them when I build locally and add it to angular 12 project manually.

sean-perkins:
Ok, I'll continue to research and see if there's a way to exclude the generated output getting barreled from ViewEngine or at least prevent it from colliding with other generated output. 

Angular's documentation still refers library authors to not ship Ivy builds to support Angular 9, so there's got to be a way to get around this issue.

waterplea:
I guess ivy extracts a lot of things into separate exports during compilation of node modules and they all are getting named `ɵ0`. I will go through all of them and make another PR separating them into named exports. I don't think it's an issue with * imports anymore.

waterplea:
Looks like this is caused by Webpack 5. Our secondary entry points do not know that they will get bundled up into higher level entry point and they make non-unique auto separated names which happen for all inlined `useValue`. I will go over all of them. @MarsiBarsi we would need to come up with some way to guard against those, maybe search `dist` for `ɵ0` as a first of those auto generated names. This also might increase bundle size since I would have to use clear names for those separate constants. Anyway, I'll make it work and then we will figure out how to move forward. I'll close this PR, @sean-perkins thank you for your help!

PR #402, opened 2021-05-15
feat(core): base skeleton vibe animation on opacity instead of backgroundColor

filipwiniarski:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [x] Feature

Skeleton animation uses backgroundColor property which causes page repaint each frame. With lots of content on a page, many lazy loaded images, or heavy tasks in the background, this can cause noticeable performance issue.

[https://developers.google.com/web/fundamentals/performance/rendering/stick-to-compositor-only-properties-and-manage-layer-count](https://developers.google.com/web/fundamentals/performance/rendering/stick-to-compositor-only-properties-and-manage-layer-count)

Closes #

## What is the new behavior?

tuiSkeletonBackgroundVibe has been replaced with tuiSkeletonVibe animation which only uses opacity property instead. Skeleton elements now have base background color, and the "vibe" effect is achieved by manipulating opacity.

To make it work for TuiLazyLoadingDirective, new host binding for background color had to be made, making the opacity animation visible. Since opacity animation would still be visible after image load, there had to be a _load_ event detection followed by an animation cancel. 

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/kit/directives/lazy-loading/lazy-loading.directive.ts
+++ b/projects/kit/directives/lazy-loading/lazy-loading.directive.ts
@@ -44,16 +44,12 @@ export class TuiLazyLoadingDirective {
     ) {
         fromEvent(this.elementRef.nativeElement, 'load')
             .pipe(takeUntil(destroy$), watch(changeDetectorRef))
-            .subscribe(() => this.cancelSkeletonAnimation());
+            .subscribe(() => (this.animation = ''));
 
         if (!this.supported) {
             this.src$.subscribe(src => {
                 this.src = src;
             });
         }
-    }
-
-    private cancelSkeletonAnimation() {
-        this.animation = '';
     }
 }
--- a/projects/kit/directives/lazy-loading/lazy-loading.directive.ts
+++ b/projects/kit/directives/lazy-loading/lazy-loading.directive.ts
@@ -7,7 +7,7 @@ import {
     Input,
 } from '@angular/core';
 import {IntersectionObserverService} from '@ng-web-apis/intersection-observer';
-import {TuiDestroyService} from '@taiga-ui/cdk';
+import {TuiDestroyService, watch} from '@taiga-ui/cdk';
 import {fromEvent} from 'rxjs';
 import {takeUntil} from 'rxjs/operators';
 import {TuiLazyLoadingService} from './lazy-loading.service';
@@ -37,14 +37,14 @@ export class TuiLazyLoadingDirective {
     constructor(
         @Inject(TuiDestroyService)
         private readonly destroy$: TuiDestroyService,
-        @Inject(ChangeDetectorRef) private readonly changeDetectorRef: ChangeDetectorRef,
         @Inject(TuiLazyLoadingService)
         private readonly src$: TuiLazyLoadingService,
         @Inject(ElementRef)
         private readonly elementRef: ElementRef<HTMLImageElement>,
+        @Inject(ChangeDetectorRef) changeDetectorRef: ChangeDetectorRef,
     ) {
         fromEvent(this.elementRef.nativeElement, 'load')
-            .pipe(takeUntil(this.destroy$))
+            .pipe(takeUntil(this.destroy$), watch(changeDetectorRef))
             .subscribe(() => this.cancelSkeletonAnimation());
 
         if (!this.supported) {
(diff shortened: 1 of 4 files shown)


projects/kit/directives/lazy-loading/lazy-loading.directive.ts
```diff
@@ -1,14 +1,29 @@
...
 import {TuiLazyLoadingService} from './lazy-loading.service';
 
+import {fromEvent} from 'rxjs';
+import {takeUntil} from 'rxjs/operators';
+
 @Directive({
     selector: 'img[loading="lazy"]',
-    providers: [TuiLazyLoadingService, IntersectionObserverService, TuiDestroyService],
+    providers: [
+        TuiDestroyService,
```
waterplea, line 19: DestroyService is here already

projects/kit/directives/lazy-loading/lazy-loading.directive.ts
```diff
@@ -18,21 +33,35 @@ export class TuiLazyLoadingDirective {
         this.src$.next(src);
     }
 
+    animation = 'tuiSkeletonVibe ease-in-out 1s infinite alternate';
```
waterplea, line 36: I think it would be more obvious to use explicit HostBinding here since now this thing changes

projects/kit/directives/lazy-loading/lazy-loading.directive.ts
```diff
@@ -1,14 +1,29 @@
...
+    ChangeDetectorRef,
+    Directive,
+    ElementRef,
+    HostBinding,
+    Inject,
+    Input,
+} from '@angular/core';
 import {IntersectionObserverService} from '@ng-web-apis/intersection-observer';
 import {TuiDestroyService} from '@taiga-ui/cdk';
 import {TuiLazyLoadingService} from './lazy-loading.service';
```
waterplea, line 12: Let's remove this empty line so all imports are connected ;)

projects/kit/directives/lazy-loading/lazy-loading.directive.ts
```diff
@@ -18,21 +33,35 @@ export class TuiLazyLoadingDirective {
...
+        @Inject(TuiDestroyService)
+        private readonly destroy$: TuiDestroyService,
+        @Inject(ChangeDetectorRef) private readonly changeDetectorRef: ChangeDetectorRef,
         @Inject(TuiLazyLoadingService)
         private readonly src$: TuiLazyLoadingService,
         @Inject(ElementRef)
         private readonly elementRef: ElementRef<HTMLImageElement>,
     ) {
+        fromEvent(this.elementRef.nativeElement, 'load')
+            .pipe(takeUntil(this.destroy$))
```
waterplea, line 53: you can use `watch(changeDetectorRef)` inside pipe so you don't have to store it on `this`. Also since that would make `cancelSkeletonAnimation` just one line let's have it here directly.

waterplea (changes requested):
Thank you for implementing this! A few little comments there ;)

projects/kit/directives/lazy-loading/lazy-loading.directive.ts
herodot-filipwiniarski: That is so neat! 😎

projects/kit/directives/lazy-loading/lazy-loading.directive.ts
filipwiniarski: Good catch! Done

projects/kit/directives/lazy-loading/lazy-loading.directive.ts
filipwiniarski: Done

projects/kit/directives/lazy-loading/lazy-loading.directive.ts
filipwiniarski: Finding "Import sources within a group must be alphabetized." combination was actually harder than expected 😅 Done

projects/kit/directives/lazy-loading/lazy-loading.directive.ts
filipwiniarski: That is so neat, done! Also I've removed unnecessary class scope from `destroy$` as well.

waterplea (approved):

PR #432, opened 2021-05-22
fix(core): remove calendar jumps

artemv01:
## PR Checklist

## PR Type

-   [x] Bugfix
-   [ ] Feature

When user selects a date in calendarRange and current range limit exceeds number of days in either left / right side calendars, we do jump on another month.

Closes https://github.com/TinkoffCreditSystems/taiga-ui/issues/418

## What is the new behavior?

When user selects a date in calendarRange and current range limit exceeds number of days in either left / right side calendars, we do not jump on another month.

-   [ ] Yes
-   [x] No

## Other information

Not sure if we need the foolproof in ngOnInit, lmk if i need to remove that.

### DIFF
--- a/projects/core/components/calendar/calendar.component.ts
+++ b/projects/core/components/calendar/calendar.component.ts
@@ -3,6 +3,7 @@ import {
     Component,
     EventEmitter,
     Input,
+    OnInit,
     Output,
 } from '@angular/core';
 import {
@@ -29,7 +30,7 @@ import {TuiMarkerHandler} from '@taiga-ui/core/types';
     styleUrls: ['./calendar.style.less'],
     changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class TuiCalendarComponent implements TuiWithOptionalMinMax<TuiDay> {
+export class TuiCalendarComponent implements TuiWithOptionalMinMax<TuiDay>, OnInit {
     @Input()
     @tuiDefaultProp()
     month = TuiMonth.currentLocal();
@@ -81,25 +82,21 @@ export class TuiCalendarComponent implements TuiWithOptionalMinMax<TuiDay> {
 
     year: TuiYear | null = null;
 
+    ngOnInit(): void {
+        const {month, computedMinViewedMonth, computedMaxViewedMonth} = this;
+
+        if (month.monthBefore(computedMinViewedMonth)) {
+            this.month = computedMinViewedMonth;
+        } else if (month.monthAfter(computedMaxViewedMonth)) {
+            this.month = computedMaxViewedMonth;
+        }
+    }
+
     readonly disabledItemHandlerMapper: TuiMapper<
         TuiBooleanHandler<TuiDay>,
         TuiBooleanHandler<TuiDay>
     > = (disabledItemHandler, min: TuiDay, max: TuiDay) => item =>
         item.dayBefore(min) || item.dayAfter(max) || disabledItemHandler(item);
-
-    get computedMonth(): TuiMonth {
-        const {month, computedMinViewedMonth, computedMaxViewedMonth} = this;
-
-        if (month.monthBefore(computedMinViewedMonth)) {
-            return computedMinViewedMonth;
-        }
-
-        if (month.monthAfter(computedMaxViewedMonth)) {
-            return computedMaxViewedMonth;
-        }
-
-        return month;
-    }
 
     get computedMinViewedMonth(): TuiMonth {
         return this.minViewedMonth.monthSameOrAfter(this.min)
@@ -119,7 +116,7 @@ export class TuiCalendarComponent implements TuiWithOptionalMinMax<TuiDay> {
 
     onPickerYearClick({year}: TuiYear) {
         this.year = null;
-        this.updateViewedMonth(new TuiMonth(year, this.computedMonth.month));
+        this.updateViewedMonth(new TuiMonth(year, this.month.month));
     }
 
     onPaginationValueChange(month: TuiMonth) {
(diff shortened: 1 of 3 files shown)


projects/core/components/calendar/calendar.component.ts
```diff
@@ -81,26 +82,22 @@ export class TuiCalendarComponent implements TuiWithOptionalMinMax<TuiDay> {
...
     year: TuiYear | null = null;
 
-    readonly disabledItemHandlerMapper: TuiMapper<
-        TuiBooleanHandler<TuiDay>,
-        TuiBooleanHandler<TuiDay>
-    > = (disabledItemHandler, min: TuiDay, max: TuiDay) => item =>
-        item.dayBefore(min) || item.dayAfter(max) || disabledItemHandler(item);
-
-    get computedMonth(): TuiMonth {
+    ngOnInit(): void {
```
waterplea, line 85: Month is an input so it can change. Therefore ngOnInit is not gonna do the trick. I think you can switch this to:

```
@Input('month')
set monthSetter(month: TuiMonth) {
... your clamping of the month to the valid value
}

month = TuiMonth.currentLocal();
```

waterplea (changes requested):
Hey, thanks for looking into it! I have one comment, please take a look :)

waterplea (approved):

projects/core/components/calendar/calendar.component.ts
```diff
@@ -30,8 +30,18 @@ import {TuiMarkerHandler} from '@taiga-ui/core/types';
     changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class TuiCalendarComponent implements TuiWithOptionalMinMax<TuiDay> {
-    @Input()
-    @tuiDefaultProp()
+    @Input('month')
+    set monthSetter(month: TuiMonth) {
+        const {computedMinViewedMonth, computedMaxViewedMonth} = this;
+
+        if (month.monthBefore(computedMinViewedMonth)) {
```
waterplea, line 37: I think it would look clearer if we moved that into separate function returning `TuiMonth`

waterplea (changes requested):
Added a small comment + looks like the tests are failing. Do you have access to test run results?

projects/core/components/calendar/calendar.component.ts
artemv01: Smth like this?
```
 @Input('month')
    set monthSetter(month: TuiMonth) {
        this.month = this.getComputedMonth(month);
    }
    month = TuiMonth.currentLocal();
...
private getComputedMonth(month: TuiMonth): TuiMonth {
        const {computedMinViewedMonth, computedMaxViewedMonth} = this;

        if (month.monthBefore(computedMinViewedMonth)) {
            return computedMinViewedMonth;
        } else if (month.monthAfter(computedMaxViewedMonth)) {
            return computedMaxViewedMonth;
        } else {
            return month;
        }
    }
```

waterplea: Right, except with returns you would not need `else if`, just 2 separate `if` and `return` at the very end with no `else` so it would look better :)

artemv01: @waterplea hmm tests error out using that approach. i think we could also add a getter so all look classy?
```
@Input()
    set month(month: TuiMonth) {
        this._month = this.getComputedMonth(month);
    }
    get month() {
        return this._month;
    }
    _month = TuiMonth.currentLocal();

```

artemv01:
yeah I do have access. I'll fix them

projects/core/components/calendar/calendar.component.ts
waterplea: I don't think we need a getter. We basically have property month and a setter just to make sure we get a valid value.

artemv01: yeah @waterplea but without the above getter these tests fail:
```
  it('if set month is less than viewed, it takes computed min viewed', () => {
        const month = TuiMonth.currentLocal();
        const monthAfter = month.append({month: 1});

        component.minViewedMonth = monthAfter;
        component.month = month;

        expect(component.month).toBe(monthAfter);
    });

    it('if set month is more than max, it takes max viewed', () => {
        component.max = TuiDay.currentLocal().append({month: -1});
        component.month = TuiDay.currentLocal();
        expect(component.month).toBe(component.max);
    });
```
Looks like .month doesn't access the setter, but the property itself.

waterplea: Yes, if you do it imperatively with code, change it out monthSetter =. Only angular templates know about input aliases

artemv01: ok @waterplea though now we have 2 ways of accessing the same value in the component: `monthSetter` and `month` which hits readability so IMO getter with `_month` could be preferable .

waterplea: Sure, that's reasonable. Make it the way you want. Underscore private props just tend to annoy me for some reason )

artemv01: @waterplea when we clamp our `month` in the setter we depend on other input values like `maxViewedMonth` , `max`, which means that we depend on the order in which the input bindings were set in the template as you now. So i don't think this cool , we could use ngOnCHanges because it doesn't look component's binding will be updated often or just remove the clamping overall.

artemv01: @waterplea removed clamping, i think this one can be merged if everything's ok

waterplea (approved):
Thank you! Sorry it turned into such a drag :)

PR #433, opened 2021-05-23
fix(kit): textMaskOptions pipe checked countryCode in previous value

### LINKED ISSUE #419: [BUG] `InputPhone` removing value leave country code
# 🐞 Bug report

### Description

When you hit backspace and remove whole number of `InputPhone`, then click away — country code remains in the input, although it is treated as empty value (label flies down). It disappears on the next focus. Can be reproduced on demo portal:
https://taiga-ui.dev/components/input-phone

imitrakov:
## PR Checklist

## PR Type

-   [x] Bugfix
-   [ ] Feature

Closes #419 

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/kit/components/input-phone/input-phone.component.ts
+++ b/projects/kit/components/input-phone/input-phone.component.ts
@@ -156,6 +156,14 @@ export class TuiInputPhoneComponent
             : this.search || '';
     }
 
+    set nativeValue(value: string) {
+        setTimeout(() => {
+            if (this.nativeFocusableElement) {
+                this.nativeFocusableElement.value = value;
+            }
+        });
+    }
+
     get inputMode(): TuiInputMode {
         return this.allowText ? TuiInputMode.Text : TuiInputMode.Numeric;
     }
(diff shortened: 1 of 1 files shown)


projects/kit/components/input-phone/input-phone.component.ts
```diff
@@ -96,14 +96,21 @@ export class TuiInputPhoneComponent
...
-        pipe: value => {
+        pipe: (value, config) => {
             if (this.allowText) {
                 return value;
             }
 
-            return value === '' && this.focused && !this.readOnly
-                ? `${this.countryCode} `
-                : value.replace(/-$/, '');
+            if (value === '' && this.focused && !this.readOnly) {
```
waterplea, line 104: I think it's better to just reset native input to empty value upon leaving the control. You can set `nativeFocusableElement.value = ''` or create a setter for `get nativeValue` and do it in `onActiveZone` at the bottom of it.

waterplea (changes requested):
Hi! Thank you for this PR. Check out my comment, I think I've spotted an easier way to fix it :)

projects/kit/components/input-phone/input-phone.component.ts
```diff
@@ -156,6 +156,14 @@ export class TuiInputPhoneComponent
             : this.search || '';
     }
 
+    set nativeValue(value: string) {
+        setTimeout(() => {
```
waterplea, line 160: @imitrakov  Why do you need the setTimeout here?

imitrakov: Or return as I did before, when check  countryCode in pipe textMask? :)

imitrakov:
Hi! Check out my comment, please :)

projects/kit/components/input-phone/input-phone.component.ts
waterplea: I see now. Previously active zone worked a little different and by the time pipe kicked in `focused` was already `false`. Looks like it's not as easy as I though 🤔

waterplea:
Hi @imitrakov. Thank you for fixing this, but it turned out to be a tip of a bigger issue we have with `ActiveZone`. If you leave control with Tab key it would work fine but clicking away causing this issue — that's because by the time `ActiveZone` fires `false` from `mousedown` the focus is still in the zone. This is one of the trickier parts of the library so I'm going to have to close this PR and make my own fixing the root cause. Thank you for helping us uncover it!

PR #449, opened 2021-05-29
perf(cdk): replace `setTimeout` with rAF and skip change detection

arturovt:
## PR Checklist

## PR Type

-   [x] Refactoring (no functional changes, no api changes)

-   [x] No

I don't wanna create breaking changes, but I'm not sure if we need to run `tick()` when the element gets focused since it's done via the native browser's API, which doesn't require Angular to run change detection. That's why I've also removed `markForCheck()` to prevent Angular running change detection down to the autofocus directive after `setTimeout` fires.


waterplea:
Hi @arturovt. Any particular reason you came up with this PR? Were you seeing any issues related to autofocus feature?

waterplea:
@arturovt pinging you to see if this is still relevant. This directive is messy as is, thanks to brilliant iOS webkit. I wouldn't want to introduce yet another bunch of workarounds unless there are strong reasons to do so.

PR #499, opened 2021-06-22
chore(demo-integrations): configure basic integration tests

vladimirpotekhin:
## PR Checklist

-   [ ] The commit message follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/)

## PR Type

-   [ ] Bugfix
-   [ ] Feature

Closes #

-   [ ] Yes
-   [ ] No


projects/demo-integrations/cypress/tests/demo-paths.ts
```diff
@@ -0,0 +1,89 @@
+export const DEMO_PATHS = [
```
waterplea, line 1: Would be great if we could automate this stuff. Can we reach for routing config?

projects/demo-integrations/cypress/tests/shared.entities.ts
```diff
@@ -0,0 +1,3 @@
+export const EXAMPLE_ID = 'tui-doc-example';
+export const PROPS_AND_METHODS =
+    '%D0%A1%D0%B2%D0%BE%D0%B9%D1%81%D1%82%D0%B2%D0%B0_%D0%B8_%D0%BC%D0%B5%D1%82%D0%BE%D0%B4%D1%8B';
```
waterplea, line 3: This looks like Russian tab name from proprietary portal. Should be `API`, I guess

projects/demo-integrations/cypress/tests/demo/exclusions.ts
```diff
@@ -0,0 +1,10 @@
+const map = new Map([
+    ['components/select', [5, 6]],
```
waterplea, line 2: Why are we excluding those?

vladimirpotekhin: I remove the 6th example from the exclusion. 5th example makes the tests unstable due to the loader.

projects/demo-integrations/cypress/tests/shared.entities.ts
vladimirpotekhin: Thanks, I delete this one. This is an unnecessary const for deep tests, I'll add it in a future PR

projects/demo-integrations/cypress/tests/demo-paths.ts
vladimirpotekhin: Now the paths are taken from the constant `pages` from demo with exclusions

waterplea (approved):

PR #503, opened 2021-06-23
fix(cdk): `TuiDialogHostComponent` opening many different types dialogs together

nsbarsukov:
## PR Checklist

## PR Type

-   [x] Bugfix
-   [ ] Feature

Closes #

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/cdk/abstract/dialog.service.ts
+++ b/projects/cdk/abstract/dialog.service.ts
@@ -37,6 +37,7 @@ export abstract class AbstractTuiDialogService<T extends {}> extends Observable<
                 completeWith,
                 $implicit: observer,
                 component: this.component,
+                createdAt: Date.now(),
                 id: this.idService.generate(),
             };
 
--- a/projects/cdk/components/dialog-host/dialog-host.component.ts
+++ b/projects/cdk/components/dialog-host/dialog-host.component.ts
@@ -2,7 +2,8 @@ import {ChangeDetectionStrategy, Component, Inject} from '@angular/core';
 import {TUI_PARENT_ANIMATION} from '@taiga-ui/cdk/constants';
 import {TuiAriaDialogContext} from '@taiga-ui/cdk/interfaces';
 import {TUI_DIALOGS} from '@taiga-ui/cdk/tokens';
-import {merge, Observable} from 'rxjs';
+import {combineLatest, Observable} from 'rxjs';
+import {map} from 'rxjs/operators';
 
 // @dynamic
 @Component({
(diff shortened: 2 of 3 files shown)


projects/cdk/components/dialog-host/dialog-host.component.ts
```diff
@@ -13,10 +14,12 @@ import {merge, Observable} from 'rxjs';
     animations: [TUI_PARENT_ANIMATION],
 })
 export class TuiDialogHostComponent<T extends TuiAriaDialogContext> {
-    readonly dialogs$ = merge(...this.dialogs);
+    readonly dialogs$ = combineLatest(this.allTypesDialogs).pipe(
+        map(allTypesDialogs => new Array<T>().concat(...allTypesDialogs)),
```
nsbarsukov, line 18: i've understood right now that it is bad solution: it doesn't take order of dialog opening into consideration =/

I move PR to draft and think more

nsbarsukov: fixed

projects/cdk/abstract/dialog.service.ts
```diff
@@ -37,6 +37,7 @@ export abstract class AbstractTuiDialogService<T extends {}> extends Observable<
                 completeWith,
                 $implicit: observer,
                 component: this.component,
+                createdAt: Date.now(),
```
waterplea, line 40: Maybe we can just use id below? It's autogenerated based on ever increasing increment and `Date.now()`

projects/cdk/components/dialog-host/dialog-host.component.ts
```diff
@@ -13,10 +14,16 @@ import {merge, Observable} from 'rxjs';
...
+            new Array<T>()
+                .concat(...allTypesDialogs)
+                .sort((a, b) => a.createdAt - b.createdAt),
+        ),
+    );
 
     constructor(
         @Inject(TUI_DIALOGS)
-        private readonly dialogs: ReadonlyArray<Observable<ReadonlyArray<T>>>,
+        private readonly allTypesDialogs: Observable<readonly T[]>[],
```
waterplea, line 27: Let's rename it to `dialogsByType` if you want to emphasize these are arrays of different type dialogs.

projects/cdk/abstract/dialog.service.ts
nsbarsukov: i'm not sure that it is good idea.

If I use id I will need to remove `tui_` prefix from id (before converting to the number).

Moreover, does it violate the principles of Dependency Injection if component knows how  TuiIdService generate id?
TuiIdService can be changed in future that will break logic inside dialogs.

projects/cdk/components/dialog-host/dialog-host.component.ts
nsbarsukov: done

waterplea (approved):

PR #516, opened 2021-06-29
feat(i18n): add Dutch translation for countries

maartentibau:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [x] Feature

Closes #

-   [ ] Yes
-   [x] No


waterplea:
@maartentibau thank you for a quick PR! Please merge main into your branch, it had issues with imports that is now fixed. That's why checks are failing.

projects/i18n/languages/dutch/countries.ts
```diff
@@ -0,0 +1,208 @@
+import {TuiCountryIsoCode} from '../../enums/country-iso-code';
```
nsbarsukov, line 1: replace all relative imports by absolute, please
(it breaks build)

> import {TuiCountryIsoCode} from '@taiga-ui/i18n/enums';
> import {TUI_DUTCH_LANGUAGE_COUNTRIES} from '@taiga-ui/i18n/languages/dutch';

maartentibau:
@waterplea any idea why the build is still failing? I personally do not understand this error message, I already looked into the files, but cannot really figure out what is wrong.

image

projects/i18n/languages/dutch/countries.ts
nsbarsukov: @maartentibau you've fixed only the first import

maartentibau: Well the import that is done in the `core.ts` file of a language can refer to a local folder, so I do not really understand why than suddenly you need to do an import with `@taiga-ui/i18n/...`

I just looked at how it was done in the `russian/core.ts` file and there the import was just like this:

`import {TUI_RUSSIAN_LANGUAGE_COUNTRIES} from './countries';

So I assumed that for dutch this would be the same way then as

`import {TUI_DUTCH_LANGUAGE_COUNTRIES} from './countries';

waterplea:
@nsbarsukov looks like tests for international input phone are flickery or broken. They were passing before but now are failing.

projects/i18n/languages/dutch/core.ts
```diff
@@ -1,5 +1,5 @@
 import {LanguageCore} from '@taiga-ui/i18n/interfaces';
-import {TUI_ENGLISH_LANGUAGE_COUNTRIES} from '@taiga-ui/i18n/languages/english';
+import {TUI_DUTCH_LANGUAGE_COUNTRIES} from '@taiga-ui/i18n/languages/dutch';
```
waterplea, line 2: Inside one entry point paths must be relative. They way it is written now is for cross points references. That's why you are getting cycling dependency on itself. This should be relative here.

projects/i18n/languages/russian/core.ts
```diff
@@ -1,5 +1,5 @@
 import {LanguageCore} from '@taiga-ui/i18n/interfaces';
-import {TUI_RUSSIAN_LANGUAGE_COUNTRIES} from './countries';
+import {TUI_RUSSIAN_LANGUAGE_COUNTRIES} from '@taiga-ui/i18n/languages/russian';
```
waterplea, line 2: This should be changed back to the way it was

waterplea (changes requested):
Please have a look at my comments, this should fix building issues

projects/i18n/languages/dutch/countries.ts
waterplea: Yes, it was correct because it is in the same entry point. You need to do this across entry points, like in the case of enum.

nsbarsukov: My mistake ;(
@maartentibau  sorry for confusing

projects/i18n/languages/russian/core.ts
maartentibau: Done

waterplea:
Now it's an issue with autogenerated private exports. I will merge it and solve myself so next translations will just have a reference to copy-paste. This happens due to lots of secondary entry points and re-exports. Thank you, @maartentibau!

waterplea:
For the record this was missing for Russian:
`export * from './countries';`
so it was added as a private export which clashed with the same name with some other export:
https://github.com/TinkoffCreditSystems/taiga-ui/pull/518

PR #561, opened 2021-07-16
feat(kit): `MultiSelect` + `InputTag` size `s`

nsbarsukov:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [x] Feature

Closes #

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/kit/components/input-tag/input-tag.component.ts
+++ b/projects/kit/components/input-tag/input-tag.component.ts
@@ -212,7 +212,7 @@ export class TuiInputTagComponent
     get labelOutside(): boolean {
         const {size, labelOutside} = this.controller;
 
-        return size !== 's' ? labelOutside : true;
+        return size === 's' || labelOutside;
     }
 
     get hasCleaner(): boolean {
--- a/projects/kit/components/input-tag/input-tag.component.ts
+++ b/projects/kit/components/input-tag/input-tag.component.ts
@@ -210,7 +210,9 @@ export class TuiInputTagComponent
 
     @HostBinding('class._label-outside')
     get labelOutside(): boolean {
-        return this.controller.labelOutside;
+        const {size, labelOutside} = this.controller;
+
+        return size !== 's' ? labelOutside : true;
     }
 
     get hasCleaner(): boolean {
--- a/projects/kit/components/input-tag/input-tag.style.less
+++ b/projects/kit/components/input-tag/input-tag.style.less
@@ -1,5 +1,7 @@
 @import 'taiga-ui-local';
 
+@tag-vertical-space: 0.125rem;
+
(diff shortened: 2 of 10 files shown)


projects/kit/components/input-tag/input-tag.style.less
```diff
@@ -66,9 +71,14 @@
     padding: 0;
     align-items: flex-start;
 
+    :host[data-tui-host-size='s'] &,
     :host[data-tui-host-size='m'] & {
         padding-left: 0.75rem;
     }
+
+    :host[data-tui-host-size='s'] & {
+        padding-right: 0;
```
waterplea, line 80: Do we need it? It looks like it's 0 anyway

waterplea (commented):
Spacings look a bit off when it turns to multiline:
image
Also it probably make sense to force switch into labelOutside = true on small size with some getter to avoid this:
image

projects/kit/components/input-tag/input-tag.style.less
nsbarsukov: yes, we need it because of [this line](https://github.com/TinkoffCreditSystems/taiga-ui/blob/main/projects/core/styles/mixins/textfield.less#L37)

nsbarsukov:
@waterplea 

> to force switch into labelOutside = true on small size with some getter

✅

> Spacings look a bit off when it turns to multiline

✅

projects/kit/components/input-tag/input-tag.component.ts
```diff
@@ -210,7 +211,9 @@ export class TuiInputTagComponent
 
     @HostBinding('class._label-outside')
     get labelOutside(): boolean {
-        return this.controller.labelOutside;
+        const {size, labelOutside} = this.controller;
+
+        return sizeBigger(size, 's') ? labelOutside : true;
```
waterplea, line 216: I think we can just use `size === 's'` here

nsbarsukov: done

projects/kit/components/input-tag/input-tag.component.ts
```diff
@@ -210,7 +210,9 @@ export class TuiInputTagComponent
 
     @HostBinding('class._label-outside')
     get labelOutside(): boolean {
-        return this.controller.labelOutside;
+        const {size, labelOutside} = this.controller;
+
+        return size !== 's' ? labelOutside : true;
```
waterplea, line 215: `size === 's' || labelOutside` 😎

nsbarsukov: done

waterplea (approved):

PR #566, opened 2021-07-19
feat(kit): `decimalSeparator` for input value decimals formatting

shoom3301:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [x] Feature

Closes #

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/kit/components/input-number/input-number.component.ts
+++ b/projects/kit/components/input-number/input-number.component.ts
@@ -25,6 +25,7 @@ import {
     maskedNumberStringToNumber,
     tuiCreateAutoCorrectedNumberPipe,
     tuiCreateNumberMask,
+    TuiDecimalSymbol,
     TuiDecimalT,
     TuiPrimitiveTextfieldComponent,
     TuiTextMaskOptions,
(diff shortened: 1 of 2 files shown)


waterplea:
I feel this should rather be a i18n token of numbers internationalization of some kind. We probably need to carefully read this stuff :)
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

waterplea:
@vladimirpotekhin @nsbarsukov @MarsiBarsi what do you think? Maybe add a token `TUI_NUMBER_FORMAT` with an object of one key `decimalSeparator` for the time being? I don't think people need to change it on the fly, right @shoom3301? It would also be used in `tui-money` component as well.

MarsiBarsi:
> @vladimirpotekhin @nsbarsukov @MarsiBarsi what do you think? Maybe add a token `TUI_NUMBER_FORMAT` with an object of one key `decimalSeparator` for the time being? I don't think people need to change it on the fly, right @shoom3301? It would also be used in `tui-money` component as well.

Sure, I like the idea of one DI token because we can set these rules for the whole app once

waterplea:
Looks like `core` package is a good place for it to go. @shoom3301 will you add it please? Here are all the places where it should be used:

- `Money` component
- `InputNumber` component
- `InputSlider` component
- `InputRange` component
- `FormatNumber` pipe, default value for the decimal separator

shoom3301:
https://github.com/TinkoffCreditSystems/taiga-ui/pull/676

PR #586, opened 2021-07-28
fix(cdk): make `markControlAsTouchedAndValidate` work with empty FormArray and FormGroup

uberpwner:
## PR Checklist

## PR Type

-   [x] Bugfix
-   [ ] Feature

Currently `markControlAsTouchedAndValidate` is not setting `touched: true` on FormArray or FormGroup with empty controls

## What is the new behavior?

Set `touched: true` on empty FormArray or FormGroup

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/cdk/utils/miscellaneous/mark-control-as-touched-and-validate.ts
+++ b/projects/cdk/utils/miscellaneous/mark-control-as-touched-and-validate.ts
@@ -1,6 +1,9 @@
 import {AbstractControl, FormArray, FormGroup} from '@angular/forms';
 
 export function markControlAsTouchedAndValidate(control: AbstractControl) {
+    control.markAsTouched();
+    control.updateValueAndValidity();
+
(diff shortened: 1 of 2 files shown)


projects/cdk/utils/miscellaneous/mark-control-as-touched-and-validate.ts
```diff
@@ -16,7 +19,4 @@ export function markControlAsTouchedAndValidate(control: AbstractControl) {
 
         return;
```
waterplea, line 20: Unnecessary `return` now. I wonder if order is important. Maybe we should just remove both `return` so that for every array/group it would start with deep nested controls and after all child controls are processed it would process the group/array itself.

uberpwner: You're right, that would be shorter

waterplea (approved):

PR #635, opened 2021-08-20
feat(toggle): allow configuration of default options

KarimovDev:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [x] Feature

Partially implements #133

## What is the new behavior?

Default toggle options can be set via a provider. This is a partial implementation of #133

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/kit/components/toggle/toggle-options.ts
+++ b/projects/kit/components/toggle/toggle-options.ts
@@ -1,0 +1,28 @@
+import {InjectionToken} from '@angular/core';
+import {TuiContextWithImplicit} from '@taiga-ui/cdk';
+import {TuiSizeL} from '@taiga-ui/core';
+import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
+
+export interface ToggleOptions {
+    readonly icons: Readonly<{
+        toggleOff: PolymorpheusContent<TuiContextWithImplicit<TuiSizeL>>;
+        toggleOn: PolymorpheusContent<TuiContextWithImplicit<TuiSizeL>>;
+    }>;
+}
+
+/** Default values for the toggle options. */
+export const TUI_TOGGLE_DEFAULT_OPTIONS: ToggleOptions = {
+    icons: {
+        toggleOff: ({$implicit}) =>
+            $implicit === 'm' ? 'tuiIconToggleOff' : 'tuiIconToggleOffLarge',
+        toggleOn: ({$implicit}) =>
+            $implicit === 'm' ? 'tuiIconToggleOn' : 'tuiIconToggleOnLarge',
+    },
+};
+
+export const TUI_TOGGLE_OPTIONS = new InjectionToken<ToggleOptions>(
+    'Default parameters for toggle component',
+    {
+        factory: () => TUI_TOGGLE_DEFAULT_OPTIONS,
+    },
+);
--- a/projects/kit/components/toggle/toggle-options.ts
+++ b/projects/kit/components/toggle/toggle-options.ts
@@ -8,16 +8,24 @@ export interface ToggleOptions {
         toggleOff: PolymorpheusContent<TuiContextWithImplicit<TuiSizeL>>;
         toggleOn: PolymorpheusContent<TuiContextWithImplicit<TuiSizeL>>;
     }>;
+    readonly singleColor: boolean;
+    readonly showIcons: boolean;
+    readonly size: TuiSizeL;
 }
 
 /** Default values for the toggle options. */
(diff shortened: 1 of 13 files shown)


projects/kit/components/toggle/toggle-options.ts
```diff
@@ -0,0 +1,28 @@
...
+    readonly icons: Readonly<{
+        toggleOff: PolymorpheusContent<TuiContextWithImplicit<TuiSizeL>>;
+        toggleOn: PolymorpheusContent<TuiContextWithImplicit<TuiSizeL>>;
+    }>;
+}
+
+/** Default values for the toggle options. */
+export const TUI_TOGGLE_DEFAULT_OPTIONS: ToggleOptions = {
+    icons: {
+        toggleOff: ({$implicit}) =>
```
waterplea, line 16: Please change it to actual function instead of an arrow (see [here](https://github.com/TinkoffCreditSystems/taiga-ui/blob/main/projects/core/components/primitive-checkbox/checkbox-options.ts#L29)). Otherwise Angular creates a separate private export with auto-generated name for it and we get naming conflict because of many secondary entry points.

projects/kit/components/toggle/toggle-options.ts
```diff
@@ -0,0 +1,28 @@
+import {InjectionToken} from '@angular/core';
+import {TuiContextWithImplicit} from '@taiga-ui/cdk';
+import {TuiSizeL} from '@taiga-ui/core';
+import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
+
+export interface ToggleOptions {
+    readonly icons: Readonly<{
```
waterplea, line 7: Please also include other parameters: `singleColor`, `showIcons` and `size`. We probably do not need to include `showLoader` because it's a value people toggle on and off dynamically.

projects/kit/components/toggle/toggle.template.html
```diff
@@ -7,33 +7,52 @@
...
-            [showLoader]="checked"
-        ></tui-loader>
-        <tui-svg
-            *ngIf="showIcons && !showLoader"
-            class="icon"
-            automation-id="tui-toggle__check-icon"
-            [src]="iconOn"
-        ></tui-svg>
+        <span
+            polymorpheus-outlet
```
waterplea, line 11: Right now custom content makes it impossible to use `showLoader`. This should only be applied to the `tui-svg` with icon below.

projects/kit/components/toggle/toggle.template.html
```diff
@@ -7,33 +7,52 @@
...
+            class="icon-wrapper"
+            [content]="iconOn"
+            [context]="context"
```
waterplea, line 14: I recently discovered that Angular optimizes inline objects, so we can just write `{ $implicit: size }` here and drop getter and pure private method.

projects/kit/components/toggle/toggle.template.html
```diff
@@ -7,33 +7,52 @@
...
-            [showLoader]="!checked"
-        ></tui-loader>
-        <tui-svg
-            *ngIf="showIcons && !showLoader"
-            class="icon icon_off"
-            automation-id="tui-toggle__cancel-icon"
-            [src]="iconOff"
-        ></tui-svg>
+        <span
+            polymorpheus-outlet
```
waterplea, line 35: Same two comments are applied here

waterplea (changes requested):
Thank you for this PR! Please check out my comments ;)

projects/kit/components/toggle/toggle-options.ts
KarimovDev: Interesting behavior. I've changed it

projects/kit/components/toggle/toggle-options.ts
KarimovDev: Done

projects/kit/components/toggle/toggle.template.html
KarimovDev: Ops, my fault. I've corrected )

projects/kit/components/toggle/toggle.template.html
KarimovDev: Wow! That's cool. Changed

projects/kit/components/toggle/toggle.template.html
KarimovDev: Done

KarimovDev:
@waterplea, also I've added tests for new token

waterplea (approved):

vladimirpotekhin (approved):

PR #636, opened 2021-08-21
feat(notification-alert): allow configuration of default options

KarimovDev:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [x] Feature

Partially implements #133

## What is the new behavior?

Default toggle options can be set via a provider. This is a partial implementation of #133

-   [ ] Yes
-   [x] No


projects/core/modules/notifications/notification-alert/notification-alert-options.ts
```diff
@@ -0,0 +1,22 @@
+import {InjectionToken} from '@angular/core';
+import {TuiContextWithImplicit} from '@taiga-ui/cdk';
+import {TuiNotification} from '@taiga-ui/core/enums';
+import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
+
+export interface NotificationAlertOptions {
+    readonly autoClose: boolean | number;
+    readonly label: PolymorpheusContent<TuiContextWithImplicit<TuiNotification>>;
```
waterplea, line 8: It was designed to be plain string. Do you have a need to use templates there?

projects/core/modules/notifications/notification-alert/notification-alert-options.ts
```diff
@@ -0,0 +1,22 @@
+import {InjectionToken} from '@angular/core';
+import {TuiContextWithImplicit} from '@taiga-ui/cdk';
+import {TuiNotification} from '@taiga-ui/core/enums';
+import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
+
+export interface NotificationAlertOptions {
```
waterplea, line 6: Please add `hasIcons`, `status` and `hasCloseButton` too

projects/core/modules/notifications/notification-alert/notification-alert.template.html
```diff
@@ -4,13 +4,17 @@
...
-        *ngIf="safeItem.label"
-        automation-id="tui-notification-alert__heading"
-        class="heading"
-    >
-        {{safeItem.label}}
-    </label>
+    <span polymorpheus-outlet [content]="label" [context]="context">
+        <ng-template let-label>
+            <label
+                *ngIf="label"
```
waterplea, line 10: I think we should just add `polymorpheus-outlet` to this `label`. That is, if we end up expanding to polymorpheus here.

waterplea (changes requested):
Thank you for your help! Please check the comments and also reuse this token in `Notification` component for the inputs that overlap.

projects/core/modules/notifications/notification-alert/notification-alert-options.ts
KarimovDev: Nope. I thought about ability to change label depending on status. But your idea is good. I've made it polymorpheus )

What do you think about adding an ability to change autoClose timeout depending on status also? How can we implement this with polymorpheus? Or, in this case, should we just use custom primitive and function type?

projects/core/modules/notifications/notification-alert/notification-alert-options.ts
KarimovDev: Done

projects/core/modules/notifications/notification-alert/notification-alert.template.html
KarimovDev: Done

KarimovDev:
> Thank you for your help! Please check the comments and also reuse this token in `Notification` component for the inputs that overlap.

Done

projects/core/modules/notifications/notification-alert/Notification-alert.ts
```diff
@@ -8,15 +9,17 @@ import {
 } from '../notification-options';
 
 export class NotificationAlert<O, I> {
-    readonly status: TuiNotification;
+    readonly status: TuiNotification | undefined;
```
waterplea, line 12: Can we just add `?:` ?

KarimovDev: Yeap, done )

projects/core/modules/notifications/notification-alert/Notification-alert.ts
```diff
@@ -8,15 +9,15 @@ import {
 } from '../notification-options';
 
 export class NotificationAlert<O, I> {
-    readonly status: TuiNotification;
+    readonly status?: TuiNotification;
```
waterplea, line 12: Let's not change these to optional but instead let's inject the defaults into `TuiNotificationsService` and use it in `show` method:
```ts
const notification = new NotificationAlert(observer, content, {...this.options, ...options});
```
Also a good TODO for v3 is to make all keys of options interface mandatory and use `Partial` utility type where they all are optional. Right now we can add `Required` utility type to constructor parameter in this file.

waterplea: I wonder if it also reads better to make options `private readonly` and use them here instead of restructuring on constructor 🤔
```ts
readonly hasIcon = this.options.hasIcon;

readonly autoClose = this.options.autoClose;
//...
```

waterplea (changes requested):
@KarimovDev I'm coming back to this PR. Please see new comments :)

projects/core/modules/notifications/notification-alert/notification-alert-options.ts
waterplea: @vladimirpotekhin @nsbarsukov @MarsiBarsi what do you think about adding `(status: TuiNotification) => number | boolean` type to `autoClose`? I can see it as both overengineering and a reasonable addition so people can, say, make errors stick while regular notifications would autoclose. What are your thoughts?

projects/core/modules/notifications/notification-alert/notification-alert.template.html
```diff
@@ -1,15 +1,17 @@
...
+    *ngIf="hasCloseButton else noClose"
+    [status]="status"
+    [hasIcon]="hasIcon"
     (close)="closeNotification()"
 >
     <label
-        *ngIf="safeItem.label"
+        *ngIf="label"
+        polymorpheus-outlet
+        [content]="label"
```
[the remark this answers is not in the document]
KarimovDev: Done

waterplea (changes requested):
A couple small comments and let's go ahead and add `(status: TuiNotification) => number | boolean` to the `autoClose` type.

projects/core/tokens/notification-options.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export interface NotificationTokenOptions {
+    readonly autoClose: boolean | number;
+    readonly label: PolymorpheusContent<TuiContextWithImplicit<TuiNotification>>;
+    readonly status: TuiNotification;
+    readonly hasIcon: boolean;
+    readonly hasCloseButton: boolean;
+}
+
+/** Default values for the toggle options. */
```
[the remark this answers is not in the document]
KarimovDev: Oops, my fault

projects/core/modules/notifications/notification-alert/Notification-alert.ts
```diff
@@ -27,25 +26,13 @@ export class NotificationAlert<O, I> {
     constructor(
         observer: Observer<O>,
```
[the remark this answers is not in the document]
KarimovDev: Done

waterplea:
> > A couple small comments and let's go ahead and add `(status: TuiNotification) => number | boolean` to the `autoClose` type.
> 
> Done, with handler execution inside NotificationAlert class. @waterplea What do you think about moving notification-options to interfaces? In this case I can also create shared `TuiNotificationAutoClose` type and share it between token and input options

@KarimovDev yeah, let's add a dedicated type to autoClose

KarimovDev:
> > > A couple small comments and let's go ahead and add `(status: TuiNotification) => number | boolean` to the `autoClose` type.
> > 
> > 
> > Done, with handler execution inside NotificationAlert class. @waterplea What do you think about moving notification-options to interfaces? In this case I can also create shared `TuiNotificationAutoClose` type and share it between token and input options
> 
> @KarimovDev yeah, let's add a dedicated type to autoClose

Done

projects/core/interfaces/notification-options.ts
```diff
@@ -0,0 +1,20 @@
+import {TuiContextWithImplicit} from '@taiga-ui/cdk';
+import {TuiNotification} from '@taiga-ui/core/enums';
+import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
+
+export type TuiNotificationAutoClose =
```
waterplea, line 5: There's a folder `types` for types :) @vladimirpotekhin @nsbarsukov @Smirnov906 I wonder if it makes sense to keep this distinction 🤔

waterplea (approved):

projects/core/interfaces/notification-options.ts
vladimirpotekhin: I think it makes no sense

splincode: @waterplea It all depends on the style guides of the project )) I don't know yet what the rules are

PR #662, opened 2021-09-01
fix(addon-doc): fix overlapping by sticky header while using anchor links

nsbarsukov:
## PR Checklist

-   [X] The commit message follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/)

## PR Type

-   [ ] Bugfix
-   [ ] Feature
-   [X] Other... Please describe:

1. use any anchor link (for example, https://taiga-ui.dev/getting-started#options)
2. beginning of content overlapped by header

Closes #

## What is the new behavior?
No overlapping

-   [ ] Yes
-   [X] No

### DIFF
--- a/projects/addon-doc/src/components/main/main.style.less
+++ b/projects/addon-doc/src/components/main/main.style.less
@@ -1,5 +1,11 @@
 @import 'taiga-ui-local';
 
+@sticky-header-height: 4rem;
+
+html {
+    scroll-padding-top: @sticky-header-height;
+}
+
 @keyframes tuiShaking {
     from,
     to {


vladimirpotekhin (approved):

projects/demo/src/styles.less
```diff
@@ -3,6 +3,11 @@
 @import 'taiga-ui-theme';
 
 @sidebar-width: 15.5rem;
+@sticky-header-height: 4rem;
+
+html {
+    scroll-padding-top: @sticky-header-height;
```
waterplea, line 9: Wow, nice stuff!

waterplea (approved):

waterplea (changes requested):
This probably needs to be applied in `addon-doc` so it works everywhere

nsbarsukov:
@waterplea i applied it in `addon-doc` library

waterplea (approved):

hakimio:
@nsbarsukov I don't think this a good solution/workaround. Proper fix would be to have scrollbar at the right place: `tui-doc-page` instead of top level like it's done in [Angular Material documentation page](https://material.angular.io/components/card/overview).

waterplea:
@hakimio what concerns are you having? From my experience keeping global scroll global is a good idea. `scroll-padding-top` seems to be *exactly* the right solution, based on it's definition :)
> The `scroll-padding-top` property defines offsets for the top of the optimal viewing region of the scrollport: the region used as the target region for placing things in view of the user. This allows the author to exclude regions of the scrollport that are obscured by other content.

hakimio:
@waterplea you have three separate regions on the page: top header, left navigation menu and the "doc page". It makes sense that each of those regions have their own scrollbars. "doc-page" scrollbar extending over the top header looks like a bug, not a feature.

waterplea:
@hakimio That's not how I see it. I see it as the web page that has content and it is scrollable. And there's a sticky header which is always visible. And a sticky sidebar which is also always visible. And this CSS property exists for this particular case.

When you are making some region of the page scrollable, rather than the whole page you will have issues with scroll position restoration, you need to implement your own `ViewportScroller` for Angular apps. You will also need to write your own tools for automated scrolling of Cypress and e2e tests. And I'm pretty sure it's better from accessibility point of view. If you check angular.io and not the material demo portal they also have similar layout. 

So if you do not have concrete reasons to back up this concern let's go with this solution.

PR #675, opened 2021-09-06
feat(kit): allow configuration of default options for input-count

KarimovDev:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [x] Feature

Partially implements #133

## What is the new behavior?

Default input-count options can be set via a provider. This is a partial implementation of #133

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/kit/components/input-count/input-count-options.ts
+++ b/projects/kit/components/input-count/input-count-options.ts
@@ -1,0 +1,42 @@
+import {InjectionToken} from '@angular/core';
+import {TuiContextWithImplicit} from '@taiga-ui/cdk';
+import {TuiSizeL} from '@taiga-ui/core';
+import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
+
+export interface InputCountOptions {
+    readonly icons: Readonly<{
+        up: PolymorpheusContent<TuiContextWithImplicit<TuiSizeL>>;
+        down: PolymorpheusContent<TuiContextWithImplicit<TuiSizeL>>;
+    }>;
+    appearance: string;
+    hideButtons: boolean;
+    min: number;
+    max: number;
+    step: number;
+    postfix: string;
+}
+
+// TODO: remove in ivy compilation
+export const PASSWORD_ICON_UP = 'tuiIconPlus';
+export const PASSWORD_ICON_DOWN = 'tuiIconMinus';
+
+/** Default values for the input password options. */
+export const TUI_INPUT_COUNT_DEFAULT_OPTIONS: InputCountOptions = {
+    icons: {
+        up: PASSWORD_ICON_UP,
+        down: PASSWORD_ICON_DOWN,
+    },
+    appearance: 'textfield',
+    hideButtons: false,
+    min: 0,
+    max: Infinity,
+    step: 1,
+    postfix: '',
+};
+
+export const TUI_INPUT_COUNT_OPTIONS = new InjectionToken<InputCountOptions>(
+    'Default parameters for input count component',
+    {
+        factory: () => TUI_INPUT_COUNT_DEFAULT_OPTIONS,
+    },
+);
--- a/projects/kit/components/input-count/input-count-options.ts
+++ b/projects/kit/components/input-count/input-count-options.ts
@@ -18,7 +18,7 @@ export interface InputCountOptions {
 export const PASSWORD_ICON_UP = 'tuiIconPlus';
 export const PASSWORD_ICON_DOWN = 'tuiIconMinus';
 
-/** Default values for the input password options. */
+/** Default values for the input count options. */
 export const TUI_INPUT_COUNT_DEFAULT_OPTIONS: InputCountOptions = {
     icons: {
         up: PASSWORD_ICON_UP,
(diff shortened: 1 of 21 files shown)


projects/kit/components/input-count/input-count-options.ts
```diff
@@ -0,0 +1,42 @@
+import {InjectionToken} from '@angular/core';
+import {TuiContextWithImplicit} from '@taiga-ui/cdk';
+import {TuiSizeL} from '@taiga-ui/core';
+import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
+
+export interface InputCountOptions {
+    readonly icons: Readonly<{
+        up: PolymorpheusContent<TuiContextWithImplicit<TuiSizeL>>;
```
waterplea, line 8: Buttons are rather tiny, they always have size "s". How about we simplify it by keeping context `{}` — this way you can just pass it to buttons' `icon` input, as it is already polymorpheus.

projects/kit/components/input-count/input-count-options.ts
```diff
@@ -0,0 +1,42 @@
...
+export const PASSWORD_ICON_DOWN = 'tuiIconMinus';
+
+/** Default values for the input password options. */
+export const TUI_INPUT_COUNT_DEFAULT_OPTIONS: InputCountOptions = {
+    icons: {
+        up: PASSWORD_ICON_UP,
+        down: PASSWORD_ICON_DOWN,
+    },
+    appearance: 'textfield',
+    hideButtons: false,
```
waterplea, line 30: Right now we force hide buttons in tables by comparing appearance to table. That's rather hardcode. @vladimirpotekhin  @nsbarsukov what do you think, it probably makes sense to alter this token in `Table` directive now? With a factory, skipSelf and spread with `hideButtons` set to `true`? Sounds like that would be more idiomatic.

projects/kit/components/input-count/input-count-options.ts
KarimovDev: But buttons' height gets smaller when input-count uses "s" size. And this polymorpheus context is taken into account textfield size

projects/kit/components/input-count/input-count-options.ts
KarimovDev: In this case someone can rewrite token if they use nested components inside the table. Or am I wrong?

waterplea: In this case, if somebody uses component inside table and for some reason want to change it, they can create a directive that would override it, yes. I don't think that's bad.

projects/kit/components/input-count/input-count-options.ts
waterplea: I meant that you probably cannot fit a big icon anyway, even for a big input size. Template complexity really increases now and I don't want this token option feature to become an overkill.

projects/kit/components/input-count/input-count-options.ts
KarimovDev: Ok. I've already done this as an example. @vladimirpotekhin @nsbarsukov you can check it right here

projects/kit/components/input-count/input-count-options.ts
KarimovDev: Ok. I agree about template complexity. I've changed it to {} context

projects/kit/components/input-count/input-count.component.ts
```diff
@@ -118,7 +126,7 @@ export class TuiInputCountComponent
 
     @HostBinding('class._has-buttons')
     get hasButtons(): boolean {
```
waterplea, line 128: Unnecessary getter, let's revert what CSS does and bind it to input directly

KarimovDev: Done

projects/kit/components/input-count/input-count-options.ts
KarimovDev: @waterplea there is also tuiTableMode directive with TUI_TEXTFIELD_APPEARANCE token. Should we add TUI_INPUT_COUNT_OPTIONS here as well? In this case should we move TUI_INPUT_COUNT_OPTIONS into core package?

projects/kit/components/input-count/input-count-options.ts
KarimovDev: @waterplea there is also tuiTableMode directive with TUI_TEXTFIELD_APPEARANCE token. Should we add TUI_INPUT_COUNT_OPTIONS here as well? In this case, should we move TUI_INPUT_COUNT_OPTIONS into core package?

waterplea: No, that directive was there before we introduced table component. I believe we might end up just dropping it in the next major release.

waterplea (approved):

waterplea (approved):

projects/kit/components/input-count/input-count-options.ts
waterplea: Oops, @KarimovDev I see this directive is still actually in use on the demo and that fails the screenshots. Let's copy-paste this factory for the time being then.

KarimovDev: Done

waterplea (approved):

waterplea (approved):

PR #798, opened 2021-09-29
ci: `safe to test`

IKatsuba:


.github/workflows/firebase-hosting-pull-request.yml
```diff
@@ -2,13 +2,16 @@
...
+on:
+    pull_request_target:
+        types: [labeled]
 
 permissions: read-all
 
 jobs:
     build_and_preview:
         runs-on: ubuntu-latest
+        if: contains(github.event.pull_request.labels.*.name, 'safe to test')
```
waterplea, line 14: Let's name it more friendly. Since it's only about deployments, how about we add a label called "Deploy"?

PR #924, opened 2021-10-23
feat(i18n): add language switcher

artemv01:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [x] Feature

https://github.com/TinkoffCreditSystems/taiga-ui/pull/846#discussion_r725963229
Users can't change i18n for taiga-ui on the fly.

Closes #

Using special service users can change i18n on the fly. The selected locale saved to LOCAL_STORAGE and retrieved on page reload.
Usage example (will be update in https://github.com/TinkoffCreditSystems/taiga-ui/pull/846)
```
  TuiLanguageService,
    {
        provide: TUI_LANGUAGE,
        deps: [TUI_STORED_LANGUAGE],
        useExisting: TuiLanguageService,
    },
    {
        provide: TUI_DEFAULT_LANGUAGE,
        useValue: TUI_ITALIAN_LANGUAGE,
    },
    {
        provide: TUI_I18N_STORAGE_KEY,
        useValue: 'tui_key_i18n',
    }
```

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/i18n/constants/language-code.ts
+++ b/projects/i18n/constants/language-code.ts
@@ -1,0 +1,14 @@
+export enum TUI_LANGUAGE_CODE {
+    ENGLISH = 'en',
+    RUSSIAN = 'ru',
+    DUTCH = 'nl',
+    FRENCH = 'fr',
+    GERMAN = 'de',
+    POLISH = 'pl',
+    PORTUGUESE = 'pt',
+    SPANISH = 'es',
+    TURKISH = 'tr',
+    UKRANIAN = 'ua',
+    VIETNAMESE = 'vi',
+    ITALIAN = 'it',
+}
--- a/projects/i18n/constants/language-code.ts
+++ b/projects/i18n/constants/language-code.ts
@@ -1,4 +1,4 @@
-export enum TUI_LANGUAGE_CODE {
+export enum TuiLanguageCode {
     ENGLISH = 'en',
     RUSSIAN = 'ru',
     DUTCH = 'nl',
--- a/projects/i18n/tools/default-lanuage.ts
+++ b/projects/i18n/tools/default-lanuage.ts
@@ -5,8 +5,6 @@ import {TUI_ENGLISH_LANGUAGE} from '@taiga-ui/i18n/languages';
 export const TUI_DEFAULT_LANGUAGE = new InjectionToken<Language>(
     `Default language for Taiga UI libraries i18n`,
     {
-        factory: () => {
-            return TUI_ENGLISH_LANGUAGE;
-        },
+        factory: () => TUI_ENGLISH_LANGUAGE,
     },
 );
--- a/projects/i18n/tools/default-lanuage.ts
+++ b/projects/i18n/tools/default-lanuage.ts
@@ -1,0 +1,12 @@
+import {InjectionToken} from '@angular/core';
+import {Language} from '@taiga-ui/i18n/interfaces';
+import {TUI_ENGLISH_LANGUAGE} from '@taiga-ui/i18n/languages';
+
+export const TUI_DEFAULT_LANGUAGE = new InjectionToken<Language>(
+    `Default language for Taiga UI libraries i18n`,
+    {
+        factory: () => {
+            return TUI_ENGLISH_LANGUAGE;
+        },
+    },
+);
--- a/projects/i18n/tools/language.service.ts
+++ b/projects/i18n/tools/language.service.ts
@@ -1,0 +1,19 @@
+import {inject, Injectable} from '@angular/core';
+import {LOCAL_STORAGE} from '@ng-web-apis/common';
+import {Language} from '@taiga-ui/i18n';
+import {TUI_I18N_STORAGE_KEY, TUI_STORED_LANGUAGE} from '@taiga-ui/i18n/tools';
+import {ReplaySubject} from 'rxjs';
+@Injectable()
+export class TuiLanguageService extends ReplaySubject<Language> {
+    storedLanguage = inject(TUI_STORED_LANGUAGE);
+    storageKey = inject(TUI_I18N_STORAGE_KEY);
+    storage = inject(LOCAL_STORAGE);
+    constructor() {
+        super(1);
+        this.next(this.storedLanguage);
+    }
+    setLang(lang: Language) {
+        this.storage.setItem(this.storageKey, lang.code);
+        super.next(lang);
+    }
+}
--- a/projects/i18n/tools/language.service.ts
+++ b/projects/i18n/tools/language.service.ts
@@ -1,8 +1,9 @@
 import {Inject, Injectable} from '@angular/core';
 import {LOCAL_STORAGE} from '@ng-web-apis/common';
-import {Language} from '..';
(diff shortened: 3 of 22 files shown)


projects/i18n/tools/language.ts
```diff
@@ -9,3 +21,19 @@ export const TUI_LANGUAGE = new InjectionToken<Observable<Language>>(
         factory: () => of(TUI_ENGLISH_LANGUAGE),
     },
 );
+
+export const TUI_LANGUAGE_MAP = new Map<string, Language>(
+    Object.entries({
```
waterplea, line 26: This way all the translations get into bundle

waterplea:
This seems a little too much, but I'm also not sure how to make it simpler. Without storing selected language in localStorage this makes no sense. And to store it in localStorage we need to add all those tokens. I need to think on this a little :)

projects/i18n/tools/default-lanuage.ts
```diff
@@ -0,0 +1,12 @@
+import {InjectionToken} from '@angular/core';
+import {Language} from '@taiga-ui/i18n/interfaces';
+import {TUI_ENGLISH_LANGUAGE} from '@taiga-ui/i18n/languages';
+
+export const TUI_DEFAULT_LANGUAGE = new InjectionToken<Language>(
+    `Default language for Taiga UI libraries i18n`,
+    {
+        factory: () => {
```
waterplea, line 8: unnecessary { and return

projects/i18n/tools/language.service.ts
```diff
@@ -0,0 +1,19 @@
+import {inject, Injectable} from '@angular/core';
+import {LOCAL_STORAGE} from '@ng-web-apis/common';
+import {Language} from '@taiga-ui/i18n';
+import {TUI_I18N_STORAGE_KEY, TUI_STORED_LANGUAGE} from '@taiga-ui/i18n/tools';
+import {ReplaySubject} from 'rxjs';
+@Injectable()
+export class TuiLanguageService extends ReplaySubject<Language> {
+    storedLanguage = inject(TUI_STORED_LANGUAGE);
```
waterplea, line 8: We do not write `inject` like that but rather add them with `@Inject` in constructor

projects/i18n/tools/language.service.ts
```diff
@@ -0,0 +1,19 @@
+import {inject, Injectable} from '@angular/core';
+import {LOCAL_STORAGE} from '@ng-web-apis/common';
+import {Language} from '@taiga-ui/i18n';
+import {TUI_I18N_STORAGE_KEY, TUI_STORED_LANGUAGE} from '@taiga-ui/i18n/tools';
+import {ReplaySubject} from 'rxjs';
+@Injectable()
+export class TuiLanguageService extends ReplaySubject<Language> {
```
waterplea, line 7: It feels more like a `BehaviorSubject` with `super(value)` rather than `ReplaySubject(1)` with `next` in constructor.

projects/i18n/tools/language.service.ts
```diff
@@ -0,0 +1,19 @@
+import {inject, Injectable} from '@angular/core';
+import {LOCAL_STORAGE} from '@ng-web-apis/common';
+import {Language} from '@taiga-ui/i18n';
+import {TUI_I18N_STORAGE_KEY, TUI_STORED_LANGUAGE} from '@taiga-ui/i18n/tools';
+import {ReplaySubject} from 'rxjs';
+@Injectable()
```
waterplea, line 6: providedIn: root

projects/i18n/tools/stored-i18n-key.ts
```diff
@@ -0,0 +1,12 @@
+import {InjectionToken} from '@angular/core';
+
+const TUI_I18N_STORAGE_DEFAULT_KEY = 'taiga_ui_i18n';
+
+export const TUI_I18N_STORAGE_KEY = new InjectionToken<string>(
+    `Storage key for language of Taiga UI libraries i18n`,
+    {
+        factory: () => {
```
waterplea, line 8: Unnecessary { and return

projects/i18n/tools/language.service.ts
```diff
@@ -0,0 +1,19 @@
...
+    storedLanguage = inject(TUI_STORED_LANGUAGE);
+    storageKey = inject(TUI_I18N_STORAGE_KEY);
+    storage = inject(LOCAL_STORAGE);
+    constructor() {
+        super(1);
+        this.next(this.storedLanguage);
+    }
+    setLang(lang: Language) {
+        this.storage.setItem(this.storageKey, lang.code);
+        super.next(lang);
```
waterplea, line 17: Either this method should be just `next` or this should be `this.next` and not `super.next`

projects/i18n/tools/stored-language.ts
```diff
@@ -0,0 +1,26 @@
...
+    `Stored language for Taiga UI libraries i18n`,
+    {
+        factory: () => {
+            const storage = inject(LOCAL_STORAGE);
+            const defaultLanguage = inject(TUI_DEFAULT_LANGUAGE);
+            const storageKey = inject(TUI_I18N_STORAGE_KEY);
+            const storedLanguageCode = storage.getItem(storageKey);
+            const storedLanguage =
+                storedLanguageCode &&
+                TUI_LANGUAGE_MAP.has(storedLanguageCode) &&
```
waterplea, line 18: Typescript will still think it might be undefined so this line does nothing, really.

projects/i18n/tools/stored-language.ts
```diff
@@ -0,0 +1,26 @@
...
+                TUI_LANGUAGE_MAP.get(storedLanguageCode);
+
+            return storedLanguage
+                ? storedLanguage
+                : defaultLanguage || TUI_ENGLISH_LANGUAGE;
```
waterplea, line 23: `return storedLanguage ?? defaultLanguage` as there can not NOT be `defaultLanguage`

waterplea:
Please go through all my comments and I will take a second look. Seems like a good solution, but we need to figure out a way to not pull all the translations into the bundle, only the needed ones.

projects/i18n/constants/language-code.ts
```diff
@@ -0,0 +1,14 @@
+export enum TUI_LANGUAGE_CODE {
```
waterplea, line 1: That's not really a constant :)

projects/i18n/tools/stored-language.ts
artemv01: TS gives me a null check error when I delete that)

artemv01:
> Please go through all my comments and I will take a second look. Seems like a good solution, but we need to figure out a way to not pull all the translations into the bundle, only the needed ones.

hi @MarsiBarsi sorry for the delay, i pushed all code review changes. I have an idea about the way we could sort out the above: we could make users provide the `TUI_LANGUAGE_MAP` , and we make a default injector for that which consists of english translation only.

waterplea:
@artemv01 maybe we can use something like that to lazy-load the translation?

```ts
function getTranslation(language: string): Observable<TuiLanguage> {
    return from(import(`@taiga-ui/i18n/languages/${language}`));
}
```

splincode:
fixed - https://github.com/Tinkoff/taiga-ui/pull/2263

PR #1002, opened 2021-11-15
fix(core): baseline alignment for postfix in value-decoration

splincode:
## PR Checklist

-   [x] The commit message follows [Conventional Commits](https://www.conventionalcommits.org/en/)

## PR Type

-   [x] Bugfix
-   [ ] Feature
-   [ ] Refactoring
-   [ ] Code style update

Closes https://github.com/TinkoffCreditSystems/taiga-ui/issues/989

image

## What is the new behavior?

image

image

- value is empty

image

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/core/components/primitive-textfield/value-decoration/value-decoration.style.less
+++ b/projects/core/components/primitive-textfield/value-decoration/value-decoration.style.less
@@ -19,6 +19,23 @@
     text-overflow: clip;
 }
 
+.t-postfix {
+    display: inline-block;
+    line-height: 0;
+    margin-top: 1.15ch;
+
+    &:not(&--is-empty):before {
+        /**
+         @note(splincode): emulate native whitespace before input value
+         */
+        content: '\00a0';
+    }
+}
+
+.t-ghost:empty + .t-postfix:before {
+    content: '';
+}
+
 .t-filler {
     color: var(--tui-text-03);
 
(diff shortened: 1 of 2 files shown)


projects/core/components/primitive-textfield/value-decoration/value-decoration.style.less
```diff
@@ -19,6 +19,23 @@
...
+.t-postfix {
+    display: inline-block;
+    line-height: 0;
+    margin-top: 1.15ch;
+
+    &:not(&--is-empty):before {
+        /**
+         @note(splincode): emulate native whitespace before input value
+         */
+        content: '\00a0';
```
waterplea, line 31: Currently whitespace is added through the getter so this makes two whitespaces. And the reason it is added with a getter is `%` sign has to be shown without space according to our typographic guidelines :) If you really want to move it to CSS it can be done though

projects/core/components/primitive-textfield/value-decoration/value-decoration.template.html
```diff
@@ -1,4 +1,4 @@
 <span #pre subtree characterData waMutationObserver>{{prefix}}</span>
-<span class="t-ghost">{{value}}</span>
 <span class="t-filler">{{filler}}</span>
```
waterplea, line 2: Filler has to go after ghost:
image

You can still achieve your desired CSS effect with `~` instead of `+`.

waterplea:
Like I said, we tried and fixes that work on some resolutions, DPIs and zooms fail on others
image

PR #1012, opened 2021-11-16
chore: harnesses first draft

waterplea:
This is a first attempt at covering components with Harnesses. More on this concept:
https://material.angular.io/cdk/test-harnesses/overview#api-for-test-authors

In future this should allow us to write better tests and drop `automation-id` attributes to decrease bundle size + it would also allow everybody to use our component in tests more comfortably, including our internal teams.

Harness environment for Cypress can be found here:
https://github.com/jscutlery/devkit


projects/testing/src/core/loader/loader.harness.ts
```diff
@@ -0,0 +1,16 @@
...
+
+    async isLoading(): Promise<boolean> {
+        return !!(await this.loader());
+    }
+
+    async getText(): Promise<string> {
+        const loader = await this.loader();
+
+        return loader ? loader.text() : '';
+    }
```
splincode, line 15:
```ts
return loader?.text() ?? '';
```

splincode (approved):

vladimirpotekhin (approved):

splincode:
@waterplea What do you think?
`chore: harnesses first draft` -> `feat(testing): add harnesses utils`

waterplea:
> @waterplea What do you think? `chore: harnesses first draft` -> `feat(testing): add harnesses utils`

We do not yet publish the testing package so this does not yet belong in the changelog :)

PR #1026, opened 2021-11-21
feat(core): allow configuration of default button options

### LINKED ISSUE #133: 🚀 - Use providers to inject default configurations

vladimirpotekhin:
## PR Checklist

-   [ ] The commit message follows [Conventional Commits](https://www.conventionalcommits.org/en/)

## PR Type

-   [ ] Bugfix
-   [x] Feature
-   [ ] Refactoring
-   [ ] Code style update

Closes #133 

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/core/components/button/button-options.ts
+++ b/projects/core/components/button/button-options.ts
@@ -1,0 +1,27 @@
+import {InjectionToken} from '@angular/core';
+import {TuiAppearance} from '@taiga-ui/core/enums';
+import {TuiSizeXL, TuiSizeXS} from '@taiga-ui/core/types';
+import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
+
+export interface ButtonOptions {
+    readonly size: TuiSizeXS | TuiSizeXL;
+    readonly appearance: TuiAppearance | string;
+    readonly shape: 'square' | 'rounded' | null;
+    readonly icon: PolymorpheusContent;
+    readonly iconRight: PolymorpheusContent;
+}
+
+export const TUI_BUTTON_DEFAULT_OPTIONS: ButtonOptions = {
+    size: 'l',
+    shape: null,
+    appearance: TuiAppearance.Primary,
+    icon: '',
+    iconRight: '',
+};
+
+export const TUI_BUTTON_OPTIONS = new InjectionToken<ButtonOptions>(
+    'Default parameters for button component',
+    {
+        factory: () => TUI_BUTTON_DEFAULT_OPTIONS,
+    },
+);
--- a/projects/core/components/button/button-options.ts
+++ b/projects/core/components/button/button-options.ts
@@ -1,22 +1,17 @@
 import {InjectionToken} from '@angular/core';
 import {TuiAppearance} from '@taiga-ui/core/enums';
 import {TuiSizeXL, TuiSizeXS} from '@taiga-ui/core/types';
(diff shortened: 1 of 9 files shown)


projects/core/components/button/button-options.ts
```diff
@@ -0,0 +1,27 @@
+import {InjectionToken} from '@angular/core';
+import {TuiAppearance} from '@taiga-ui/core/enums';
+import {TuiSizeXL, TuiSizeXS} from '@taiga-ui/core/types';
+import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
+
+export interface ButtonOptions {
+    readonly size: TuiSizeXS | TuiSizeXL;
+    readonly appearance: TuiAppearance | string;
+    readonly shape: 'square' | 'rounded' | null;
+    readonly icon: PolymorpheusContent;
```
waterplea, line 10: I don't think it makes sense to configure icon with DI. It is different for each instance anyway.

projects/core/components/button/button.component.ts
```diff
@@ -22,11 +22,10 @@ import {
...
 } from '@taiga-ui/cdk';
-import {TuiAppearance} from '@taiga-ui/core/enums';
-import {TuiSizeS, TuiSizeXL, TuiSizeXS} from '@taiga-ui/core/types';
-import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
+import {TuiSizeS} from '@taiga-ui/core/types';
 import {Observable} from 'rxjs';
 import {takeUntil} from 'rxjs/operators';
+import {ButtonOptions, TUI_BUTTON_OPTIONS} from './button-options';
 
 @Component({
```
waterplea, line 30: Let's add interface implementation so the types are more apparent.

projects/demo/src/modules/components/button/examples/5/index.html
```diff
@@ -0,0 +1 @@
+<button tuiButton>Default options</button>
```
waterplea, line 1: Isn't it exactly NOT the default options? 🙂

projects/core/components/button/button.component.ts
vladimirpotekhin: done

projects/demo/src/modules/components/button/examples/5/index.html
vladimirpotekhin: u r right) renamed

projects/core/components/button/button-options.ts
vladimirpotekhin: removed 👍

splincode (approved):

waterplea (approved):

Issue #648, opened 2021-08-25
[FEATURE] `addon-doc` Long menu items


vladimirpotekhin:
# 🚀 Feature request

### Is your feature request related to a problem?

Long menu items don't fit in the side-menu

### Describe the solution you'd like

Perhaps we should show each menu item in one line with elipsis, or make it customizable

### Additional context

![image](https://user-images.githubusercontent.com/46284632/130826624-ae936a90-eb9e-4bd3-aa8e-93c431c856ac.png)

yurakhomitsky:
To implement this feature all we have to do is to introduce the "max-width": 100% property in the tuiLink component and add these styles to the content block

```
white-space: nowrap;
text-overflow: ellipsis;
width: 100%;
overflow: hidden;
vertical-align: top;

```

Let me know if that makes sense to you so I can go ahead and introduce this changes.

waterplea:
Can you add max-width inplace too, not on every tuiLink? I'd like to build in as little styles as possible in those low level components.

PR #1080, opened 2021-12-08
feat(addon-doc): truncate link content

yurakhomitsky:
## PR Checklist

-   [X] The commit message follows [Conventional Commits](https://www.conventionalcommits.org/en/)

## PR Type

-   [ ] Bugfix
-   [X] Feature
-   [ ] Refactoring
-   [ ] Code style update

Closes [#648](https://github.com/TinkoffCreditSystems/taiga-ui/issues/648)

-   [ ] Yes
-   [X] No

### DIFF
--- a/projects/addon-doc/src/components/navigation/navigation.style.less
+++ b/projects/addon-doc/src/components/navigation/navigation.style.less
@@ -55,20 +55,17 @@
     .clearbtn();
     font: var(--tui-font-text-m);
     margin-left: 0.5rem;
-    line-height: 2.75rem;
+    padding-top: 1rem;
 
     &_small {
         margin-left: 1rem;
         line-height: 1.5rem;
-
-        &:not(:first-child) {
-            margin-top: 1rem;
-        }
+        padding-top: 0.5rem;
     }
 
     &_subsection {
         margin-left: 0;
-        line-height: 2.75rem;
+        line-height: 1.6rem;
     }
 
     &_active {


waterplea (changes requested):
I looked closer at this issue and I think it would be better if we overflow it to the second line. Sublinks already look fine:
image

The problem is with the main links which have a big line-height:
image

So instead of cutting content let's remove `line-height` and add vertical `padding` instead. Something around `0.6rem` should look roughly the same. Just need to make sure all kinds of links look right, regular, those that have subitems and subitems themselves.

projects/addon-doc/src/components/navigation/navigation.style.less
```diff
@@ -55,7 +55,7 @@
     .clearbtn();
     font: var(--tui-font-text-m);
     margin-left: 0.5rem;
-    line-height: 2.75rem;
+    padding-top: 0.6rem;
```
waterplea, line 58: This is not enough. There's also `&_small` below and even if I add the same fix, scaling it according to smaller line-height I get this:
image

Please try multiline in all possible situations for the navigation menu:
image

waterplea (changes requested):

projects/addon-doc/src/components/navigation/navigation.style.less
yurakhomitsky: I should decrease the line-height on each of these menu items or I'm missing something?

I don't quite understand how the result should look like

yurakhomitsky: For example, for the **Common** accordion item, we will have to remove the text-overflow styles because it won't allow doing multiline
```
.title {
    .text-overflow();
    margin-right: 0;
    flex-grow: 1;
}
```

yurakhomitsky: This is how it looks like now
image

waterplea: Let's make it 1rem for `sublink` and 0.5rem for `sublink_small`. And we also need to get rid of this rule:
```css
        &:not(:first-child) {
            margin-top: 1rem;
        }
```
And then I believe we're good to merge

yurakhomitsky: To which property apply 1rem and 0.5rem?
0.5rem line-height not gonna work for `sublink_small`
image

waterplea: Yeah, I meant padding, sorry

waterplea (approved):
Looks great now! Thank you :)

MarsiBarsi (approved):

MarsiBarsi:
🚀

splincode (approved):

Issue #816, opened 2021-10-03
🚀 - `Accordion` support eager content


rvalitov:
### Which @taiga-ui/* package(s) are relevant/releated to the feature request?

core

### Description

I use Accordion and Expand components to show/hide a big search form with search results. I want to use the components to allow the user temporarily show/hide the content (both search form, and search results) so it does not consume space on screen when not needed. However, after the user hides and then opens again the Expand or Accordion, the inner content of the components is reset, i.e. it does not preserve the state, the filled fields of the form and the search results. Everything is reinitialized.

I suspect the cause of this reason is that both of these components use `[*ngIf]` to show/hide the content. And it results in such behavior. In order to preserve the state and contents, the `[hide]` should be used.

Example of my code:
```html
<tui-expand [expanded]="expandedClientSelect">
          <ng-template tuiExpandContent>
            <app-clients-list
              [allowCreate]="true"
              [allowSearch]="true"
              [allowCancel]="false"
              (onSelect)="onClientSelect($event)"
            >
            </app-clients-list>
          </ng-template>
        </tui-expand>
```

It would be great if there would be an option like `[useHide]` that define the behavior of the Expand and Accordion components.

waterplea:
This isn't a bug, it's by design. You can drop `ng-template` from `tui-expand` and it would work the way you want. However it's not possible in `tui-accordion`. This could probably be turned into feature request for Accordion.

rvalitov:
@waterplea thank you for a workaround. Yes, after dropping `ng-template` it seems to work properly. But in the browser console now I get error every time the Expand is closed:
```
ERROR Error: NG0100: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value for '_focused': 'true'. Current value: 'false'.. Find more at https://angular.io/errors/NG0100
```

Is it something serious?

waterplea:
Not sure, I'll probably need a StackBlitz to dig into this.

waterplea:
For accordion feature we can probably add another directive to put on content that would project it without template directly into the expand.

PR #1081, opened 2021-12-08
feat(kit): add support for eager content for accordion-item

yurakhomitsky:
## PR Checklist

-   [X] The commit message follows [Conventional Commits](https://www.conventionalcommits.org/en/)

## PR Type

-   [ ] Bugfix
-   [X] Feature
-   [ ] Refactoring
-   [ ] Code style update

Closes [#816](https://github.com/TinkoffCreditSystems/taiga-ui/issues/816)

-   [ ] Yes
-   [X] No

### DIFF
--- a/projects/demo/src/modules/components/accordion/accordion.component.ts
+++ b/projects/demo/src/modules/components/accordion/accordion.component.ts
@@ -1,6 +1,7 @@
 import {DOCUMENT} from '@angular/common';
 import {Component, ElementRef, Inject, ViewChild} from '@angular/core';
 import {changeDetection} from '@demo/emulate/change-detection';
+import {TuiDocExample} from '@taiga-ui/addon-doc';
 import {tuiCustomEvent} from '@taiga-ui/cdk';
 import {TUI_EXPAND_LOADED, TuiSizeS} from '@taiga-ui/core';
 
@@ -54,6 +55,12 @@ export class ExampleTuiAccordionComponent {
         TypeScript: example3Ts,
         HTML: example3Html,
         LESS: example3Less,
+    };
+
+    readonly example4: TuiDocExample = {
+        TypeScript: import('!!raw-loader!./examples/4/index.html'),
+        HTML: import('!!raw-loader!./examples/4/index.html'),
+        LESS: import('!!raw-loader!./examples/4/index.less'),
     };
 
     readonly bordersVariants = ['all', 'top-bottom'];
--- a/projects/kit/components/accordion/accordion-item/accordion-item-content.directive.ts
+++ b/projects/kit/components/accordion/accordion-item/accordion-item-content.directive.ts
@@ -1,9 +1,8 @@
 import {ChangeDetectorRef, Directive, Inject, Self, TemplateRef} from '@angular/core';
 import {PolymorpheusTemplate} from '@tinkoff/ng-polymorpheus';
 
-// TODO: do not rebuild the template, think about "lazy" flag
 @Directive({
-    selector: '[tuiAccordionItemContent]',
+    selector: 'ng-template[tuiAccordionItemContent]',
 })
 export class TuiAccordionItemContentDirective extends PolymorpheusTemplate<{}> {
     constructor(
--- a/projects/kit/components/accordion/accordion-item/accordion-item-eager-content.directive.ts
+++ b/projects/kit/components/accordion/accordion-item/accordion-item-eager-content.directive.ts
@@ -1,0 +1,6 @@
+import {Directive} from '@angular/core';
+
+@Directive({
+    selector: '[tuiAccordionItemContent]:not(ng-template)',
+})
+export class TuiAccordionItemEagerContentDirective {}
--- a/projects/kit/components/accordion/accordion-item/accordion-item.component.ts
+++ b/projects/kit/components/accordion/accordion-item/accordion-item.component.ts
@@ -24,6 +24,7 @@ import {MODE_PROVIDER, TUI_MODE, TuiBrightness, TuiSizeS} from '@taiga-ui/core';
 import {Observable} from 'rxjs';
 
 import {TuiAccordionItemContentDirective} from './accordion-item-content.directive';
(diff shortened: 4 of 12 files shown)


projects/kit/components/accordion/accordion-item/accordion-item-eager-content.directive.ts
```diff
@@ -0,0 +1,16 @@
+import {ChangeDetectorRef, Directive, Inject, Self, TemplateRef} from '@angular/core';
+import {PolymorpheusTemplate} from '@tinkoff/ng-polymorpheus';
+
+@Directive({
+    selector: '[tuiAccordionItemEagerContent]',
+})
+export class TuiAccordionItemEagerContentDirective extends PolymorpheusTemplate<{}> {
```
waterplea, line 7: If it's eager content, there's no need to extend template, it will not be a template, you would put it on a plain DOM element.

projects/kit/components/accordion/accordion-item/accordion-item-eager-content.directive.ts
```diff
@@ -0,0 +1,16 @@
+import {ChangeDetectorRef, Directive, Inject, Self, TemplateRef} from '@angular/core';
+import {PolymorpheusTemplate} from '@tinkoff/ng-polymorpheus';
+
+@Directive({
+    selector: '[tuiAccordionItemEagerContent]',
```
waterplea, line 5: I like this idea, how about we keep the same selector but write here `[tuiAccordionItemContent]:not(ng-template)` and on the other, lazy directive we will change selector to `ng-template[tuiAccordionItemContent]`. This way people will not need to remember two selectors, it will be the same but for eager you put it on, say, a `div` and for lazy content you put it on a `ng-template` (it would also work with a star syntax: `<div *tuiAccordionItemContent>` for lazy and `<div tuiAccordionItemContent>` for eager).

waterplea (changes requested):
Thank you for your PR, I like the idea. Please check my comments and it would be great if you add another example in the demo to show how to use it.

waterplea:
BTW, to fix failing build I believe you need to add your new directive to `index.ts` inside `accordion` folder.

projects/kit/components/accordion/accordion-item/accordion-item-eager-content.directive.ts
yurakhomitsky: Yeah, this make sense, thank you

projects/kit/components/accordion/accordion-item/accordion-item-eager-content.directive.ts
yurakhomitsky: Updated directive selectors to suggested

projects/kit/components/accordion/accordion-item/accordion-item.template.html
```diff
@@ -28,12 +28,16 @@
...
         <ng-template tuiExpandContent>
             <div
+                *ngIf="lazyContent"
                 polymorpheus-outlet
                 automation-id="tui-accordion__item-content"
                 class="content"
-                [content]="content"
+                [content]="lazyContent"
             ></div>
         </ng-template>
```
yurakhomitsky, line 42: I introduced `ngIf` because `ng-template` is always projected in the `expand` component therefore we ended up with duplicated styles from the `.content` class. 

I'm not sure if it is a correct implementation

projects/kit/components/accordion/accordion-item/accordion-item-content.directive.ts
```diff
@@ -3,7 +3,7 @@ import {PolymorpheusTemplate} from '@tinkoff/ng-polymorpheus';
 
 // TODO: do not rebuild the template, think about "lazy" flag
```
waterplea, line 4: This `TODO` can now be removed :)

projects/kit/components/accordion/accordion-item/accordion-item.template.html
waterplea: I don't see a better solution so let's keep it that way 👍

waterplea (approved):

projects/demo/src/modules/components/accordion/accordion.component.ts
```diff
@@ -56,6 +59,12 @@ export class ExampleTuiAccordionComponent {
         LESS: example3Less,
     };
 
+    readonly example4: FrontEndExample = {
+        TypeScript: example4Ts,
```
splincode, line 63: please, use import directly `TypeScript: import('!!raw-loader!./examples/4/index.html'), ...`

projects/kit/components/accordion/accordion-item/accordion-item.component.ts
```diff
@@ -88,8 +89,11 @@ export class TuiAccordionItemComponent
     @Output()
     readonly openChange = new EventEmitter<boolean>();
 
+    @ContentChild(TuiAccordionItemEagerContentDirective)
+    readonly eagerContent?: TuiAccordionItemEagerContentDirective;
+
     @ContentChild(TuiAccordionItemContentDirective)
-    readonly content?: TuiAccordionItemContentDirective;
+    readonly lazyContent?: TuiAccordionItemContentDirective;
```
splincode, line 96: breaking changes?

waterplea: No, this is internal, it's only public because it is used in the template.

yurakhomitsky:
@splincode I'm unable to use import directly in the object

image

splincode:
@yurakhomitsky use `TuiDocExample` instead `FrontEndExample` interface

waterplea:
> @splincode I'm unable to use import directly in the object
> 
> image

Yes, you need to change type of that object to `TuiDocExample`. This is a new shorter way of importing code for examples which we recently implemented and most of our examples still use the old one, but eventually we will migrate :)

splincode (approved):

waterplea (approved):

PR #1130, opened 2021-12-20
chore(core,kit): make components dependent on token instead of media 

vladimirpotekhin:
…query

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring
- [ ] Code style update

Closes #

- [ ] Yes
- [x] No

### DIFF
--- a/projects/core/components/root/root.component.ts
+++ b/projects/core/components/root/root.component.ts
@@ -2,15 +2,24 @@ import {
     ChangeDetectionStrategy,
     Component,
     ElementRef,
+    HostBinding,
+    HostListener,
     Inject,
+    NgZone,
     Optional,
 } from '@angular/core';
-import {TUI_DIALOGS, TUI_IS_MOBILE, tuiAssert} from '@taiga-ui/cdk';
+import {WINDOW} from '@ng-web-apis/common';
+import {TUI_DIALOGS, TUI_IS_MOBILE, tuiAssert, tuiZoneOptimized} from '@taiga-ui/cdk';
 import {VERSION} from '@taiga-ui/core/constants';
+import {TuiMedia} from '@taiga-ui/core/interfaces';
 import {TuiNotificationsHostComponent} from '@taiga-ui/core/modules/notifications';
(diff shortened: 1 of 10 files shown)


splincode (approved):

projects/core/components/root/root.component.ts
```diff
@@ -24,6 +33,15 @@ import {map} from 'rxjs/operators';
     },
 })
 export class TuiRootComponent {
+    @HostBinding('$.class._mobile')
```
waterplea, line 36: It will not work this way, because app module does not have NO_ERRORS_SCHEMA on projects. Make all this pipe a private provider, inject elementref and add this property manually (like we do with TUI_MODE) and only use hostlistener + move it to host in Component decorator, otherwise it has to be callable. Again, see how we do it with mode.

projects/core/providers/is-mobile-resolution.provider.ts
```diff
@@ -0,0 +1,29 @@
...
+    ngZone: NgZone,
+): Observable<boolean> {
+    const mobile$ = fromEvent(windowRef, 'resize').pipe(
+        startWith(null),
+        map(() => windowRef.innerWidth < mobile),
+        distinctUntilChanged(),
+        tuiZoneOptimized(ngZone),
+    );
+
+    return (nativeElement['$.class._mobile'] = mobile$);
```
waterplea, line 22: Isn't that too ninja? Maybe let's use up an extra line but write it clearly :)

projects/core/tokens/is-mobile-resolution.ts
```diff
@@ -0,0 +1,5 @@
+import {InjectionToken} from '@angular/core';
+
+export const TUI_IS_MOBILE_RES = new InjectionToken<boolean>(
```
waterplea, line 3: Observable<boolean>

projects/core/providers/is-mobile-resolution.provider.ts
vladimirpotekhin: ok:)

waterplea (approved):

PR #1213, opened 2022-01-19
fix(kit): `InputSlider` fix decimal limit

### LINKED ISSUE #1212: 🐞 - `InputSlider` mask cuts more than 2 decimals

kit

_No response_

### Description

It is currently not possible to enter a value with 3 or more decimals cause the default decimalLimit in the mask is 2

### Angular version

9>

### Taiga UI version

2.30

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

vladimirpotekhin:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #1212 

- [ ] Yes
- [x] No

### DIFF
--- a/projects/kit/abstract/input-slider.ts
+++ b/projects/kit/abstract/input-slider.ts
@@ -145,7 +145,7 @@ export abstract class AbstractTuiInputSlider<T>
     }
 
     private get decimalLimit(): number {
-        return (1 / this.quantum - 1).toString().split('.')[0].length;
+        return this.quantum.toString().split('.')[1]?.length || 2;
     }
 
     protected valueGuard(value: number): number {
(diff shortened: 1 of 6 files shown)


projects/kit/abstract/input-slider.ts
```diff
@@ -143,6 +144,10 @@ export abstract class AbstractTuiInputSlider<T>
         this.updateHovered(hovered);
     }
 
+    private get decimalLimit(): number {
```
splincode, line 147:
```ts
private get decimalLimit(): number {
   const [, fractionPart = ''] = this.quantum.toString().split('.');
   return fractionPart.length || 2;
}
```

`2` - Do we need to move/provide this value to a constant?

splincode (approved):

projects/kit/abstract/input-slider.ts
vladimirpotekhin: @splincode  thanks, done 👍  

@waterplea  getter is only called from TuiMapper depending on quantum. So I think another tuiPure would be redundant. What do you think?

waterplea: Right, it's already memoized. Let's change it to a function instead of getter, that gets `quantum` as an argument to keep mapping function pure.

waterplea (approved):

PR #1222, opened 2022-01-20
feat(kit): `<tui-slider />` => `<input tuiSlider>`

nsbarsukov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring
- [ ] Code style update

Closes # 

- [ ] Yes
- [X] No


projects/demo/src/modules/components/slider-old/examples/1/index.html
```diff
@@ -0,0 +1 @@
+<tui-slider [formControl]="testValue" [max]="10"></tui-slider>
```
nsbarsukov, line 1: unfortunately, git did not properly detect diff :(

All files with old slider's version (their paths includes `*/slider-old/*`) were not touched in this PR.

projects/demo/src/modules/components/slider/examples/2/index.html
```diff
@@ -1,8 +1,3 @@
-<tui-slider
-    [formControl]="testValue"
-    [keySteps]="keySteps"
-    [min]="5000"
-    [max]="1000000"
-    [segments]="3"
-    [steps]="3"
-></tui-slider>
+<input tuiSlider value="65" class="first" />
```
waterplea, line 1: We need examples with ngModel and with formControl too.

projects/demo/src/modules/components/slider/examples/3/index.html
```diff
@@ -0,0 +1,5 @@
+<input tuiSlider value="1" max="4" size="m" [showTicks]="true" />
```
waterplea, line 1: I think ticks should not necessarily be tied to step, one can show ticks ever 2 steps or every 4. Lets keep previous API of showing a number of segments?

projects/kit/components/slider/slider.component.ts
```diff
@@ -1,84 +1,93 @@
...
-    }
+    @Input()
+    @tuiDefaultProp()
+    @HostBinding('class._ticks')
+    showTicks = false;
 
-    get focused(): boolean {
-        return isNativeFocused(this.nativeFocusableElement);
+    @HostBinding('attr.type')
+    readonly type = 'range';
```
waterplea, line 39: Maybe move it to `host` of `@Component` decorator since it's never going to change?

projects/kit/components/slider/slider.component.ts
```diff
@@ -1,84 +1,93 @@
...
+
+    get min(): number {
+        return +this.elementRef.nativeElement.min;
```
waterplea, line 42: Are we ok with this way of converting? Maybe it's more explicit to write `Number(this.elementRef.nativeElement.min)`?

projects/kit/components/slider/slider.component.ts
```diff
@@ -1,84 +1,93 @@
...
 
-    protected processValue(value: number) {
-        this.updateValue(this.valueGuard(value));
+    @HostBinding('class._old-edge')
+    get isOldEdge(): boolean {
+        return isEdgeOlderThan(CHROMIUM_EDGE_START_VERSION, this.userAgent);
     }
+
+    get value(): number {
+        return +this.control?.value || +this.elementRef.nativeElement.value || 0;
```
waterplea, line 73: I think we do not need to convert control value and expect it to be number, like we usually do.

projects/kit/components/slider/slider.style.less
```diff
@@ -1,12 +1,129 @@
...
-.wrapper {
-    border-left: none;
+:host:not(._old-edge) {
+    appearance: none;
+    height: @track-height;
+    padding: 0.4375rem 0; /* To catch click events nearby THIN input's track */
+    background: @track-color;
+    background-clip: content-box; /* To clip filling of paddings */
+
+    &:focus {
```
waterplea, line 70: No need to put it in `:focus`, just leave at `:host`. It's like people often put `cursor: pointer` in `:hover` :D

projects/kit/components/slider/slider.style.less
```diff
@@ -1,12 +1,129 @@
...
+    border: none;
+    border-radius: 50%;
+    height: @thumb-heights[ @@size];
+    width: @thumb-heights[ @@size];
+
+    :not(:disabled)& {
+        cursor: ew-resize;
+    }
+
+    :focus& {
```
waterplea, line 41: Let's use `:focus-visible`. It recently shipped in Safari as well and it can be considered graceful degradation if it doesn't work.

projects/demo/src/modules/app/app.routes.ts
```diff
@@ -777,6 +777,16 @@ export const ROUTES = [
             title: 'Slider',
         },
     },
+    {
+        path: 'components/slider-old',
```
splincode, line 781: @waterplea @nsbarsukov What do you think about path as `slider[legacy]` or `slider[deprecated]`instead of `slider-old`

projects/demo/src/modules/components/slider-old/examples/import/declare-form.txt
```diff
@@ -0,0 +1,12 @@
+import {FormControl, FormGroup} from '@angular/forms';
```
splincode, line 1: what do you think about using markdown? as it is in the example with the avatar demo

projects/demo/src/modules/components/slider-old/examples/import/import-module.txt
```diff
@@ -0,0 +1,13 @@
+import {FormsModule, ReactiveFormsModule} from '@angular/forms';
```
splincode, line 1: what do you think about using markdown? as it is in the example with the avatar demo

projects/demo/src/modules/components/slider-old/examples/import/insert-template.txt
```diff
@@ -0,0 +1,6 @@
+<tui-slider [formControl]="control"
```
splincode, line 1: what do you think about using markdown? as it is in the example with the avatar demo

projects/demo/src/modules/components/slider/examples/import/declare-form.txt
```diff
@@ -1,12 +1,10 @@
-import {FormControl, FormGroup} from '@angular/forms';
+import {FormControl} from '@angular/forms';
```
splincode, line 1: what do you think about using markdown? as it is in the example with the avatar demo

projects/demo/src/modules/components/slider/examples/import/insert-template.txt
```diff
@@ -1,6 +1 @@
-<tui-slider [formControl]="control"
-            max="10"></tui-slider>
-
-<tui-slider [formControl]="testValue"
-            [max]="10"></tui-slider>
-
+<input tuiSlider [formControl]="control"/>
```
splincode, line 1: what do you think about using markdown? as it is in the example with the avatar demo

projects/demo/src/modules/components/slider-old/slider-old.component.ts
```diff
@@ -0,0 +1,85 @@
...
+
+import {FrontEndExample} from '../../interfaces/front-end-example';
+
+@Component({
+    selector: 'example-slider',
+    templateUrl: './slider-old.template.html',
+    changeDetection,
+})
+export class ExampleTuiSliderComponent {
+    readonly exampleImportModule = exampleImportModule;
```
splincode, line 23: can we replace all instructions with `imports(...)` in this file?

projects/demo/src/modules/components/slider-old/examples/import/declare-form.txt
nsbarsukov: it is `slider-old`, let leave legacy files as it is (they will be deleted in next major version).

But i will fix all new files, thanks

splincode: thank you)

projects/demo/src/modules/app/app.routes.ts
waterplea: I don't really care because it's going away in 3.0 anyway

projects/demo/src/modules/components/slider/examples/2/index.html
nsbarsukov: fixed

Now we have examples with ngModel, with FormControl and with native attributes

projects/demo/src/modules/components/slider/examples/3/index.html
nsbarsukov: I made input-property `[segments]` (as it was in previous API)

projects/kit/components/slider/slider.component.ts
nsbarsukov: fixed

projects/kit/components/slider/slider.component.ts
nsbarsukov: fixed

projects/kit/components/slider/slider.style.less
nsbarsukov: fixed

projects/kit/components/slider/slider.style.less
nsbarsukov: done

projects/demo/src/modules/app/app.routes.ts
nsbarsukov: I have changed route path to `slider[deprecated]`

projects/demo/src/modules/app/app.routes.ts
```diff
@@ -797,6 +797,16 @@ export const ROUTES = [
             title: 'Slider',
         },
     },
+    {
+        path: 'components/slider-deprecated',
+        loadChildren: () =>
+            import(`../components/slider-old/slider-old.module`).then(
```
splincode, line 803: what do you think about rename?
`slider-old` -> `slider-deprecated`

projects/demo/src/modules/components/slider-old/slider-old.component.ts
```diff
@@ -0,0 +1,85 @@
...
+import {TuiKeySteps} from '@taiga-ui/kit';
+
+import {default as example1Html} from '!!raw-loader!./examples/1/index.html';
+import {default as example1Ts} from '!!raw-loader!./examples/1/index.ts';
+import {default as example2Html} from '!!raw-loader!./examples/2/index.html';
+import {default as example2Ts} from '!!raw-loader!./examples/2/index.ts';
+import {default as exampleForm} from '!!raw-loader!./examples/import/declare-form.txt';
+import {default as exampleModule} from '!!raw-loader!./examples/import/import-module.txt';
+import {default as exampleHtml} from '!!raw-loader!./examples/import/insert-template.txt';
+
```
splincode, line 14: use imports directly

projects/kit/components/slider/slider.component.ts
```diff
@@ -1,84 +1,93 @@
...
+    segments = 1;
+
+    get min(): number {
+        return Number(this.elementRef.nativeElement.min);
     }
 
-    get nativeFocusableElement(): TuiNativeFocusableElement | null {
-        return this.dotRight ? this.dotRight.nativeElement : null;
+    get max(): number {
+        return Number(this.elementRef.nativeElement.max) || 100;
```
splincode, line 51: can we move `100` to options token?

projects/kit/components/slider/slider.component.ts
```diff
@@ -1,84 +1,93 @@
...
     }
 
-    get focused(): boolean {
-        return isNativeFocused(this.nativeFocusableElement);
+    get step(): number {
+        return Number(this.elementRef.nativeElement.step) || 1;
```
splincode, line 55: can we move `100` to options token?

splincode (approved):

projects/kit/components/slider/slider.component.ts
nsbarsukov: these default values are the same as in [native \<input type="range"/>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/range#validation).

> The value won't be less than [min](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-min). **The default is 0.**
The value won't be greater than [max](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-max). **The default is 100.**
The value will be a multiple of [step](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-step). **The default is 1.**

To be honest, this `|| 100` never be invoked because browser will set above mentioned values (if no values was set by user).

nsbarsukov: Thus, `|| 100` is just for the case if something goes wrong)

nsbarsukov:
Partially resolves https://github.com/Tinkoff/taiga-ui/issues/256

PR #1228, opened 2022-01-20
refactor: unify determining if control is interactive

### LINKED ISSUE #1183: 🚀 - Add `interactive` getter to `AbstractTuiInteractive`

cdk

### Description

Many of our components rely in their logic on combination of not being `readOnly` and not being `disabled`. Let's add this:

```ts
get interactive(): boolean {
  return !this.computedDisabled && !this.readOnly;
}
```

to `AbstractTuiInteractive` and refactor everything to use that getter.

**Note:** We need to keep in mind it might be written in different order and sometimes it can be `disabled` and sometimes it can be `computedDisabled`. The best approach would be to search whole project for `this.readOnly` and replace code with that new getter wherever possible.

GeorgeSokolovsky:
Refs: #1183

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring
- [ ] Code style update

Closes #1183  

## What is the new behavior?

The same, but with less code 😄 

The only thing I did differently from what was proposed at #1183  is I added a new getter to the `AbstractTuiControl` class instead of  `AbstractTuiInteractive`. The reason is that not all inheritors of `AbstractTuiInteractive` should have `readOnly` prop (for example, `TuiActionComponent` or `TuiAccordionItemComponent`). On the other hand, any control has `disabled` and `readOnly` states, and it makes `AbstractTuiControl` a perfect candidate for keeping this logic.

- [ ] Yes
- [x] No


splincode (approved):

projects/cdk/abstract/control.ts
```diff
@@ -98,6 +97,10 @@ export abstract class AbstractTuiControl<T>
         return this.safeNgControlData<boolean>(({disabled}) => disabled, false);
     }
 
+    get interactive(): boolean {
+        return !this.readOnly && !this.disabled;
```
waterplea, line 101: Let's use `computedDisabled`, components have ability to override it like radio buttons do.

projects/kit/components/combo-box/combo-box.component.ts
```diff
@@ -153,7 +153,7 @@ export class TuiComboBoxComponent<T>
     }
 
     get canOpen(): boolean {
```
waterplea, line 155: Since `interactive` is already a public getter we can just use it wherever we used `canOpen`.

projects/kit/components/input-date-time/input-date-time.component.ts
```diff
@@ -221,7 +221,7 @@ export class TuiInputDateTimeComponent
     }
 
     get canOpen(): boolean {
```
waterplea, line 223: Same here

projects/kit/components/input-file/input-file.component.ts
```diff
@@ -141,7 +141,7 @@ export class TuiInputFileComponent
     }
 
     get allowDelete(): boolean {
```
waterplea, line 143: Same here

projects/kit/components/input-month-range/input-month-range.component.ts
```diff
@@ -113,7 +113,7 @@ export class TuiInputMonthRangeComponent
     }
 
     get canOpen(): boolean {
```
waterplea, line 115: And here :)

projects/kit/components/input-month/input-month.component.ts
```diff
@@ -92,7 +92,7 @@ export class TuiInputMonthComponent
     }
 
     get canOpen(): boolean {
```
waterplea, line 94: And here :)

projects/kit/components/input-password/input-password.component.ts
```diff
@@ -117,7 +117,7 @@ export class TuiInputPasswordComponent
     }
 
     get hasEyeIcon(): boolean {
```
waterplea, line 119: And here :)

projects/kit/components/select/select.component.ts
```diff
@@ -120,7 +120,7 @@ export class TuiSelectComponent<T>
     }
 
     get canOpen(): boolean {
```
waterplea, line 122: And here :)

waterplea (commented):
Thank you! Left a few comments

waterplea (approved):

waterplea:
Thank you!

PR #1249, opened 2022-01-24
feat(core): default options configuration for primitive-textfield

mullinsmikey:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

Issue #133 is not fully implemented.

## What is the new behavior?

Partial implementation of #133 for primitive-textfield component.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/core/components/primitive-textfield/primitive-textfield-options.ts
+++ b/projects/core/components/primitive-textfield/primitive-textfield-options.ts
@@ -5,7 +5,6 @@ import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
 export interface PrimitiveTextfieldOptions {
     readonly iconAlign: TuiHorizontalDirection;
     readonly iconCleaner: PolymorpheusContent;
-    readonly postfix: string;
 }
 
 // TODO: remove in ivy compilation
@@ -15,7 +14,6 @@ export const PRIMITIVE_TEXTFIELD_ICON_CLEANER = 'tuiIconCloseLarge';
 export const TUI_PRIMITIVE_TEXTFIELD_DEFAULT_OPTIONS: PrimitiveTextfieldOptions = {
     iconAlign: 'right',
     iconCleaner: PRIMITIVE_TEXTFIELD_ICON_CLEANER,
-    postfix: '',
 };
 
 export const TUI_PRIMITIVE_TEXTFIELD_OPTIONS =
--- a/projects/core/components/primitive-textfield/primitive-textfield-options.ts
+++ b/projects/core/components/primitive-textfield/primitive-textfield-options.ts
@@ -1,0 +1,27 @@
+import {InjectionToken} from '@angular/core';
+import {TuiHorizontalDirection} from '@taiga-ui/core/types';
+import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
+
+export interface PrimitiveTextfieldOptions {
+    readonly iconAlign: TuiHorizontalDirection;
+    readonly iconCleaner: PolymorpheusContent;
+    readonly postfix: string;
+}
+
+// TODO: remove in ivy compilation
+export const PRIMITIVE_TEXTFIELD_ICON_CLEANER = 'tuiIconCloseLarge';
+
+/** Default values for primitive textfield options */
+export const TUI_PRIMITIVE_TEXTFIELD_DEFAULT_OPTIONS: PrimitiveTextfieldOptions = {
+    iconAlign: 'right',
+    iconCleaner: PRIMITIVE_TEXTFIELD_ICON_CLEANER,
+    postfix: '',
+};
+
+export const TUI_PRIMITIVE_TEXTFIELD_OPTIONS =
+    new InjectionToken<PrimitiveTextfieldOptions>(
+        'Default parameters for primitive textfield component',
+        {
+            factory: () => TUI_PRIMITIVE_TEXTFIELD_DEFAULT_OPTIONS,
+        },
+    );
--- a/projects/core/components/primitive-textfield/primitive-textfield-options.ts
+++ b/projects/core/components/primitive-textfield/primitive-textfield-options.ts
@@ -2,22 +2,22 @@ import {InjectionToken} from '@angular/core';
 import {TuiHorizontalDirection} from '@taiga-ui/core/types';
 import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
 
-export interface PrimitiveTextfieldOptions {
+export interface TuiPrimitiveTextfieldOptions {
     readonly iconAlign: TuiHorizontalDirection;
     readonly iconCleaner: PolymorpheusContent;
 }
 
 // TODO: remove in ivy compilation
-export const PRIMITIVE_TEXTFIELD_ICON_CLEANER = 'tuiIconCloseLarge';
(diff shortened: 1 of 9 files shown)


projects/core/components/primitive-textfield/primitive-textfield-options.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+// TODO: remove in ivy compilation
+export const PRIMITIVE_TEXTFIELD_ICON_CLEANER = 'tuiIconCloseLarge';
+
+/** Default values for primitive textfield options */
+export const TUI_PRIMITIVE_TEXTFIELD_DEFAULT_OPTIONS: PrimitiveTextfieldOptions = {
+    iconAlign: 'right',
+    iconCleaner: PRIMITIVE_TEXTFIELD_ICON_CLEANER,
+    postfix: '',
```
waterplea, line 18: Not sure `postfix` makes sense in DI config. It is something that most likely changes from control to control and being able to set something else as default, other than an empty string seems excessive to me. Do you have a use case?

projects/demo/src/modules/components/primitive-textfield/primitive-textfield.component.ts
```diff
@@ -51,6 +53,11 @@ export class ExampleTuiPrimitiveTextfieldComponent extends AbstractExampleTuiInt
         LESS: example1Less,
     };
 
+    readonly example2: FrontEndExample = {
+        TypeScript: example2Ts,
```
splincode, line 57: use `import` directly, please

projects/demo/src/modules/components/primitive-textfield/examples/2/index.ts
```diff
@@ -0,0 +1,66 @@
...
+} from '@taiga-ui/core';
+
+@Component({
+    selector: 'tui-primitive-textfield-example-2',
+    templateUrl: './index.html',
+    changeDetection,
+    encapsulation,
+    providers: [
+        {
+            provide: TUI_PRIMITIVE_TEXTFIELD_OPTIONS, // <-- You are looking for this token
```
splincode, line 26: nit: redundant comment

projects/demo/src/modules/components/primitive-textfield/examples/2/index.ts
```diff
@@ -0,0 +1,66 @@
...
+            },
+        },
+    ],
+})
+export class TuiPrimitiveTextfieldExample2 extends AbstractTuiControl<string> {
+    @ViewChild(TuiPrimitiveTextfieldComponent)
+    private readonly textfield?: TuiPrimitiveTextfieldComponent;
+
+    constructor(
+        @Optional() @Self() @Inject(NgControl) control: NgControl | null,
```
splincode, line 39: nit: need use `prettier` here

projects/demo/src/modules/components/primitive-textfield/examples/2/index.ts
mullinsmikey: I definitely agree with your point but for the demo purposes everything else in this file is just to get the component up&running and I thought it'd be nice to point out the exact piece of code related to this demo. Anyway, if you'd like me to remove this, I'll do that by this evening

projects/core/components/primitive-textfield/primitive-textfield-options.ts
mullinsmikey: Not really and I guess you're right about that. Will be fixed up today

projects/demo/src/modules/components/primitive-textfield/examples/2/index.ts
mullinsmikey: I also noticed that it didn't take care of everything in these commits as it usually would but thought that it's due to configuration change. I'll look into it

projects/core/components/primitive-textfield/primitive-textfield.template.html
```diff
@@ -87,13 +87,18 @@
                 <tui-svg class="custom-icon" [src]="iconSrc"></tui-svg>
             </ng-template>
         </div>
-        <tui-svg
+        <span
```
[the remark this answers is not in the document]
waterplea: I personally prefer `span` when flex or other styles are applied because `div` is invalid content for some containers whereas `span` can be put almost everywhere and layout would be valid.

projects/demo/src/modules/components/primitive-textfield/examples/2/index.ts
splincode: it would be great if you find a bug in our prettier configuration

projects/demo/src/modules/components/primitive-textfield/examples/2/index.ts
waterplea: I think we can keep it, there's a lot of code in this example and it might be helpful to point to the important part.

projects/demo/src/modules/components/primitive-textfield/primitive-textfield.component.ts
waterplea: @mullinsmikey Look for an example I posted in your other PR 😉

projects/demo/src/modules/components/primitive-textfield/examples/2/index.ts
mullinsmikey: @splincode, I fixed my local husky config and that fixed a lot of issues with the pre-commit hook but I still can't figure out why this particular line of code is not being taken care of, except that it's only 73 chars long

waterplea: Perhaps it's correct from prettier point of view? :)

splincode: oh)) I see, this is the correct line for prettier

image

please, add manually break line for decorators))

splincode: @mullinsmikey also, please, tell me, what did you have to do with the husky to make it work?)

waterplea: The point of prettier is to never add manual breaks and accept prettier as objective reality and a higher power :D

splincode: prettier does breaks only if the line exceeds the maximum length of 80 characters for typescript code )) it does not analyze what kind of construct is used in the code, decorators or something else.

mullinsmikey: @splincode, sure.
When you use nvm sometimes npx is not present in your path. Usually this is not a problem, but GUI git clients may produce an error like "npx command not found". Here's an [official way](https://typicode.github.io/husky/#/?id=command-not-found) to fix this on your machine, but there's a [quick&dirty way](https://github.com/typicode/husky/issues/912#issuecomment-817522060) to persist this behavior in the repo.

splincode: @mullinsmikey could you add please `quick@dirty` or `official` way to fix it in another PR?

splincode (approved):

waterplea (approved):

PR #1250, opened 2022-01-24
feat(core): default options configuration for tooltip

mullinsmikey:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

Issue #133 is not fully implemented.

## What is the new behavior?

Partial implementation of #133 for tooltip component.

- [ ] Yes
- [x] No


projects/core/components/tooltip/tooltip-options.ts
```diff
@@ -0,0 +1,30 @@
+import {InjectionToken} from '@angular/core';
+import {TuiDirection, TuiHintModeT} from '@taiga-ui/core/types';
+import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
+
+export interface TooltipOptions {
```
waterplea, line 5: Everything, except for the icon is actually more relevant to `HintDirective`. How about we move it there and keep the icon there as well? I know icon is irrelevant for the directive, but I think it better to have a single source of default values for both `Tooltip` and `Hint` and making a separate token just for the icon might be an overkill.

projects/demo/src/modules/components/tooltip/tooltip.component.ts
```diff
@@ -40,6 +42,11 @@ export class ExampleTuiTooltipComponent {
         LESS: example3Less,
     };
 
+    readonly example4: FrontEndExample = {
```
waterplea, line 45: We now support promises in examples, so please rewrite it like so:
```ts
readonly example4 = {
    TypeScript: import('!!raw-loader!./examples/4/index.ts'),
    HTML: import('!!raw-loader!./examples/4/index.html'),
};
```
Rewriting the rest of the examples in this file and in your other relevant PRs would be super helpful too! 👍

waterplea (changes requested):
Hint in textfield actually follows `TuiHintControllerDirective` so you would need to reuse these defaults there as well. And if directive is missing, default value comes from `TUI_HINT_CONTROLLER` so you would need to add it there as well.

projects/core/components/tooltip/tooltip.template.html
```diff
@@ -1,11 +1,16 @@
-<tui-svg
-    src="tuiIconTooltipLarge"
+<span
```
splincode, line 1: @waterplea what do you think, `span` is better than `div` in terms of html semantics?

waterplea: Yes, like I said in another PR, if we use `div` here just putting `tui-tooltip` in a `p` tag will make it invalid:
image

Not that we follow it everywhere but still it's better to keep that in mind.

splincode: sure, thank you for explain ))

projects/core/components/tooltip/tooltip-options.ts
mullinsmikey: I'm not sure this is the best solution that I've come up with. Check the last commit please, I'm sure there is a better way.

projects/core/directives/hint/hint-options.ts
```diff
@@ -0,0 +1,30 @@
+import {InjectionToken} from '@angular/core';
+import {
+    AbstractHintOptions,
+    TUI_ABSTRACT_HINT_DEFAULT_OPTIONS,
+} from '@taiga-ui/core/abstract';
+import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
+
+export interface HintOptions extends AbstractHintOptions {
```
splincode, line 8: `TuiHintOptions`

projects/core/directives/manual-hint/manual-hint-options.ts
```diff
@@ -0,0 +1,19 @@
+import {InjectionToken} from '@angular/core';
+import {
+    AbstractHintOptions,
+    TUI_ABSTRACT_HINT_DEFAULT_OPTIONS,
+} from '@taiga-ui/core/abstract';
+
+export type ManualHintOptions = AbstractHintOptions;
```
splincode, line 7: `TuiManualHintOptions`

projects/core/directives/pointer-hint/pointer-hint-options.ts
```diff
@@ -0,0 +1,24 @@
+import {InjectionToken} from '@angular/core';
+import {
+    AbstractHintOptions,
+    TUI_ABSTRACT_HINT_DEFAULT_OPTIONS,
+} from '@taiga-ui/core/abstract';
+
+export interface PointerHintOptions extends AbstractHintOptions {
```
splincode, line 7: `TuiPointerHintOptions`

projects/core/directives/hint/hint-options.ts
```diff
@@ -0,0 +1,30 @@
...
+    readonly tuiHintShowDelay: number;
+    readonly tuiHintHideDelay: number;
+    readonly tooltipIcon: PolymorpheusContent;
+}
+
+// TODO: remove in ivy compilation
+export const TOOLTIP_ICON = 'tuiIconTooltipLarge';
```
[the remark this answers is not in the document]
mullinsmikey: I used #675 as a reference for this naming and honestly didn't notice that something's off. The fix will be here in a moment.

splincode (approved):

waterplea (commented):
I'm not sure we need those extra tokens, maybe let's use one for all hint directives for now? @splincode @vladimirpotekhin what do you think? @mullinsmikey also don't forget about `TuiHintControllerDirective`

splincode:
@waterplea @vladimirpotekhin I thought it was a very good customization way for tooltip, but if you think it's redundant, then okay

mullinsmikey:
@waterplea, as I can see the way to reduce the number of excessive tokens is to move all the properties (including tooltipIcon) to AbstractHint which I personally see as an overkill.

waterplea:
Ok, let's keep it as is. Let's just drop the new token for controller and use `TUI_HINT_OPTIONS` there.

mullinsmikey:
That does make sense. The latest commit fixes that

waterplea (approved):

Issue #1229, opened 2022-01-21
🚀 - Support 12h am/pm input in InputTime


pwespi:
### Which @taiga-ui/* package(s) are relevant/releated to the feature request?

kit

### Description

Support 12h am/pm input in InputTime.

Some ideas: 
* minimal solution is probably to allow setting max to 12:00, then one can manually add an am/pm toggle or radio input
* nicer solution would be to allow setting the mode to 12h, and then am/pm toggle is shown and handled under the hood
  * instead of a toggle or radio input for am/pm, it could also be part of the masked input, so one would type e.g. `09:00 am`

Related to https://github.com/Tinkoff/taiga-ui/issues/70, https://github.com/Tinkoff/taiga-ui/issues/270.

Issue #270, opened 2021-03-06
🚀 - `InputTime` support higher input times than 23:59


0xphilipp:
# 🚀 Feature request

Support Input Time with higher value than 23:59. It would be great if the max value could be specified. Then the control can also be used to select a time range like in X hours, Y minutes (50 hours, 30 minutes)
Or even a different control like TuiInputTimeRange.

nsbarsukov:
Our internal teams also requires `min`/`max` input props for this input

waterplea:
The linked issue https://github.com/Tinkoff/taiga-ui/issues/1229 is probably going to be fixed under this ticket, but I'll keep it open and separate just in case.

waterplea:
I believe the issue is fixed now, however I'll keep this ticket open for 3 reasons:

1. No mention of `min`/`max` on the API page
2. Arrow navigation is still limited to 24 hours and 60 minutes/seconds
3. It needs to be revisited with new mask

shatanov:
is this issues relevant? Is it possible to try to take his?

PR #1256, opened 2022-01-25
feat(kit): configurable maxValues for input-time

mullinsmikey:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

Issue #270 is not fully implemented.
Issue #1229 is not fully implemented.

## What is the new behavior?

Partial implementation of #270.
Workaround implementation of #1229.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/demo/src/modules/components/input-time/examples/3/index.ts
+++ b/projects/demo/src/modules/components/input-time/examples/3/index.ts
@@ -2,7 +2,11 @@ import {Component} from '@angular/core';
 import {FormControl, FormGroup} from '@angular/forms';
 import {changeDetection} from '@demo/emulate/change-detection';
 import {encapsulation} from '@demo/emulate/encapsulation';
-import {TUI_INPUT_TIME_OPTIONS, tuiCreateTimePeriods} from '@taiga-ui/kit';
+import {
+    TUI_INPUT_TIME_DEFAULT_OPTIONS,
+    TUI_INPUT_TIME_OPTIONS,
+    tuiCreateTimePeriods,
+} from '@taiga-ui/kit';
 
 @Component({
     selector: 'tui-input-time-example-3',
@@ -15,6 +19,7 @@ import {TUI_INPUT_TIME_OPTIONS, tuiCreateTimePeriods} from '@taiga-ui/kit';
             useValue: {
                 icon: 'tuiIconCheckCircleLarge',
                 mode: 'HH:MM:SS',
+                maxValues: {...TUI_INPUT_TIME_DEFAULT_OPTIONS.maxValues, SS: 165},
                 itemSize: 's',
             },
         },
--- a/projects/demo/src/modules/components/input-time/examples/5/index.ts
+++ b/projects/demo/src/modules/components/input-time/examples/5/index.ts
@@ -2,14 +2,10 @@ import {Component} from '@angular/core';
 import {FormControl, FormGroup} from '@angular/forms';
 import {changeDetection} from '@demo/emulate/change-detection';
 import {encapsulation} from '@demo/emulate/encapsulation';
-import {
-    TUI_INPUT_TIME_DEFAULT_OPTIONS,
-    TUI_INPUT_TIME_OPTIONS,
-    tuiCreateTimePeriods,
-} from '@taiga-ui/kit';
+import {TUI_INPUT_TIME_DEFAULT_OPTIONS, TUI_INPUT_TIME_OPTIONS} from '@taiga-ui/kit';
 
 @Component({
-    selector: 'tui-input-time-example-3',
+    selector: 'tui-input-time-example-5',
     templateUrl: './index.html',
     changeDetection,
     encapsulation,
@@ -18,17 +14,19 @@ import {
             provide: TUI_INPUT_TIME_OPTIONS,
             useValue: {
                 ...TUI_INPUT_TIME_DEFAULT_OPTIONS,
-                icon: 'tuiIconCheckCircleLarge',
-                mode: 'HH:MM:SS',
-                itemSize: 's',
+                mode: 'HH:MM',
+                maxValues: {HH: 11, MM: 59},
             },
         },
     ],
 })
-export class TuiInputTimeExample3 {
+export class TuiInputTimeExample5 {
     readonly testForm = new FormGroup({
         testValue: new FormControl(null),
+        isPm: new FormControl(false),
     });
 
-    items1 = tuiCreateTimePeriods();
+    get postfix(): string {
+        return this.testForm.value?.isPm ? 'PM' : 'AM';
+    }
 }
--- a/projects/demo/src/modules/components/input-time/examples/5/index.ts
+++ b/projects/demo/src/modules/components/input-time/examples/5/index.ts
@@ -16,6 +16,7 @@ import {TUI_INPUT_TIME_DEFAULT_OPTIONS, TUI_INPUT_TIME_OPTIONS} from '@taiga-ui/
                 ...TUI_INPUT_TIME_DEFAULT_OPTIONS,
                 mode: 'HH:MM',
                 maxValues: {HH: 11, MM: 59},
+                strict12h: true,
             },
         },
     ],
--- a/projects/demo/src/modules/components/input-time/examples/5/index.ts
+++ b/projects/demo/src/modules/components/input-time/examples/5/index.ts
@@ -16,7 +16,6 @@ import {TUI_INPUT_TIME_DEFAULT_OPTIONS, TUI_INPUT_TIME_OPTIONS} from '@taiga-ui/
                 ...TUI_INPUT_TIME_DEFAULT_OPTIONS,
                 mode: 'HH:MM',
                 maxValues: {HH: 11, MM: 59},
-                strict12h: true,
             },
         },
     ],
--- a/projects/kit/components/input-time/input-time-options.ts
+++ b/projects/kit/components/input-time/input-time-options.ts
@@ -10,7 +10,6 @@ export interface InputTimeOptions {
     readonly mode: TuiTimeMode;
     readonly postfix: string;
     readonly maxValues: Record<TuiTimeFormatParts, number>;
-    readonly strict12h: boolean;
     readonly itemSize: TuiSizeS | TuiSizeL;
 }
 
@@ -23,7 +22,6 @@ export const TUI_INPUT_TIME_DEFAULT_OPTIONS: InputTimeOptions = {
     mode: 'HH:MM',
     postfix: '',
     maxValues: MAX_TIME_VALUES,
-    strict12h: false,
     itemSize: 'm',
 };
 
--- a/projects/kit/components/input-time/input-time-options.ts
+++ b/projects/kit/components/input-time/input-time-options.ts
@@ -8,6 +8,7 @@ import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
 export interface InputTimeOptions {
     readonly icon: PolymorpheusContent<TuiContextWithImplicit<TuiSizeS | TuiSizeL>>;
     readonly mode: TuiTimeMode;
+    readonly postfix: string;
     readonly maxValues: Record<TuiTimeFormatParts, number>;
     readonly itemSize: TuiSizeS | TuiSizeL;
 }
@@ -19,6 +20,7 @@ export const INPUT_TIME_ICON = ({$implicit}: any) =>
 export const TUI_INPUT_TIME_DEFAULT_OPTIONS: InputTimeOptions = {
     icon: INPUT_TIME_ICON,
     mode: 'HH:MM',
+    postfix: '',
     maxValues: MAX_TIME_VALUES,
     itemSize: 'm',
 };
--- a/projects/kit/components/input-time/input-time-options.ts
+++ b/projects/kit/components/input-time/input-time-options.ts
@@ -1,6 +1,7 @@
 import {InjectionToken} from '@angular/core';
 import {TuiContextWithImplicit, TuiTimeMode} from '@taiga-ui/cdk';
 import {TuiSizeL, TuiSizeS} from '@taiga-ui/core';
+import {MAX_TIME_VALUES} from '@taiga-ui/kit/constants';
 import {TuiTimeFormatParts} from '@taiga-ui/kit/types';
 import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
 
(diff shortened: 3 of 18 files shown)


projects/kit/utils/mask/create-time-mask.ts
```diff
@@ -1,8 +1,26 @@
...
+        ...(mode.includes('HH:MM:SS')
+            ? [':', ...Array(maxValues.SS.toString().length).fill(TUI_DIGIT_REGEXP)]
+            : []),
+        ...(mode === 'HH:MM:SS.MSS'
+            ? ['.', ...Array(maxValues.MS.toString().length).fill(TUI_DIGIT_REGEXP)]
+            : []),
+    ];
+}
+
+export function tuiCreateTimeMaskBak(mode: TuiTimeMode): TuiTextMaskList {
```
splincode, line 23: `tuiCreateTimeMaskBak` is it necessary??

projects/kit/components/input-time/input-time-options.ts
```diff
@@ -1,11 +1,13 @@
 import {InjectionToken} from '@angular/core';
 import {TuiContextWithImplicit, TuiTimeMode} from '@taiga-ui/cdk';
 import {TuiSizeL, TuiSizeS} from '@taiga-ui/core';
+import {TuiTimeFormatParts} from '@taiga-ui/kit/types';
 import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
 
 export interface InputTimeOptions {
     readonly icon: PolymorpheusContent<TuiContextWithImplicit<TuiSizeS | TuiSizeL>>;
     readonly mode: TuiTimeMode;
+    readonly maxValues: Record<TuiTimeFormatParts, number>;
```
splincode, line 10: `Record<TuiTimeFormatParts, number>` can we use more concrete the type?

mullinsmikey: Could you be more specific please? AFAIR, `Record` enforces us to specify all keys

splincode: sorry, I see that now

projects/kit/utils/mask/create-auto-corrected-time-pipe.ts
```diff
@@ -1,23 +1,24 @@
...
 import {TuiTextMaskPipeHandler} from '@taiga-ui/core';
+import {TuiTimeFormatParts} from '@taiga-ui/kit/types';
 
 /**
  * Adjusts the entered time by omitting only suitable values for hours and minutes
  * @returns time as a string
  */
 export function tuiCreateAutoCorrectedTimePipe(
     timeMode: TuiTimeMode = 'HH:MM',
+    maxValues: Record<TuiTimeFormatParts, number> = {HH: 23, MM: 59, SS: 59, MS: 999},
```
splincode, line 11: `{HH: 23, MM: 59, SS: 59, MS: 999}` maybe move to special constant for reuse?

mullinsmikey: Also thought about it. Thanks for supporting this idea

projects/kit/utils/mask/create-time-mask.ts
```diff
@@ -1,8 +1,26 @@
...
+        ':',
+        ...Array(maxValues.MM.toString().length).fill(TUI_DIGIT_REGEXP),
+        ...(mode.includes('HH:MM:SS')
+            ? [':', ...Array(maxValues.SS.toString().length).fill(TUI_DIGIT_REGEXP)]
+            : []),
+        ...(mode === 'HH:MM:SS.MSS'
+            ? ['.', ...Array(maxValues.MS.toString().length).fill(TUI_DIGIT_REGEXP)]
+            : []),
+    ];
+}
```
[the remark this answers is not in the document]
mullinsmikey: I see that this function looks clumsy. Do you have any ideas on how could it be split into separate functions?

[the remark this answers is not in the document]
splincode:
```ts
export function getMaskPositionsBy(formatPartValue: number): RegExp[]  {
 return Array.from({ length: formatPartValue.toString().length }).fill(TUI_DIGIT_REGEXP);
}
```

`getMaskPositionsBy` - not sure about that is correct name

```ts
export function createTimeMask(mode, maxValues): TuiTextMaskList {
  const HH = getMaskPositionsBy(maxValues.HH);
  const MM =  getMaskPositionsBy(maxValues.MM);
  const SS_OR_MS = [];
  
  if (mode.includes('HH:MM:SS')) {
     SS_OR_MS = [':', ...getMaskPositionsBy(maxValues.SS)];
  } else if (mode === 'HH:MM:SS.MSS') {
     SS_OR_MS = ['.', ...getMaskPositionsBy(maxValues.MS)];
  }

  return [...HH, ...MM, ...SS_OR_MS];
}
```


I see it like this) maybe you have better ideas

projects/kit/utils/mask/create-auto-corrected-time-pipe.ts
```diff
@@ -1,23 +1,25 @@
...
+    const timeFormatArray: TuiTimeFormatParts[] = ['HH', 'MM', 'SS', 'MS'];
 
     return (conformedValue: string) => {
         const indexesOfPipedChars: number[] = [];
         const conformedValueArr = conformedValue.split('');
 
         timeFormatArray.forEach(format => {
             const position = timeMode.indexOf(format);
-            const maxFirstDigit = parseInt(maxValue[format].toString().substr(0, 1), 10);
+            const maxFirstDigit = parseInt(maxValues[format].toString().substr(0, 1), 10);
```
splincode, line 22: `maxValues[format].`

can we change to `maxValues[format].?` for exclude null pointer in the future and create unit test where maxValues pass as `{}`

projects/kit/utils/mask/create-time-mask.ts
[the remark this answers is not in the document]
mullinsmikey: It's not seconds _or_ milliseconds but I still get your idea. Give me some time to prepare a commit

projects/kit/utils/mask/create-auto-corrected-time-pipe.ts
mullinsmikey: I'll look into that

projects/demo/src/modules/components/input-time/examples/3/index.ts
```diff
@@ -15,6 +19,7 @@ import {TUI_INPUT_TIME_OPTIONS, tuiCreateTimePeriods} from '@taiga-ui/kit';
             useValue: {
                 icon: 'tuiIconCheckCircleLarge',
                 mode: 'HH:MM:SS',
+                maxValues: {...TUI_INPUT_TIME_DEFAULT_OPTIONS.maxValues, SS: 165},
```
splincode, line 22: Could you create a different `example-4` instead changes here?

and add example for current case https://github.com/Tinkoff/taiga-ui/issues/1229

projects/kit/utils/mask/create-time-mask.ts
```diff
@@ -3,19 +3,34 @@ import {TUI_DIGIT_REGEXP, TuiTextMaskList} from '@taiga-ui/core';
...
+    prefix?: string,
+): RegExp[] {
+    const partLength =
+        typeof maxPartValue === 'number' ? maxPartValue.toString().length : 0;
+
+    if (!partLength) {
+        return [];
+    }
+
+    const regExp = Array(partLength).fill(TUI_DIGIT_REGEXP);
```
splincode, line 17: please, use `Array.from({ length: partLength })` instead of `Array(partLength)` 

for example - https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-new-array.md

projects/kit/utils/mask/create-time-mask.ts
```diff
@@ -3,19 +3,34 @@ import {TUI_DIGIT_REGEXP, TuiTextMaskList} from '@taiga-ui/core';
 import {MAX_TIME_VALUES} from '@taiga-ui/kit/constants';
 import {TuiTimeFormatParts} from '@taiga-ui/kit/types';
 
+function tuiCreateTimePartMask(
```
splincode, line 6: please, create unit test for this function

projects/kit/utils/mask/create-auto-corrected-time-pipe.ts
```diff
@@ -1,23 +1,28 @@
...
 
     return (conformedValue: string) => {
         const indexesOfPipedChars: number[] = [];
         const conformedValueArr = conformedValue.split('');
 
         timeFormatArray.forEach(format => {
             const position = timeMode.indexOf(format);
-            const maxFirstDigit = parseInt(maxValue[format].toString().substr(0, 1), 10);
+            const maxFirstDigit =
+                format in maxValues
```
waterplea, line 23: How can `format` not be in `maxValues` if `Record` enforces all formats to be present?

projects/kit/utils/mask/create-auto-corrected-time-pipe.ts
```diff
@@ -26,10 +31,11 @@ export function tuiCreateAutoCorrectedTimePipe(
             }
         });
 
-        const isInvalid = timeFormatArray.some(
-            format =>
-                parseInt(conformedValue.substr(timeMode.indexOf(format), 2), 10) >
-                maxValue[format],
+        const isInvalid = timeFormatArray.some(format =>
+            format in maxValues
```
waterplea, line 35: Same here.

projects/kit/utils/mask/create-time-mask.ts
```diff
@@ -1,13 +1,39 @@
 import {TuiTimeMode} from '@taiga-ui/cdk';
 import {TUI_DIGIT_REGEXP, TuiTextMaskList} from '@taiga-ui/core';
-import {TUI_TIME_MASK} from '@taiga-ui/kit/constants';
+import {MAX_TIME_VALUES} from '@taiga-ui/kit/constants';
+import {TuiTimeFormatParts} from '@taiga-ui/kit/types';
 
-export function tuiCreateTimeMask(mode: TuiTimeMode): TuiTextMaskList {
+function tuiCreateTimePartMask(
+    maxPartValue: number | undefined,
```
waterplea, line 7: Again, looks like it cannot be `undefined`.

projects/kit/utils/mask/create-time-mask.ts
```diff
@@ -1,13 +1,39 @@
...
+    prefix?: string,
+): Array<string | RegExp> {
+    const partLength =
+        typeof maxPartValue === 'number' ? maxPartValue.toString().length : 0;
```
waterplea, line 11: `String(x)` is a slightly shorter way to write `x.toString()` ;)

projects/kit/utils/mask/create-time-mask.ts
```diff
@@ -1,13 +1,39 @@
...
+
+    if (!partLength) {
+        return [];
+    }
+
+    const regExp = Array.from<undefined, RegExp | string>(
```
waterplea, line 17: `Array(length).fill(TUI_DIGIT_REGEXP)`

splincode: Oh, just yesterday I recommended using a different format
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-new-array.md

@mullinsmikey anyway, sorry about

projects/kit/utils/mask/create-auto-corrected-time-pipe.ts
mullinsmikey: Yesterday @splincode specifically asked me to perform this check in this code and this wouldn't be necessary if such kind of enforcement worked with tokens. Right now, one can provide `maxValues: {}` through DI.

projects/kit/utils/mask/create-time-mask.ts
mullinsmikey: I'm aware of it, thanks for your concern) this form is just a little more explicit

projects/kit/utils/mask/create-auto-corrected-time-pipe.ts
splincode: Yes, we need to exclude non-obvious behavior at runtime with null pointer

waterplea: I'd leave it. We use Typescript and enforce strict types everywhere. If somebody provides wrong type for token that's their fault. That was the philosophy behind Taiga UI from the beginning. It's unfortunate that tokens are not type safe, that's true. But that's true across ALL tokens.

waterplea: We can use `Partial` type for token and add this to make us safer:
```ts
const max = {...MAX_TIME_VALUES, ...maxValues};
```

mullinsmikey: Nice idea, I'll work on this today

projects/kit/utils/mask/create-time-mask.ts
waterplea: I'm skeptical about most of such rules :D

projects/demo/src/modules/components/input-time/examples/5/index.ts
```diff
@@ -0,0 +1,32 @@
...
+    templateUrl: './index.html',
+    changeDetection,
+    encapsulation,
+    providers: [
+        {
+            provide: TUI_INPUT_TIME_OPTIONS,
+            useValue: {
+                ...TUI_INPUT_TIME_DEFAULT_OPTIONS,
+                mode: 'HH:MM',
+                maxValues: {HH: 11, MM: 59},
```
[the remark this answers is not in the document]
mullinsmikey: The solution could've been better if I spent more time playing with the mask but since this field is time-only, I suppose the current implementation is somewhat acceptable.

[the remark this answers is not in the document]
waterplea: Ok, let's keep the am/pm issue open then to improve this later.

projects/kit/components/input-time/input-time-options.ts
```diff
@@ -1,11 +1,16 @@
...
+import {MAX_TIME_VALUES} from '@taiga-ui/kit/constants';
+import {TuiTimeFormatParts} from '@taiga-ui/kit/types';
 import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
 
 export interface InputTimeOptions {
     readonly icon: PolymorpheusContent<TuiContextWithImplicit<TuiSizeS | TuiSizeL>>;
     readonly mode: TuiTimeMode;
+    readonly postfix: string;
+    readonly maxValues: Record<TuiTimeFormatParts, number>;
+    readonly strict12h: boolean;
```
waterplea, line 13: Once added, the flags cannot be taken back 🤔 Let's not add it for now. Maybe once we add `minValues` we will not need it?

waterplea (changes requested):
Please remove the strict12h flag for now, otherwise looks good to me, thank you very much!

waterplea (approved):

splincode (approved):

Issue #1337, opened 2022-02-08
🚀 - Expose native input in tricky cases


waterplea:
### Which @taiga-ui/* package(s) are relevant/releated to the feature request?

kit

### Description

Current `Textfield` directive, exposing native inputs from Taiga UI controls also needs to work with `TextArea`, `InputInline` and non-`PrimitiveTextfield`-based controls (`InputSlider`, `InputRange`, maybe `InputTag`/`MultiSelect`...)

### Cleaning up

Also fix change detection in `InputMonth` and `InputMonthRange` and consider `maxLength` in `InputNumber` and `inputMode` in `InputPhone`.

waterplea:
All issues are fixed and `textarea` is exposed. The rest we can expose if we need so in the future as separate tasks.

PR #1415, opened 2022-02-17
fix(kit): `InputMonth` fix change detection in `Textfield` mode

waterplea:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

Part of #1337

- [ ] Yes
- [ ] No


projects/kit/components/input-month-range/input-month-range.directive.ts
```diff
@@ -10,16 +20,53 @@ import {TuiInputMonthRangeComponent} from './input-month-range.component';
...
+    ) {
+        super(host);
+
+        this.value$
+            .pipe(
+                distinctUntilChanged(),
+                switchMap((value: TuiMonthRange | null) =>
+                    combineLatest([
+                        formatter(value && value.from),
+                        formatter(value && value.to),
```
splincode, line 48: `formatter(value?.from)`
`formatter(value?.to)`

splincode (approved):

projects/kit/components/input-month-range/input-month-range.directive.ts
splincode: why not `TuiHandler<TuiMonth | null | undefined, ...`?)

waterplea: Because it is meant to work with a form control and lack of value is `null` in Angular forms. I don't really want to extend it to support everything and besides `?.` is transpiled to a more verbose structure anyway.

projects/demo/src/modules/components/input-month-range/examples/import/insert-template.md
```diff
@@ -0,0 +1,5 @@
+```html
+<form [formGroup]="testForm">
+  <tui-input-month-range formControlName="testValue">Choose a range of months</tui-input-month-range>
```
nsbarsukov, line 3: `<input tuiTextfield />` ?

projects/kit/components/input-month-range/input-month-range.directive.ts
```diff
@@ -10,16 +20,53 @@ import {TuiInputMonthRangeComponent} from './input-month-range.component';
...
+                    ]),
+                ),
+                takeUntil(destroy$),
+            )
+            .subscribe(localizedValue => {
+                this.localizedValue = localizedValue;
```
nsbarsukov, line 54: **nit**

Can `@HostListener('change'...)` do the same thing (without subscription) ?

nsbarsukov (approved):

vladimirpotekhin (approved):

projects/demo/src/modules/components/input-month-range/examples/import/insert-template.md
waterplea: It's optional, only if you need to declare something native like `placeholder`. In future we will just have `Textfield` inside `PrimitiveTextfield` as a fallback so it would remain optional in 3.0

projects/kit/components/input-month-range/input-month-range.directive.ts
waterplea: We cannot because formatting relies on i18n which is `Observable`.

PR #1425, opened 2022-02-18
fix(kit): `Slider` change detection problem with `[(ngModel)]`

nsbarsukov:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

The ValueAccessor.writeValue method is called twice on any value accessor during component initialization, when a control is bound using `[(ngModel)]`, first time with a phantom `null` value.
With `changeDetection: ChangeDetectionStrategy.OnPush` the second call of writeValue with real value don't re-render the view.

See this issue https://github.com/angular/angular/issues/14988

- [ ] Yes
- [X] No

### DIFF
--- a/projects/kit/components/slider/slider.component.ts
+++ b/projects/kit/components/slider/slider.component.ts
@@ -1,5 +1,6 @@
 import {
     ChangeDetectionStrategy,
+    ChangeDetectorRef,
     Component,
     ElementRef,
     HostBinding,
@@ -16,6 +17,7 @@ import {
     tuiDefaultProp,
 } from '@taiga-ui/cdk';
 import {TuiSizeS} from '@taiga-ui/core';
+import {take} from 'rxjs/operators';
 
 @Component({
     /**
@@ -60,6 +62,10 @@ export class TuiSliderComponent {
         return Number(this.elementRef.nativeElement.step) || 1;
     }
 
+    get value(): number {
+        return Number(this.elementRef.nativeElement.value) || 0;
+    }
+
     @HostBinding('style.--tui-slider-fill-percentage.%')
     get fillPercentage(): number {
         return (100 * this.value) / (this.max - this.min);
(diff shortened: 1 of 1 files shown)


projects/kit/components/slider/slider.component.ts
```diff
@@ -75,20 +81,26 @@ export class TuiSliderComponent {
...
+            /**
+             * The ValueAccessor.writeValue method is called twice on any value accessor during component initialization,
+             * when a control is bound using [(ngModel)], first time with a phantom null value.
+             * With `changeDetection: ChangeDetectionStrategy.OnPush` the second call of writeValue with real value don't re-render the view.
+             * ___
+             * See this {@link https://github.com/angular/angular/issues/14988 issue}
+             */
+            control.valueChanges
+                ?.pipe(take(1))
+                .subscribe(() => this.changeDetectorRef.markForCheck());
```
splincode, line 103: Why not only `detectChanges`? Why should we notify parents? ))

splincode (approved):

projects/kit/components/slider/slider.component.ts
waterplea: Not sure it would work due to all the HostBindings

projects/kit/components/slider/slider.component.ts
```diff
@@ -75,20 +81,26 @@ export class TuiSliderComponent {
...
-
-        return controlValue || Number(this.elementRef.nativeElement.value) || 0;
-    }
-
     constructor(
         @Optional()
         @Self()
         @Inject(NgControl)
-        private readonly control: NgControl | null,
+        readonly control: NgControl | null,
```
waterplea, line 88: Why `control` is no longer `private`? I think you meant to just remove all modifiers. And `ChangeDetectorRef` is also not needed on the class. I prefer `watch(changeDetectorRef)` operator in the chain and empty `subscribe` but that's up to you.

projects/kit/components/slider/slider.component.ts
splincode: But if we need update only full parent view, we can access to parent ChangeDetectorRef

```ts
constructor(@SkipSelf() protected parentCdRef ChangeDetectorRef) {
```

and then trigger only update view `this.parentCdRef.detectChanged()`

projects/kit/components/slider/slider.component.ts
nsbarsukov: fixed

projects/kit/components/slider/slider.component.ts
```diff
@@ -75,20 +82,24 @@ export class TuiSliderComponent {
...
+    ) {
+        if (control instanceof NgModel) {
+            /**
+             * The ValueAccessor.writeValue method is called twice on any value accessor during component initialization,
+             * when a control is bound using [(ngModel)], first time with a phantom null value.
+             * With `changeDetection: ChangeDetectionStrategy.OnPush` the second call of writeValue with real value don't re-render the view.
+             * ___
+             * See this {@link https://github.com/angular/angular/issues/14988 issue}
+             */
+            control.valueChanges?.pipe(watch(changeDetectorRef), take(1)).subscribe();
```
splincode, line 102: forgot `takeUntil`?

projects/kit/components/slider/slider.component.ts
nsbarsukov: @splincode detectChanges doesnt helps here(

projects/kit/components/slider/slider.component.ts
nsbarsukov: first value will emit anyway if you use `ngModel`
At least, `undefined` will be emitted

waterplea (approved):

PR #1467, opened 2022-03-01
[WIP] feat(core): make global styles optional

### LINKED ISSUE #906: 🚀 - Make global styles optional

core

### Description

In future, we want to make all our components work without global styles, just with a theme.

waterplea:
Closes #906


splincode (approved):

splincode:
I'm not sure why e2e tests fell, maybe you need to run them locally

projects/cdk/services/directive-styles.service.ts
```diff
@@ -10,11 +18,26 @@ import {TUI_DEFAULT_RENDERER} from '@taiga-ui/cdk/tokens';
     providedIn: 'root',
 })
 export class TuiDirectiveStylesService {
+    private readonly map = new Map<Type<unknown>, unknown>();
```
splincode, line 21: why not WeakMap?

waterplea: Would it make any difference here? 🤔

splincode: I think this will protect against memory leaks at the time of HMR

waterplea: But the key is just a class, nobody stores it anywhere. Wouldn't that cause this `WeakMap` to drop the reference?

splincode: just wondering what you think, okey

waterplea: I think we don't really need this and can even drop this map altogether 🙂 it would behave as if empty component was attached to the same selector and act naturally except the component is created manually and doesn't take up place of that single component that can match a selector. But I figured we might as well save some performance since styles are attached only the first time.

projects/core/components/group/group.directive.ts
```diff
@@ -33,6 +35,14 @@ export class TuiGroupDirective {
     @tuiDefaultProp()
     size: TuiSizeL = 'm';
 
+    absoluteLoh = 'loh';
```
vladimirpotekhin, line 38: no loh

waterplea: Looool

vladimirpotekhin (approved):

PR #1512, opened 2022-03-13
feat(core): add type inference to `icon` properties

splincode:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

## What is the new behavior?

image

- [ ] Yes
- [x] No


nsbarsukov (approved):

projects/core/components/svg/svg.component.ts
```diff
@@ -48,7 +48,7 @@ const FAILED_EXTERNAL_ICON = 'Failed to load external SVG';
 export class TuiSvgComponent {
     private readonly src$ = new ReplaySubject<void>(1);
     private readonly isIE = isIE(this.userAgent);
-    private icon = '';
+    private icon: keyof typeof import('@taiga-ui/icons') | string = '';
```
vladimirpotekhin, line 51: `keyof typeof import('@taiga-ui/icons')` seems so long. How about type alias?

vladimirpotekhin, line 51: `keyof typeof import('@taiga-ui/icons')` seems so long. How about type alias?

splincode: Where do you think it's best to place an alias? `kit` or `cdk` package or export from `icons` package?

vladimirpotekhin: Hmm.. I think cdk is the best place

waterplea: Definitely not `cdk` and also this makes packages dependent. It would cause issues with proprietary icons because there are more names and because in proprietary environment we do not install public icons package. I think this whole PR is doubtful.

PR #1518, opened 2022-03-14
feat(kit): `TUI_TAG_OPTIONS` default options configuration for `Tag`

zarghamkhandev:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

Partially implements https://github.com/Tinkoff/taiga-ui/issues/133

## What is the new behavior?

Default tag options can be set via a provider. This is a partial implementation of https://github.com/Tinkoff/taiga-ui/issues/133

- [ ] Yes
- [x] No

### DIFF
--- a/projects/kit/components/tag/tag-options.ts
+++ b/projects/kit/components/tag/tag-options.ts
@@ -5,9 +5,6 @@ import {TuiStatusT} from '@taiga-ui/kit/types';
 export interface TuiTagOptions {
     readonly size: TuiSizeS | TuiSizeL;
     readonly status: TuiStatusT;
-    readonly hoverable: boolean;
-    readonly removable: boolean;
-    readonly disabled: boolean;
     readonly autoColor: boolean;
 }
 
@@ -15,9 +12,6 @@ export interface TuiTagOptions {
 export const TUI_TAG_DEFAULT_OPTIONS: TuiTagOptions = {
     size: 'm',
     status: 'default',
-    hoverable: false,
-    removable: false,
-    disabled: false,
     autoColor: false,
 };
 
--- a/projects/kit/components/tag/tag-options.ts
+++ b/projects/kit/components/tag/tag-options.ts
@@ -5,7 +5,6 @@ import {TuiStatusT} from '@taiga-ui/kit/types';
 export interface TuiTagOptions {
     readonly size: TuiSizeS | TuiSizeL;
     readonly status: TuiStatusT;
-    readonly showLoader: boolean;
     readonly hoverable: boolean;
     readonly removable: boolean;
     readonly disabled: boolean;
@@ -16,7 +15,6 @@ export interface TuiTagOptions {
 export const TUI_TAG_DEFAULT_OPTIONS: TuiTagOptions = {
     size: 'm',
     status: 'default',
-    showLoader: false,
     hoverable: false,
     removable: false,
     disabled: false,
--- a/projects/kit/components/tag/tag-options.ts
+++ b/projects/kit/components/tag/tag-options.ts
@@ -1,0 +1,38 @@
+import {InjectionToken, ValueProvider} from '@angular/core';
+import {TuiSizeL, TuiSizeS} from '@taiga-ui/core/types';
+import {TuiStatusT} from '@taiga-ui/kit/types';
+
+export interface TuiTagOptions {
+    readonly size: TuiSizeS | TuiSizeL;
+    readonly status: TuiStatusT;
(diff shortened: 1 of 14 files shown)


projects/kit/components/tag/tag-options.ts
```diff
@@ -0,0 +1,36 @@
...
+    readonly removable: boolean;
+    readonly disabled: boolean;
+    readonly autoColor: boolean;
+}
+
+/** Default values for the tag options. */
+export const TUI_TAG_DEFAULT_OPTIONS: TuiTagOptions = {
+    size: 'm',
+    status: 'default',
+    hoverable: false,
```
waterplea, line 18: I have doubts about `hoverable`, `removable` and `disabled`, what do you guys think? @splincode @vladimirpotekhin @nsbarsukov 

It looks like it doesn't make sense to change defaults for those since they are very situational and people enable them only in places they need them.

waterplea (approved):
Thank you for this PR! I wonder if we can somehow deprecate `TUI_TAG_STATUS` token now 🤔

waterplea:
To fix build please add explicit types to tag inputs.

zarghamkhandev:
> Thank you for this PR! I wonder if we can somehow deprecate `TUI_TAG_STATUS` token now 🤔

meaning that we use TUI_TAG_DEFAULT_OPTIONS instead ?

projects/kit/components/tag/tag-options.ts
zarghamkhandev: Should i go for it and remove these props ?

nsbarsukov: > I have doubts about hoverable, removable and disabled, what do you guys think?

Sorry for the late response!
I agree that all interactive props (`hoverable`, `removable`, `disabled`) should be always `false` by default (without customization by DI-token).

**Argumentation:**
If user set them to true by default (for example, in root `app.module.ts`), it can forget about handlers somewhere deeper in the app. And it will break UX (imagine, tag has removal icon but do nothing when user click on this icon).
I think with explicit setting of this interactive props, there is less probability of this mistake.


@zarghamkhandev is there a need to customize these interactive props (`hoverable`, `removable`, `disabled`) by the token in your project ?  If no, could we temporarily remove them from `TUI_TAG_OPTIONS`. If there is demand on these, we will add them later.

nsbarsukov:
> > Thank you for this PR! I wonder if we can somehow deprecate `TUI_TAG_STATUS` token now 🤔
> 
> meaning that we use TUI_TAG_DEFAULT_OPTIONS instead ?

@zarghamkhandev he means that old token `TUI_TAG_STATUS` can become part of new `TUI_TAG_OPTIONS`.
But it is difficult to do without breaking changes (because both token are optional and have default value).

Could you please add here comment `//TODO use TUI_TAG_OPTIONS instead of it in v3.0` ?
https://github.com/Tinkoff/taiga-ui/blob/9b6113f6c280f017179440663eb72a9d2556ba91/projects/kit/tokens/tag-status.ts#L3-L4

nsbarsukov (approved):
Thanks for the PR, it is awesome!

Could you please rename PR to 
```
feat(kit): `TUI_TAG_OPTIONS` default options configuration for `Tag`
```

waterplea:
> Thanks for the PR, it is awesome!
> 
> Could you please rename PR to
> 
> ```
> feat(kit): `TUI_TAG_OPTIONS` default options configuration for `Tag`
> ```

We can just change squash commit message when we merge it. @zarghamkhandev please remove those options we discussed and add TODO for TUI_TAG_STATUS token and we can merge it once all jobs pass.

splincode (approved):

splincode:
@nsbarsukov @zarghamkhandev any progress? what else is left to do?

projects/kit/components/tag/tag-options.ts
zarghamkhandev: removed ✅

zarghamkhandev:
I'm sorry, been busy a bit lately. Just pushed the changes.

Remove interactive options (hoverable, removable, disabled) ✅

Add TODO comment for TUI_TAG_STATUS token ✅

nsbarsukov (approved):

vladimirpotekhin (approved):

zarghamkhandev:
any ideas on how to fix failing test ?

splincode:
@zarghamkhandev don't worry about e2e-tests

PR #1530, opened 2022-03-18
fix(kit): `InputDateRange` fix initial months

### LINKED ISSUE #1519: 🐞 - InputDateRange incorrect display of the max date

kit

https://stackblitz.com/edit/taiga-starter-qrjcjg?file=src%2Fapp%2Fapp.component.ts

### Description

If set the max value before the current date, then the current and next month are always displayed, even if they are disabled. It would be more correct to display the maximum available month in the right column.

### Angular version

12.2.4

### Taiga UI version

2.37.0

- [X] Chrome
- [X] Firefox
- [X] Safari
- [X] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

vladimirpotekhin:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #1519 

## What is the new behavior?

Previously initial months was corrected by getter in calendar, but it was deleted in #432 due to jumps (using maxLength property)

Now we just once correcting initial value

- [ ] Yes
- [x] No


projects/kit/internal/primitive-calendar-range/primitive-calendar-range.component.ts
```diff
@@ -110,6 +110,7 @@ export class TuiPrimitiveCalendarRangeComponent implements OnInit {
 
     ngOnInit() {
         this.updateViewedMonths();
+        this.setInitialMonths();
```
nsbarsukov, line 113:
```ts
this.updateViewedMonths();
```
This old function set values to `this.userViewedMonthFirst` and `this.userViewedMonthSecond` too (as your new function `this.setInitialMonths();`).

This functions are both synchronous, so the second function overrides the second one.
Do we need both functions ?

nsbarsukov (commented):
nit: tests ?

waterplea (approved):
Approved so you can merge today after you resolve @nsbarsukov comment

projects/kit/internal/primitive-calendar-range/primitive-calendar-range.component.ts
vladimirpotekhin: yes, `this.updateViewedMonths()` set values to `this.userViewedMonthFirst` and `this.userViewedMonthSecond` every value change. But `setInitialValues` we should call only once (onInit) to clamp `this.userViewedMonthFirst` and `this.userViewedMonthSecond` between max/min

vladimirpotekhin: I'll add unit for this case 👍

waterplea: Then maybe we do not need `this.updateViewedMonths();` in `ngOnInit`?

vladimirpotekhin: thanks, `this.updateViewedMonths();` deleted from `ngOnInit`

PR #1575, opened 2022-03-30
Migrations

vladimirpotekhin:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

- [ ] Yes
- [x] No


splincode (approved):

projects/cdk/schematics/ng-update/constants/types.ts
```diff
@@ -0,0 +1,17 @@
+interface TypeToRename {
+    from: string;
+    to: string;
+    moduleSpecifier?: string | string[];
+}
+export const TYPES_TO_RENAME: readonly TypeToRename[] = [
+    {
+        from: 'ButtonOptions',
+        to: 'TuiButtonOptions',
+        moduleSpecifier: ['@taiga-ui/core', '@taiga-ui/core/components'],
```
waterplea, line 10: There should be no deep imports in projects. What is the second specifier for? By the way we can probably add a schematic to remove deep imports if projects have any.

projects/cdk/schematics/ng-update/constants/enums.ts
```diff
@@ -0,0 +1,61 @@
+interface ReplacementEnum {
+    name: string;
+    replaceValues: Record<string, string>;
+}
+
+export const ENUMS_FOR_REPLACE: ReplacementEnum[] = [
```
waterplea, line 6: Feels like there were more enums to replace 🤔

projects/cdk/schematics/utils/get-named-import-references.ts
```diff
@@ -0,0 +1,22 @@
+import {arrayFlat, getImports, Node} from 'ng-morph';
+
+export function getNamedImportReferences(
+    namedImport: string,
+    moduleSpecifier?: string | string[],
+): Node[] {
+    const importDeclarations = getImports('**/**', {
+        namedImports: [namedImport],
+        moduleSpecifier: moduleSpecifier || '**/**',
```
waterplea, line 9: Add fallback directly to the attribute so you can just write `moduleSpecifier,` here

projects/cdk/schematics/ng-update/tests/schematic-replace-enums.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+    afterEach(() => {
+        resetActiveProject();
+    });
+});
+
+function createMainFiles() {
+    createSourceFile(
+        'test/app/app.component.ts',
+        `import {Component} from '@angular/core';
+import {TuiTextAlign, some} from '@taiga-ui/cdk';
```
waterplea, line 60: You think there's a chance to properly auto-format such cases?

projects/cdk/schematics/ng-update/constants/enums.ts
vladimirpotekhin: added more enums 👍

projects/cdk/schematics/ng-update/tests/schematic-replace-enums.spec.ts
vladimirpotekhin: No, unfortunately . I think it technically possible to run prettier after schematic, but I don't think that is is right way

projects/cdk/schematics/ng-update/constants/types.ts
vladimirpotekhin: We need to take not all ButtonOptions, only from concrete package (described in `moduleSpecifier`). I can remove second specifier but who knows how ButtonOptions imported in a random project?)

vladimirpotekhin:
> 

I added `keepAsType` option, depending on this options enums used as type and their imports will not remove

nsbarsukov (approved):

splincode:
why is `chore` instead of `feat` prefix?

vladimirpotekhin:
> why is `chore` instead of `feat` prefix?

I don't want to add this to the public changelog now. It's just a part of bigger migration and it doesn't matter yet

waterplea (approved):

PR #1583, opened 2022-04-01
[WIP] feat(kit): `InputRange` refactor (use `InputNumber` inside)

nsbarsukov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

Closes # 

- [ ] Yes
- [ ] No


projects/kit/components/input-range/input-range.style.less
```diff
@@ -1,143 +1,52 @@
...
-        :host[data-mode='onDark']._disabled & {
-            color: var(--tui-text-03-night);
-        }
-    }
-}
-
-.t-native {
-    .clearinput();
+.t-text-input {
     flex: 1;
```
waterplea, line 12: Left input moves the right one away with a long placeholder. I wonder if it makes sense to absolutely position inputs instead and make first one have 100% of width and the second to just overlay on top with 50% right: 0
image

projects/kit/components/input-range/input-range.style.less
```diff
@@ -1,143 +1,52 @@
...
-    &._segmented {
-        // TODO: research if it can be used in rem units
-        /* stylelint-disable-next-line */
-        border-bottom: 26px solid transparent;
-    }
-
-    &[data-mode='onDark'] {
-        color: var(--tui-base-01);
+tui-input-number ::ng-deep tui-wrapper {
+    &:after {
```
waterplea, line 2: Maybe we can just add this to appearance="none" so we don't resort to ng-deep?

waterplea (approved):

projects/kit/components/input-range/input-range.template.html
```diff
@@ -101,6 +81,30 @@
         [disabled]="readOnly || disabled"
         [ngModel]="value"
         (ngModelChange)="onRangeValue($event)"
-        (mousedown.prevent)="onMouseDown()"
+        (tuiPressedChange)="focusToTextInput()"
```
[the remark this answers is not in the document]
nsbarsukov: we should move focus on mouseUp (instead of mouseDown) because otherwise `InputNumber` cannot deal with it

projects/kit/components/input-range/input-range.style.less
nsbarsukov: good point, thanks!

But if make them absolutely positioned, we can face with problem of overlapping long values.
May be we should make `max-width: 80%` for left input ?

projects/kit/components/input-range/input-range.style.less
nsbarsukov: ✅

projects/kit/components/input-range/input-range.component.ts
```diff
@@ -43,270 +65,255 @@ import {AbstractTuiInputSlider} from '@taiga-ui/kit/abstract';
...
 
-        if (value[0] !== this.value[0] || value[1] !== this.value[1]) {
-            this.updateValue(value);
+        if (textInputRef) {
+            textInputRef.nativeValue = textInputRef.getFormattedValue(value);
         }
     }
 }
+
+@Directive({
```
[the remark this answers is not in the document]
nsbarsukov: ✅

projects/core/styles/theme/wrapper/base.less
```diff
@@ -1,8 +1,8 @@
 @import '../../taiga-ui-local.less';
 
 /* stylelint-disable order/order */
-tui-wrapper,
-[tuiWrapper] {
+tui-wrapper:not([data-appearance='none']),
+[tuiWrapper]:not([data-appearance='none']) {
```
waterplea, line 5: Gotta be careful not to mess the specificity of some rules that override those. Just keep that in mind, screenshot tests might not catch some tricky case

projects/kit/components/input-range/input-range.style.less
waterplea: That's up to you to test extreme values/placeholders and pick best solution. In my head absolute positioning causes less troubles :)

projects/kit/components/input-range/input-range.template.html
```diff
@@ -101,6 +81,31 @@
...
+        (activeThumbChange)="lastActiveSide = $event"
```
waterplea, line 85: Let's make it a function

projects/core/styles/theme/wrapper/base.less
nsbarsukov: You are right.
The increased specificity breaks other appearance.

Luckily, our screenshot tests works))

splincode (approved):

projects/core/styles/theme/wrapper/base.less
nsbarsukov: fixed

projects/kit/components/input-range/input-range.style.less
nsbarsukov: fixed ✅

Also, I added screenshot tests for this case
01-long-placeholder_basic snap

___
02-long-placeholder_value-content snap

___
03-long-placeholder_pluralize snap

vladimirpotekhin (approved):

PR #1641, opened 2022-04-17
fix(kit): incorrect alignment of icons in multiselect

### LINKED ISSUE #1636: 🐞 - `MultiSelect` incorrect arrow's vertical alignment inside `Dialog`

kit

https://stackblitz.com/edit/select-arrow-vertical-alignment-bug?file=src/app/app.component.ts

### Description

1. Open stackblitz example
2. Press button to open dialog
Screenshot 2022-04-15 at 11 02 48

### Angular version

any

### Taiga UI version

2.36.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

volvachev:
Fixes a bug in `multiselect` when the above elements have a `data-size` attribute.

Fixes #1636

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #1636

- [ ] Yes
- [x] No

### DIFF
--- a/projects/kit/components/multi-select/multi-select.style.less
+++ b/projects/kit/components/multi-select/multi-select.style.less
@@ -29,15 +29,15 @@
         opacity: var(--tui-disabled-opacity);
     }
 
-    :host-context([data-size='s']) & {
+    :host[data-size='s'] & {
         height: var(--tui-height-s);
     }
 
-    :host-context([data-size='m']) & {
+    :host[data-size='m'] & {
         height: var(--tui-height-m);
     }
 
-    :host-context([data-size='l']) & {
+    :host[data-size='l'] & {
         height: var(--tui-height-l);
     }
 }
--- a/projects/demo-integrations/cypress/tests/kit/multi-select/multi-select.spec.ts
+++ b/projects/demo-integrations/cypress/tests/kit/multi-select/multi-select.spec.ts
@@ -23,6 +23,29 @@ describe('MultiSelect', () => {
                 .wait(DEFAULT_TIMEOUT_BEFORE_ACTION)
                 .matchImageSnapshot(`01-not-overflow-by-tags`);
         });
+
+        ['s', 'm', 'l'].forEach((size, index) => {
+            it(`multiselect inside dialog with tuiTextfieldSize=${size}`, () => {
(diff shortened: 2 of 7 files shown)


projects/kit/components/multi-select/multi-select.style.less
```diff
@@ -29,15 +29,15 @@
         opacity: var(--tui-disabled-opacity);
     }
 
-    :host-context([data-size='s']) & {
+    :host[data-size='s'] & {
```
splincode, line 32: Hello, please add e2e test for this case

volvachev: Hi, do I need to create a new `doc-example` for this case? On which page should I create a new example (`dialog` \ `multiselect`)?
Can I create a new integration test without a new `doc-example' block?

splincode: Hello! 

> do I need to create a new doc-example for this case? 

Does the bug reproduce on the page `multi-select/API?tuiMode=null`? If yes, then the test should be done on this page, without creating an example. Otherwise, you can create your own example.

> Can I create a new integration test without a new `doc-example' block?

You can add tests here - https://github.com/Tinkoff/taiga-ui/blob/main/projects/demo-integrations/cypress/tests/kit/multi-select/multi-select.spec.ts

volvachev: > Does the bug reproduce on the page `multi-select/API?tuiMode=null`? If yes, then the test should be done on this page, without creating an example. Otherwise, you can create your own example.

I've created the tests, but I'm not sure if I wrote them right.

splincode:
Do you think it is possible to reproduce this bug in the current showcase without creating an additional example?

image

image

volvachev:
> Do you think it is possible to reproduce this bug in the current showcase without creating an additional example?

I tried it, but I didn't manage to reproduce it on `api` page. 
We should wrap `multiselect` component with some container that will have `data-size` attribute.

splincode (approved):

waterplea (approved):

PR #1654, opened 2022-04-19
ci: enable some rxjs rules for eslint

splincode:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update


projects/cdk/observables/drag-and-drop-from.ts
```diff
@@ -16,7 +16,7 @@ export class TuiDragState {
     constructor(readonly stage: TuiDragStage, readonly event: MouseEvent) {}
 }
 
-export function dragAndDropFrom(element: Element): Observable<TuiDragState> {
+export function dragAndDropFrom(element: Element): Observable<TuiDragState> | never {
```
waterplea, line 19: Why is that `never`? Isn't throwing ensures we return what we expect or we cause an explicit error stating this is not the intended situation?

splincode: `export function dragAndDropFrom(element: Element): Observable<TuiDragState> | never {` 

in this case, then it would be necessary to rewrite code in such a way that the error will be returned through the stream

image

splincode: for 

image

instead of

image

waterplea: For external user we are not supposed to know the function throws. It basically serves as a typecast but with an explicit message for the unexpected situation.

splincode: fixed

nsbarsukov (approved):

waterplea (approved):

PR #1669, opened 2022-04-22
fix(kit): emit null from mode provider instead cancel stream

splincode:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

- [ ] Yes
- [x] No

### DIFF
--- a/projects/kit/components/input-tag/input-tag.component.ts
+++ b/projects/kit/components/input-tag/input-tag.component.ts
@@ -185,9 +185,7 @@ export class TuiInputTagComponent
         this.initScrollerSubscription(scroller);
     }
 
-    status$: Observable<TuiStatusT> = this.mode$.pipe(
-        map(mode => (mode ? 'default' : this.tagStatus)),
-    );
+    status$: Observable<TuiStatusT> = this.mode$.pipe(map(() => this.status));
 
     open = false;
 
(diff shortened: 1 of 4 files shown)


waterplea (changes requested):
I think it makes more sense to change `EMPTY` to `of(null)` in `MODE_PROVIDER`

projects/kit/components/input-tag/input-tag.component.ts
```diff
@@ -295,12 +295,8 @@ export class TuiInputTagComponent
         return this.hasCleaner || this.hasTooltip || this.iconAlignRight;
     }
 
-    /**
-     * @deprecated: use `status$ | async` instead
-     * TODO: remove in v3.0
-     */
     get status(): TuiStatusT {
-        return this.modeDirective && this.modeDirective.mode ? 'default' : this.tagStatus;
+        return this.modeDirective?.mode ? 'default' : this.tagStatus;
```
nsbarsukov, line 299: we have 2 variables about status in different parts of file.

Can we place this logic inside stream `status$` (and drop this getter) ?
Something like this
```ts
status$: Observable<TuiStatusT> = this.mode$.pipe(
        map(mode => (mode ? 'default' : this.tagStatus)),
        map(status => status || this.modeDirective?.mode ? 'default' : this.tagStatus)
    );
```

splincode: it's probably async pipe is never emitted and return null

splincode: I changed to 

```
status$: Observable<TuiStatusT> = this.mode$.pipe(map(() => this.status));
```

waterplea (approved):

splincode:
@waterplea @nsbarsukov sorry, updated

nsbarsukov (approved):

PR #1703, opened 2022-04-29
fix(kit): `Slider[keySteps]` when `min`-property equals to `max`-property

nsbarsukov:
## PR Checklist

- [X] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

This code
```html
<tui-input-slider
    [min]="25"
    [max]="25"
    [ngModel]="25"
></tui-input-slider>
```
causes this error
Screenshot 2022-04-29 at 15 03 08

## What is the new behavior?
It works in the same way as native sliders

- [ ] Yes
- [X] No


projects/kit/components/slider/slider-key-steps.directive.ts
```diff
@@ -96,7 +98,18 @@ export class TuiSliderKeyStepsDirective
             return;
         }
 
-        this.slider.value = this.transformToNativeValue(controlValue);
+        const clampedControlValue = clamp(controlValue, this.min, this.max);
+
+        if (controlValue !== clampedControlValue) {
+            this.control?.setValue(clampedControlValue);
```
waterplea, line 104: Components should not affect control value unless through user actions. Form control component is a GUI for user to work with control, we can correct use actions but not externally set value. If you want — you can keep the assertion but manipulating form control value like that is something we avoid as it is philosophically wrong.

waterplea (changes requested):

projects/kit/components/slider/slider-key-steps.directive.ts
nsbarsukov: fixed ✅

PR #1727, opened 2022-05-05
feat(core): `Lists` add triangle bullet list type

n1k1t0ss:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [x] Code style update

Tinkoff add new core style for bullet list - it appears as triangle - "◤" but taiga doesn't have it.

## What is the new behavior?
Adds new style without changing others. It has same usage as regular linear: `<ul class="tui-list tui-list_triangle">`

Screenshot_89

- [ ] Yes
- [x] No

### DIFF
--- a/projects/core/styles/markup/tui-list.less
+++ b/projects/core/styles/markup/tui-list.less
@@ -75,6 +75,20 @@
                 background-color: transparent;
             }
         }
+        
+        .tui-list_triangle & {
+            padding-left: @space * 7;
+
+            &:before {
+                content: '◤';
+                left: 0;
+                top: auto;
+                width: auto;
+                height: auto;
+                background-color: transparent;
+                color: #9299a2;
+            }
+        }
 
         .tui-list_ordered & {
             padding-left: @space * 5;
--- a/projects/core/styles/markup/tui-list.less
+++ b/projects/core/styles/markup/tui-list.less
@@ -75,20 +75,6 @@
                 background-color: transparent;
             }
         }
-        
-        .tui-list_triangle & {
-            padding-left: @space * 7;
-
-            &:before {
-                content: '◤';
-                left: 0;
-                top: auto;
-                width: auto;
-                height: auto;
-                background-color: transparent;
-                color: #9299a2;
-            }
-        }
 
         .tui-list_ordered & {
             padding-left: @space * 5;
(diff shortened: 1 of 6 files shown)


projects/core/styles/markup/tui-list.less
```diff
@@ -75,6 +75,20 @@
...
+            padding-left: @space * 7;
+
+            &:before {
+                content: '◤';
+                left: 0;
+                top: auto;
+                width: auto;
+                height: auto;
+                background-color: transparent;
+                color: #9299a2;
```
nsbarsukov, line 89: ```var(--tui-base-06)``` ?

projects/core/styles/markup/tui-list.less
```diff
@@ -75,6 +75,20 @@
                 background-color: transparent;
             }
         }
+        
+        .tui-list_triangle & {
+            padding-left: @space * 7;
+
+            &:before {
+                content: '◤';
```
nsbarsukov, line 83: is it well supported (i am not sure) ?

Or better to use css-entity `\25E4` ?

nsbarsukov (changes requested):
Could you, please add example of new list to 
https://taiga-ui.dev/lists ?

projects/core/styles/markup/tui-list.less
n1k1t0ss: am ok with `\25E4`

projects/core/styles/markup/tui-list.less
n1k1t0ss: Thanks! It is pretty close to `#9299A2`, so I think, it would be better choise

projects/core/styles/markup/tui-list.less
n1k1t0ss: I added an example, as you requested:
Screenshot_92

I noticed 2 small bugs while working on this ticket:
- Edit on Stackblitz button doesn't work in lists example (check buttons on url https://taiga-ui.dev/lists)
- styles `.tui-list_ordered &` and `.tui-list_linear &` is incorrect, because they change all of their nodes, not only childrens. Correct should be `.tui-list_ordered > &` and `.tui-list_linear > &`

waterplea: Feel free to fix the second issue. Global styles is something that is often gets overlooked since we mostly work on Angular components and related stuff.

nsbarsukov (approved):

projects/core/styles/markup/tui-list.less
nsbarsukov: > it is pretty close to #9299A2

In our proprietary kit `--tui-base-06` is equal to `#9299A2` :D

PR #1737, opened 2022-05-07
fix(kit): make tree render children trees only when expanded

### LINKED ISSUE #1665: 🐞 - A tree renders all children(and children all nested children) event if parents are not expanded.

kit

_No response_

### Description

The tree should render children only if it's expanded. If my trees are made of an array of 3000 items, it will leak a lot of memory and a colossal performance downgrade. Especially if I render it in a dropdown.

I guess the fix is as simple as it gets.

We need to add `ngIf` in `tree.component.html`

```html
<tui-tree-item
    *tuiLet="children$ | async as children"
    #view
    [tuiTreeNode]="value"
    [expandable]="isExpandable"
>
    <div
        *ngIf="value !== children"
        polymorpheus-outlet
        [content]="content"
        [context]="{$implicit: value, node: view}"
    ></div>

     <ng-container *ngIf="item?.isExpanded" ngProjectAs="tui-tree">
        <tui-tree
           *ngFor="let child of children"
           [value]="child"
            [content]="content"
        ></tui-tree>
    </ng-container>
</tui-tree-item>
```

add a getter in `tree.component.ts` (because we have a handler to check whether it is expandable or not)

```ts
get isExpandable(): boolean {
    return !!this.handler(this.value).length;
  }
```

add an input to the `tree-item.component.ts` and remove `@ContentChildren nested` due to its needless.

Also I'd like to have an ability to pass `trackBy` to nested trees.

### Angular version

13.1.3

### Taiga UI version

2.34.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

EricPoul:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #1665

- [ ] Yes
- [ ] No

### DIFF
--- a/projects/kit/components/tree/components/tree-item/tree-item.component.ts
+++ b/projects/kit/components/tree/components/tree-item/tree-item.component.ts
@@ -7,6 +7,7 @@ import {
     forwardRef,
     HostBinding,
     Inject,
+    Optional,
     QueryList,
 } from '@angular/core';
 import {EMPTY_QUERY} from '@taiga-ui/cdk';
@@ -21,6 +22,7 @@ import {
     TUI_TREE_LEVEL,
     TUI_TREE_NODE,
 } from '../../misc/tree.tokens';
+import {TuiTreeComponent} from '../tree/tree.component';
 import {TUI_TREE_ITEM_PROVIDERS} from './tree-item.providers';
 
 @Component({
@@ -51,6 +53,9 @@ export class TuiTreeItemComponent implements DoCheck {
     );
 
     constructor(
+        @Optional()
+        @Inject(TuiTreeComponent)
+        private readonly parent: TuiTreeComponent<unknown> | null,
         @Inject(ElementRef)
         private readonly elementRef: ElementRef<HTMLElement>,
         @Inject(forwardRef(() => TUI_TREE_CONTROLLER))
@@ -63,7 +68,7 @@ export class TuiTreeItemComponent implements DoCheck {
 
     @HostBinding('class._expandable')
     get isExpandable(): boolean {
-        return !!this.nested.length;
+        return this.parent?.isExpandable ?? !!this.nested.length;
     }
 
     get isExpanded(): boolean {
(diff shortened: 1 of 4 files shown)


EricPoul:
@waterplea  I make a draft to show an intermediate result. I came back to `ngProjectAs` since without it the change detection problem pops up and works only by manually calling `detectChanges`(it's too much for me). Also the big problem that I'm struggling with comes up. Since it's always a new instance of the tree old maps that check the expansion of the tree don't work anymore. I haven't found a better solution yet(locally, haven't pushed anything connected to the maps).

projects/kit/components/tree/components/tree-item/tree-item.component.ts
```diff
@@ -51,6 +53,9 @@ export class TuiTreeItemComponent implements DoCheck {
     );
 
     constructor(
+        @Optional()
+        @Inject(TuiTreeComponent)
```
waterplea, line 57: Please use `forwardRef` just in case

EricPoul:
@waterplea 
So there are my thoughts:
I don't see any ways to implement this tree lazy rendering since we can't you `WeakMap` and I haven't found any other 100% workable options to track the `isExpand` of the tree after recreating it(while collapsing/expanding the parent tree). 
I tried to create unique ids(using the index of itself in the parent view) in a `Map` that have a relation to the parents recursively but it works not so well. If the items in the array are sorted or filtered and the index is changed those items will use someone`s else `isExpand` value. Another variant is to use `value`, it's similar but still, there're some issues: the value can be changed or it could be 2 or more equal strings as values.
As a real deal, I see only one option now(not pretty). Make an input `lazy`(kinda naming) and if you set it true you must pass one more input `trackBy`(to determine unique ids) or you'll have an error. It's the most workable one unless you pass an array style tree(strange one) as a value.

Please give a note what you think about all of these. I hope you won't offer such an elegant and easy way to manage it with it because I'll be shamed. 😄

waterplea:
@EricPoul sorry, didn't have time to think about it yet.

PR #1793, opened 2022-05-21
feat(core): add InjectionToken to config scrollbars

### LINKED ISSUE #1525: 🚀 - ability to disable or alter custom scollbar

_No response_

### Description

I think the title is pretty self-explanatory

ddubrava:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #1525

## What is the new behavior?

`TUI_DISABLE_CUSTOM_SCROLLBAR` token is added to disable a custom scrollbar and enable a native one.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/core/components/root/root.component.ts
+++ b/projects/core/components/root/root.component.ts
@@ -1,12 +1,11 @@
 import {DOCUMENT} from '@angular/common';
 import {ChangeDetectionStrategy, Component, ElementRef, Inject} from '@angular/core';
-import {TUI_DIALOGS, TUI_IS_MOBILE, tuiAssert, tuiPure} from '@taiga-ui/cdk';
+import {TUI_DIALOGS, TUI_IS_MOBILE, tuiAssert} from '@taiga-ui/cdk';
 import {VERSION} from '@taiga-ui/core/constants';
 import {TUI_IS_MOBILE_RES_PROVIDER} from '@taiga-ui/core/providers';
 import {
     TUI_ANIMATIONS_DURATION,
     TUI_ASSERT_ENABLED,
-    TUI_DISABLE_CUSTOM_SCROLLBAR,
     TUI_IS_MOBILE_RES,
     TUI_THEME,
 } from '@taiga-ui/core/tokens';
@@ -27,18 +26,10 @@ import {map} from 'rxjs/operators';
     },
 })
 export class TuiRootComponent {
-    @tuiPure
-    get isScrollbarShown$(): Observable<boolean> {
-        if (this.isMobile || this.customScrollbarDisabled) {
-            return of(false);
-        }
-
-        if (!this.dialogs.length) {
-            return of(true);
-        }
-
-        return merge(...this.dialogs).pipe(map(({length}) => !length));
-    }
+    readonly scrollbars$: Observable<boolean> =
+        this.dialogs.length && !this.isMobile
+            ? merge(...this.dialogs).pipe(map(({length}) => !length))
+            : of(!this.isMobile);
 
     constructor(
         @Inject(TUI_ANIMATIONS_DURATION) readonly duration: number,
@@ -50,24 +41,8 @@ export class TuiRootComponent {
         @Inject(TUI_IS_MOBILE_RES) readonly isMobileRes$: Observable<boolean>,
         @Inject(DOCUMENT) {body}: Document,
         @Inject(TUI_THEME) theme: string,
-        @Inject(TUI_DISABLE_CUSTOM_SCROLLBAR)
-        private readonly customScrollbarDisabled: boolean,
     ) {
         tuiAssert.enabled = enabled;
         body.setAttribute('data-tui-theme', theme.toLowerCase());
-        this.switchScrollbars(body);
-    }
-
-    /**
-     * Switches between native and custom scrollbars.
-     * If TUI_DISABLE_CUSTOM_SCROLLBAR is false, adds a class to disable a native scrollbar.
-     * If TUI_DISABLE_CUSTOM_SCROLLBAR is true, removes a class that disables a native scrollbar.
-     */
-    private switchScrollbars(body: HTMLElement): void {
-        const cssClass = 'tui-zero-scrollbar';
-
-        this.customScrollbarDisabled
-            ? body.classList.remove(cssClass)
-            : body.classList.add(cssClass);
     }
 }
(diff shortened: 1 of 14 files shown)


ddubrava:
The idea is to toggle `tui-zero-scrollbar` to control a native scrollbar.
Showing/hiding `tui-scroll-controls` helps to control a custom scrollbar.

Please, check `isScrollbarShown$` getter, I'm not sure I understood the logic.

projects/core/components/root/root.component.ts
```diff
@@ -26,10 +27,18 @@ import {map} from 'rxjs/operators';
     },
 })
 export class TuiRootComponent {
-    readonly scrollbars$: Observable<boolean> =
-        this.dialogs.length && !this.isMobile
-            ? merge(...this.dialogs).pipe(map(({length}) => !length))
-            : of(!this.isMobile);
+    @tuiPure
+    get isScrollbarShown$(): Observable<boolean> {
```
splincode, line 31: I can't agree that it's better

What do you think about make a custom pipe?

```html
<tui-scroll-controls
  *ngIf="dialogs | tuiScrollbars | async"
```

```ts
@Pipe({ name: 'tuiScrollbars' })
export class TuiScrollbarPipe implements PipeTransform {
   constructor(
       @Inject(DOCUMENT) {body}: Document,
       @Inject(TUI_SCROLLBAR_MODE)
       private readonly tuiScrollbar: TuiScrollbarMode,,
       @Inject(TUI_IS_MOBILE) private readonly isMobile: boolean,
    ) {
       // ps: not sure mutate body is good for disable scrollbar
        const cssClass = 'tui-zero-scrollbar';
        this.tuiScrollbar.enable
            ? body.classList.remove(cssClass)
            : body.classList.add(cssClass);
   }

   transform<T>(dialogs: ReadonlyArray<Observable<readonly T[]>>): Observable<boolean> {
      if (this.isMobile || !this.tuiScrollbar.enable) {
            return of(false);
        }

        if (!this.dialogs.length) {
            return of(true);
        }

        return merge(...this.dialogs).pipe(map(({length}) => !length));
   }
}
```

projects/core/tokens/disable-custom-scrollbar.ts
```diff
@@ -0,0 +1,8 @@
+import {InjectionToken} from '@angular/core';
+
+export const TUI_DISABLE_CUSTOM_SCROLLBAR = new InjectionToken(
```
splincode, line 3: What do you think about 

```ts

interface TuiScrollbarMode {
  enabled: boolean
}

export const TUI_SCROLLBAR_MODE = new InjectionToken<TuiScrollbarMode>(
'custom options for scrollbar,
    {
        factory: () => { enabled: true }, // in the future it will be possible to expand with other properties
    },
);
```

ddubrava: I would have liked to implement a config to control all scrollbars, including local bars. But thought that it’s too complicated to toggle them. However, it seems that if we implement some extra css classes, it’s possible to achieve this

Anyway, make sense, I’ll change the token

splincode: `But thought that it’s too complicated to toggle them`

You can create handy helper like this:
https://github.com/Tinkoff/taiga-ui/blob/main/projects/kit/components/rating/rating-options.ts#L24

projects/core/components/root/root.component.ts
ddubrava: > not sure mutate body is good for disable scrollbar

any ideas? I tried multiple but thought it was the best.

1. I couldn’t apply the mixin in main.less depending on children state
2. I tried to control it in tui-root + ng deep html, but it looked like a hack

ddubrava: it’s not clear to me why we name it ‘tuiScrollbars’ while it returns a boolean. It sounds like it should return an array of scrollbars 🤔

splincode: `it’s not clear to me why we name it ‘tuiScrollbars’`

you can suggest any solution, it was just an example

`any ideas? I tried multiple but thought it was the best.`

you can leave it like that, maybe `@waterplea` offer a better solution

projects/core/tokens/disable-custom-scrollbar.ts
ddubrava: the problem is in how to manage css styles, I mean

waterplea: I'm not sure we need to control `tui-scrollbar`. Why would somebody use it when they don't need a custom scrollbar?

ddubrava: So we don’t need to implement options or this feature at all?

Don’t we use ‘tui-scrollbar’ in some components? So we can’t control it

waterplea: Right, this makes sense. Let's make a token that is used in all scrollbars then. We can have an object there:
```ts
{
  readonly mode: 'native' | 'custom';
}
```
for now. Later we might add `hidden` and `auto-hide` and maybe `size` or something, so having an object is reasonable. The way I see it we can add factory to `TUI_SCROLL_REF` and use `inject(DOCUMENT).scrollingElement` as default value. Then in `tui-scroll-controls` component we can set necessary styles to injected `TUI_SCROLL_REF`. Looks like this would also simplify `scrollbars` getter inside `TuiScrollControlsComponent`.

ddubrava: There are some issues with `inject(DOCUMENT).scrollingElement`. It might return `null` in some cases: 

> When in quirks mode, the scrollingElement attribute returns the HTML body element if it exists and is potentially scrollable , otherwise it returns null. 

Should we handle this in factory or just return as it is?

The second problem is that `scrollingElement` is `Element`, should we cast it to `HTMLElement` in factory or in components?

ddubrava:
@waterplea @splincode 

Hey there, take a look, please!

I left a comment in the last thread, why I didn't use a factory.

The second moment is that `tui-scroll-controls` component logic might be confusing now since it handles toggling of native scrollbar. As a result, we can't hide this component using `*ngIf` otherwise native scrollbar is visible. Any other ideas how can we toggle native scrollbar?

splincode:
@ddubrava sorry for waiting, we will contact you shortly to discuss the details

splincode:
@ddubrava @waterplea What do you think about this, is it worth further development of this PR?

ddubrava:
> @ddubrava @waterplea What do you think about this, is it worth further development of this PR?

I think at some point it can be useful. if you need this feature, I can work on this. But the current logic is a little bit complicated, there are some hacks. So we need to discuss what we have now and come up with the best solution.

splincode:
@ddubrava can we refactor existing code without extra options?

splincode:
I think it will be necessary to call the whole team next year and discuss this

PR #1867, opened 2022-06-06
fix(kit): `InputNumber` prefix is overlapped by value

### LINKED ISSUE #1862: 🐞 - `InputNumber` prefix is overlapped by value

kit

https://stackblitz.com/edit/input-number-prefix-bug?file=src/app/app.component.html

Non-Blocking

### Description

1. Open stackblitz reproduction. It contains these lines:
```html
<tui-input-number
    [formControl]="control"
    prefix="$"
>
    Type a sum
</tui-input-number>
```
inputNumberBug

### Angular version

any 

### Taiga UI version

2.47.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

nsbarsukov:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

**Possible reproduction:**
If any textfield-component projected via `<ng-content />` (and this content projection is attached by some async logic), these lines
https://github.com/Tinkoff/taiga-ui/blob/3c7c53b6a707718c9d4441f447991dd79b4f4bba/projects/core/components/primitive-textfield/value-decoration/value-decoration.component.ts#L34-L38
calculates `offsetWidth` as zero (because the component was initialized but it is not attached to the DOM yet). 

Closes #1862

- [ ] Yes
- [X] No

### DIFF
--- a/projects/demo/src/modules/components/input-number/input-number.template.html
+++ b/projects/demo/src/modules/components/input-number/input-number.template.html
@@ -11,37 +11,50 @@
             A component to input numbers. Control value is also of number type.
         </div>
 
-        <p i18n>
-            Component can be used to input money if you choose a currency symbol
-            as a postfix or prefix (see
-            <code>Currency</code>
-            pipe samples).
-        </p>
-
-        <p i18n>
-            There are also other components to input numbers:
-            <a
-                tuiLink
-                routerLink="/components/input-count"
-            >
-                <code>InputCount</code>
-            </a>
-            for integers,
-            <a
-                tuiLink
-                routerLink="/components/slider"
-            >
-                <code>Slider</code>
-            </a>
-            and
-            <a
-                tuiLink
-                routerLink="/components/input-slider"
-            >
-                <code>InputSlider</code>
-            </a>
-            for inputting with a slider
-        </p>
+        <section i18n>
+            <h3>There are also other components to input numbers:</h3>
+            <ul class="tui-list tui-list_small">
+                <li class="tui-list__item">
+                    <a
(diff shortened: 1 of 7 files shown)


splincode (approved):

projects/demo/src/modules/components/input-number/input-number.template.html
```diff
@@ -11,37 +11,50 @@
...
-            <a
-                tuiLink
-                routerLink="/components/input-slider"
-            >
-                <code>InputSlider</code>
-            </a>
-            for inputting with a slider
-        </p>
+        <section i18n>
+            <h3>There are also other components to input numbers:</h3>
```
vladimirpotekhin, line 15: btw I remember there was "SeeAlso" component in 1.x taiga that showed related components. Do you know why we don't use it now? 🤔

vladimirpotekhin (approved):

projects/demo/src/modules/components/input-number/input-number.template.html
nsbarsukov: I dont even know about this feature :)

I have investigated this question.
> Do you know why we don't use it now? 

We use it. But it does not work because of the bug (i will fix it in a separate PR).

vladimirpotekhin: Oh got it, thanks) So I think it's not so important, we live without that feature long time))

PR #1998, opened 2022-06-24
fix(cdk): correct selection from textfield in firefox

splincode:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

### DIFF
--- a/projects/cdk/utils/dom/get-selected-text.ts
+++ b/projects/cdk/utils/dom/get-selected-text.ts
@@ -1,0 +1,18 @@
+import {tuiIsFormFieldElement} from './is-form-field';
+
+/**
+ * @description:
+ * cross browser way to get selected text
+ *
+ * History:
+ * BUG - window.getSelection() fails when text selected in a form field
+ * https://bugzilla.mozilla.org/show_bug.cgi?id=85686
+ */
+export function tuiGetSelectedText({getSelection, document}: Window): string | null {
+    return tuiIsFormFieldElement(document.activeElement)
+        ? document.activeElement.value.slice(
+              document.activeElement.selectionStart || 0,
+              document.activeElement.selectionEnd || 0,
+          )
+        : getSelection()?.toString() || null;
+}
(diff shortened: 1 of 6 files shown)


vladimirpotekhin (approved):

projects/cdk/utils/dom/get-selection-text.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * @description:
+ * cross browser way to get selected text
+ *
+ * History:
+ * BUG - window.getSelection() fails when text selected in a form field
+ * https://bugzilla.mozilla.org/show_bug.cgi?id=85686
+ */
+export function tuiGetSelectionText(windowRef: Window, target?: Target): string | null {
```
waterplea, line 11: What about copying not from an input field? There's got to be target in that event as well.

projects/cdk/utils/dom/get-selection-text.ts
```diff
@@ -0,0 +1,23 @@
...
+    const selection = windowRef.getSelection();
```
waterplea, line 12: How about we pass `document` instead so we only have one argument?
https://developer.mozilla.org/en-US/docs/Web/API/Document/getSelection
We can then use `activeElement` to test if we are in an input field:
```js
document.activeElement.value.substring(
  document.activeElement.selectionStart,
  document.activeElement.selectionEnd
)
```

splincode: fixed

projects/cdk/utils/dom/get-selection-text.ts
splincode: fixed

projects/cdk/utils/dom/get-selected-text.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * History:
+ * BUG - window.getSelection() fails when text selected in a form field
+ * https://bugzilla.mozilla.org/show_bug.cgi?id=85686
+ */
+export function tuiGetSelectedText(windowRef: Window): string | null {
+    const selection = windowRef.getSelection();
+    const element = windowRef.document.activeElement as
+        | HTMLInputElement
+        | HTMLTextAreaElement
```
nsbarsukov, line 13: let use type guard
```ts
function isFormFieldElement(
    element: Target,
): element is HTMLInputElement | HTMLTextAreaElement {
    return (
        !!element &&
        'value' in element &&
        'selectionStart' in element &&
        'selectionEnd' in element
    );
}
```

splincode: fixed

nsbarsukov (approved):

PR #2188, opened 2022-07-23
fix(core): add more customization for `font-size` in buttons

### LINKED ISSUE #2172: 🚀 - Why does Taiga uses less vars for font size on data-size instead of the defined --tui-font-text-x ?

core

### Description

I'm trying to override the overal theme of TaigaUI and doing so I redefined all the --tui-font-text vars like this

``` 
...
--tui-font-text-m: normal 1.25rem var(--tui-font-text);
... etc
```

Then the S button `<button tuiButton size='s'></button>`  uses `--tui-font-text-s` but the M button `<button tuiButton size='m'></button>` doesn't use `--tui-font-text-m` but uses the less var `@font-size-large`. 
Why is that ? I am not able to override this value.

**Resulting in compiled**
Inkedimage (4)_LI

**Code responsible for this : [button.style.less](https://github.com/Tinkoff/taiga-ui/blob/main/projects/core/components/button/button.style.less)**

image

splincode:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #2172

### DIFF
--- a/projects/core/components/button/button.style.less
+++ b/projects/core/components/button/button.style.less
@@ -6,8 +6,6 @@
 @padding-l: 2.25rem;
 @padding-xl: 2.5rem;
 @icon-margin: 0.5rem;
-@font-size-large: 0.9375rem;
-@font-size-extra-large: 1.0625rem;
 
 :host {
     .clearbtn();
@@ -19,7 +17,6 @@
     border-radius: var(--tui-radius-m);
     user-select: none;
     text-decoration: none;
-    font-weight: bold;
     outline: none;
 
     &[disabled] {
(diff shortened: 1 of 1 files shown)


waterplea (approved):

waterplea:
Don't forget to check the proprietary theme

projects/core/components/button/button.style.less
```diff
@@ -81,7 +79,7 @@
 
     &[data-size='xl'] {
         height: 3.75rem;
-        font-size: @font-size-large;
+        font-size: var(--tui-font-size-m);
```
waterplea, line 82: Why not just use existing font variables?

splincode: fixed

nsbarsukov (approved):

PR #2210, opened 2022-07-26
feat(addon-table): `TableFilters` add new module

### LINKED ISSUE #2198: 🚀 - Filter in table.

kit

### Description

User selects one or more columns of text or date(and datetime) for filtering in the table settings in UI.  
The columns to filter can be selected before being used in code. In this case, we don't show the filter button in the UI if no columns are selected for the filter.

**Requirements**
1. For all selected text columns we show multiselect between unique existing values. 
- Null-values also if they exist. 
- Values are sorted by alphabetically. 
- If the filter isn't set, this works the same as if all values in the list were selected
2. For date(datetime) columns we use calendar for select date interval. The start date is always less than or equal to the end date.
3. Columns of another type work like "text". 
4. The title of the multiselect or calendar is equal to the name of the column.
5. Filter fields are arranged in the order of the corresponding columns. Multiple lines if required. 

**How does the filter work?**

1. When user sets a filter in a field, it is applied with a standard delay.
2. The field displays the selected values. 
3. User cans clear any filter field 
4. User can set several filters and use the search at the same time: the columns are between each other and the search is applied through "AND"

waterplea:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #2198

- [ ] Yes
- [ ] No


projects/addon-table/interfaces/check.ts
```diff
@@ -0,0 +1,3 @@
+export interface TuiCheck {
```
splincode, line 1: `TuiCheck<I = any, V = any>`

`check(item: I, value: V): boolean;`

splincode (approved):

projects/addon-table/directives/table-filters/table-filters.directive.ts
```diff
@@ -0,0 +1,40 @@
+import {Directive} from '@angular/core';
+import {identity, merge, Observable, ReplaySubject} from 'rxjs';
+import {map, startWith, switchMap} from 'rxjs/operators';
+
+import {TuiTableFilterDirective} from './table-filter.directive';
```
nsbarsukov, line 5: I am not sure but it can produce problem in Angular 11.

You import from `table-filter.directive.ts` here. And inside `table-filter.directive.ts` you import from `table-filters.directive.ts`.
Possible problem: https://angular.io/errors/NG3003

Could you please locally merge your branch with `v3.x`-branch and check it ?

P.S. There is chance that no error will produce (because you import entities only for typings).

projects/demo/src/modules/tables/table-filters/examples/import/insert-template.md
```diff
@@ -0,0 +1,49 @@
...
+        tuiTh
+      >
+        Name
+      </th>
+      <th
+        *tuiHead="'balance'"
+        tuiTh
+      >
+        <tui-input-number
+          tuiTableFilter="balance"
```
nsbarsukov, line 20: VERY small nit:

We can inject `TuiHeadDirective` (with `@Optional()`) inside `tuiTableFilter` and get its name.
It can allow to write in this way)) 
```html
 <th
    *tuiHead="'balance'"
    tuiTh
>
    <tui-input-number
        tuiTableFilter
        [tuiGenericFilter]="filter"
        [formControl]="balance"
    >
          Minimal balance
    </tui-input-number>
</th>
```

projects/addon-table/directives/table-filters/table-filters.directive.ts
waterplea: I'll just add interface :)

nsbarsukov (approved):

vladimirpotekhin (approved):

PR #2248, opened 2022-07-29
fix(cdk): removed `overflow` property as not animatable

### LINKED ISSUE #2247: 🐞 - The following provided properties are not animatable: overflow

core, kit, cdk

_No response_

Non-Blocking

### Description

image

Warnings in browser console. There is a selection on the page where the layout of the option template is larger than the block size.

### Angular version

14.0.0

### Taiga UI version

2.54.1

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #2247

- [ ] Yes
- [x] No

### DIFF
--- a/projects/cdk/constants/parent-animation.ts
+++ b/projects/cdk/constants/parent-animation.ts
@@ -1,8 +1,5 @@
-import {animateChild, query, style, transition, trigger} from '@angular/animations';
+import {animateChild, query, transition, trigger} from '@angular/animations';
 
 export const TUI_PARENT_ANIMATION = trigger(`tuiParentAnimation`, [
-    transition(`* => void`, [
-        style({overflow: `hidden`}),
-        query(`@*`, [animateChild()], {optional: true}),
-    ]),
+    transition(`:leave`, [query(`@*`, [animateChild()], {optional: true})]),
 ]);


projects/cdk/constants/parent-animation.ts
```diff
@@ -1,8 +1,5 @@
-import {animateChild, query, style, transition, trigger} from '@angular/animations';
+import {animateChild, query, transition, trigger} from '@angular/animations';
 
 export const TUI_PARENT_ANIMATION = trigger(`tuiParentAnimation`, [
-    transition(`* => void`, [
```
waterplea, line 4: It's actually is animatable and is necessary, I believe in some cases. Do not merge this until I investigate.

waterplea (changes requested):

projects/cdk/constants/parent-animation.ts
splincode: But the overflow property can't be animated with CSS :((
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_animated_properties

waterplea: It's not on the list, but it works, at least in Chrome:
https://codepen.io/waterplea/pen/PoRJEvP

I believe we need it so scrollbars don't appear in dialogs during animation, or something like that

PR #2260, opened 2022-08-01
feat(i18n): add switcher for language

splincode:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

### DIFF
--- a/projects/addon-doc/src/components/language-switcher/language-switcher.component.ts
+++ b/projects/addon-doc/src/components/language-switcher/language-switcher.component.ts
@@ -1,0 +1,47 @@
+import {ChangeDetectionStrategy, Component, Inject} from '@angular/core';
+import {WINDOW} from '@ng-web-apis/common';
+import {TuiLanguageName, TuiLanguageSwitcher} from '@taiga-ui/i18n';
+
+@Component({
+    selector: `tui-language-switcher`,
+    templateUrl: `./language-switcher.component.html`,
+    styleUrls: [`./language-switcher.component.less`],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TuiLanguageSwitcherComponent {
+    language: TuiLanguageName = this.switcher.lang;
+
+    flags: Map<TuiLanguageName, string> = new Map([
+        [`chinese`, `🇨🇳 chinese`],
+        [`dutch`, `🇳🇱 dutch`],
+        [`english`, `🇬🇧 english`],
+        [`french`, `🇫🇷 french`],
+        [`german`, `🇫🇷 german`],
+        [`german`, `🇩🇪 german`],
+        [`italian`, `🇩🇪 italian`],
+        [`polish`, `🇵🇱 polish`],
+        [`portuguese`, `🇵🇹 portuguese`],
+        [`russian`, `🇷🇺 russian`],
+        [`spanish`, `🇪🇸 spanish`],
+        [`turkish`, `🇹🇷 turkish`],
+        [`ukrainian`, `🇺🇦 ukrainian`],
+        [`vietnamese`, `🇻🇳 vietnamese`],
+    ]);
+
+    languages: TuiLanguageName[] = Array.from(this.flags.keys());
+
+    constructor(
+        @Inject(WINDOW) private readonly windowRef: Window,
+        @Inject(TuiLanguageSwitcher) private readonly switcher: TuiLanguageSwitcher,
+    ) {}
+
+    set(lang: TuiLanguageName): void {
+        this.switcher.setLang(lang);
+        this.windowRef.location.reload();
+    }
+
+    clear(): void {
+        this.switcher.removeLang();
+        this.windowRef.location.reload();
+    }
+}
--- a/projects/addon-doc/src/components/main/main.template.html
+++ b/projects/addon-doc/src/components/main/main.template.html
@@ -17,6 +17,7 @@
                 (ngModelChange)="onMode($event)"
             ></tui-toggle>
         </label>
+        <tui-language-switcher></tui-language-switcher>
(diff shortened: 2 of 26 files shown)


projects/cdk/utils/nodejs/normalize-common-js-import.ts
```diff
@@ -0,0 +1,6 @@
+export async function tuiNormalizeCommonJSImport<T>(
+    importPromise: Promise<T>,
+): Promise<T> {
+    // CommonJS's `module.exports` is wrapped as `default` in ESModule.
+    return importPromise.then((m: any) => (m.default || m) as T);
```
nsbarsukov, line 5: Let's also use this new util here
https://github.com/Tinkoff/taiga-ui/blob/83a562e0bff72ad72f601a199e008d9696a7d287/projects/addon-doc/src/utils/raw-load.ts#L7-L9

projects/i18n/switch/language.switcher.ts
```diff
@@ -0,0 +1,46 @@
...
+import {TuiLanguage, TuiLanguageName} from '@taiga-ui/i18n/interfaces';
+import {TUI_DEFAULT_LANGUAGE} from '@taiga-ui/i18n/tools';
+import {BehaviorSubject, from, Observable} from 'rxjs';
+import {map} from 'rxjs/operators';
+
+import {TUI_LANG_DEFAULT_KEY, TUI_LANG_STORAGE_KEY} from './options';
+
+// @dynamic
+@Injectable({providedIn: `root`})
+export class TuiLanguageSwitcher extends BehaviorSubject<Observable<TuiLanguage>> {
```
nsbarsukov, line 13: You extends `BehaviorSubject` but dont implement its behaviour.

It is good idea to use `BehaviorSubject` but let's do it in the same way as here (no public method except `next`):
https://github.com/Tinkoff/taiga-ui/blob/83a562e0bff72ad72f601a199e008d9696a7d287/projects/demo/src/modules/app/theme.service.ts#L7-L19

projects/i18n/tools/language.ts
```diff
@@ -1,11 +1,18 @@
-import {InjectionToken} from '@angular/core';
-import {Language} from '@taiga-ui/i18n/interfaces';
+import {inject, InjectionToken} from '@angular/core';
+import {TuiLanguage} from '@taiga-ui/i18n/interfaces';
 import {TUI_ENGLISH_LANGUAGE} from '@taiga-ui/i18n/languages';
 import {Observable, of} from 'rxjs';
 
-export const TUI_LANGUAGE = new InjectionToken<Observable<Language>>(
-    `Language for Taiga UI libraries i18n`,
+export const TUI_DEFAULT_LANGUAGE = new InjectionToken<Observable<TuiLanguage>>(
```
nsbarsukov, line 6: Default token for default value of another token ? Do we really need it ?

projects/demo/src/modules/customization/i18n/i18n.component.ts
```diff
@@ -16,4 +17,21 @@ export class I18nComponent {
             default: string;
         }>,
     ).pipe(map(readme => readme.default.split(`Supported languages:`)[1]));
+
+    exampleModule = import(`!!raw-loader!./app.module.md`);
+
+    example1: TuiDocExample = {
+        'language-switcher.component.html': import(
+            `!!raw-loader!../../../../../addon-doc/src/components/language-switcher/language-switcher.component.html`
```
nsbarsukov, line 25: Can we use absolute paths here ?

projects/cdk/utils/nodejs/normalize-common-js-import.ts
splincode: But we are return `*.default` every time for this

projects/demo/src/modules/customization/i18n/i18n.component.ts
splincode: I'm not sure how

projects/i18n/tools/language.ts
splincode: Yes, because we are override `TUI_LANGUAGE` by service and return fallback `TUI_DEFAULT_LANGUAGE` when storage is empty

projects/addon-doc/src/components/language-switcher/language-switcher.component.ts
```diff
@@ -0,0 +1,47 @@
...
+    languages: TuiLanguageName[] = Array.from(this.flags.keys());
+
+    constructor(
+        @Inject(WINDOW) private readonly windowRef: Window,
+        @Inject(TuiLanguageSwitcher) private readonly switcher: TuiLanguageSwitcher,
+    ) {}
+
+    set(lang: TuiLanguageName): void {
+        this.switcher.setLang(lang);
+        this.windowRef.location.reload();
```
vladimirpotekhin, line 40: Can we make a language change on the fly without a reload? It seems that for this we did "i18n" on observables

projects/addon-doc/src/components/main/main.template.html
```diff
@@ -17,6 +17,7 @@
                 (ngModelChange)="onMode($event)"
             ></tui-toggle>
         </label>
+        <tui-language-switcher></tui-language-switcher>
```
vladimirpotekhin, line 20: Now it looks like the component switches the language of the entire site (including navigation etc) but it doesn't. Maybe move the switcher closer to components?

splincode: I'm not sure 
but yes, part of the functionality is translated ))

image

splincode: image

PR #2311, opened 2022-08-08
feat(styles): add `@taiga-ui/styles`

splincode:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update


angular.json
```diff
@@ -815,6 +815,28 @@
...
+            "root": "projects/styles",
+            "sourceRoot": "projects/styles",
+            "projectType": "library",
+            "prefix": "tui",
+            "architect": {
+                "build": {
+                    "builder": "@angular-devkit/build-ng-packagr:build",
+                    "outputs": ["dist/styles"],
+                    "options": {
+                        "tsConfig": "tsconfig.build.json",
```
nsbarsukov, line 828: Do we really need build process for this package ?

Or we can just copy files to `dist/`-directory ?

waterplea:
Mixins and such should stay in `core` package, we want to move only the global styles, that are optional, to be able to delegate its support in future.

vladimirpotekhin:
this way we will have to support the same styles in two places in 2.x. How about we make this change in 3.0 and remove the old styles from the core?

PR #2487, opened 2022-08-27
fix(kit): change min/max from Infinity to min/max integers

zarghamkhandev:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

Closes # https://github.com/Tinkoff/taiga-ui/issues/2421

- [ ] Yes
- [x] No

### DIFF
--- a/projects/kit/components/input-number/input-number.component.ts
+++ b/projects/kit/components/input-number/input-number.component.ts
@@ -70,11 +70,11 @@ export class TuiInputNumberComponent
 
     @Input()
     @tuiDefaultProp()
-    min = -Infinity;
+    min = Number.MIN_SAFE_INTEGER;
 
     @Input()
     @tuiDefaultProp()
-    max = Infinity;
+    max = Number.MAX_SAFE_INTEGER;
 
     @Input()
     @tuiDefaultProp()


zarghamkhandev:
Do we also want to limit the input to the component ? So that it does not exceed MAX/MIN safer integers ? I only changed the default values. @waterplea @MarsiBarsi

waterplea:
> Do we also want to limit the input to the component ? So that it does not exceed MAX/MIN safer integers ? I only changed the default values. @waterplea @MarsiBarsi 

I don't think that extra safety is necessary.

projects/kit/components/input-number/input-number.component.ts
```diff
@@ -70,11 +70,11 @@ export class TuiInputNumberComponent
 
     @Input()
     @tuiDefaultProp()
-    min = -Infinity;
+    min = Number.MIN_SAFE_INTEGER;
```
splincode, line 73: What do you think, will you be able to add a unit test or a demo example with set max value inside for check on e2e?

splincode (approved):

projects/kit/components/input-number/input-number.component.ts
zarghamkhandev: sure, will do that.

waterplea (approved):

PR #2490, opened 2022-08-28
feat: add tokens for customizing different dialogs that extend `AbstractTuiDialogService`

festimprebreza:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

Does not exist.

Closes https://github.com/Tinkoff/taiga-ui/issues/2267

## What is the new behavior?
A token is provided so that the user can configure all dialogs, without the need to provide the same configuration each time a new dialog is opened.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/addon-mobile/components/mobile-dialog/mobile-dialog.tokens.ts
+++ b/projects/addon-mobile/components/mobile-dialog/mobile-dialog.tokens.ts
@@ -1,0 +1,27 @@
+import {InjectionToken, ValueProvider} from '@angular/core';
+
+import {TuiMobileDialogOptions} from './mobile-dialog-options';
+
+type TuiMobileDialogDefaultOptions = Omit<TuiMobileDialogOptions<unknown>, 'data'>;
+
+export const TUI_MOBILE_DIALOG_DEFAULT_OPTIONS: TuiMobileDialogDefaultOptions = {
+    label: ``,
+    actions: [`OK`],
+};
+
+export const TUI_MOBILE_DIALOG_OPTIONS =
+    new InjectionToken<TuiMobileDialogDefaultOptions>(
+        `[TUI_MOBILE_DIALOG_OPTIONS]: Default parameters for mobile dialog component`,
+        {
+            factory: () => TUI_MOBILE_DIALOG_DEFAULT_OPTIONS,
+        },
+    );
+
+export function tuiMobileDialogOptionsProvider(
+    options: Partial<TuiMobileDialogDefaultOptions>,
+): ValueProvider {
+    return {
+        provide: TUI_MOBILE_DIALOG_OPTIONS,
+        useValue: {...TUI_MOBILE_DIALOG_DEFAULT_OPTIONS, ...options},
+    };
+}
--- a/projects/core/components/dialog/dialog.service.ts
+++ b/projects/core/components/dialog/dialog.service.ts
@@ -1,24 +1,20 @@
-import {Injectable} from '@angular/core';
+import {inject, Injectable} from '@angular/core';
 import {AbstractTuiDialogService} from '@taiga-ui/cdk';
-import {TuiDialogOptions} from '@taiga-ui/core/interfaces';
+import type {TuiDialogOptions} from '@taiga-ui/core/interfaces';
 import {PolymorpheusComponent} from '@tinkoff/ng-polymorpheus';
 
 import {TuiDialogComponent} from './dialog.component';
+import {TUI_DIALOG_OPTIONS} from './dialog.tokens';
 
 const DIALOG = new PolymorpheusComponent(TuiDialogComponent);
-const DEFAULT_OPTIONS = {
-    size: `m`,
-    required: false,
-    closeable: true,
-    dismissible: true,
-    label: ``,
-    header: ``,
-} as const;
 
 @Injectable({
     providedIn: `root`,
 })
 export class TuiDialogService extends AbstractTuiDialogService<TuiDialogOptions<any>> {
     protected readonly component = DIALOG;
-    protected readonly defaultOptions: TuiDialogOptions<any> = DEFAULT_OPTIONS as any;
+    protected readonly defaultOptions: TuiDialogOptions<any> = {
+        ...inject(TUI_DIALOG_OPTIONS),
+        data: undefined,
+    };
 }
--- a/projects/core/components/dialog/dialog.tokens.ts
+++ b/projects/core/components/dialog/dialog.tokens.ts
@@ -1,0 +1,29 @@
+import {InjectionToken, ValueProvider} from '@angular/core';
+import type {TuiDialogOptions} from '@taiga-ui/core/interfaces';
+
+type TuiDialogDefaultOptions = Omit<TuiDialogOptions<unknown>, 'data'>;
+
+export const TUI_DIALOG_DEFAULT_OPTIONS: TuiDialogDefaultOptions = {
+    size: `m`,
+    required: false,
+    closeable: true,
+    dismissible: true,
+    label: ``,
+    header: ``,
+};
+
+export const TUI_DIALOG_OPTIONS = new InjectionToken<TuiDialogDefaultOptions>(
+    `[TUI_DIALOG_OPTIONS]: Default parameters for dialog component`,
+    {
+        factory: () => TUI_DIALOG_DEFAULT_OPTIONS,
+    },
+);
+
+export function tuiDialogOptionsProvider(
+    options: Partial<TuiDialogDefaultOptions>,
+): ValueProvider {
+    return {
+        provide: TUI_DIALOG_OPTIONS,
+        useValue: {...TUI_DIALOG_DEFAULT_OPTIONS, ...options},
+    };
+}
--- a/projects/core/components/dialog/test/dialog.component.spec.ts
+++ b/projects/core/components/dialog/test/dialog.component.spec.ts
@@ -1,0 +1,58 @@
+import {Component, DebugElement} from '@angular/core';
+import {ComponentFixture, TestBed} from '@angular/core/testing';
+import {NoopAnimationsModule} from '@angular/platform-browser/animations';
+import {TuiIdService} from '@taiga-ui/cdk';
+import {configureTestSuite, TuiPageObject} from '@taiga-ui/testing';
+
+import {TuiRootModule} from '../../root/root.module';
+import {TuiDialogModule} from '../dialog.module';
+import {TuiDialogService} from '../dialog.service';
+import {tuiDialogOptionsProvider} from '../dialog.tokens';
+
+describe(`Dialog with TUI_DIALOG_OPTIONS`, () => {
+    @Component({
+        template: `
+            <tui-root></tui-root>
+        `,
+    })
+    class TestComponent {}
+
+    const closeable = false;
+
+    let fixture: ComponentFixture<TestComponent>;
+    let tuiDialogService: TuiDialogService;
+    let pageObject: TuiPageObject<TestComponent>;
+
+    function getCloseButton(): DebugElement {
+        return pageObject.getByAutomationId(`tui-dialog__close`)!;
+    }
+
+    configureTestSuite(() => {
+        TestBed.configureTestingModule({
+            imports: [NoopAnimationsModule, TuiRootModule, TuiDialogModule],
+            declarations: [TestComponent],
+            providers: [
+                tuiDialogOptionsProvider({closeable}),
+                {
+                    provide: TuiDialogService,
+                    useFactory: () => new TuiDialogService(TestBed.inject(TuiIdService)),
+                },
+            ],
+        });
+    });
+
+    beforeEach(() => {
+        fixture = TestBed.createComponent(TestComponent);
+        tuiDialogService = TestBed.inject(TuiDialogService);
+        pageObject = new TuiPageObject(fixture);
+    });
+
+    describe(`close button`, () => {
+        it(`when closeable = false is absent`, () => {
+            tuiDialogService.open(`Test`).subscribe();
+            fixture.detectChanges();
+
+            expect(getCloseButton()).toBeNull();
+        });
+    });
+});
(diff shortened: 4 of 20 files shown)


festimprebreza:
I added that constructor to `TuiDialogServcie` because otherwise the test would fail. Without a constructor, Angular could not find `TuiIdService`, dependency which is needed for `TuiDialogService` to work correctly. 

Injecting `TuiDialogService` from the `TestBed` method, would give `undefined` instead of a `TuiIdService` instance. Manually creating a `TuiDialogService` instance does not work in this case either, because its parent constructor is protected. I debugged it a little, and saw that Angular could not find necessary the dependencies for `TuiDialogService`, because it's generated constructor would not match the patterns for "a child class that calls the parent constructor". Might be a Karma issue, did not go any further.

You can test for yourself here: https://github.com/festimprebreza/taiga-ui/tree/dialog-options-using-inject-method

projects/core/components/dialog/dialog.service.ts
```diff
@@ -1,24 +1,24 @@
-import {Injectable} from '@angular/core';
-import {AbstractTuiDialogService} from '@taiga-ui/cdk';
+import {Injectable, Inject, inject} from '@angular/core';
```
splincode, line 1: @festimprebreza hello, please run autofix to sort these imports!

projects/core/components/dialog/dialog.service.ts
```diff
@@ -1,24 +1,24 @@
...
 })
 export class TuiDialogService extends AbstractTuiDialogService<TuiDialogOptions<any>> {
     protected readonly component = DIALOG;
-    protected readonly defaultOptions: TuiDialogOptions<any> = DEFAULT_OPTIONS as any;
+    protected readonly defaultOptions: TuiDialogOptions<any> = {
+        ...inject(TUI_DIALOG_OPTIONS),
+        data: undefined,
+    };
+
+    constructor(@Inject(TuiIdService) idService: TuiIdService) {
```
waterplea, line 21: Is this constructor required? 🤔

waterplea (commented):
Please don't skip precommit hooks :) All those format issues are autocorrected inside those

projects/core/components/dialog/dialog.service.ts
festimprebreza: Can you please check the comment I made above

festimprebreza:
Sorry, no idea why hooks did not kick in automatically, so I ran the commands manually now

projects/core/components/dialog/dialog.service.ts
```diff
@@ -1,24 +1,24 @@
...
-    closeable: true,
-    dismissible: true,
-    label: ``,
-    header: ``,
-} as const;
 
 @Injectable({
     providedIn: `root`,
 })
 export class TuiDialogService extends AbstractTuiDialogService<TuiDialogOptions<any>> {
```
splincode, line 14: it turns out this token will work only for this class?

image

splincode: I mean we have to do the same everywhere?

image

splincode: what do you think if we can just override some properties if we need it
image

splincode: without additional changes in inherited classes

projects/core/components/dialog/dialog.service.ts
waterplea: Got it, thank you.

projects/core/components/dialog/dialog.service.ts
festimprebreza: While it is a very good idea for reducing code duplication, I don't think having only one token is feasible. There would be the issue of conflicting property names between different dialog options?

waterplea: > While it is a very good idea for reducing code duplication, I don't think having only one token is feasible. There would be the issue of conflicting property names between different dialog options?

That's right. This token is for exactly this dialog implementation.

projects/core/components/dialog/dialog.service.ts
```diff
@@ -1,24 +1,24 @@
...
     protected readonly component = DIALOG;
-    protected readonly defaultOptions: TuiDialogOptions<any> = DEFAULT_OPTIONS as any;
+    protected readonly defaultOptions: TuiDialogOptions<any> = {
+        ...inject(TUI_DIALOG_OPTIONS),
+        data: undefined,
+    };
+
+    constructor(@Inject(TuiIdService) idService: TuiIdService) {
```
splincode, line 21: please, add `@note: explain why` before merge

splincode (approved):

splincode:
@festimprebreza can we implement options for other dialogs (mobile, confirm, etc) too?

festimprebreza:
Sure, I'll do it, would that be a different PR or just add changes to this one? I'll have to think of a general solution though, I'll discuss it here after I draft some ideas

vladimirpotekhin (approved):

projects/core/tokens/dialog-options.ts
```diff
@@ -0,0 +1,30 @@
...
+export const TUI_DIALOG_DEFAULT_OPTIONS: TuiDialogDefaultOptions = {
+    size: `m`,
+    required: false,
+    closeable: true,
+    dismissible: true,
+    label: ``,
+    header: ``,
+};
+
+export const TUI_DIALOG_OPTIONS = new InjectionToken<TuiDialogDefaultOptions>(
```
nsbarsukov, line 16: Most of the tokens `TUI_SOMETHING_OPTIONS` we place nearby the entity which this token customises.

Could you move this file `projects/core/tokens/dialog-options.ts` => `projects/core/components/dialog/dialog-options.ts`, please.

nsbarsukov (changes requested):

projects/core/tokens/dialog-options.ts
festimprebreza: Actually, I put it there because a similar token `TUI_NOTIFICATION_OPTIONS` was in that directory too

splincode: @nsbarsukov

projects/core/components/dialog/dialog.service.ts
```diff
@@ -1,24 +1,29 @@
...
     providedIn: `root`,
 })
 export class TuiDialogService extends AbstractTuiDialogService<TuiDialogOptions<any>> {
     protected readonly component = DIALOG;
-    protected readonly defaultOptions: TuiDialogOptions<any> = DEFAULT_OPTIONS as any;
+    protected readonly defaultOptions: TuiDialogOptions<any> = {
+        ...inject(TUI_DIALOG_OPTIONS),
+        data: undefined,
+    };
+
```
festimprebreza, line 20: @splincode so, since having a token for each dialog is non-negotiable, the only thing that could help a little is a helper function that would look something like this: 

image

but even this does not help too much, as doing this manually is rather simple and trivial anyways. So, what do you think? Should I use this function, or just go the simple way?

splincode: just go the simple way))

projects/core/tokens/dialog-options.ts
splincode: @festimprebreza I think it's better to move next to the files, as the `tokens ` subpackage will be removed in the future

projects/core/components/dialog/test/dialog.component.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+    configureTestSuite(() => {
+        TestBed.configureTestingModule({
+            imports: [NoopAnimationsModule, TuiRootModule, TuiDialogModule],
+            declarations: [TestComponent],
+            providers: [
+                tuiDialogOptionsProvider({closeable}),
+                {
+                    provide: TuiDialogService,
+                    useFactory: () => new TuiDialogService(TestBed.inject(TuiIdService)),
+                },
```
festimprebreza, line 39: Two things that need your attention:
1. Since the abstract dialog service's constructor is not protected anymore, I created the dialog instances manually on the tests, so then I removed the constructor from `dialog.service.ts`. What do you think? This way, or just add constructors to each dialog implementation?
2. `mobile-dialog.service.ts` did not need this work around. I investigated a bit, and saw that the issue I mentioned before happens in testing because of code coverage functionality. Now this does not happen for Mobile Dialog, because `addon-mobile` package could be misconfigured. In `project.json` if you remove the `/src` from `sourceRoot`, the test for Mobile Dialog fails too. So at the moment there is no code coverage report for this package, I don't know if that is intentional or not. Please tell me if I need to modify anything

splincode (approved):

projects/kit/components/push/test/push.component.spec.ts
```diff
@@ -0,0 +1,60 @@
+import {Component, DebugElement} from '@angular/core';
+import {ComponentFixture, TestBed} from '@angular/core/testing';
+import {NoopAnimationsModule} from '@angular/platform-browser/animations';
+import {TuiIdService} from '@taiga-ui/cdk';
+import {TuiRootModule} from '@taiga-ui/core/components/root/root.module';
```
nsbarsukov, line 5:
```suggestion
import {TuiRootModule} from '@taiga-ui/core';
```

nsbarsukov (approved):

projects/core/components/dialog/dialog.tokens.ts
```diff
@@ -0,0 +1,29 @@
...
+    size: `m`,
+    required: false,
+    closeable: true,
+    dismissible: true,
+    label: ``,
+    header: ``,
+};
+
+export const TUI_DIALOG_OPTIONS = new InjectionToken<TuiDialogDefaultOptions>(
+    `Default parameters for dialog component`,
```
nsbarsukov, line 16: We have new eslint rule for injection tokens's description.

Rebase is required :(

splincode:
@festimprebreza please squash all your commits and rebase
image

splincode:
@festimprebreza @waterplea hello, any news?

festimprebreza:
Can you please see why `E2E tests/ demo (pull_request)` check failed? I analyzed it a little and I think it has something to do with your architecture, can you maybe re-run them?

splincode:
@festimprebreza please, rebase with main branch

projects/core/components/dialog/dialog.service.ts
```diff
@@ -1,24 +1,20 @@
-import {Injectable} from '@angular/core';
+import {inject, Injectable} from '@angular/core';
 import {AbstractTuiDialogService} from '@taiga-ui/cdk';
-import {TuiDialogOptions} from '@taiga-ui/core/interfaces';
+import type {TuiDialogOptions} from '@taiga-ui/core';
```
waterplea, line 3: Looks like this should still be from the `interfaces` entry point.

projects/core/components/dialog/dialog.tokens.ts
```diff
@@ -0,0 +1,29 @@
+import {InjectionToken, ValueProvider} from '@angular/core';
+import type {TuiDialogOptions} from '@taiga-ui/core';
```
waterplea, line 2: Same here

projects/addon-mobile/components/mobile-dialog/mobile-dialog.tokens.ts
```diff
@@ -0,0 +1,27 @@
+import {InjectionToken, ValueProvider} from '@angular/core';
+
+import {TuiMobileDialogOptions} from './mobile-dialog-options';
+
+export type TuiMobileDialogDefaultOptions = Omit<TuiMobileDialogOptions<unknown>, 'data'>;
```
waterplea, line 5: I'm not sure we want to export such types. Maybe keep them local? That `data` gives all sorts of headaches for types :D I just don't want to have a bunch of confusing interfaces.

projects/core/components/dialog/dialog.service.ts
splincode: image

festimprebreza: This import apparently was auto-fixed by this rule: @taiga-ui/no-deep-imports. 
So right now, I can't commit this fix. If we leave it like this, there is the `dependency cycle` issue. So should I just `eslint-ignore` the `no-deep-imports` for now?

splincode: In theory it should work
`import {TuiDialogOptions} from '@taiga-ui/core/interfaces';`

otherwise it's a bug in our eslint plugin

festimprebreza: @splincode have a look
image

festimprebreza: However, same thing happens for other files that are already commited. I checked with this other file that was edited today in this PR: https://github.com/Tinkoff/taiga-ui/pull/2955/files and below you can see the same error. However in the PR workflow checks it does not show this error at all

image

festimprebreza: Turns out there is actually a bug in the `no-deep-imports.js` plugin. When searching for `currentProject` token, you search in the file absolute path. So, if by any chance, someone like me has the `taiga-ui` library inside a parent directory named `projects`, the script never finds the correct library name :smile: 

You can see in the pictures above that the library was inside `/home/festimprebreza/projects/...`

splincode: could you fix this bug?

splincode: @nsbarsukov fyi

festimprebreza: @splincode sure :+1:

festimprebreza: @waterplea can you please review the changes. I rebased and squashed directly since they were small changes

waterplea (approved):

waterplea:
Huge thanks for all this work! 💪

PR #2511, opened 2022-08-29
feat(testing): card component harness test

zarghamkhandev:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

Partially implements https://github.com/Tinkoff/taiga-ui/issues/476

## What is the new behavior?
Introduces ComponentHarness for card component.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/addon-commerce/components/card/test/card.component.spec.ts
+++ b/projects/addon-commerce/components/card/test/card.component.spec.ts
@@ -55,14 +55,6 @@ describe(`Card`, () => {
 
             expect(hasPaymentSystemLogo).toBeTruthy();
         });
-
-        it(`returns empty string if paymentSystem is not inputed`, () => {
-            testComponent.paymentSystem = null;
-
-            fixture.detectChanges();
-
-            expect(testComponent.component.paymentSystemLogo).toBe(``);
-        });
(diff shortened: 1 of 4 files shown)


zarghamkhandev:
The current implementation of [`with` ](https://github.com/Tinkoff/taiga-ui/blob/84129b9e9b2a07df1e16706a33e4c54390e3ace6/projects/testing/utils/helpers.ts#L41) method loses the type information of harness class.

```
    const tuiCard = await loader.getHarness(TuiCardHarness);             // TuiCardHarness
    
    const tuiCard2 = await loader.getHarness(TuiCardHarness.with({selector: '1234'}));            // ComponentHarness

```

Any ideas on how to get around this ?

zarghamkhandev:
Instead of mixin, I would vote for the inheritance approach, but not sure if that's possible at this point.

waterplea:
> Instead of mixin, I would vote for the inheritance approach, but not sure if that's possible at this point.

I think I broke my head over this already and this is the best I could come up with other than to write that `with` manually well the time 😄 I'm open for suggestions.

PR #2688, opened 2022-09-13
feat(eslint-plugin): create prefer-inject-decorator

### LINKED ISSUE #2662: 🤖 - ESLint rule `prefer-inject-decorator`
### Description

Our Taiga UI libraries have a strict code style rule: always use `Inject()` for Dependency Injection.

Why?
You can read it here: https://twitter.com/Waterplea/status/1376527548486725632

Write eslint-rule (with auto-fix, optionally) for this case

ddubrava:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #2662

## What is the new behavior?

ESLint forces to use Inject() for Dependency Injection.

- [ ] Yes
- [X] No

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -102,5 +102,6 @@ module.exports = {
                 ],
             },
         ],
+        '@taiga-ui/prefer-inject-decorator': 'error',
     },
 };
--- a/projects/cdk/directives/for/test/for.directive.spec.ts
+++ b/projects/cdk/directives/for/test/for.directive.spec.ts
@@ -1,4 +1,4 @@
-import {Component, ElementRef} from '@angular/core';
+import {Component, ElementRef, Inject} from '@angular/core';
 import {ComponentFixture, TestBed} from '@angular/core/testing';
 import {configureTestSuite} from '@taiga-ui/testing';
 import {Subject} from 'rxjs';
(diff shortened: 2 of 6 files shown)


ddubrava:
Hey! I tried not to hardcode a lot but in some cases we can't escape this.

Anyway, there are some questions:

1. Can we use typescript? I looked at some articles, and I don't think it's too complicated to implement it. The problem is that `.?` stuff comes not from typings but from real experience and AST explorer. It's unreliable. I believe typings over AST should cover such cases.
2. Can we implement testing? It's out of the box in TypeScript ESLint. I think this rule is a little bit complicated and testing can help us.
3. There are [TypeScript ESLint Utils](https://www.npmjs.com/package/@typescript-eslint/utils), can we install it? I need an AST NODE TYPES enum not to hardcode them.

nsbarsukov:
@ddubrava thanks for your contribution! 🚀 
Let's discuss your questions.
___

> Can we use typescript?

I really want to use Typescript too.
But this new `eslint-plugin` is not only for publication but also is used for local usage.
Is there a convenient way to use this plugin locally without compiling Typescript to Javascript (and always keeping `dist`-folder) ?
I don't know this way that is why I use JS. May be you know a good practice for mentioned problem.
___

> Can we implement testing?

Of course, we love testing ! And we will really appreciate if you add them!

___

> There are [TypeScript ESLint Utils](https://www.npmjs.com/package/@typescript-eslint/utils), can we install it?

Yep, you can install it.
@splincode could you help with advice ?
Should it be as `peer-dependency` of package `@taiga-ui/eslint-plugin` ?

splincode:
```
Can we use typescript?
```

We need use `tsc` in later

```
There are [TypeScript ESLint Utils](https://www.npmjs.com/package/@typescript-eslint/utils), can we install it?

Yep, you can install it.
@splincode could you help with advice ?
```

As dependency

splincode (approved):

splincode:
@ddubrava `--fix`
image

PR #2784, opened 2022-10-01
docs: add flag to show edit button only if TS & HTML exist

### LINKED ISSUE #2766: 🐞 -  'Edit on Stackblitz' doesn't work in Grid chapter

Don't known / other

Non-Blocking

### Description

Open [Grid doc page](https://taiga-ui.dev/grid) and click on button 'Edit on Stackblitz' in any example. There will be no a new tab with example code in Stackblitz.

### Angular version

14.2.3

### Taiga UI version

3.4.0

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [X] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

ddubrava:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #2766

## What is the new behavior?

An `Edit on` link is not visible for examples without HTML or TypeScript files.

An `Edit on` link doesn't work in the linked issue since the example doesn't have a TypeScript property, but `TuiStackblitzService.edit` method requires it. The possible solution is just to add this missing property to the example object. But there are other examples where the `Edit on` feature also doesn't work, e.g. [variables](https://taiga-ui.dev/variables). I guess that we can't force to use HTML & TypeScript properties all the time. So I implemented logic to hide an `Edit on ...` link if an example has neither HTML nor TypeScript properties. Otherwise, the button is visible but an example still cannot be opened due to the if condition in `TuiStackblitzService.edit`.

image

- [ ] Yes
- [x] No

### DIFF
--- a/projects/addon-doc/src/components/example/example.template.html
+++ b/projects/addon-doc/src/components/example/example.template.html
@@ -48,14 +48,14 @@
             </tui-tabs-with-more>
 
             <button
-                *ngIf="codeEditor"
+                *ngIf="files | tuiMapper: visible"
                 tuiButton
                 appearance="flat"
                 size="s"
                 [showLoader]="!!(loading$ | async)"
                 (click)="edit(files)"
             >
-                Edit on {{ codeEditor.name }}
+                Edit on {{ codeEditor!.name }}
             </button>
         </div>
 
--- a/projects/addon-doc/src/components/example/example.component.ts
+++ b/projects/addon-doc/src/components/example/example.component.ts
@@ -18,7 +18,7 @@ import {BehaviorSubject, Observable, Subject} from 'rxjs';
 import {map, switchMap} from 'rxjs/operators';
 
 import {TuiCodeEditor} from '../../interfaces/code-editor';
-import {TuiDocExample} from '../../interfaces/page';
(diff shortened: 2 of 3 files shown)


projects/addon-doc/src/components/example/example.template.html
```diff
@@ -48,14 +48,14 @@
             </tui-tabs-with-more>
 
             <button
-                *ngIf="codeEditor"
+                *ngIf="editButtonVisible$ | async"
```
splincode, line 51: what do you think about

```ts
 readonly visible = (files) => this.codeEditor &&
                        files[TUI_EXAMPLE_PRIMARY_FILE_NAME.TS] &&
                        files[TUI_EXAMPLE_PRIMARY_FILE_NAME.HTML];
}
```

```html
<button
                *ngIf="processor$ | async | tuiMapper: visible"
```

ddubrava: yep, looks better, thanks!

splincode (approved):

waterplea (approved):

PR #2785, opened 2022-10-01
feat(kit): add date input options DI token for custom icons

### LINKED ISSUE #2756: 🚀 - Ability to override calendar icon on tui-input-date

kit

### Description

image
Provide a way to override the calendar icon, I've tried with `tuiTextfieldCustomContent` but it doesn't replace it.

This is needed when using custom icons from custom design system

alvaromartmart:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #2756

## What is the new behavior?

This PR adds a `TUI_INPUT_DATE_OPTIONS` token that can be used to customize `iconCalendar` & `iconCalendarLarge` values, which are used to display the appropriate icon.
The implementation is based on what's seen on https://github.com/Tinkoff/taiga-ui/blob/main/projects/kit/components/select/select-options.ts, I followed the same approach to allow overriding options.

- [ ] Yes
- [x] No

### DIFF
--- a/projects/kit/components/input-date-range/input-date-range.component.ts
+++ b/projects/kit/components/input-date-range/input-date-range.component.ts
@@ -42,7 +42,6 @@ import {
     TuiDialogService,
     TuiMarkerHandler,
     TuiPrimitiveTextfieldComponent,
-    tuiSizeBigger,
     TuiSizeL,
     TuiSizeS,
     TuiTextfieldSizeDirective,
@@ -232,7 +231,7 @@ export class TuiInputDateRangeComponent
     }
 
     get size(): TuiSizeL | TuiSizeS {
-        return tuiSizeBigger(this.textfieldSize.size) ? `l` : `s`;
+        return this.textfieldSize.size;
     }
 
     set nativeValue(value: string) {
--- a/projects/kit/components/input-date-range/input-date-range.component.ts
+++ b/projects/kit/components/input-date-range/input-date-range.component.ts
@@ -38,12 +38,9 @@ import {
 } from '@taiga-ui/cdk';
 import {
     TUI_DEFAULT_MARKER_HANDLER,
-    TUI_TEXTFIELD_SIZE,
     TuiDialogService,
     TuiMarkerHandler,
     TuiPrimitiveTextfieldComponent,
-    tuiSizeBigger,
-    TuiTextfieldSizeDirective,
     TuiTextMaskOptions,
     TuiWithOptionalMinMax,
 } from '@taiga-ui/core';
@@ -142,8 +139,6 @@ export class TuiInputDateRangeComponent
         @Optional()
         @Inject(TUI_MOBILE_CALENDAR)
         private readonly mobileCalendar: Type<object> | null,
-        @Inject(TUI_TEXTFIELD_SIZE)
-        private readonly textfieldSize: TuiTextfieldSizeDirective,
         @Inject(TUI_DATE_FORMAT) readonly dateFormat: TuiDateMode,
         @Inject(TUI_DATE_SEPARATOR) readonly dateSeparator: string,
         @Inject(TUI_DATE_TEXTS)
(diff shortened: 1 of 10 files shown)


alvaromartmart:
Please let me know if the approach is the right one for supporting this customization.

splincode (approved):

waterplea (changes requested):
Thank you for this PR, a few comments:

1. Let's reuse this token in `InputDateRange` and `InputDateTime` as well.
2. We can add `min` and `max` to this token too.
3. Please take a look at [how it is done](https://github.com/Tinkoff/taiga-ui/blob/main/projects/kit/components/input-time/input-time-options.ts#L17) in `InputTime`, where we have single icon, but it's `PolymorpheusContent` where size comes as context.

alvaromartmart:
> Thank you for this PR, a few comments:
> 
> 1. Let's reuse this token in `InputDateRange` and `InputDateTime` as well.

Shall I move the `input-date-options.ts` to `kit/tokens` folder then since now it's reused by several components?

> 2. We can add `min` and `max` to this token too.

👍 

> 3. Please take a look at [how it is done](https://github.com/Tinkoff/taiga-ui/blob/main/projects/kit/components/input-time/input-time-options.ts#L17) in `InputTime`, where we have single icon, but it's `PolymorpheusContent` where size comes as context.

Will do, thanks

waterplea:
> Shall I move the input-date-options.ts to kit/tokens folder then since now it's reused by several components?

Yes, let's move it so entry points are less entangled.

alvaromartmart:
Applied suggested changes

waterplea:
I believe you need to change templates too to enable that size based polymorpheus content for icons.

alvaromartmart:
> I believe you need to change templates too to enable that size based polymorpheus content for icons.

Done

waterplea (changes requested):
Great! Thank you. One last thing, let's remove those `size` getters and just pass size from the controller. Our current design has the same icon for `m` and `l`, but that might not be the case for others who would want to override default icons.

alvaromartmart:
> Great! Thank you. One last thing, let's remove those `size` getters and just pass size from the controller. Our current design has the same icon for `m` and `l`, but that might not be the case for others who would want to override default icons.

I'm not sure I understand what you mean, are you referring to the ternary expression (` return tuiSizeBigger(this.textfieldSize.size) ? "l" : "s"`) or to the getter altogether? I used a `size` getter based on other instances in the code that implemented the same behaviour. Example:

https://github.com/alvaromartmart/taiga-ui/blob/8fd65e76655f398d0856d56b5e7a7d69ef9a98c6/projects/kit/components/input-tag/input-tag.component.ts#L219-L221

waterplea:
> > Great! Thank you. One last thing, let's remove those `size` getters and just pass size from the controller. Our current design has the same icon for `m` and `l`, but that might not be the case for others who would want to override default icons.
> 
> I'm not sure I understand what you mean, are you referring to the ternary expression (` return tuiSizeBigger(this.textfieldSize.size) ? "l" : "s"`) or to the getter altogether? I used a `size` getter based on other instances in the code that implemented the same behaviour. Example:
> 
> https://github.com/alvaromartmart/taiga-ui/blob/8fd65e76655f398d0856d56b5e7a7d69ef9a98c6/projects/kit/components/input-tag/input-tag.component.ts#L219-L221

Yes, the link you sent is just a getter to access private controller, it returns size "as is". In your code you narrow down size to only 's' or 'l':
```ts
    get size(): TuiSizeL | TuiSizeS {
        return tuiSizeBigger(this.textfieldSize.size) ? `l` : `s`;
    }
```
Change it to just returning `this.textfieldSize.size`, because otherwise it's impossible to override icon for 'm' size specifically.

alvaromartmart:
Hi, I believe this is done, please let me know if there's still anything pending

waterplea (approved):

PR #2800, opened 2022-10-03
ci: support import/no-cycle

### LINKED ISSUE #2750: 🤖 - `eslint-plugin` add `no-import-circular-deps`
## Problem

**projects/testing/core/badge.harness.ts:**
```ts
import {TuiComponentHarness} from '@taiga-ui/testing';
```
It breaks build because it produces circular dependency:
```console
Entry point @taiga-ui/testing has a circular dependency on @taiga-ui/testing/core.
```
To fix build it should be corrected to:
```ts
import {TuiComponentHarness} from '@taiga-ui/testing/utils';
```

## Proposed solution
Write eslint-rule (with autofix!).

ddubrava:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [x] Code style update

Closes #2750

## What is the new behavior?

`import/no-cycle` is added

- [ ] Yes
- [x] No

### DIFF
--- a/projects/addon-editor/extensions/image-editor/image-editor.component.ts
+++ b/projects/addon-editor/extensions/image-editor/image-editor.component.ts
@@ -5,7 +5,7 @@ import {TuiDestroyService, tuiTypedFromEvent} from '@taiga-ui/cdk';
 import {merge} from 'rxjs';
 import {takeUntil} from 'rxjs/operators';
 
-import {TuiEditableImage} from './image-editor.extension';
+import type {TuiEditableImage} from './image-editor.extension';
 
 export const TUI_EDITOR_MIN_IMAGE_WIDTH = new InjectionToken<number>(
     `[TUI_EDITOR_MIN_IMAGE_WIDTH]: Min size of resizable image inside editor`,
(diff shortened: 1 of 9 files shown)


splincode:
@nsbarsukov @ddubrava 

image

ddubrava:
ye, I've seen all these errors. I'm not sure I can fix them properly, architecture stuff.
I can give it a try.

waterplea:
Seems like they could be fixed by `import type`, can they? At least some of them.

ddubrava:
Hi! I fixed some errors but not all. There are a lot of issues with importing components, directives, constants etc.

splincode:
@waterplea I think we will not get profit, maybe we should close this PR

ddubrava:
If it's possible it's better to avoid circular dependencies. They lead to problems in most cases, e.g. domino effect and so on

ddubrava:
> @ddubrava can we fix this situation?
> 
> line-chart.component.ts -> line-chart-hint.directive.ts -> line-chart.component.ts
> 
> https://github.com/Tinkoff/taiga-ui/blob/main/projects/addon-charts/components/line-chart/line-chart.component.ts#L35 https://github.com/Tinkoff/taiga-ui/blob/main/projects/addon-charts/components/line-chart/line-chart-hint.directive.ts#L28

Ye, I'm not sure about the correct approach but you can inject the directive by an injection token. And the injection token is stored in the another file. I think it will fix the issue but I'm not sure that it can improve sth. And it's probably just a hack, I mean It fixes the issue but fixes it in the wrong way.

And I think it might be ok to have circular dependencies if you have two entities that are strongly connected. I mean you can't use this hint without the component and vice versa.

splincode:
I think we can close this PR and investigate [this](https://github.com/Tinkoff/taiga-ui/pull/2865/files) one in the future

PR #2825, opened 2022-10-06
fix(styles): pseudo elements `:after` aren't removed in Safari 16

splincode:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

### DIFF
--- a/projects/styles/basic/main.less
+++ b/projects/styles/basic/main.less
@@ -71,3 +71,16 @@ html,
 ::-ms-reveal {
     display: none;
 }
+
+// @note: only safari 16+
+@supports (-webkit-hyphens: none) and (text-align-last: right) {
+    /**
+      * @descriptions:
+      * Safari 16 has bug when for some reason,
+      * the ::after blocks aren't removed
+      * after the destruction of the parent class
+     */
+    :after {
+        content: none;
+    }
+}


projects/styles/basic/main.less
```diff
@@ -71,3 +71,15 @@ html,
...
+
+@supports (-webkit-hyphens: none) {
+    /**
+      * @descriptions:
+      * Safari 16 has bug when for some reason,
+      * the ::after blocks aren't removed
+      * after the destruction of the parent class
+     */
+    :after {
+        content: none;
```
nsbarsukov, line 83: I am afraid that it can break something(

Should we use `:where()` (for 0 specificity) ?
Safari 16 supports it
https://caniuse.com/mdn-css_selectors_where

Guys, what do you think about it ?

nsbarsukov: I mean something like this

```css
@supports (-webkit-hyphens: none) {
    /**
      * @descriptions:
      * Safari 16 has bug when for some reason,
      * the ::after blocks aren't removed
      * after the destruction of the parent class
     */
    :where(:after) {
        content: none;
    }
}
```

splincode: didn't work

vladimirpotekhin (approved):

projects/styles/basic/main.less
waterplea: If it works, we should go for it. Such hack should be as narrow as possible with lowest possible specificity. Also we need to be sure there are no default before/after elements on some native tags that could suffer from this hack.

waterplea:
We narrowed styles to Safari with hyphen hack, can we narrow it even more to Safari 16 with `@supports (text-indent: each-line)`:
https://caniuse.com/css-text-indent

waterplea (approved):

PR #2854, opened 2022-10-10
fix(styles): move required styles from `styles` package to tui-root

### LINKED ISSUE #2633: 🐞 - `@taiga-ui/styles` issues with the new "optional" package

Don't known / other

Non-Blocking

### Description

There are a couple of issues when the new `@taiga-ui/styles` package global styles are not included:

- You get double main page scrollbar because the styles which hide the default scrollbar are [provided by `@taiga-ui/styles`](https://github.com/Tinkoff/taiga-ui/blob/main/projects/styles/basic/main.less#L60)
image

- Text color in dark theme is black. Had to add the following CSS to fix it:
```css
body {
    color: var(--tui-text-01);
}
```

Also, the GitHub issue template doesn't have `@taiga-ui/styles` package selection in the "select package(s)" dropdown.

### Angular version

14.2

### Taiga UI version

3

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

vladimirpotekhin:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #2633 

- [ ] Yes
- [ ] No

### DIFF
--- a/projects/core/components/root/root.style.less
+++ b/projects/core/components/root/root.style.less
@@ -1,4 +1,5 @@
 @import '../../styles/mixins/mixins.less';
+@import '../../styles/mixins/text.less';
 
 // Required by some Taiga UI components
 @keyframes tuiPresent {
@@ -7,19 +8,28 @@
     }
 }
 
-:host {
+html,
+.tui-zero-scrollbar {
+    .scrollbar-hidden();
+}
+
+body {
+    .text-basic();
+}
+
+tui-root {
     position: relative;
     display: block;
     flex: 1;
     -webkit-tap-highlight-color: transparent;
 }
 
-.t-scrollbar {
+tui-root .t-root-scrollbar {
     .fullsize(fixed, inset);
     margin: 0;
 }
 
-.t-content {
+.t-root-content {
     .createStackingContext();
     height: 100%;
 }
--- a/projects/core/components/root/root.style.less
+++ b/projects/core/components/root/root.style.less
@@ -15,6 +15,7 @@ html,
 
 body {
     .text-basic();
+    margin: 0;
 }
 
 tui-root {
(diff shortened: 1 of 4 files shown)


splincode (approved):

waterplea (approved):
If we go for separate styles component let's also move this from `root.styles.less`:
```less
// Required by some Taiga UI components
@keyframes tuiPresent {
    to {
        content: '1';
    }
}
```

projects/core/components/root/root.style.less
```diff
@@ -7,19 +8,28 @@
...
+
+tui-root {
     position: relative;
     display: block;
     flex: 1;
     -webkit-tap-highlight-color: transparent;
 }
 
-.t-scrollbar {
+tui-root .t-root-scrollbar {
```
waterplea, line 27: `>` if you want to be safe, otherwise let's just drop `tui-root`.

projects/core/components/root/root.style.less
```diff
@@ -7,19 +8,28 @@
     }
 }
 
-:host {
+html,
+.tui-zero-scrollbar {
+    .scrollbar-hidden();
+}
+
+body {
```
waterplea, line 16: Let's remove default margin too

projects/core/components/root/root.style.less
vladimirpotekhin: it doesn't work without tui-root, I added `>`

splincode (approved):

oayoub84:
Hi, is there a way to override styles on this file, adding a margin: 0 without using a css variable is not flexible. In our case we use taiga ui inside a micro frontend which lives with other micro frontends in the same page, so we don't have the rights to change body styles.

waterplea:
> Hi, is there a way to override styles on this file, adding a margin: 0 without using a css variable is not flexible. In our case we use taiga ui inside a micro frontend which lives with other micro frontends in the same page, so we don't have the rights to change body styles.

Please file an issue so we can think of something.

PR #2904, opened 2022-10-14
feat(core): `PrimitiveCalendar` add ability to change the color of the days

### LINKED ISSUE #2606: 🚀 - Calendar should provide way to customize weekends

_No response_

### Description

- The days on which weekends fall should be customizable, they aren't Sat/Sun in all locales.
- Weekend days shouldn't automatically be coloured as `tui-negative`.

vladimirpotekhin:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #2606 

## What is the new behavior?
Снимок экрана 2022-10-14 в 17 40 31

- [ ] Yes
- [x] No

### DIFF
--- a/projects/core/components/primitive-calendar/primitive-calendar.template.html
+++ b/projects/core/components/primitive-calendar/primitive-calendar.template.html
@@ -28,7 +28,7 @@
                     <div
                         automation-id="tui-primitive-calendar__item"
                         class="t-item"
-                        [attr.data-type]="getDataType(item)"
+                        [attr.data-type]="item | tuiMapper: getDataType:dayTypeHandler"
                         [class.t-item_unavailable]="itemIsUnavailable(item)"
                     >
                         {{ item.day }}
--- a/projects/core/components/primitive-calendar/primitive-calendar.template.html
+++ b/projects/core/components/primitive-calendar/primitive-calendar.template.html
@@ -28,7 +28,7 @@
                     <div
                         automation-id="tui-primitive-calendar__item"
                         class="t-item"
-                        [attr.data-type]="item | tuiMapper: getDataType:dayTypeHandler"
+                        [attr.data-type]="item | tuiMapper: dayTypeHandler"
                         [class.t-item_unavailable]="itemIsUnavailable(item)"
                     >
                         {{ item.day }}
--- a/projects/core/components/primitive-calendar/primitive-calendar.template.html
+++ b/projects/core/components/primitive-calendar/primitive-calendar.template.html
@@ -16,7 +16,6 @@
                 <div
                     *ngIf="!itemIsUnavailable(item) || showAdjacent"
                     automation-id="tui-primitive-calendar__cell"
-                    style.color="var(--tui-day-of-week-{{ getDayIndex(colIndex) }})"
                     class="t-cell"
                     [class.t-cell_today]="itemIsToday(item)"
                     [class.t-cell_interval]="itemIsInterval(item)"
@@ -29,6 +28,7 @@
                     <div
                         automation-id="tui-primitive-calendar__item"
                         class="t-item"
+                        [attr.data-type]="getDataType(item)"
                         [class.t-item_unavailable]="itemIsUnavailable(item)"
                     >
                         {{ item.day }}
--- a/projects/core/components/primitive-calendar/primitive-calendar.template.html
+++ b/projects/core/components/primitive-calendar/primitive-calendar.template.html
@@ -16,8 +16,8 @@
                 <div
                     *ngIf="!itemIsUnavailable(item) || showAdjacent"
                     automation-id="tui-primitive-calendar__cell"
+                    style.color="var(--tui-day-of-week-{{ colIndex + 1 }})"
                     class="t-cell"
-                    [class.t-cell_weekend]="item.isWeekend"
                     [class.t-cell_today]="itemIsToday(item)"
                     [class.t-cell_interval]="itemIsInterval(item)"
                     [attr.data-range]="getItemRange(item)"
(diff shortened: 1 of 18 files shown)


splincode (approved):

projects/core/styles/theme/variables.less
```diff
@@ -148,4 +148,10 @@
     --tui-chart-2: var(--tui-support-21);
     --tui-chart-3: var(--tui-support-11);
     --tui-chart-4: var(--tui-base-05);
+    // Days of week
+    --tui-day-of-week-6: var(--tui-negative);
+    --tui-day-of-week-7: var(--tui-negative);
+    each(range(1, 5, 1), {
```
waterplea, line 154: Didn't know you can do loops so easily 👍

projects/core/components/primitive-calendar/primitive-calendar.template.html
```diff
@@ -16,8 +16,8 @@
                 <div
                     *ngIf="!itemIsUnavailable(item) || showAdjacent"
                     automation-id="tui-primitive-calendar__cell"
+                    style.color="var(--tui-day-of-week-{{ colIndex + 1 }})"
```
waterplea, line 19: Would be nice if it was not `colIndex` but day index so you can color sunday red, no matter if it's the first or the last day.

projects/core/components/primitive-calendar/primitive-calendar.template.html
```diff
@@ -16,7 +16,7 @@
                 <div
                     *ngIf="!itemIsUnavailable(item) || showAdjacent"
                     automation-id="tui-primitive-calendar__cell"
-                    style.color="var(--tui-day-of-week-{{ colIndex + 1 }})"
+                    style.color="var(--tui-day-of-week-{{ (colIndex + this.firstDayOfWeek || 7) % 7 || 7 }})"
```
splincode, line 19: what do you think?

```html
style.color="var(--tui-day-of-week-{{ computeDay(colIndex) }})"
```

```ts
computeDay(colIndex) {
  return (colIndex + this.firstDayOfWeek || 7) % 7 || 7;
}
```

projects/core/components/primitive-calendar/primitive-calendar.template.html
vladimirpotekhin: I am not sure if this is convenient. Don't you think if you choose TUI FIRST DAY as Sunday it's more intuitive to colorize it with `--tui-day-of-week-1` ? 🤔

Anyway I pushed fix, take a look (commit has a little math hack due to Sunday index 0 in our `TuiDayOfWeek` enum)

evantrimboli:
I think that the concept of a weekend class/marker being applied should still be retained here, otherwise the theming is intrinsically tied to the locale. For example in momentjs, the 'en-in' locale has 6 weekdays:

https://github.com/moment/moment/blob/develop/src/locale/en-in.js#L12

evantrimboli:
I think it would be good to leave the weekend markers in here and allow locales to determine what the weekends are. Consider switching from a locale where the workweek goes from Mon -> Fri, to a locale where it's Mon -> Sat.

projects/core/components/primitive-calendar/primitive-calendar.template.html
waterplea: In the states they would make Sunday the first day, but still would keep Sunday and Saturday as weekends/colored red. This seems more semantic to me.

waterplea:
> I think it would be good to leave the weekend markers in here and allow locales to determine what the weekends are. Consider switching from a locale where the workweek goes from Mon -> Fri, to a locale where it's Mon -> Sat.

I'm not sure, what if somebody wants to color Saturdays and Sundays differently? I'm not sure there's a real use case for that, but more freedom in this case could be good.

evantrimboli:
I would say keeping the distinction between weekdays/weekends is good because it's fairly common to want to know that in for business apps. It's a pretty reasonable default to at least mark those cells as such.

On top of that, is it possible to allow users to decorate cells with data attributes? For example, if I could pass some function:

```typescript
const decorate = (d: Date) => {
  const attrs = [];
  if (isWeekendDay(d)) {
    attrs.push('weekend');
  }
  if (isPublicHoliday(d)) {
    attrs.push('holiday');
  }
  if (isSomeonesBirthday(d)) {
    attrs.push('birthday');
  }
  return attrs;
}
```
The underlying cell would then have `<div class="cell" data-weekend="true" data-holiday="true">` and you could target specific cells, but the core wouldn't need to bake anything in, just the ability to add the attributes.

projects/core/components/primitive-calendar/primitive-calendar.template.html
```diff
@@ -29,6 +28,7 @@
                     <div
                         automation-id="tui-primitive-calendar__item"
                         class="t-item"
+                        [attr.data-type]="getDataType(item)"
```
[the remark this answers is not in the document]
waterplea: Indeed, let's use `tuiMapper` here, because we cannot be sure how performant this handler is.

projects/core/tokens/day-type-handler.ts
```diff
@@ -0,0 +1,9 @@
+import {InjectionToken} from '@angular/core';
+import {TuiDay, TuiHandler} from '@taiga-ui/cdk';
+
+export const TUI_DAY_TYPE_HANDLER = new InjectionToken<TuiHandler<TuiDay, string>>(
+    `[TUI_DAY_TYPE_HANDLER]: [TUI_DAY_TYPE]: Token for adding data-type attribute to calendar cell`,
```
waterplea, line 5: Something's wrong with the name

projects/core/components/primitive-calendar/primitive-calendar.template.html
```diff
@@ -29,6 +28,7 @@
                     <div
                         automation-id="tui-primitive-calendar__item"
                         class="t-item"
+                        [attr.data-type]="item | tuiMapper: getDataType:dayTypeHandler"
```
waterplea, line 31: Just `item | tuiMapper : dayTypeHandler`

vladimirpotekhin: fixed

waterplea (approved):

PR #2966, opened 2022-10-26
fix(kit): icon in date-related components occupies space when hidden

### LINKED ISSUE #2960: 🐞 - Calendar icon in date inputs can be hidden, but it still occupies its space

kit

https://stackblitz.com/edit/angular-owymhi?file=src/app/app.component.ts

Blocking

### Description

I'd like to move the icon to the left, and I'm kinda able to do it via `providers`, but there is a visual bug that stops me from using this solution. I can open a PR and add an `*ngIf` to completely hide the icon when it's not present (e.g. [here](https://github.com/Tinkoff/taiga-ui/blob/1f44e39fb4fba6b579107cae2094d7e5a228e0d7/projects/kit/components/input-date/input-date.template.html#L44))

And/or we can open another issue for moving the icon left

### Angular version

14.2.7

### Taiga UI version

3.8.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

theorlovsky:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #2960 

## What is the new behavior?

Icon is completely hidden when removed via DI

- [ ] Yes
- [x] No

### DIFF
--- a/projects/kit/components/input-date-range/input-date-range.template.html
+++ b/projects/kit/components/input-date-range/input-date-range.template.html
@@ -11,7 +11,7 @@
         automation-id="tui-input-date-range__textfield"
         tuiValueAccessor
         class="t-textfield"
-        [tuiTextfieldIcon]="computedMobile ? mobileIconContent : iconContent"
+        [tuiTextfieldIcon]="calendarIcon && iconContent"
         [pseudoFocus]="innerPseudoFocused"
         [pseudoHover]="pseudoHover"
         [pseudoActive]="pseudoActive"
@@ -31,21 +31,13 @@
         ></ng-content>
     </tui-primitive-textfield>
 
-    <ng-template #mobileIconContent>
+    <ng-template #iconContent>
         <tui-svg
             *polymorpheusOutlet="calendarIcon as src; context: {$implicit: size}"
             automation-id="tui-input-date-range__icon"
             [class.t-icon]="!computedDisabled"
             [src]="src"
-            (click)="onMobileClick()"
-        ></tui-svg>
-    </ng-template>
-
-    <ng-template #iconContent>
-        <tui-svg
-            *polymorpheusOutlet="calendarIcon as src; context: {$implicit: size}"
-            [class.t-icon]="!computedDisabled"
-            [src]="src"
+            (click)="onIconClick()"
         ></tui-svg>
     </ng-template>
 
--- a/projects/kit/components/input-date-time/input-date-time.template.html
+++ b/projects/kit/components/input-date-time/input-date-time.template.html
@@ -9,7 +9,7 @@
         automation-id="tui-input-date-time__textfield"
         tuiValueAccessor
         class="t-textfield"
-        [tuiTextfieldIcon]="iconContent"
+        [tuiTextfieldIcon]="calendarIcon && iconContent"
         [pseudoFocus]="pseudoFocus"
         [pseudoHover]="pseudoHover"
         [invalid]="computedInvalid"
--- a/projects/kit/components/input-date/input-date.component.ts
+++ b/projects/kit/components/input-date/input-date.component.ts
@@ -225,10 +225,8 @@ export class TuiInputDateComponent
         return this.activeItem ? `` : filler;
     }
 
-    onMobileClick(): void {
-        if (!this.mobileCalendar) {
-            this.open = !this.open;
-
+    onIconClick(): void {
+        if (!this.computedMobile || !this.mobileCalendar) {
             return;
         }
 
--- a/projects/kit/components/input-date-range/input-date-range.component.ts
+++ b/projects/kit/components/input-date-range/input-date-range.component.ts
@@ -253,10 +253,8 @@ export class TuiInputDateRangeComponent
         return this.activePeriod ? `` : this.getDateRangeFiller(dateFiller);
     }
 
-    onMobileClick(): void {
-        if (!this.mobileCalendar) {
-            this.toggle();
-
+    onIconClick(): void {
+        if (!this.computedMobile || !this.mobileCalendar) {
             return;
         }
 
--- a/projects/kit/components/input-date/input-date.template.html
+++ b/projects/kit/components/input-date/input-date.template.html
@@ -10,7 +10,7 @@
         automation-id="tui-input-date-range__textfield"
         tuiValueAccessor
         class="t-textfield"
-        [tuiTextfieldIcon]="computedMobile ? mobileIconContent : iconContent"
+        [tuiTextfieldIcon]="calendarIcon && iconContent"
         [pseudoFocus]="pseudoFocus"
         [pseudoHover]="pseudoHover"
         [invalid]="computedInvalid"
@@ -31,21 +31,13 @@
         ></ng-content>
     </tui-primitive-textfield>
 
-    <ng-template #mobileIconContent>
+    <ng-template #iconContent>
         <tui-svg
             *polymorpheusOutlet="calendarIcon as src; context: {$implicit: size}"
             automation-id="tui-input-date-range__icon"
             [class.t-icon]="!computedDisabled"
             [src]="src"
-            (click)="onMobileClick()"
-        ></tui-svg>
-    </ng-template>
-
-    <ng-template #iconContent>
-        <tui-svg
-            *polymorpheusOutlet="calendarIcon as src; context: {$implicit: size}"
-            [class.t-icon]="!computedDisabled"
-            [src]="src"
+            (click)="onIconClick()"
         ></tui-svg>
     </ng-template>
 


projects/kit/components/input-date-range/input-date-range.template.html
```diff
@@ -11,7 +11,7 @@
         automation-id="tui-input-date-range__textfield"
         tuiValueAccessor
         class="t-textfield"
-        [tuiTextfieldIcon]="computedMobile ? mobileIconContent : iconContent"
+        [tuiTextfieldIcon]="hasCalendarIcon ? (computedMobile ? mobileIconContent : iconContent) : null"
```
splincode, line 14: what do you think about implementing the pipe?

```html
[tuiTextfieldIcon]="calendarIcon | tuiPresent ? (computedMobile ? mobileIconContent : iconContent) : null"
```

theorlovsky: looks like a good idea

waterplea: There no need for a pipe like that, it only complicates things, besides empty `string` is also lack of icon, not just `null`. There's also no need for nesting a ternary operator, you can write it like that:
```html
[tuiTextfieldIcon]="computedMobile ? mobileIconContent : (calendarIcon && iconContent)"
```
If `calendarIcon` is falsy (empty `string` or `null`) it is passed to `tuiTextfieldIcon`, disabling the icon, otherwise `iconContent` is used.

projects/kit/components/input-date-time/input-date-time.template.html
```diff
@@ -9,7 +9,7 @@
         automation-id="tui-input-date-time__textfield"
         tuiValueAccessor
         class="t-textfield"
-        [tuiTextfieldIcon]="iconContent"
+        [tuiTextfieldIcon]="hasCalendarIcon ? iconContent : null"
```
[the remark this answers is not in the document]
waterplea: `[tuiTextfieldIcon]="calendarIcon && iconContent"`

projects/kit/components/input-date-range/input-date-range.template.html
theorlovsky: yes, but what about `mobileIconContent`? it should not be present as well

theorlovsky: besides, an empty string still renders `tui-svg`, it just has nothing to show. while `null` doesn't render anything

theorlovsky: while I agree that it looks more complicated, but using `null` instead of an empty string is a right decision imho. and then it comes down to this simple choice: method or pipe?

`hasCalendarIcon ? (computedMobile ? mobileIconContent : iconContent) : null`
vs
`(calendarIcon | tuiIsPresent) ? (computedMobile ? mobileIconContent : iconContent) : null`

theorlovsky:
I forgot to add docs for `tuiIsPresent` pipe, will do it now

splincode (approved):

projects/cdk/pipes/is-present/is-present.pipe.ts
```diff
@@ -0,0 +1,10 @@
+import {Pipe, PipeTransform} from '@angular/core';
+
+import {tuiIsPresent} from '../../utils';
+
+@Pipe({name: `tuiIsPresent`})
+export class TuiIsPresentPipe implements PipeTransform {
+    transform<T>(value: T | null): value is T {
```
splincode, line 7: `transform<T>(value?: T | null): value is T {`

theorlovsky: yeah, my bad

splincode: image

projects/kit/components/input-date-range/input-date-range.template.html
waterplea: Ok. 
I don't see any difference between mobile and desktop icon content except for the click handler, but it has an early return anyway. How about we get rid of it and write:
```
[tuiTextfieldIcon]="calendarIcon && iconContent"
```
and change `iconContent` to:
```html
<ng-template #iconContent>
        <tui-svg
            *polymorpheusOutlet="calendarIcon as src; context: {$implicit: size}"
            automation-id="tui-input-date-range__icon"
            [class.t-icon]="!computedDisabled"
            [src]="src"
            (click)="onIconClick()"
        ></tui-svg>
</ng-template>
```
and just rename `onMobileClick` to `onIconClick` + don't forget to remove `open` toggle from the early return.

waterplea: As for the pipe, it would be a useful addition to the `cdk` but I don't think we really need it here.

projects/cdk/pipes/is-present/is-present.pipe.ts
```diff
@@ -0,0 +1,9 @@
+import {Pipe, PipeTransform} from '@angular/core';
+import {tuiIsPresent} from '@taiga-ui/cdk';
```
waterplea, line 2: This needs to be `from '@taiga-ui/cdk/utils';`

projects/demo/src/modules/app/pages.ts
```diff
@@ -1153,6 +1153,12 @@ export const pages: TuiDocPages = [
                 keywords: `format, форматирование, преобразование, phone, телефон, пайп, pipe`,
                 route: `/pipes/format-phone`,
             },
+            {
+                section: $localize`Tools`,
+                title: `IsPresent`,
+                keywords: `present, существует, ngif, if, пайп, pipe`,
```
waterplea, line 1159: Let's add `null, undefined`

projects/kit/components/input-date-range/input-date-range.template.html
theorlovsky: @waterplea do you want me to move the pipe to a separate PR?

theorlovsky: as for this one
```
[tuiTextfieldIcon]="calendarIcon && iconContent"
```
I still don't think that this is a right solution. if `calendarIcon` is an empty string, then the expression `'' && iconContent` becomes just `''` and you'll get an empty `tui-svg` rendered, as shown on the screenshot above. so the bug with an empty space next to a cleaner won't be fixed (which is a purpose of this PR)

theorlovsky: and I've just figured out that because of this very reason we don't need the pipe here 🙂
because it filters out only `nil` values, but we need an empty string to be removed as well

theorlovsky: how do you handle such cases in Taiga?

```ts
onIconClick(): void {
    if (!this.computedMobile) { // <-- this.isMobile && !!this.mobileCalendar
        return;
    }

    this.dialogService
        .open<TuiDay>(new PolymorpheusComponent(this.mobileCalendar, this.injector), {
        // 'Type<object> | null' is not assignable... ^^^^^^^^^^^^
}
```

I could add `!` to `mobileCalendar`, but it's not type safe, and the linter is also not happy with this solution

just check it again?

```ts
if (!this.computedMobile || !this.mobileCalendar) {
    return;
}
```

theorlovsky:
moved `tuiIsPresent` pipe to a separate PR #3002

projects/kit/components/input-month-range/input-month-range.template.html
```diff
@@ -12,7 +12,7 @@
         [disabled]="computedDisabled"
         [nativeId]="nativeId"
         [readOnly]="readOnly"
-        [tuiTextfieldIcon]="calendarIcon"
+        [tuiTextfieldIcon]="calendarIcon ? calendarIcon : null"
```
splincode, line 15: what about :))

```html
[tuiTextfieldIcon]="calendarIcon | tuiFallback: null"
```

where

```ts
export class TuiFallbackPipe {
  transform<T, K = T>(value?: T|null, fallback: K): T | K {
    return tuiIsPresent(value) ? value : fallback;
  } 
}
```

splincode (approved):

projects/kit/components/input-date/input-date.component.ts
```diff
@@ -225,10 +225,8 @@ export class TuiInputDateComponent
         return this.activeItem ? `` : filler;
     }
 
-    onMobileClick(): void {
```
splincode, line 228: breaking changes?

theorlovsky: a method name changed, yes, but other than that it should work as before, I didn't notice any difference

projects/kit/components/input-month-range/input-month-range.template.html
theorlovsky: I don't mind, but I want to here @waterplea's opinion on this one, since it's our favourite topic in this PR 😄

theorlovsky: but again, it doesn't cover a [case](https://github.com/Tinkoff/taiga-ui/pull/2966#discussion_r1009408564) with an empty string

splincode: Oh, sorry, you can use simple solution

```

```html
[tuiTextfieldIcon]="calendarIcon ?? null"
```
```

Comment (theorlovsky) [reply projects/kit/components/input-month-range/input-month-range.template.html:15]: oh right, I didn't even think about it for some reason

**Comment by theorlovsky** [comment]
@splincode @waterplea so, everything got a lot simpler. I hope no one will come with an idea of hiding the icon using an empty string, and if they do, I wouldn't consider the result a bug on the Taiga side. so we're free to write as you suggested:

```html
[tuiTextfieldIcon]="calendarIcon && iconContent"
```

as for this one

```html
[tuiTextfieldIcon]="calendarIcon ? calendarIcon : null"
```

it can easily be reverted to the original

```html
[tuiTextfieldIcon]="calendarIcon"
```

theorlovsky:
@splincode since Alex is unavailable now, can we get someone else involved to get two approvals?

splincode:
`[tuiTextfieldIcon]="calendarIcon"`

theorlovsky:
> `[tuiTextfieldIcon]="calendarIcon"`

@splincode what do you mean?

splincode:
@theorlovsky sorry, I mean use `[tuiTextfieldIcon]="calendarIcon && iconContent"`

theorlovsky:
> @theorlovsky sorry, I mean use `[tuiTextfieldIcon]="calendarIcon && iconContent"`

@splincode I already use it

splincode:
@theorlovsky sorry, it's good now

theorlovsky:
@splincode how can we get a second approval here?

splincode:
@vladimirpotekhin @nsbarsukov fyi

waterplea (approved):

PR #3002, opened 2022-11-01
feat(cdk): implement `tuiIsPresent` pipe

theorlovsky:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

- [ ] Yes
- [x] No

### DIFF
--- a/projects/demo/src/modules/pipes/is-present/examples/1/component.ts
+++ b/projects/demo/src/modules/pipes/is-present/examples/1/component.ts
@@ -1,32 +1,22 @@
 import {Component} from '@angular/core';
-import {FormControl} from '@angular/forms';
 import {changeDetection} from '@demo/emulate/change-detection';
 import {encapsulation} from '@demo/emulate/encapsulation';
+import {of, Subject} from 'rxjs';
+import {delayWhen} from 'rxjs/operators';
 
 @Component({
     selector: `tui-is-present-example1`,
     templateUrl: `./template.html`,
-    styleUrls: [`./style.less`],
     changeDetection,
     encapsulation,
 })
 export class TuiIsPresentExample1 {
-    readonly items = [`String`, `null`, `undefined`];
-    readonly control = new FormControl(null);
+    private readonly loadCountSubject = new Subject<void>();
 
-    get value(): string | null | undefined {
-        return this.objectifyValue(this.control.value ?? ``);
-    }
+    readonly count$ = of(0).pipe(delayWhen(() => this.loadCountSubject));
 
-    private objectifyValue(value: string): string | null | undefined {
-        if (value === `null`) {
-            return null;
-        }
-
-        if (value === `undefined`) {
-            return undefined;
-        }
-
-        return value;
+    loadCount(): void {
+        this.loadCountSubject.next();
+        this.loadCountSubject.complete();
     }
 }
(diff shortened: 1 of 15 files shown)


projects/demo/src/modules/pipes/is-present/examples/1/component.ts
```diff
@@ -0,0 +1,36 @@
...
+import {tuiIsPresent} from '@taiga-ui/cdk';
+
+@Component({
+    selector: `tui-is-present-example1`,
+    templateUrl: `./template.html`,
+    styleUrls: [`./style.less`],
+    changeDetection,
+    encapsulation,
+})
+export class TuiIsPresentExample1 {
```
splincode, line 14: not sure I understand this example

theorlovsky: it's basically a copy-paste from the existing [`tuiIsPresent`](https://github.com/Tinkoff/taiga-ui/blob/main/projects/demo/src/modules/utils/miscellaneous/examples/5/index.ts) function example

splincode: but why should we show how the pipe works in this way?

instead

```html
{{ value | tuiPresent }} = value | tuiPresent
```

theorlovsky: oh, I see. you're right

theorlovsky: updated

splincode (approved):

splincode:
@nsbarsukov @vladimirpotekhin

waterplea:
I think a nice example would be something like:
```html
<div *ngIf="count$ | async | tuiIsPresent as value">
```
showing how this can be used for falsy values like `0`, but not for `null` while stream has not returned any value.

projects/demo/src/modules/pipes/is-present/examples/1/component.ts
```diff
@@ -0,0 +1,32 @@
...
+    readonly items = [`String`, `null`, `undefined`];
+    readonly control = new FormControl(null);
+
+    get value(): string | null | undefined {
+        return this.objectifyValue(this.control.value ?? ``);
+    }
+
+    private objectifyValue(value: string): string | null | undefined {
+        if (value === `null`) {
```
waterplea, line 22: nit: Looks like a job for `switch` :)

waterplea (approved):

projects/demo/src/modules/pipes/is-present/examples/1/component.ts
theorlovsky: do you want to keep this example or completely replace it with the one above?

theorlovsky: I went ahead and replaced it since the new example is more useful 🙂

theorlovsky: @waterplea please check if it's not too complicated

splincode (approved):

waterplea (approved):

projects/demo/src/modules/pipes/is-present/examples/1/component.ts
waterplea: Looks great. Thank you!

PR #3092, opened 2022-11-20
fix(kit): correct default active month based on `min/max`

### LINKED ISSUE #3090: 🐞 - InputDate opens current month and not the month of the maximum date

kit

https://taiga-ui.dev/components/input-date/API?tuiMode=null&max$=1

Non-Blocking

### Description

If setting the maximum date for tui-input-date < the current month, then when you open the calendar, the current month of the current year is still opened, where all the days are locked (as they should be). Clicking on the arrow to go to the previous month immediately opens the month of the set maximum date, where you can choose something. So, in fact, one can still use the input, but the UX doesn’t seem to be great here.
It seems like it would be much nicer if the calendar would open the maximum possible month instead of the current one (when max month < current).

An example for this issue:
https://taiga-ui.dev/components/input-date/API?tuiMode=null&min$=1

image

### Angular version

11.2.13

### Taiga UI version

2.57.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #3090

- [ ] Yes
- [x] No


projects/demo/src/modules/components/input-date/input-date.component.ts
```diff
@@ -65,7 +65,12 @@ export class ExampleTuiInputDateComponent extends AbstractExampleTuiControl {
         ),
     };
 
-    minVariants = [TUI_FIRST_DAY, new TuiDay(2017, 2, 5), new TuiDay(1900, 0, 1)];
+    minVariants = [
+        TUI_FIRST_DAY,
+        new TuiDay(2017, 2, 5),
+        new TuiDay(1900, 0, 1),
+        TuiDay.normalizeOf(new Date().getFullYear() + 3, 1, 1),
```
waterplea, line 72: nit: why not just `new`?

waterplea (approved):

vladimirpotekhin (approved):

projects/demo-integrations/cypress/tests/desktop/kit/input-date-range/input-date-range.cy.ts
```diff
@@ -1,4 +1,4 @@
-describe(`InputDateRange`, () => {
+describe(`InputDateTime`, () => {
```
waterplea, line 1: file is called `input-date-range` 🤔

waterplea (approved):

projects/demo-integrations/cypress/tests/desktop/kit/input-date-range/input-date-range.cy.ts
splincode: fixed

projects/cdk/utils/calendar/compute-active-year-or-month.ts
```diff
@@ -0,0 +1,17 @@
+import {TuiYear} from '@taiga-ui/cdk/date-time';
+
+export function tuiComputeActiveYearOrMonth<T extends TuiYear>(
```
waterplea, line 3: Let's call it `tuiDateClamp`.

splincode: fixed

waterplea (approved):

PR #3105, opened 2022-11-23
feat(addon-table): add paginationChange event

### LINKED ISSUE #2044: 🐞 - table-pagination - sizeChange souldn't call pageChange

addon-table

Blocking

### Description

Hi there, at first, thank you for amazing UI Library. It's the greatest!
A few days ago I decided to use TaigaUI for my project and met a proplem. 

In my application router query params and pagination are synced, so query string is a source of truth for the pagination. 
It looks like when a user change something on the page (pageIndex, perPage, etc), at first, I use router to change the query string, and then, state from query string changes the pagination on a page. 
Unfortunately, it doesn't work well, because when I change page size, table-pagination unexpectedly calls pageChange event and router.navigate calls twice (For page size changing and then for current page changing). But angular can't handle it (two navigates an once) and I got an error
`router.mjs:5931 Router Event: NavigationCancel`
`router.mjs:5932 NavigationCancel(id: 6, url: '/crud?page=1&per_page=20')`
`router.mjs:5933 NavigationCancel {id: 6, url: '/crud?page=1&per_page=20', reason: 'Navigation ID 6 is not equal to the current navigation id 7'}`
It means that angular can't update query params, because another navigation has started.

I think it heppens [here](https://github.com/Tinkoff/taiga-ui/blob/main/projects/addon-table/components/table-pagination/table-pagination.component.ts#L77). As we can see thera are two call at once.

`this.sizeChange.emit`
`this.pageChange.emit(this.page)`

In my opinion this behavior is not trasparent and I'd like to have an ability to change the page on my own with page size (at once) to prevent this error.

Thank you!

### Angular version

13.2.0

### Taiga UI version

2.50.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

xvs32x:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #2044 

## What is the new behavior?
This PR adding an opportunity to subscribe to both of parameters (page and size) at once and let us to prevent behavior from the issue

- [ ] Yes
- [x] No

### DIFF
--- a/projects/addon-table/components/table-pagination/pagination.model.ts
+++ b/projects/addon-table/components/table-pagination/pagination.model.ts
@@ -1,0 +1,4 @@
+export interface TuiPagination {
+    page: number;
+    size: number;
+}
--- a/projects/addon-table/components/table-pagination/table-pagination.component.ts
+++ b/projects/addon-table/components/table-pagination/table-pagination.component.ts
@@ -74,7 +74,7 @@ export class TuiTablePaginationComponent {
         return {
             page: this.page,
             size: this.size,
-        }
+        };
     }
 
     onItem(size: number): void {
--- a/projects/addon-table/components/table-pagination/table-pagination.component.ts
+++ b/projects/addon-table/components/table-pagination/table-pagination.component.ts
@@ -4,11 +4,11 @@ import {tuiDefaultProp} from '@taiga-ui/cdk';
 import {TUI_SPIN_TEXTS} from '@taiga-ui/core';
 import {Observable} from 'rxjs';
 
+import {TuiPagination} from './pagination.model';
 import {
     TUI_TABLE_PAGINATION_OPTIONS,
     TuiTablePaginationOptions,
 } from './table-pagination-options';
-import {TuiPagination} from './pagination.model';
 
 @Component({
     selector: `tui-table-pagination`,
--- a/projects/addon-table/components/table-pagination/table-pagination.component.ts
+++ b/projects/addon-table/components/table-pagination/table-pagination.component.ts
@@ -32,9 +32,17 @@ export class TuiTablePaginationComponent {
     @tuiDefaultProp()
     size = this.items[0];
 
+    /**
+     * TODO: Remove in 4.0
+     * @deprecated use paginationChange
+     */
     @Output()
     readonly pageChange = new EventEmitter<number>();
 
+    /**
+     * TODO: Remove in 4.0
+     * @deprecated use paginationChange
+     */
     @Output()
     readonly sizeChange = new EventEmitter<number>();
 
--- a/projects/addon-table/components/table-pagination/table-pagination.component.ts
+++ b/projects/addon-table/components/table-pagination/table-pagination.component.ts
@@ -8,6 +8,7 @@ import {
     TUI_TABLE_PAGINATION_OPTIONS,
     TuiTablePaginationOptions,
 } from './table-pagination-options';
+import {TuiPagination} from './pagination.model';
 
 @Component({
     selector: `tui-table-pagination`,
@@ -36,6 +37,9 @@ export class TuiTablePaginationComponent {
 
     @Output()
     readonly sizeChange = new EventEmitter<number>();
+
+    @Output()
+    readonly paginationChange = new EventEmitter<TuiPagination>();
 
     open = false;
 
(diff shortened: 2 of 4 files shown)


xvs32x:
Hi, @splincode. What do you think?

splincode (approved):

projects/addon-table/components/table-pagination/table-pagination.component.ts
```diff
@@ -37,6 +38,9 @@ export class TuiTablePaginationComponent {
     @Output()
     readonly sizeChange = new EventEmitter<number>();
 
+    @Output()
+    readonly paginationChange = new EventEmitter<TuiPagination>();
```
nsbarsukov, line 42: Update API page, please:

https://taiga-ui.dev/components/table-pagination/API

https://github.com/Tinkoff/taiga-ui/blob/main/projects/demo/src/modules/tables/table-pagination/table-pagination.template.html

projects/addon-table/components/table-pagination/table-pagination.component.ts
```diff
@@ -66,6 +70,13 @@ export class TuiTablePaginationComponent {
         return this.end === this.total;
     }
 
+    get pagination(): TuiPagination {
+        return {
+            page: this.page,
+            size: this.size,
```
nsbarsukov, line 76: Run prettier for this file

projects/addon-table/components/table-pagination/table-pagination.component.ts
xvs32x: done

projects/addon-table/components/table-pagination/table-pagination.component.ts
xvs32x: done

nsbarsukov:
[Lint](https://github.com/Tinkoff/taiga-ui/actions/runs/3575585008/jobs/6023726505)-job fails with error:

```
Run npm run prettier -- --check

> @taiga-ui/components@3.11.0 prettier
> prettier '**/*.{svg,yml,js,ts,html,md,less,json}' --cache node_modules/.cache/prettier --check

Checking formatting...
[warn] projects/demo/src/modules/tables/table-pagination/table-pagination.template.html
[warn] Code style issues found in the above file. Forgot to run Prettier?
```

waterplea:
I think we should also deprecate original outputs since this is the preferred way now, right?

xvs32x:
> I think we should also deprecate original outputs since this is the preferred way now, right?

Yeah, it makes sense. Could you add some examples, how i can deprecate the props?

waterplea:
> > I think we should also deprecate original outputs since this is the preferred way now, right?
> 
> Yeah, it makes sense. Could you add some examples, how I can deprecate the props?

Just write:
```ts
/**
 * TODO: Remove in 4.0
 * @deprecated use paginationChange
 */
```
over those outputs

projects/addon-table/components/table-pagination/pagination.model.ts
```diff
@@ -0,0 +1,4 @@
+export interface TuiPagination {
```
splincode, line 1: `pagination.model.ts` -> `table-pagination.ts`

projects/addon-table/components/table-pagination/table-pagination.component.ts
```diff
@@ -4,6 +4,7 @@ import {tuiDefaultProp} from '@taiga-ui/cdk';
 import {TUI_SPIN_TEXTS} from '@taiga-ui/core';
 import {Observable} from 'rxjs';
 
+import {TuiPagination} from './pagination.model';
```
splincode, line 7: `TuiTablePagination`

projects/demo/src/modules/tables/table-pagination/table-pagination.template.html
```diff
@@ -123,6 +123,19 @@
                 <code>size</code>
                 -property (user selects new amount of lines per page).
             </ng-template>
+            <ng-template
+                #documentationPropertyPaginationChange="documentationProperty"
+                i18n
+                documentationPropertyName="paginationChange"
+                documentationPropertyMode="output"
+                documentationPropertyType="TuiPagination"
```
splincode, line 131: `TuiTablePagination`

xvs32x: done

PR #3141, opened 2022-11-29
feat(core): replace `TUI_DROPDOWN_OFFSET` with injection token

bondarvladislave:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [x] Refactoring
- [ ] Code style update

In different design systems, the `tui-data-list` may contain paddings other than `4px`, which are set by the `TUI_DROPDOWN_OFFSET` constant. I would like to do this as an injection token so that the dropdown does not cover the content.

image

- [ ] Yes
- [x] No

### DIFF
--- a/projects/core/directives/dropdown/dropdown-options.directive.ts
+++ b/projects/core/directives/dropdown/dropdown-options.directive.ts
@@ -16,17 +16,13 @@ import {
 } from '@taiga-ui/core/types';
 import {tuiOverrideOptions} from '@taiga-ui/core/utils';
 
-/**
- * Safe space around host and screen edges
- */
-export const TUI_DROPDOWN_OFFSET = 4;
-
 export interface TuiDropdownOptions {
     readonly align: TuiHorizontalDirection;
     readonly direction: TuiVerticalDirection | null;
     readonly limitWidth: TuiDropdownWidth;
     readonly minHeight: number;
     readonly maxHeight: number;
+    readonly offset: number;
 }
 
 /** Default values for dropdown options */
@@ -36,6 +32,7 @@ export const TUI_DROPDOWN_DEFAULT_OPTIONS: TuiDropdownOptions = {
     limitWidth: `auto`,
     maxHeight: 400,
     minHeight: 80,
+    offset: 4,
 };
 
 export const TUI_DROPDOWN_OPTIONS = new InjectionToken<TuiDropdownOptions>(
@@ -57,7 +54,7 @@ export const tuiDropdownOptionsProvider: (
 });
 
 @Directive({
-    selector: `[tuiDropdownAlign], [tuiDropdownDirection], [tuiDropdownLimitWidth], [tuiDropdownMinHeight], [tuiDropdownMaxHeight]`,
+    selector: `[tuiDropdownAlign], [tuiDropdownDirection], [tuiDropdownLimitWidth], [tuiDropdownMinHeight], [tuiDropdownMaxHeight], [tuiDropdownOffset]`,
     providers: [
         {
             provide: TUI_DROPDOWN_OPTIONS,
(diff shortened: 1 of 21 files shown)


splincode (approved):

waterplea:
How about we add it to options? And also maybe it makes sense to make it two values: horizontal and vertical?

bondarvladislave:
> How about we add it to options? And also maybe it makes sense to make it two values: horizontal and vertical?

@waterplea 

Yes, it really doesn't make sense to me if we don't add vertical and horizontal ones.

If added to options then we need to extend `TuiDropdownOptionsDirective` with selectors `tuiDropdownHorizontalOffset` and `tuiDropdownVerticalOffset` ?

But I would consider adding this to all places in the api documentation.
Because horizontal offset will work as expected only in `tuiDropdownSided`.

And for a regular `tuiDropdown`, it is difficult to describe in the api example why nothing happens until a specific number. (**on screenshot**)
I don’t know how to properly separate the logic so as not to confuse users.

And maybe you remember why you created `[tuiDropdownCustomPosition]`?
Maybe this was the beginning for some future solution on this topic


telegram-cloud-document-2-5206398513428047338

waterplea:
Hmmm. True. Then it doesn't make sense to add both, if only one would matter :) Let's keep it one and add to directive/options token.

bondarvladislave:
@waterplea 
I don't know how to add a translation in `projects/demo/src/modules/components/abstract/dropdown-documentation/dropdown-documentation.template.html` 
phrase `Dropdown vertical offset`

waterplea:
> @waterplea 
> I don't know how to add a translation in `projects/demo/src/modules/components/abstract/dropdown-documentation/dropdown-documentation.template.html` 
> phrase `Dropdown vertical offset`
> 
> 

We only support English docs now, don't have resources for multiple languages, so don't worry about it.

waterplea:
@bondarvladislave will you finish this PR? In discussion we figured out there's no need for 2 values because only one is used at any given time.

bondarvladislave:
@waterplea 

`tuiDropdownSided` uses two at once. And they are needed, I would like to avoid this vertical shift while increasing the horizontal one. That's why I need both.
telegram-cloud-photo-size-2-5260493291969954490-m

It's just that in cases other than `tuiDropdownSided` it works in a very non-obvious way for the user, and therefore I would not want to add it to the Demo for everyone

waterplea:
@bondarvladislave I'm not sure horizontal offset makes sense for traditional dropdown, if more complicated positioning is required, it's best to make your own `PositionAccessor`. As for vertical offset in case of sided dropdown, we need to think. This slight offset we currently have is meant to compensate for padding to align things like this:
image
Although currently it doesn't work that well, especially since that padding changes, depending on the size of `DataList`. Maybe we can have it a separate option for `TuiDropdownPositionSidedDirective`?

bondarvladislave:
@waterplea Should I create a separate file for sided dropdown options or combine with all options this way?

telegram-cloud-photo-size-2-5265189692679243222-y

Then `[tuiDropdownOffset]` will do nothing on sided because only `sidedVerticalOffset` and `sidedHorizontalOffset` will be used in the calculations.
And I don’t know if it is necessary and where exactly to describe it in the documentation and show it in the demo.

image

waterplea:
How about we only make single offset in this PR, use it for vertical in regular and horizontal in sided, and we will address sided vertical position in another PR?

bondarvladislave:
@waterplea
I hope I understood you correctly.
I removed the vertical calculations for the side opening.
What confuses me is that this will change this component for everyone who uses it at the moment, the side ones will be shifted 4px vertically.
And about another PR, will you do it yourself? Because I don't see a solution yet.

waterplea:
> @waterplea
> I hope I understood you correctly.
> I removed the vertical calculations for the side opening.
> What confuses me is that this will change this component for everyone who uses it at the moment, the side ones will be shifted 4px vertically.
> And about another PR, will you do it yourself? Because I don't see a solution yet.

Yes, I'll do the other one myself next week. We will release new version on Monday, then merge your PR and I'll cover the sided offset so there will not be a version where vertical offset is wrong.

waterplea (approved):

bondarvladislave:
@waterplea 
Should I do something else?
I received your changes via rebase.

waterplea:
> @waterplea Should I do something else? I received your changes via rebase.

No, I was just sick for a couple of days. Thank you for this PR.

PR #3150, opened 2022-11-30
feat(core): support override icon cleaner for proprietary kit

splincode:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update


waterplea:
Sheet, Dialog, Alert and Push are not related to textfield, they just happen to use the same icon, we shouldn't have them here. Also why do we change the icon in proprietary kit all of a sudden? If we updated icons and new ones don't look right, it's a problem with proprietary icons, not customization. Proprietary icon for that name should be the one that we use in inputs to avoid unnecessary alteration of default behavior.

PR #3192, opened 2022-12-07
feat(kit): use `tuiIconClose` for size `s`

splincode:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

UIKIT-6192

- [ ] Yes
- [x] No

### DIFF
--- a/projects/core/components/primitive-textfield/primitive-textfield.component.ts
+++ b/projects/core/components/primitive-textfield/primitive-textfield.component.ts
@@ -24,8 +24,10 @@ import {
 import {TuiHintOptionsDirective} from '@taiga-ui/core/directives/hint';
 import {
     TEXTFIELD_CONTROLLER_PROVIDER,
+    TUI_TEXTFIELD_OPTIONS,
     TUI_TEXTFIELD_WATCHED_CONTROLLER,
     TuiTextfieldController,
+    TuiTextfieldOptions,
 } from '@taiga-ui/core/directives/textfield-controller';
 import {MODE_PROVIDER} from '@taiga-ui/core/providers';
 import {TUI_MODE, TUI_TEXTFIELD_APPEARANCE} from '@taiga-ui/core/tokens';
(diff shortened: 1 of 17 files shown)


projects/core/components/primitive-textfield/primitive-textfield.component.ts
```diff
@@ -282,6 +290,13 @@ export class TuiPrimitiveTextfieldComponent
         this.updateAutofilled(autofilled);
     }
 
+    @tuiPure
+    private getDefaultIconCleaner(size: TuiSizeL | TuiSizeS): PolymorpheusContent {
+        return typeof this.options.iconCleaner === `function`
```
waterplea, line 295: You don't need to compute that and context should not be just size, but `TuiContextWithImplicit<TuiSizeL | TuiSizeS>`. Context passing should be handled by polymorpheus outlet:
```html
<tui-svg
    *polymorpheusOutlet="iconCleaner as src; context: { $implicit: size }"
    [src]="src"
></tui-svg>
```

splincode: image

Sorry, I don't understand then how we can set icon for size `S` by default for all inputs?

splincode: thank you! fixed

projects/core/components/primitive-textfield/primitive-textfield.component.ts
```diff
@@ -77,7 +75,7 @@ export class TuiPrimitiveTextfieldComponent
 
     @Input()
     @tuiDefaultProp()
-    iconCleaner: TuiPrimitiveTextfieldOptions['iconCleaner'] = this.options.iconCleaner;
```
waterplea, line 80: I think we should probably deprecate it as input. Do you see a case where it would be dynamically changed?

waterplea (approved):

vladimirpotekhin (approved):

projects/core/components/primitive-textfield/primitive-textfield.component.ts
splincode: fixed

PR #3196, opened 2022-12-07
fix(addon-table): don't update direction order when updating sorter programmatically

### LINKED ISSUE #2969: 🐞 - disable automatic trigger for `sort$` and `direction$`

addon-table

https://taiga-ui.dev/components/table

Non-Blocking

### Description

I store table settings in url - sortOrder, sortBy, page, size. When parameters change (for example, when navigating back), I update the table settings.

Moreover, if I simultaneously call both sort$ and direction$ (which are passed to the direction and tuiSortBy inputs), then the table itself will trigger the change of direction$ to asc (1) when sort$ changes. This is inconvenient, because sometimes I need to restore the direction desc.

I would generally expect that if I manage the settings programmatically, then this automatic change of direction to asc will not happen.

### Angular version

latest

### Taiga UI version

latest

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

vladimirpotekhin:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #2969 

- [ ] Yes
- [x] No

### DIFF
--- a/projects/addon-table/components/table/directives/table.directive.ts
+++ b/projects/addon-table/components/table/directives/table.directive.ts
@@ -63,21 +63,30 @@ export class TuiTableDirective<
     @tuiDefaultProp()
     sorter: TuiComparator<T> = () => 0;
 
-    updateSorter(sorter: TuiComparator<T> | null): void {
+    updateSorterAndDirection(sorter: TuiComparator<T> | null): void {
         if (this.sorter === sorter) {
-            this.direction = this.direction === 1 ? -1 : 1;
-            this.directionChange.emit(this.direction);
+            this.updateDirection(this.direction === 1 ? -1 : 1);
         } else {
-            this.sorter = sorter || (() => 0);
-            this.sorterChange.emit(this.sorter);
-            this.direction = 1;
-            this.directionChange.emit(1);
+            this.updateSorter(sorter);
+            this.updateDirection(1);
         }
-
-        this.change$.next();
     }
 
     ngAfterViewInit(): void {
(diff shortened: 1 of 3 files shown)


splincode (approved):

projects/addon-table/components/table/directives/table.directive.ts
```diff
@@ -63,21 +63,30 @@ export class TuiTableDirective<
...
-
-        this.change$.next();
     }
 
     ngAfterViewInit(): void {
         this.changeDetectorRef.detectChanges();
     }
+
+    updateSorter(sorter: TuiComparator<T> | null): void {
+        if (sorter !== this.sorter) {
```
nsbarsukov, line 80: If I've not mistaken, this condition is checked twice (some lines above and here).

Do we really need this double check ?

nsbarsukov (approved):

projects/addon-table/components/table/directives/table.directive.ts
vladimirpotekhin: oh this function also called independently, from `Sortable` [directive](https://github.com/Tinkoff/taiga-ui/blob/335455cd644d15dad156e6fbe51e3d7246257325/projects/addon-table/components/table/directives/sortable.directive.ts#L36)
But you are right, `Sortable` has the same check, so I removed 👍

PR #3751, opened 2023-02-22
fix(core): escape event doesn't work in Safari

### LINKED ISSUE #3741: 🐞 - Escape event doesn't work in Safari for TuiHostedDropdown

core

https://stackblitz.com/edit/angular-sjpzht-t2uafm?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.ts&hideExplorer=1

Blocking

### Description

Open an interface with tui-hosted-dropdown component in Safari and try close it with press on Esc key.

**Expected result:**
Add posibillity to close dropdown with press on Esc key in Safari.

### Angular version

14.2.7

### Taiga UI version

3.2

- [ ] Chrome
- [ ] Firefox
- [X] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Code style update

Closes #3741

- [ ] Yes
- [x] No

### DIFF
--- a/projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
+++ b/projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
@@ -31,6 +31,7 @@ import {
     TuiDropdownHoverDirective,
 } from '@taiga-ui/core/directives/dropdown';
 import {tuiIsEditingKey} from '@taiga-ui/core/utils/miscellaneous';
+import {shouldCall} from '@tinkoff/ng-event-plugins';
 import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
 import {BehaviorSubject, EMPTY, merge} from 'rxjs';
 import {distinctUntilChanged, skip} from 'rxjs/operators';
@@ -40,6 +41,15 @@ import {TuiHostedDropdownConnectorDirective} from './hosted-dropdown-connector.d
 export interface TuiHostedDropdownContext
     extends TuiContextWithImplicit<TuiActiveZoneDirective> {
     close(): void;
+}
+
+function shouldClose(this: TuiHostedDropdownComponent, {key}: KeyboardEvent): boolean {
+    return (
+        key.toLowerCase() === 'escape' &&
+        this.canOpen &&
+        this.open &&
+        !this.dropdown?.nextElementSibling
+    );
 }
 
 /* eslint-disable @typescript-eslint/member-ordering */
@@ -159,12 +169,9 @@ export class TuiHostedDropdownComponent implements TuiFocusableElementAccessor {
         }
     }
 
-    @HostListener('keydown.esc', ['$event'])
+    @shouldCall(shouldClose)
+    @HostListener('document:keydown.silent', ['$event'])
     onKeyDownEsc(event: Event): void {
-        if (!this.canOpen || !this.open) {
-            return;
-        }
-
         event.stopPropagation();
         this.closeDropdown();
     }
--- a/projects/core/components/hosted-dropdown/hosted-dropdown.template.html
+++ b/projects/core/components/hosted-dropdown/hosted-dropdown.template.html
@@ -17,7 +17,6 @@
     >
         <div
             #wrapper
-            (keydown.esc)="onKeyDownEsc($event)"
             (keydown)="onKeydown($event)"
         >
             <ng-container
--- a/projects/demo-integrations/cypress/tests/core/dropdown/dropdown.cy.ts
+++ b/projects/demo-integrations/cypress/tests/core/dropdown/dropdown.cy.ts
@@ -1,22 +1,56 @@
 describe(`Dropdown`, () => {
-    beforeEach(() => {
-        cy.viewport(720, 720);
+    describe(`720x720`, () => {
+        beforeEach(() => cy.viewport(720, 720));
+
+        it(`TuiDropdown`, () => {
+            cy.tuiVisit(`/directives/dropdown`);
+
+            cy.get(`tui-dropdown-example-1 button`).first().click().wait(1000);
+            cy.matchImageSnapshot(`1`);
+
+            cy.get(`tui-dropdown-example-2 input`).first().click().wait(1000);
+            cy.matchImageSnapshot(`2`);
+        });
+
+        it(`Hosted Dropdown`, () => {
+            cy.tuiVisit(`/components/hosted-dropdown`);
+
+            cy.get(`tui-hosted-dropdown-example-1 button`).first().click().wait(1000);
+            cy.matchImageSnapshot(`3`);
+        });
     });
 
-    it(`TuiDropdown`, () => {
-        cy.tuiVisit(`/directives/dropdown`);
+    describe(`a12y`, () => {
+        beforeEach(() => cy.viewport(1280, 720));
 
-        cy.get(`tui-dropdown-example-1 button`).first().click().wait(1000);
-        cy.matchImageSnapshot(`1`);
+        it(`Esc -> Hosted Dropdown`, () => {
+            cy.tuiVisit(`/components/hosted-dropdown`);
 
-        cy.get(`tui-dropdown-example-2 input`).first().click().wait(1000);
-        cy.matchImageSnapshot(`2`);
-    });
+            cy.get(`tui-hosted-dropdown-example-2 button`)
+                .eq(1)
+                .tuiWaitBeforeAction()
+                .click();
 
-    it(`Hosted Dropdown`, () => {
-        cy.tuiVisit(`/components/hosted-dropdown`);
+            cy.tuiWaitBeforeScreenshot().matchImageSnapshot(`01-opened-hosted-dropdown`, {
+                capture: `viewport`,
+            });
 
-        cy.get(`tui-hosted-dropdown-example-1 button`).first().click().wait(1000);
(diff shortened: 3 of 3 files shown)


projects/eslint-plugin/rules/accessibility.js
```diff
@@ -0,0 +1,41 @@
...
+
+                    for (const argument of args) {
+                        const invalidEscape =
+                            argument.value === `keydown.esc` ||
+                            argument.value === `document:keydown.esc`;
+
+                        if (invalidEscape) {
+                            context.report({
+                                node,
+                                message: `Don't use "*:keydown.esc" instead of "*:keydown.escape"`,
```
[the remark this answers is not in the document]
splincode: fixed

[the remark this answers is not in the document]
waterplea: Safari is true IE of our times. Only IE has it called `escape` :)

projects/addon-commerce/components/input-card-grouped/input-card-grouped.component.ts
```diff
@@ -303,7 +303,7 @@ export class TuiInputCardGroupedComponent
         return this.cardPrefilled ? `*${this.card.slice(-4)}` : '*';
     }
 
-    @HostListener('keydown.esc')
+    @HostListener('document:keydown.escape')
```
waterplea, line 306: This would cause change detection on every escape keypress on the document. While this component is rarely used, I'd still like to avoid that.

projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
```diff
@@ -159,7 +159,7 @@ export class TuiHostedDropdownComponent implements TuiFocusableElementAccessor {
         }
     }
 
-    @HostListener('keydown.esc', ['$event'])
+    @HostListener('document:keydown.escape', ['$event'])
```
waterplea, line 162: This would cause change detection on every escape keypress on the document. This is a very common component, we definitely want to avoid that.

waterplea (changes requested):

projects/addon-commerce/components/input-card-grouped/input-card-grouped.component.ts
splincode: fixed

projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
splincode: fixed

projects/eslint-plugin/configs/all.json
```diff
@@ -4,6 +4,7 @@
         "@taiga-ui/injection-token-description": "error",
         "@taiga-ui/prefer-inject-decorator": "error",
         "@taiga-ui/prefer-self-destroy-service": "error",
+        "@taiga-ui/accessibility": "error",
```
splincode, line 7: @nsbarsukov @vladimirpotekhin what do you think about name?

nsbarsukov: Maybe it is better to call it directly `@taiga-ui/no-esc-keydown` ?

nsbarsukov (approved):

projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
waterplea: Maybe we can use `.silent` and `@shouldCall`? But they don't work with angular keydown effects, so we would need to filter for `escape` button inside `@shouldCall` predicate.

projects/eslint-plugin/configs/all.json
splincode: fixed

projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
splincode: `.silent` works as well

vladimirpotekhin (approved):

projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
waterplea: It will not work, because Angular re-enters the zone in those keydown events, so it works, but it's not «silent».

splincode: @waterplea you are right, sorry for that

projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
```diff
@@ -42,6 +43,10 @@ export interface TuiHostedDropdownContext
     close(): void;
 }
 
+function activeZoneFilter(this: TuiHostedDropdownComponent): boolean {
+    return this.isLastOpenedDropdown && this.activeZone.contains(this.wrapperElement);
```
waterplea, line 47: This is not what we need to be testing here. We need to test if the key pressed is escape and that the dropdown is open (`this.open`).

projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
```diff
@@ -159,7 +172,8 @@ export class TuiHostedDropdownComponent implements TuiFocusableElementAccessor {
         }
     }
 
-    @HostListener('keydown.esc', ['$event'])
+    @shouldCall(activeZoneFilter)
+    @HostListener('document:keydown.escape', ['$event'])
```
waterplea, line 176: Right now there's no point in this code. `@shouldCall` is used for filtering calls and **re-entering** the zone. We did not leave the zone, so all we did is write a very complex early return. We need `@HostListener('document:keydown.silent')` and inside `@shouldCall` we need to filter for escape key, because Angular re-enters zone if we rely on its built-in event manager plugin (`keydown.escape`). This can also be reused in `InputCardGrouped`.

waterplea (changes requested):

projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
splincode: image

And so, we have 2 open dropdown box. Expected behavior after the escape only the last one will be closed and so that it is:

image

projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
splincode: `Right now there's no point in this code. ` it would be great if there would be a eslint rule for that

splincode: fixed

splincode:
@waterplea @nsbarsukov @vladimirpotekhin hello, please could you review me?

splincode:
@waterplea updated

projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
```diff
@@ -42,6 +43,17 @@ export interface TuiHostedDropdownContext
     close(): void;
 }
 
+function shouldEscape(this: TuiHostedDropdownComponent, event: KeyboardEvent): boolean {
+    const wrapperElement = this.wrapperElement;
+
+    return (
+        event.key.toLowerCase() === 'escape' &&
+        this.isLastOpenedDropdown &&
+        !!wrapperElement &&
```
waterplea, line 52: What's the point of `wrapperElement` and especially checking if it's inside the zone? It **cannot** be outside the zone if it exists. Besides there's an early return for `!this.open` in `onKeyDownEsc` which already handles the case of ignoring escape when dropdown is closed (this early return should probably be moved in `shouldEscape` so that early return is handled in one place).

projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
```diff
@@ -106,6 +118,14 @@ export class TuiHostedDropdownComponent implements TuiFocusableElementAccessor {
         return this.openChange.value;
     }
 
+    get isLastOpenedDropdown(): boolean {
+        return !!this.dropdown && !this.dropdown?.nextElementSibling;
```
waterplea, line 122: Unnecessary test for `!!this.dropdown` if we will test for `this.open` anyway. `dropdown` is already public, let's just move this test for `nextElementSibling` inside `shouldEscape`. `shouldClose` is also probably a better name for that predicate :)

waterplea (changes requested):

projects/addon-commerce/components/input-card-grouped/input-card-grouped.component.ts
```diff
@@ -303,7 +315,8 @@ export class TuiInputCardGroupedComponent
         return this.cardPrefilled ? `*${this.card.slice(-4)}` : '*';
     }
 
-    @HostListener('keydown.esc')
+    @shouldCall(shouldEscape)
+    @HostListener('document:keydown.silent', ['$event'])
```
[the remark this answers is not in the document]
splincode: fixed

projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
splincode: fixed

waterplea (approved):

PR #3983, opened 2023-03-24
feat(kit): `MultiSelect` add ability to use native dropdown on mobile

vladimirpotekhin:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Code style update

Closes # 

- [ ] Yes
- [x] No


projects/kit/components/multi-select/multi-select.template.html
```diff
@@ -31,7 +31,21 @@
         (click.prevent)="onClick(inputTag)"
     >
         <ng-content></ng-content>
+        <ng-template #select>
+            <ng-container *ngIf="isMobile">
```
waterplea, line 35: Can we put this `ngIf` on `ng-content`?

projects/kit/components/multi-select/multi-select.template.html
```diff
@@ -31,7 +31,21 @@
...
+                <ng-content
+                    select="select"
+                    ngProjectAs="select"
```
waterplea, line 38: Do we need `ngProjectAs="select" both here a few lines below?

projects/kit/components/multi-select/multi-select.template.html
```diff
@@ -52,8 +66,8 @@
                 tuiWrapper
                 appearance="icon"
                 automation-id="tui-multi-select__arrow"
-                tuiPreventDefault="mousedown"
```
waterplea, line 55: Why we no longer need to prevent it?

projects/kit/components/multi-select/multi-select.component.ts
```diff
@@ -254,8 +265,15 @@ export class TuiMultiSelectComponent<T>
         this.updateValue(value.map(({item}) => item));
     }
 
+    onValueChange(value: readonly T[]): void {
+        this.updateValue(value);
+    }
+
     onSearch(search: string | null): void {
-        this.open = true;
+        if (!this.nativeDropdownMode) {
```
waterplea, line 273: When does `onSearch` trigger in native dropdown mode and why does it matter not to change `open`? Don't we already have `canOpen` that should not allow it to open?

projects/kit/components/multi-select/multi-select.directive.ts
```diff
@@ -0,0 +1,20 @@
...
+import {AbstractTuiTextfieldHost, tuiAsTextfieldHost} from '@taiga-ui/core';
+
+import {TuiMultiSelectComponent} from './multi-select.component';
+
+@Directive({
+    selector: 'tui-multi-select',
+    providers: [tuiAsTextfieldHost(TuiMultiSelectDirective)],
+})
+export class TuiMultiSelectDirective extends AbstractTuiTextfieldHost<
+    TuiMultiSelectComponent<unknown>
```
waterplea, line 11: Shouldn't it be `string`?

projects/kit/components/multi-select/native-multi-select/native-select.ts
```diff
@@ -0,0 +1,42 @@
...
+    get id(): string {
+        return this.elementRef.nativeElement.id || this.idService.generate();
+    }
+
+    selectedMapper: TuiMapper<string, true | null> = (option, value) =>
+        value.includes(option) || null;
+
+    onValueChange(target: HTMLSelectElement): void {
+        this.host.onValueChange(
+            Array.from(target.selectedOptions).map(option => option.value),
```
waterplea, line 39: Destruct?

projects/kit/components/multi-select/native-multi-select/native-select.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    @HostBinding(`id`)
+    get id(): string {
+        return this.elementRef.nativeElement.id || this.idService.generate();
+    }
+
+    selectedMapper: TuiMapper<string, true | null> = (option, value) =>
+        value.includes(option) || null;
+
+    onValueChange(target: HTMLSelectElement): void {
```
waterplea, line 37: Maybe use `this.elementRef.nativeElement` instead? Since event target is `EventTarget` and we do not have any errors just because `host` inside `@Component` is not type checked.

projects/kit/components/multi-select/native-multi-select/native-select.template.html
```diff
@@ -0,0 +1,13 @@
+<ng-container *ngIf="control.control?.valueChanges | async"></ng-container>
+
+<tui-data-list-wrapper
+    *tuiDataList
+    [items]="items"
+></tui-data-list-wrapper>
+<option
+    *ngFor="let option of items"
+    [attr.selected]="option | tuiMapper : selectedMapper : host.value"
+    [value]="option"
```
waterplea, line 10: We should also allow items to be disabled. And add it to the demo stand please. This applies to plain Select too.

projects/kit/components/multi-select/native-multi-select/native-select.ts
```diff
@@ -0,0 +1,42 @@
...
+
+@Directive()
+export abstract class AbstractTuiNativeSelect {
+    @ViewChild(TuiDataListDirective, {read: TemplateRef, static: true})
+    readonly datalist: TemplateRef<any> | null = null;
+
+    constructor(
+        @Inject(TUI_TEXTFIELD_HOST) readonly host: TuiTextfieldHost,
+        @Inject(AbstractTuiControl) readonly control: AbstractTuiControl<unknown>,
+        @Inject(ElementRef) private readonly elementRef: ElementRef<HTMLInputElement>,
```
waterplea, line 21: `HTMLSelectElement`

projects/kit/components/multi-select/native-multi-select/native-select-group.component.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import {AbstractTuiNativeSelect} from './native-select';
+
+@Component({
+    selector: 'select[multiple][tuiSelect][labels]',
+    templateUrl: './native-select-group.template.html',
+    providers: [
+        {
+            provide: TuiDataListDirective,
+            deps: [TuiNativeSelectGroupComponent],
```
waterplea, line 12: How can `useExisting` have `deps`? 🤔  If it's a mistake, maybe it's time for `tuiAsDataList` helper?

projects/kit/components/multi-select/native-multi-select/native-select-group.component.ts
```diff
@@ -0,0 +1,42 @@
...
+            useExisting: TuiNativeSelectGroupComponent,
+        },
+        {
+            provide: TemplateRef,
+            deps: [TuiNativeSelectGroupComponent],
+            useFactory: ({datalist}: TuiNativeSelectGroupComponent) => datalist,
```
waterplea, line 18: Man, that is some crazy DI stuff, I love Angular 😄

projects/core/abstract/abstract-textfield-host.ts
```diff
@@ -32,7 +32,7 @@ export abstract class AbstractTuiTextfieldHost<T extends AbstractTuiControl<any>
         return this.host.value?.toString() || ``;
     }
 
-    abstract onValueChange(value: string): void;
+    abstract onValueChange(value: string[] | string): void;
```
waterplea, line 35: I'm a bit skeptical about changing the very base class. Are you sure it's necessary and it doesn't bring any confusion to other usecases?

vladimirpotekhin: Hmm, ok I reverted changes. Plz pay attention - now we need to implement empty `onValueChange` and separate `onSelectionChange(string[])` in the `TuiMultiSelectDirective`. This way we don't touch base class, ok?

projects/kit/components/multi-select/native-multi-select/native-select.template.html
vladimirpotekhin: I'll fix plain select in a separate PR https://github.com/Tinkoff/taiga-ui/issues/4005

projects/kit/components/multi-select/multi-select.component.ts
vladimirpotekhin: I noticed a bug when clicking on cleaner: dropdown doesn't open but arrow icon rotates. But now I can't reproduce it 🤔  deleted

projects/kit/components/multi-select/multi-select.template.html
vladimirpotekhin: It was preventing native dropdown opening, and seems like it's old code and I didn't find reason for this 🤔 but now I found another solution with `pointer-events:none' and can return preventing for "desktop-dropdown" mode (just in case)

projects/kit/components/input-tag/input-tag.component.ts
```diff
@@ -157,6 +157,10 @@ export class TuiInputTagComponent
     @tuiDefaultProp()
     placeholder = '';
 
+    @Input()
+    @tuiDefaultProp()
+    removable = true;
```
waterplea, line 162: I wish we could avoid introducing flags 🤔

projects/kit/components/multi-select/multi-select.template.html
waterplea: It is there so when you are focused on input and click on arrow, focus remains inside input and does not go away and immediately back, I believe.

projects/kit/components/multi-select/native-multi-select/native-multi-select-group.template.html
```diff
@@ -0,0 +1,21 @@
...
+></tui-data-list-wrapper>
+<optgroup
+    *ngFor="let group of items; let index = index"
+    [label]="labels[index]"
+>
+    <option
+        *ngFor="let option of group"
+        [selected]="option | tuiMapper : selectedMapper : host.value"
+        [value]="option"
+        [attr.disabled]="host.disableItemHandler(option) || null"
```
waterplea, line 17: Can be property, I believe, so no need for `null` fallback.

projects/kit/components/multi-select/native-multi-select/native-multi-select.template.html
```diff
@@ -0,0 +1,15 @@
...
+<tui-data-list-wrapper
+    *tuiDataList
+    [items]="items"
+    [disabledItemHandler]="host.disableItemHandler"
+></tui-data-list-wrapper>
+<option
+    *ngFor="let option of items"
+    [selected]="option | tuiMapper : selectedMapper : host.value"
+    [value]="option"
+    [attr.disabled]="host.disableItemHandler(option) || null"
```
waterplea, line 12: Same here

projects/kit/components/multi-select/native-multi-select/native-multi-select.ts
```diff
@@ -0,0 +1,42 @@
...
+        private readonly idService: TuiIdService,
+        @Inject(ChangeDetectorRef) readonly cdr: ChangeDetectorRef,
+    ) {}
+
+    @HostBinding(`id`)
+    get id(): string {
+        return this.elementRef.nativeElement.id || this.idService.generate();
+    }
+
+    selectedMapper: TuiMapper<string, true | null> = (option, value) =>
```
waterplea, line 34: Should be `boolean` now, that you switched from attribute to property.

waterplea (approved):

projects/core/abstract/abstract-textfield-host.ts
waterplea: Ok, let's keep it that way

projects/demo/src/modules/components/multi-select/examples/10/index.html
```diff
@@ -0,0 +1,37 @@
+<tui-multi-select
+    placeholder="Ignored text"
```
nsbarsukov, line 2: Do we need this property if `[editable]="false"` ?

projects/demo/src/modules/components/multi-select/examples/10/index.html
```diff
@@ -0,0 +1,37 @@
...
+    <select
+        tuiSelect
+        tuiMultiSelectGroup
+        multiple
+        [items]="items"
+    ></select>
+</tui-multi-select>
+
+<tui-multi-select
+    placeholder="Ignored text"
```
nsbarsukov, line 21: Do we need this property if `[editable]="false"` ?

projects/kit/components/multi-select/multi-select.directive.ts
```diff
@@ -0,0 +1,28 @@
...
+> {
+    override get readOnly(): boolean {
+        return true;
+    }
+
+    disableItemHandler: TuiBooleanHandler<string> = item =>
+        this.host.disabledItemHandler(item);
+
+    onValueChange(_: string): void {
+        //
```
nsbarsukov, line 22: nit:

```ts
onValueChange = noop;
```

splincode (approved):

PR #4118, opened 2023-04-05
feat(addon-doc): add `TUI_USE_DEFAULT_NIGHT_THEME` token for ignore dark theme styles

splincode:
## PR Type

- [x] Bugfix
- [x] Feature
- [x] Refactoring

## What is the new behavior?

if we do not want to include styles
out of the box for the dark theme we can turn it off

### DIFF
--- a/projects/addon-doc/src/services/theme-night.options.ts
+++ b/projects/addon-doc/src/services/theme-night.options.ts
@@ -1,8 +1,15 @@
 import {InjectionToken} from '@angular/core';
+// eslint-disable-next-line @taiga-ui/no-deep-imports
+import {ALWAYS_TRUE_HANDLER} from '@taiga-ui/cdk/constants';
 
 export const TUI_THEME_NIGHT_STORAGE_DEFAULT_KEY = `tuiNight` as const;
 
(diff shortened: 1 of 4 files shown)


vladimirpotekhin (approved):

projects/addon-doc/src/services/theme-night.options.ts
```diff
@@ -1,8 +1,15 @@
 import {InjectionToken} from '@angular/core';
+// eslint-disable-next-line @taiga-ui/no-deep-imports
+import {ALWAYS_TRUE_HANDLER} from '@taiga-ui/cdk/constants';
```
waterplea, line 3: Why deep import is required?

waterplea (approved):

projects/addon-doc/src/services/theme-night.options.ts
splincode: from `addon-doc` we use constant in `cypress` we already use `TUI_DEFAULT_THEME_NAME`, but if you use a short path `cdk`, then `cypress` falls with cyclic dependencies

PR #4138, opened 2023-04-06
feat(cdk): `Replace` add new pipe

splincode:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring

Using `getRouterLink` or `@tuiPure getRouterLink` results in frequent calls to `replace` on strings

https://user-images.githubusercontent.com/12021443/230497337-fe5490e2-947d-4ead-acdb-db6d6e4d11c1.mov

## What is the new behavior?

https://user-images.githubusercontent.com/12021443/230497453-1cc16b69-e2e3-4c41-a746-11ffde91d82f.mov

### DIFF
--- a/projects/cdk/pipes/replace/replace.pipe.ts
+++ b/projects/cdk/pipes/replace/replace.pipe.ts
@@ -1,0 +1,18 @@
+import {Pipe, PipeTransform} from '@angular/core';
+
+@Pipe({name: `tuiReplace`})
+export class TuiReplacePipe implements PipeTransform {
+    transform(
+        value: string | null | undefined,
+        search: RegExp | string,
+        replaceValue: string | ((substring: string, ...args: any[]) => string),
+    ): string {
+        return (
+            value?.replace(
+                search,
+                // TS bug: https://github.com/microsoft/TypeScript/issues/22378
+                replaceValue as unknown as string,
+            ) ?? ``
+        );
+    }
+}
--- a/projects/addon-doc/src/components/page/page.component.ts
+++ b/projects/addon-doc/src/components/page/page.component.ts
@@ -40,6 +40,9 @@ export class TuiDocPageComponent {
 
     activeItemIndex = 0;
 
+    readonly from = / /g;
+    readonly to = '_';
+
     constructor(
         @Inject(TUI_DOC_DEFAULT_TABS) readonly defaultTabs: readonly string[],
         @Inject(PAGE_SEE_ALSO) readonly seeAlso: readonly string[],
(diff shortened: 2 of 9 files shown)


waterplea:
Replace seems to be a slow operation. Maybe we should just make a general `tuiReplace` pipe in cdk? Funnily enough, `replaceAll` with a string is slower than `replace` with `RegExt`:
image

splincode:
@waterplea fixed

projects/cdk/pipes/replace/replace.pipe.ts
```diff
@@ -0,0 +1,16 @@
+import {Pipe, PipeTransform} from '@angular/core';
+
+type Replacer = (substring: string, ...args: any[]) => string;
+
+@Pipe({name: `tuiReplace`})
+export class TuiReplacePipe implements PipeTransform {
+    transform(
+        value: string,
+        search: RegExp | string,
+        replacer: Replacer | string,
```
waterplea, line 10: Is it some kind of typescript issue? Can we make it extract type automatically from `String.prototype.replace`? Also let's make `value: string | null` to support `async` pipes since we check for falsy values anyway. BTW, I don't think it makes sense to check for falsy `search` and `replacer`, who would use a pipe like that? Just `return value && value.replace(...)` is fine, IMO.

waterplea (approved):
Don't forget to change commit message to "feat(cdk): `Replace` add new pipe"

projects/cdk/pipes/replace/replace.pipe.ts
splincode: `extract type automatically from String.prototype.replace` it doesn't work
https://github.com/microsoft/TypeScript/issues/22378

`return value && value.replace(...) is fine, IMO.` updated

splincode: fixed

projects/cdk/pipes/replace/replace.pipe.ts
```diff
@@ -0,0 +1,30 @@
...
+
+type ReplaceValue = string | ((substring: string, ...args: any[]) => string);
+
+@Pipe({name: `tuiReplace`})
+export class TuiReplacePipe implements PipeTransform {
+    transform(
+        value: null | undefined,
+        searchValue: SearchValue,
+        replaceValue: ReplaceValue,
+    ): undefined;
```
waterplea, line 13: I don't think it makes sense to overload, maybe just `return value?.replace(...) || ''`?

projects/cdk/pipes/replace/replace.pipe.ts
```diff
@@ -0,0 +1,30 @@
+import {Pipe, PipeTransform} from '@angular/core';
+
+type SearchValue = RegExp | string;
```
waterplea, line 3: Without overloads let's not extract types.

splincode: fixed

projects/cdk/pipes/replace/replace.pipe.ts
splincode: fixed

vladimirpotekhin (approved):

PR #4162, opened 2023-04-09
ci: include `.pipe.ts` for linting

splincode:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring

### DIFF
--- a/scripts/eslint/angular.js
+++ b/scripts/eslint/angular.js
@@ -1,7 +1,13 @@
 module.exports = {
     overrides: [
         {
-            files: ['*.component.ts', '*.service.ts', '*.directive.ts', '*.module.ts'],
+            files: [
+                '*.component.ts',
+                '*.service.ts',
+                '*.directive.ts',
+                '*.pipe.ts',
+                '*.module.ts',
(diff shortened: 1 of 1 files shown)


projects/kit/components/multi-select/hide-selected.pipe.ts
```diff
@@ -6,6 +6,7 @@ import {TuiMultiSelectComponent} from './multi-select.component';
 
 @Pipe({
     name: `tuiHideSelected`,
+    // eslint-disable-next-line @angular-eslint/no-pipe-impure
```
waterplea, line 9: Let's just disable this rule. Unlike `ChangeDetectionStrategy.Default`, making a pipe impure requires an explicit action by the developer which means they mean it.

waterplea (approved):

projects/kit/components/multi-select/hide-selected.pipe.ts
splincode: fixed

waterplea (approved):

PR #4216, opened 2023-04-14
fix(kit): `TuiInputFilesDirective` does not provide capture attribute if null

### LINKED ISSUE #4215: 🐞 - TuiInputFilesDirective always provide `capture` attribute
### Playground Link

-

### Description

I have a problem with file input like this.

```
<tui-input-files>
  <input tuiInputFiles accept="image/*" />
</tui-input-files>
```

On mobile devices it always opening the camera instead of file dialog.

This is because `TuiInputFilesDirective` always add the `capture` attribute.

### Problem 1

`fileInput.capture` cannot be a `null` **(on mobile browser)**. Even if there is no attribute, there is an empty string in capture.

image

So, result of expression `input.capture ?? this.options.capture` always is `''`.

### Problem 2

`@HostBinding` cannot remove attribute, even if associated getter returns `null | undefined`. In this case, `@HostBinding` transforms result to string and set it's to the attribute. For example: if getter `capture` returns null, input element has an attribute `capture="null"`.

Profs:
image_2023-04-14_08-01-15

According [documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/capture) if `capture` attribute is present (even if empty), then the camera will be opened instead the file dialog.

### Angular version

15.2.7

### Taiga UI version

3.24.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

sviat9440:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

TuiInputFilesDirective always provide capture attribute

Closes #4215

## What is the new behavior?

TuiInputFilesDirective does not provide capture attribute if `null`

- [ ] Yes
- [x] No


projects/kit/components/input-files/input-files.directive.ts
```diff
@@ -40,14 +40,14 @@ export class TuiInputFilesDirective {
         return this.el.nativeElement.multiple ?? this.options.multiple;
     }
 
-    @HostBinding('capture')
+    @HostBinding('attr.capture')
     get capture(): TuiInputFilesOptions['capture'] {
         return (
```
waterplea, line 45: Maybe this is better?
```ts
    @HostBinding('attr.capture')
    get capture(): string | null {
        return this.el.nativeElement.getAttribute('capture') || this.options.capture;
    }
```
Since we cannot really rely on the `property` due to platform specifics and we bind an `attribute` anyway.

projects/kit/components/input-files/test/input-files.directive[#4215].spec.ts
```diff
@@ -0,0 +1,100 @@
+import { Component, ElementRef, ViewChild } from '@angular/core';
```
Comment (waterplea) [inline projects/kit/components/input-files/test/input-files.directive[#4215].spec.ts:1]: I don't think this test makes sense if it's a platform specific behavior. I believe we can drop it.

Comment (sviat9440) [reply projects/kit/components/input-files/test/input-files.directive[#4215].spec.ts]: fixed

projects/kit/components/input-files/input-files.directive.ts
sviat9440: fixed

splincode (approved):

projects/kit/components/input-files/input-files.directive.ts
```diff
@@ -40,15 +40,11 @@ export class TuiInputFilesDirective {
...
-            (
-                this.el.nativeElement as HTMLInputElement & {
-                    capture: TuiInputFilesOptions['capture'];
-                }
-            ).capture ?? this.options.capture
-        );
+            this.el.nativeElement.getAttribute('capture') as TuiInputFilesOptions['capture'] | null
+        ) ?? this.options.capture;
```
splincode, line 47: `) || this.options.capture;`

waterplea (approved):

PR #4230, opened 2023-04-17
fix(kit): `input-count` should update value after blur event

### LINKED ISSUE #4053: 🐞 - `InputCount` does not emit values for {updateOn: 'blur'}
### Playground Link

https://stackblitz.com/edit/angular-fhywik?devToolsHeight=33&file=src/app/app.component.html

### Description

If you set [ngModelOptions]="{updateOn: 'blur'}", then the value does not come in $event (ngModelChange)="onValueChange($event)"
The bug is reproduced on versions: 3.22.0, 3.21.0 and not reproduced on 3.20.0

### Angular version

12.2.17

### Taiga UI version

3.22.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

Closes #4053


projects/core/interfaces/update-on-focus.ts
```diff
@@ -0,0 +1,3 @@
+export interface TuiUpdateControlValueOnFocus {
```
waterplea, line 1: I don't think we need that interface.

waterplea (approved):

projects/core/interfaces/update-on-focus.ts
splincode: fixed

waterplea (approved):

vladimirpotekhin (approved):

vladimirpotekhin:
related https://github.com/Tinkoff/taiga-ui/issues/4134

PR #4294, opened 2023-04-24
fix(kit): `CheckboxBlock` fix broken view while global height variables changing

### LINKED ISSUE #4259: 🐞 - Checkbox Block is unable to adapt to a custom height size
### Playground Link

https://stackblitz.com/edit/angular-m8v7na?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.less&hideExplorer=1

### Description

The [CheckboxBlock](https://taiga-ui.dev/components/checkbox-block) component has non-adapted checkbox positioning to a custom height size.
https://github.com/Tinkoff/taiga-ui/blob/main/projects/kit/components/checkbox-block/checkbox-block.style.less#L86-L116

It causes the following issue

checkbox-block-l

### Angular version

12.2.7

### Taiga UI version

3.25.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

constantant:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

Closes #4259

## What is the new behavior?
The `checkboxPosition` input property has been added to the component which indicates the checkbox position in case the component height is bigger then a single height.

image

```ts
    @Input()
    @HostBinding('attr.data-checkbox-position')
    @tuiDefaultProp()
    checkboxPosition: TuiVerticalDirection | null = null;
```
By default it is located in the middle

image

- [x] Yes
- [ ] No


projects/demo/src/modules/components/checkbox-block/checkbox-block.template.html
```diff
@@ -43,6 +43,7 @@
                         formControlName="testValue1"
                         class="tui-space_all-1"
                         [contentAlign]="contentAlign"
+                        [checkboxPosition]="checkboxPosition"
```
splincode, line 46: @waterplea I know, you don't like too many flags

splincode (approved):

waterplea:
Our design system requires it to be aligned - to top. I believe it can be solved by using padding-top: calc(var(--tui-height-xxx) / 2) and transform: translateY(-50%) on the checkbox.

waterplea:
Regarding being able to align it to the middle vertically, I would prefer we changed the internal HTML so that align-items: center would do the trick. Keep in mind all of these changes are applicable to RadioBlock as well

constantant:
> Our design system requires it to be aligned - to top. I believe it can be solved by using padding-top: calc(var(--tui-height-xxx) / 2) and transform: translateY(-50%) on the checkbox.

If you okay with just a fallback it can look like that:
```less
.t-view {
    margin-top: 0.75rem;
    margin-right: var(--tui-horizontal-margin-l, var(--tui-horizontal-margin-m, 0.75rem));
    align-self: flex-start;

    :host[data-align='right'] & {
        margin-left: var(--tui-horizontal-margin-l, var(--tui-horizontal-margin-m, 0.75rem));
        margin-right: 0;
    }

    :host[data-size='s'] & {
        margin-top: var(--tui-vertical-margin-s, 0.5rem);
        margin-right: var(--tui-horizontal-margin-s, 0.5rem);
    }

    :host[data-size='s'][data-align='right'] & {
        margin-left: var(--tui-horizontal-margin-s, 0.5rem);
    }

    :host[data-size='m'] & {
        margin-top: var(--tui-vertical-margin-m, 0.875rem);
    }

    :host[data-size='l'] & {
        margin-top: var(--tui-vertical-margin-l, 1rem);
    }

    :host._hidden_checkbox & {
        .visually-hidden();
    }
}
```

waterplea:
We try to be very careful about introducing new CSS variables, because that is something you cannot easily take back, much like me inputs and the rest of public APIs.

constantant:
> We try to be very careful about introducing new CSS variables, because that is something you cannot easily take back, much like new inputs and the rest of public APIs.

Any suggestions how to proceed? If you do not like the middle by default it can be easily changed to the `top` value

```ts
    @Input()
    @HostBinding('attr.data-checkbox-position')
    @tuiDefaultProp()
    checkboxPosition: TuiVerticalDirection | null = 'top';
```

waterplea:
I'm on vacation this week and unable to give any more directions than what I've outlined in previous messages.

constantant:
@waterplea 
I have made `top` by default and added an ability to customise margins as it is done in the `Button` component. 
https://github.com/Tinkoff/taiga-ui/blob/main/projects/core/components/button/button.style.less#L137-L170
If it is okay I will apply the same things to `RadioBlock`

waterplea:
Closing this as the issue was resolved by another PR

Issue #1358, opened 2022-02-10
🚀 - Ability to use native controls on mobile


waterplea:
### Which @taiga-ui/* package(s) are relevant/releated to the feature request?

addon-mobile

### Description

We need a way to use native `select` for string values so on mobile it opens a native select modal. We also need native date and time pickers and the rest, whenever applicable.

waterplea:
Remaining controls (keep in mind `min`/`max` attributes where applicable):

- [ ] `InputDate` support `min`/`max` attributes
- [ ] `MultiSelect` — `select[multiple]`
- [ ] `InputDateTime` — `input[type="datetime-local"]`
- [ ] `InputMonth` — `input[type="month"]`
- [ ] `InputTime` — `input[type="time"]`
- [ ] `InputColor` — `input[type="color"]` (does not support gradient or opacity)

PR #4295, opened 2023-04-25
feat(kit): `InputDateTime` add ability to use native picker on mobile

vladimirpotekhin:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

#1358 

- [ ] Yes
- [x] No

### DIFF
--- a/projects/kit/components/input-date-time/native-date-time/native-date-time.component.ts
+++ b/projects/kit/components/input-date-time/native-date-time/native-date-time.component.ts
@@ -1,0 +1,55 @@
+import {ChangeDetectionStrategy, Component, Inject} from '@angular/core';
+import {TuiDay, TuiTime} from '@taiga-ui/cdk';
+import {TUI_TEXTFIELD_HOST} from '@taiga-ui/core';
+
+import type {TuiInputDateTimeDirective} from '../input-date-time.directive';
+
+@Component({
+    selector: 'input[tuiDateTime]',
+    template: '',
+    host: {
+        type: 'datetime-local',
+        '[tabIndex]': '-1',
+        '[value]': 'value',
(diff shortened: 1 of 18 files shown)


splincode (approved):

projects/kit/components/input-date-time/native-date-time/native-date-time.component.ts
```diff
@@ -0,0 +1,55 @@
+import {ChangeDetectionStrategy, Component, Inject} from '@angular/core';
+import {TuiDay, TuiTime} from '@taiga-ui/cdk';
+import {TUI_TEXTFIELD_HOST} from '@taiga-ui/core';
+
+import type {TuiInputDateTimeDirective} from '../input-date-time.directive';
+
+@Component({
+    selector: 'input[tuiDateTime]',
```
waterplea, line 8: It feels like all inputs like these should be directives and styles can be applied in the corresponding control less file, since they are for internal use only.

waterplea (approved):
Looks like this time seconds do not work even on Android. Other than that seems to be working.

projects/kit/components/input-date-time/native-date-time/native-date-time.component.ts
vladimirpotekhin: > Looks like this time seconds do not work even on Android. Other than that seems to be working.

 Yes, seems like no browser supports HH:MM:SS mode ( I removed the example and added a note that it only works in HH:MM mode

> It feels like all inputs like these should be directives and styles can be applied in the corresponding control less file, since they are for internal use only.

Refactored for date and date-time 👍  other controls (select, multiselect, time) should have template

PR #4320, opened 2023-04-27
feat(cdk): add `TUI_AUTOFOCUS_CUSTOM_HANDLER` for individual focus control

### LINKED ISSUE #4281: 🐞 - Tui Editor Auto Focus Bug
### Playground Link

https://stackblitz.com/edit/angular-hmqj9z?file=src/app/app.component.ts

### Description

It does not look like the tui-editor works with the autofocus directive.  I would expect the editor to be focused so a user can start typing right now it doesn't look like the autofocus does anything when paired with the editor component

In the stackblitz attached I have a basic example with an editor being hidden with an ngIf that is toggled with a button.
When the editor is first displayed I would expect the focus to be on the editor so I could start typing in the editor. 

### Angular version

12.2.17

### Taiga UI version

3.25.0

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [X] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

Closes #4281

## What is the new behavior?

image


projects/demo-integrations/cypress/tests/addon-editor/editor-groups.cy.ts
```diff
@@ -1,4 +1,4 @@
-import {tuiGetTipTapContentSelector} from '@demo-integrations/support/editor/helpers';
+import {TUI_PROSEMIRROR_SELECTOR} from '../../../../addon-editor/components/editor/editor.constants';
```
nsbarsukov, line 1: `@taiga-ui/addon-editor` doesn't work, does it ?

projects/demo/src/modules/components/editor/starter/editor-starter.template.html
```diff
@@ -6,6 +6,7 @@
     <ng-template pageTab="API">
         <tui-doc-demo [sticky]="false">
             <tui-editor
+                tuiAutoFocus
```
nsbarsukov, line 9: I don't know if it is a significant issue but it does not work for IPhone.

https://user-images.githubusercontent.com/35179038/235081119-9da04435-80eb-4b00-b3cb-ce5a093be240.MP4

nsbarsukov (approved):

projects/demo/src/modules/components/editor/starter/editor-starter.template.html
nsbarsukov: (maybe we can create a separate issue with low priority to fix it later if there is no easy solution to do it in this PR)

splincode: this needs to be configured at the level of the custom handler

with hack again:
image

splincode:
@waterplea Hello, please give me idea, I don't know how fix without my current solution 

image

i tried to use element getter but it is always undefined, and only after a few macrotasks, it becomes a value

also changing the method content doesn't help 

image

PR #4364, opened 2023-05-03
feat(core): add `tui-for` mixin for compatibility old stackblitz less version

splincode:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring

image

### DIFF
--- a/projects/demo/src/modules/markup/breakpoints/examples/1/index.less
+++ b/projects/demo/src/modules/markup/breakpoints/examples/1/index.less
@@ -26,7 +26,7 @@
 @breakpoints: tui-mobile, tui-mobile-min, tui-mobile-interval, tui-tablet, tui-tablet-min, tui-tablet-interval,
     tui-desktop, tui-desktop-min, tui-desktop-interval, tui-desktop-lg-min;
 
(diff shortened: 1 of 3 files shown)


projects/demo/src/modules/markup/breakpoints/examples/1/index.less
```diff
@@ -26,7 +26,7 @@
 @breakpoints: tui-mobile, tui-mobile-min, tui-mobile-interval, tui-tablet, tui-tablet-min, tui-tablet-interval,
     tui-desktop, tui-desktop-min, tui-desktop-interval, tui-desktop-lg-min;
 
-each(@breakpoints, {
+.tui-for(@breakpoints, {
```
nsbarsukov, line 29: I am not sure that we need to expand public API only for Stackblitz.

Maybe we can change only this file (replace this modern `each` by legacy-way loop)?
Or should we make "magic" inside this file ?
https://github.com/Tinkoff/taiga-ui/blob/7f1fdf880f8268d0cb207a346468c81b31d4cddd/projects/demo/src/modules/app/stackblitz/utils.ts#L6

PR #4368, opened 2023-05-03
feat(kit): `InputNumber` add `step` input and deprecate `InputCount`

waterplea:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring

Closes # 

- [ ] Yes
- [ ] No


splincode (approved):

projects/kit/components/input-number/input-number.component.ts
```diff
@@ -149,6 +157,26 @@ export class TuiInputNumberComponent
...
+    get canDecrement(): boolean {
+        return this.interactive && (this.value || 0) > this.min;
+    }
+
+    get canIncrement(): boolean {
+        return this.interactive && (this.value || 0) < this.max;
+    }
+
+    @HostListener('keydown.arrowDown', ['-step'])
+    @HostListener('keydown.arrowUp', ['step'])
```
nsbarsukov, line 169: Caret jumps to the end/start on pressing this keys


https://user-images.githubusercontent.com/35179038/236197719-89472675-1dab-4d09-ae09-cd1e3ea453af.mov

nsbarsukov (approved):

projects/kit/components/input-number/input-number.component.ts
waterplea: Native input number kind of behaves the same, caret jumps at the end, just not at the beginning. Let's keep it as is for the time being and see how it will behave with the new mask first.

vladimirpotekhin (approved):

nsbarsukov (approved):

Issue #120, opened 2021-01-20
🛠 - Investigate other mask options


SergiusSidorov:
# 🚀 Feature request

### Is your feature request related to a problem?

Last release of angular2-text-mask  was about 3 years ago. Currently this project is not maintained. Using this library results in a build-time warning:
_Warning: /node_modules/@taiga-ui/kit/__ivy_ngcc__/fesm2015/taiga-ui-kit-components-input-range.js depends on 'angular2-text-mask'. CommonJS or AMD dependencies can cause optimization bailouts.
For more info see: https://angular.io/guide/build#configuring-commonjs-dependencies_

There is a possibility to switch off this warning. But I'm always frustrated when something can cause optimization bailouts. =)

### Describe the solution you'd like

Maybe it would be better to migrate to ngx-mask (https://github.com/JsDaddy/ngx-mask). This project looks alive.

ramax495:
ngx-mask doesn't work on Ionic.
Just for info)

waterplea:
Thank you for raising this issue. Yes, we would like to migrate to another mask solution at some point. 

I have refactored work with mask in input a little. Now if you were to use angular2-text-mask — you could set it directly on `tui-input`. This way Taiga UI _supports_ angular2-text-mask but doesn't enforce it on you, as end user. It is still required by our internal implementation for various masked inputs, such as InputDate, InputNumber etc. But that could be refactored without breaking changes in future when we will have resources and a good candidate library (or have written our own).

Phmager:
I can highly recommend angular-imask which uses imask.js under the hood. Which would be a bit bigger in comparison

Fafnur:
Don't use ngx-mask.

zy2ba:
> Don't use ngx-mask.

Support this opinion. We take it for our project and have some pain moments.

Primarily, ngx-mask works normally only with type=text. Mobile users be frustrated by strange keyboard type selection with it

Yohandah:
Also, this is blocking because we are unable to run Unit Tests because of this lib https://github.com/text-mask/text-mask/issues/1019

`Error: TextMaskModule does not have a module def (ɵmod property)`

hakimio:
Maybe [@ngneat/input-mask](https://github.com/ngneat/input-mask) could be a good alternative?

waterplea:
[...]

PR #4382, opened 2023-05-04
feat(kit): `InputYear` uses `Maskito` instead of legacy `text-mask`

nsbarsukov:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring

Partially solves #120

- [ ] Yes
- [X] No


projects/kit/components/input-year/input-year.component.ts
```diff
@@ -88,8 +84,20 @@ export class TuiInputYearComponent
         return this.options.icon;
     }
 
+    @tuiPure
+    getMaskOptions(max: number): MaskitoOptions {
+        const numberOptions = maskitoNumberOptionsGenerator({
```
waterplea, line 89: Looks like unnecessary const to me :)

projects/kit/components/input-year/input-year.component.ts
```diff
@@ -88,8 +84,20 @@ export class TuiInputYearComponent
...
+            max,
+            precision: 0,
+            isNegativeAllowed: false,
```
waterplea, line 92: Do `precision` and `isNegativeAllowed` matter if we override the `mask` anyway?

waterplea (approved):

projects/kit/components/input-year/input-year.component.ts
nsbarsukov: You are right, I should drop it.

nsbarsukov: fixed

projects/kit/components/input-year/input-year.component.ts
nsbarsukov: fixed

splincode (approved):

PR #4385, opened 2023-05-05
feat(addon-commerce): `InputCard` uses `Maskito` instead of legacy `text-mask`

nsbarsukov:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring

Partially solves #120

- [ ] Yes
- [X] No

### DIFF
--- a/projects/addon-commerce/components/input-card/input-card.component.ts
+++ b/projects/addon-commerce/components/input-card/input-card.component.ts
@@ -8,9 +8,10 @@ import {
     ViewChild,
 } from '@angular/core';
 import {NgControl} from '@angular/forms';
+import {MaskitoOptions, maskitoTransform} from '@maskito/core';
 import {TUI_CARD_MASK} from '@taiga-ui/addon-commerce/constants';
 import {tuiAsControl, tuiAsFocusableItemAccessor, tuiPure} from '@taiga-ui/cdk';
-import {TuiPrimitiveTextfieldComponent, TuiTextMaskOptions} from '@taiga-ui/core';
+import {TuiPrimitiveTextfieldComponent} from '@taiga-ui/core';
 
 import {AbstractTuiInputCard} from './abstract-input-card';
 import {TUI_INPUT_CARD_OPTIONS, TuiInputCardOptions} from './input-card.providers';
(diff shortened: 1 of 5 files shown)


projects/addon-commerce/components/input-card/input-card.component.ts
```diff
@@ -97,9 +96,6 @@ export class TuiInputCardComponent extends AbstractTuiInputCard<string> {
 
     @tuiPure
     private getFormattedCard(cardNumber: string): string {
-        return cardNumber
-            .split('')
-            .map((char, index) => (index && index % 4 === 0 ? ` ${char}` : char))
-            .join('');
+        return maskitoTransform(cardNumber, this.maskOptions);
```
nsbarsukov, line 99: oops, mistake

I should change it to
```ts
    private getFormattedCard(cardNumber: string, options: MaskitoOptions): string {
        return maskitoTransform(cardNumber, options);
    }
```

nsbarsukov: fixed

projects/addon-commerce/components/input-card/test/input-card.component.spec.ts
```diff
@@ -156,6 +156,22 @@ describe(`InputCard`, () => {
...
+
+        it(`7`, async () => {
+            await testFormat(`1234567`, `1234 567`);
+        });
+
+        it(`10`, async () => {
+            await testFormat(`1234567890`, `1234 5678 90`);
+        });
+
+        it(`12`, async () => {
```
splincode, line 171: The card number can have 13, 15, 16, 18 and even 19 digits. The most common option is 16 digits.
Do we have test suites for that?

nsbarsukov: Yep, few lines below this comment)

projects/addon-commerce/components/input-card/input-card.component.ts
```diff
@@ -59,7 +58,7 @@ export class TuiInputCardComponent extends AbstractTuiInputCard<string> {
     }
 
     get formattedCard(): string {
```
waterplea, line 60: This is only used in template, we can use maskito pipe.

waterplea (commented):
What about input card inside `InputCardGroup`?

projects/addon-commerce/components/input-card/input-card.component.ts
nsbarsukov: You are right, I've forgot about this pipe))

nsbarsukov:
> What about input card inside InputCardGroup?

@waterplea I will do it in next separate PR

projects/addon-commerce/components/input-card/input-card.component.ts
nsbarsukov: fixed

waterplea (approved):

vladimirpotekhin (approved):

PR #4402, opened 2023-05-06
feat(core): add `tuiSvgSrcInterceptors` for multiple source processing

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

in proprietary repo

```ts
{
        provide: TUI_SVG_SRC_PROCESSOR,
        useFactory:
            (base: string): TuiStringHandler<string> =>
            src =>
                tuiIsString(src) && src.startsWith(`tuiIconTds`)
                    ? `${base}/${src}.svg`
                    : src,
        deps: [TUI_ICONS_PLACE],
    },
```

the main problem is that your users can overwrite default behavior

## What is the new behavior?

```ts
tuiSvgSrcInterceptors((src: TuiSafeHtml) =>
    String(src).startsWith(`icons8::`)
        ? `assets/icons8/${String(src).replace(`icons8::`, ``)}.svg`
        : src,
),
tuiSvgSrcInterceptors((src: TuiSafeHtml) =>
    String(src).startsWith(`tuiIconTds`)
        ? `assets/design-tokens/${String(src)}.svg`
        : src,
)
// ...
```

### DIFF
--- a/projects/core/components/svg/svg-options.ts
+++ b/projects/core/components/svg/svg-options.ts
@@ -1,8 +1,15 @@
-import {FactoryProvider, inject, InjectionToken, Optional, SkipSelf} from '@angular/core';
-import {SafeHtml} from '@angular/platform-browser';
+import {
+    FactoryProvider,
+    inject,
+    InjectionToken,
+    Optional,
+    Provider,
+    SkipSelf,
+} from '@angular/core';
 import {
     TuiHandler,
     tuiIsString,
+    TuiSafeHtml,
     TuiStringHandler,
     tuiSvgLinearGradientProcessor,
 } from '@taiga-ui/cdk';
@@ -19,8 +26,8 @@ import {TUI_DEPRECATED_ICONS} from './deprecated-icons';
 export interface TuiSvgOptions {
     readonly path: TuiStringHandler<string>;
     readonly deprecated: TuiStringHandler<string>;
-    readonly srcProcessor: TuiHandler<SafeHtml | string, SafeHtml | string>;
-    readonly contentProcessor: TuiHandler<SafeHtml | string, SafeHtml | string>;
+    readonly srcProcessor: TuiHandler<TuiSafeHtml, TuiSafeHtml>;
+    readonly contentProcessor: TuiHandler<TuiSafeHtml, TuiSafeHtml>;
 }
 
 export const TUI_SVG_DEFAULT_OPTIONS: TuiSvgOptions = {
@@ -48,6 +55,20 @@ export const TUI_SVG_OPTIONS = new InjectionToken<TuiSvgOptions>(`[TUI_SVG_OPTIO
         contentProcessor: inject(TUI_SVG_CONTENT_PROCESSOR),
     }),
 });
+
+export const TUI_SVG_SRC_INTERCEPTORS = new InjectionToken<
+    TuiHandler<TuiSafeHtml, TuiSafeHtml>
+>(`[TUI_SVG_SRC_INTERCEPTORS]`);
+
+export function tuiSvgSrcInterceptors(
+    interceptor: TuiHandler<TuiSafeHtml, TuiSafeHtml>,
+): Provider {
+    return {
+        provide: TUI_SVG_SRC_INTERCEPTORS,
+        useValue: interceptor,
+        multi: true,
+    };
+}
 
 export const tuiSvgOptionsProvider: (
     options: Partial<Omit<TuiSvgOptions, 'path'>> & {
--- a/projects/core/components/svg/svg.component.ts
+++ b/projects/core/components/svg/svg.component.ts
@@ -14,9 +14,11 @@ import {WINDOW} from '@ng-web-apis/common';
 import {
     tuiAssert,
     tuiGetDocumentOrShadowRoot,
+    TuiHandler,
     tuiIsString,
     tuiPure,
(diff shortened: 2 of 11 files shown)


projects/core/components/svg/svg-options.ts
```diff
@@ -49,6 +56,23 @@ export const TUI_SVG_OPTIONS = new InjectionToken<TuiSvgOptions>(`[TUI_SVG_OPTIO
...
+export function tuiSvgMultiSrcProcessor(
+    handler: TuiHandler<TuiSafeHtml, TuiSafeHtml>,
+): Provider {
+    return {
+        multi: true,
+        provide: TUI_SVG_MULTI_SRC_PROCESSOR,
+        useFactory: (
+            parentHandlers: Array<TuiHandler<TuiSafeHtml, TuiSafeHtml>> | null,
+        ) => [...(parentHandlers ?? []), handler],
+        deps: [[new SkipSelf(), new Optional(), TUI_SVG_MULTI_SRC_PROCESSOR]],
```
waterplea, line 72: nit: let's move deps above, because they kind of describe what's going in the factory.

projects/core/components/svg/svg.component.ts
```diff
@@ -40,10 +43,15 @@ const FAILED_EXTERNAL_ICON = 'Failed to load external SVG';
     templateUrl: './svg.template.html',
     styleUrls: ['./svg.style.less'],
     changeDetection: ChangeDetectionStrategy.OnPush,
+    providers: [],
```
waterplea, line 46: Why?

projects/core/components/svg/svg.component.ts
```diff
@@ -40,10 +43,15 @@ const FAILED_EXTERNAL_ICON = 'Failed to load external SVG';
...
 })
 export class TuiSvgComponent {
+    private readonly srcProcessors = tuiFlatten(
```
waterplea, line 49: Why do we need to flatten them?

waterplea (commented):
Is the old processor just ignored now?

splincode:
`Is the old processor just ignored now?` no, but many users override our default `TUI_SVG_SRC_PROCESSOR`

and then in prorprietary repo we can remove it

```ts
{
        provide: TUI_SVG_SRC_PROCESSOR,
        useFactory:
            (base: string): TuiStringHandler<string> =>
            src =>
                tuiIsString(src) && src.startsWith(`tuiIconTds`)
                    ? `${base}/${src}.svg`
                    : src,
        deps: [TUI_ICONS_PLACE],
},
```

and then we can add new provider into TuiProprietaryModule

```ts
tuiSvgMultiSrcProcessor((src: TuiSafeHtml) =>
    String(src).startsWith(`tuiIconTds`)
        ? `assets/design-tokens/${String(src)}.svg`
        : src,
```

projects/core/components/svg/svg.component.ts
splincode: `Why do we need to flatten them?` without it I get array in array error

projects/core/components/svg/svg-options.ts
splincode: fixed

projects/core/components/svg/svg.component.ts
waterplea: I don't get it. How does it become multi-dimensional array?

projects/core/components/svg/svg.component.ts
splincode: fixed

projects/core/components/svg/svg.component.ts
waterplea: Got it. This way if you provide multiple processors, each of them would have all the parent processors attached. Do we really need to inherit from parent here?

splincode: `Do we really need to inherit from parent here?`  yes

otherwise component provider will override the default `tuiSvgMultiSrcProcessor` processors that can be provided by us in `proprietary-root.module.ts` or `root.module.ts`

waterplea: We can document it as a global token maybe? Is there any merit to hierarchy in its case?

splincode: `TUI_SVG_SRC_PROCESSOR` was a global token 

and thus it causes some problems where users can override the default behaviour which may be necessary

splincode: fixed

waterplea (approved):

projects/demo/src/modules/markup/icons/customization/customization-icons.component.ts
```diff
@@ -15,15 +16,19 @@ import {NgDompurifySanitizer} from '@tinkoff/ng-dompurify';
...
-                return String(src).startsWith(myCustomPrefix)
-                    ? `assets/icons8/${String(src).replace(myCustomPrefix, '')}.svg`
-                    : src;
-            },
-        }),
+        /**
+         * @note:
+         * Be careful, component has its own injector which doesn't
+         * inherit providers from parent injector. If you want to keep
+         * the module's injectors, then use the provider in the module,
```
waterplea, line 23: It would also lose global interceptors if it's a lazy module. Better to write something like "It's best to keep it global and provide all necessary interceptors at the top level of your app".

PR #4406, opened 2023-05-08
test: improve cdk coverage

splincode:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring

Bump coverage from `61%` to `70%`

image

### DIFF
--- a/projects/cdk/directives/active-zone/tests/active-zone.directive.spec.ts
+++ b/projects/cdk/directives/active-zone/tests/active-zone.directive.spec.ts
@@ -1,0 +1,154 @@
+import {CommonModule} from '@angular/common';
+import {Component} from '@angular/core';
+import {ComponentFixture, TestBed} from '@angular/core/testing';
+import {FormControl, ReactiveFormsModule} from '@angular/forms';
+import {TuiActiveZoneModule} from '@taiga-ui/cdk';
+import {configureTestSuite} from '@taiga-ui/testing';
+
+describe(`TuiActiveZoneDirective`, () => {
+    @Component({
+        template: `
+            <p id="parent-info">Parent zone: {{ parentActive }}</p>
+            <p id="child-info">Child zone: {{ childActive }}</p>
+
+            <div
+                #parent="tuiActiveZone"
+                class="active-zone"
+                [class.active-zone_active]="parentActive"
+                (tuiActiveZoneChange)="onParentActiveZone($event)"
+            >
+                <h2>Parent zone</h2>
+                <button
+                    id="button-inside-zone"
+                    type="button"
+                >
+                    A button inside zone
+                </button>
+            </div>
+
+            <div>
+                <button
+                    id="button-outside-zone"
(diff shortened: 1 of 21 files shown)


projects/cdk/utils/dom/test/is-element-editable.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+        element.setAttribute(`type`, `text`);
+        expect(tuiIsElementEditable(element)).toBe(true);
+    });
+
+    it(`should return true for editable textarea`, () => {
+        element = document.createElement(`textarea`);
+        expect(tuiIsElementEditable(element)).toBe(true);
+    });
+
+    it(`should return false for readonly textfield`, () => {
```
waterplea, line 20: We should also check it returning false for `type="checkbox"` for example.

splincode: why returning false?

splincode: image

but `tuiIsElementEditable` returning true

splincode: done

projects/cdk/directives/active-zone/tests/active-zone.directive.spec.ts
```diff
@@ -0,0 +1,154 @@
+import {CommonModule} from '@angular/common';
+import {Component} from '@angular/core';
+import {ComponentFixture, TestBed} from '@angular/core/testing';
+import {FormControl, ReactiveFormsModule} from '@angular/forms';
+import {TuiActiveZoneModule} from '@taiga-ui/cdk';
+import {configureTestSuite} from '@taiga-ui/testing';
+
+describe(`TuiActiveZoneDirective`, () => {
```
waterplea, line 8: Check out this PR:
https://github.com/Tinkoff/taiga-ui/pull/4420
I wonder if jest can properly handle that :)

splincode:
@waterplea @nsbarsukov @vladimirpotekhin hello, please review me)

projects/cdk/directives/checked/tests/checked.directive.spec.ts
```diff
@@ -0,0 +1,85 @@
+import {ElementRef, Renderer2} from '@angular/core';
+import {TuiCheckedDirective} from '@taiga-ui/cdk';
+
+describe(`TuiCheckedDirective`, () => {
+    const spyRendererSetProperty = jest.fn();
+    let directive: TuiCheckedDirective;
+    let el: ElementRef<HTMLInputElement>;
+    let renderer: Renderer2;
+
+    class MockRenderer {
```
waterplea, line 10: We should probably just drop `Renderer` from that directive. BTW, can we just do `HostBinding` instead?

splincode: fixed

PR #4409, opened 2023-05-09
docs: add custom content example to line-clamp

### LINKED ISSUE #4392: 🚀 - Add an example of how to use tui-line-clamp with custom content
### Description

Imagine that we need to display a user's email, which could be long. To truncate it, we can use tui-line-clamp. However, when the user hovers over the email, we want to show not just the full email, but also the user's ID, full name, email, and login. To achieve this, we came up with the idea of using a hint. But the issue is that tui-line-clamp and hint do not work together.

So It would be helpful if the documentation included an example of this this scenario.

Here are possible solutions:

1. Instead of using tui-line-clamp, use overflow: hidden and text-overflow: ellipsis to truncate the email.

2. Create a template for tui-line-clamp that displays the email on the first line, followed by the other user information on subsequent lines + limit the display to just one line.

ddubrava:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [x] Documentation content changes

Closes #4392

## What is the new behavior?

Added a few examples of how to show custom content in `tui-line-clamp` 

- [ ] Yes
- [x] No

### DIFF
--- a/projects/demo/src/modules/components/line-clamp/examples/5/index.html
+++ b/projects/demo/src/modules/components/line-clamp/examples/5/index.html
@@ -1,0 +1,27 @@
+<p>1. Display only the first line, in a popup display remaining lines.</p>
+
+<tui-island class="island">
+    <tui-line-clamp
+        [content]="userAdditionalInfo"
+        [linesLimit]="1"
+    ></tui-line-clamp>
+</tui-island>
+
+<p>2. Do not use `tui-line-clamp`, use `text-overflow: ellipsis` instead.</p>
+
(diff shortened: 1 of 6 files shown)


splincode (approved):

waterplea:
> hey! is it expected? is there no `getBoundingClientRect` logic?
> 
> image
> image
> 
> 

I suspect it is by design, otherwise visible parts will not align upon hovering. It is rarely the case that it is at the very bottom of the page and there's nowhere to scroll.

PR #4431, opened 2023-05-11
fix(kit): downgrade type of value in badge

splincode:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

### DIFF
--- a/projects/kit/components/badged-content/badged-content.component.ts
+++ b/projects/kit/components/badged-content/badged-content.component.ts
@@ -15,9 +15,9 @@ const BADGE_SIZE: {[key: string]: TuiSizeL | TuiSizeXS} = {
 
 @Component({
     selector: 'tui-badged-content',
-    changeDetection: ChangeDetectionStrategy.OnPush,
     templateUrl: './badged-content.template.html',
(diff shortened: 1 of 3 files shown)


projects/kit/components/badged-content/badged-content.component.ts
```diff
@@ -22,11 +22,11 @@ const BADGE_SIZE: {[key: string]: TuiSizeL | TuiSizeXS} = {
 export class TuiBadgedContentComponent {
     @Input()
     @HostBinding('class._with-top')
-    contentTop: PolymorpheusContent;
+    contentTop: PolymorpheusPrimitive;
```
waterplea, line 25: It can be full `PolymorpheusContent` here. Probably just need to change `contentTop/Bottom` to `content` in the template.

waterplea (changes requested):

projects/kit/components/badged-content/badged-content.component.ts
splincode: fixed

waterplea (approved):

PR #4437, opened 2023-05-11
chore: add title example to the demo

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring

image

## What is the new behavior?

image

image

### DIFF
--- a/projects/kit/components/input-files/input-files.directive.ts
+++ b/projects/kit/components/input-files/input-files.directive.ts
@@ -27,27 +27,22 @@ export class TuiInputFilesDirective {
 
     @HostBinding('id')
     get id(): string {
-        return this.input.id || this.idService.generate();
+        return this.el.nativeElement.id || this.idService.generate();
     }
 
     @HostBinding('accept')
     get accept(): string {
-        return this.input.accept ?? this.options.accepts;
+        return this.el.nativeElement.accept ?? this.options.accepts;
     }
 
     @HostBinding('multiple')
     get multiple(): boolean {
-        return this.input.multiple ?? this.options.multiple;
-    }
-
-    @HostBinding('title')
-    get title(): string {
-        return this.input.getAttribute('title') ?? this.options.title;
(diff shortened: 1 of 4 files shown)


projects/kit/components/input-files/input-files.template.html
```diff
@@ -31,6 +31,7 @@
             <input
                 #input
                 type="file"
+                title=""
```
nsbarsukov, line 34: `options.title` ?

nsbarsukov (approved):

projects/kit/components/input-files/input-files.template.html
splincode: @waterplea what do you think?

splincode: fixed

waterplea (commented):
Our stance is to use native `input` if you need native attributes. Let's not add it to the options.

waterplea:
Maybe just hardcode empty line for the built-in input?

splincode:
@waterplea fixed

projects/kit/components/input-files/input-files.directive.ts
```diff
@@ -27,22 +27,27 @@ export class TuiInputFilesDirective {
...
     }
 
     @HostBinding('multiple')
     get multiple(): boolean {
-        return this.el.nativeElement.multiple ?? this.options.multiple;
+        return this.input.multiple ?? this.options.multiple;
+    }
+
+    @HostBinding('title')
+    get title(): string | null {
```
waterplea, line 44: Is this necessary?

waterplea (approved):
I guess this should be called `chore: add title example to the demo` now )

projects/kit/components/input-files/input-files.directive.ts
splincode: fixed

PR #4471, opened 2023-05-16
fix(addon-editor): focus should trigger only for outside contenteditable area

splincode:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring

@waterplea sorry, but `mousedown` fixed focus, but now it doesn't respond at all
https://taiga-ui.dev/next/editor/API

image

### DIFF
--- a/projects/addon-editor/components/editor/editor.component.ts
+++ b/projects/addon-editor/components/editor/editor.component.ts
@@ -164,6 +164,15 @@ export class TuiEditorComponent
         this.editor?.unsetLink();
     }
 
+    focus(event: MouseEvent): void {
+        if (this.nativeFocusableElement?.contains(event.target as Node | null)) {
+            return;
+        }
+
+        event.preventDefault();
+        this.nativeFocusableElement?.focus();
+    }
+
     override ngOnDestroy(): void {
         this.editor?.destroy();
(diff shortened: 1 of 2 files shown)


waterplea:
It should be like in textfields, only prevent and focus if clicked outside of actual focusable part:
https://github.com/Tinkoff/taiga-ui/blob/main/projects/core/components/primitive-textfield/primitive-textfield.component.ts#L289-L298

splincode:
@waterplea fixed

projects/addon-editor/components/editor/editor.component.ts
```diff
@@ -164,6 +165,15 @@ export class TuiEditorComponent
         this.editor?.unsetLink();
     }
 
+    focus(event: MouseEvent): void {
+        if (tuiIsNativeFocused(this.nativeFocusableElement)) {
```
waterplea, line 169: You only need to act if clicked target is not inside the editable part. Otherwise first focusing click sets caret at the beginning and not at the place inside the text that was clicked.

splincode: fixed

waterplea (approved):

PR #4522, opened 2023-05-23
feat(kit): `Select` add ability to use objects in native select

vladimirpotekhin:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #

### DIFF
--- a/projects/kit/components/select/native-select/native-select-group.component.ts
+++ b/projects/kit/components/select/native-select/native-select-group.component.ts
@@ -1,7 +1,8 @@
 import {ChangeDetectionStrategy, Component, Input, TemplateRef} from '@angular/core';
-import {tuiAsDataList} from '@taiga-ui/core';
+import {tuiAsDataList, TuiValueContentContext} from '@taiga-ui/core';
 import {AbstractTuiNativeSelect} from '@taiga-ui/kit/abstract';
 import {TuiItemsHandlers} from '@taiga-ui/kit/tokens';
+import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
 
 import type {TuiSelectDirective} from '../select.directive';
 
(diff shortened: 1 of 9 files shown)


projects/kit/components/select/select.directive.ts
```diff
@@ -18,7 +19,15 @@ export class TuiSelectDirective extends AbstractTuiTextfieldHost<
         return this.host.computedValue;
     }
 
-    onValueChange(value: string): void {
+    get rawValue(): unknown {
```
waterplea, line 22: What if inject NgControl separately, instead of extending this abstract class?

projects/kit/components/select/native-select/native-select.template.html
```diff
@@ -4,9 +4,10 @@
...
 <option
-    *ngFor="let option of items"
-    [value]="option"
+    *ngFor="let option of items; let index = index"
+    [value]="stringify(option)"
+    [selected]="selected(option)"
     [disabled]="disabledItemHandler ? disabledItemHandler(option) : itemsHandlers.disabledItemHandler(option)"
 >
-    {{ option }}
+    {{ stringify(option) }}
```
waterplea, line 12: I wonder if we can not write this twice (here and in value binding). But maybe it's good just to be safe in terms of whitespace...

waterplea (approved):

projects/kit/components/select/native-select/native-select-group.component.ts
```diff
@@ -22,15 +26,35 @@ import {AbstractTuiNativeSelect} from '@taiga-ui/kit/abstract';
...
+    T
+> {
     @Input()
-    items: readonly string[][] | null = [];
+    items: readonly T[][] | null = [];
 
     @Input()
     labels: readonly string[] = [];
+
+    itemContent: PolymorpheusContent<TuiValueContentContext<T>> = ({$implicit}) =>
```
waterplea, line 44: You can use `stringify | tuiStringifyContent` in the template.

waterplea (approved):

nsbarsukov (approved):

PR #4533, opened 2023-05-25
fix: hosts of portals can throw `ExpressionChangedAfterItHasBeenCheckedError`

### LINKED ISSUE #4473: 🐞 - ExpressionChangedAfterItHasBeenCheckedError when hiding a button with hint active
### Playground Link

https://stackblitz.com/edit/angular-llisre?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.ts&hideExplorer=1

### Description

Firstly, please note that the issue is not reproduced in the linked code above. However it demonstrates roughly how my code was working.

Specifically, this came from a discussion on discord:

I was having an issue where a div containing a button with a hint was being hidden while the hint was active. The `ExpressionChangedAfterItHasBeenCheckedError` was being thrown and the error was pointing to the `ngFor` inside the `tui-hints-host` component.

From discord, @waterplea

> I suppose we need something like rx-angular push pipe, rather than async pipe in all those hosts, because since they are parallel views to the content — it might so happen that they are already checked at the time pipe calls markForCheck and it doesn't do anything. I believe I saw something like that with dialogs. Can you raise an issue quoting me in its text?

### Angular version

15.x

### Taiga UI version

3.26.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

evantrimboli:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #4473

### DIFF
--- a/projects/addon-mobile/components/sheet/components/sheets-host/sheets-host.component.ts
+++ b/projects/addon-mobile/components/sheet/components/sheets-host/sheets-host.component.ts
@@ -9,9 +9,10 @@ import {
 import {TUI_WINDOW_HEIGHT, TuiDestroyService} from '@taiga-ui/cdk';
 import {TUI_ANIMATION_OPTIONS, tuiFadeIn, tuiSlideInTop} from '@taiga-ui/core';
 import {Observable} from 'rxjs';
+import {takeUntil} from 'rxjs/operators';
+
 import {TuiSheet} from '../../sheet';
 import {TuiSheetService} from '../../sheet.service';
-import {takeUntil} from 'rxjs/operators';
 
 @Component({
     selector: 'tui-sheets-host',
@@ -21,7 +22,7 @@ import {takeUntil} from 'rxjs/operators';
     animations: [tuiSlideInTop, tuiFadeIn],
 })
 export class TuiSheetsHostComponent {
-    public sheets: readonly TuiSheet<any>[] = [];
+    sheets: ReadonlyArray<TuiSheet<any>> = [];
 
     constructor(
         @Inject(TUI_ANIMATION_OPTIONS) readonly options: AnimationOptions,
--- a/projects/addon-mobile/components/sheet/components/sheets-host/sheets-host.component.ts
+++ b/projects/addon-mobile/components/sheet/components/sheets-host/sheets-host.component.ts
@@ -1,11 +1,17 @@
 import {AnimationOptions} from '@angular/animations';
-import {ChangeDetectionStrategy, Component, Inject} from '@angular/core';
-import {TUI_WINDOW_HEIGHT} from '@taiga-ui/cdk';
+import {
+    ChangeDetectionStrategy,
+    ChangeDetectorRef,
+    Component,
+    Inject,
+    Self,
+} from '@angular/core';
+import {TUI_WINDOW_HEIGHT, TuiDestroyService} from '@taiga-ui/cdk';
 import {TUI_ANIMATION_OPTIONS, tuiFadeIn, tuiSlideInTop} from '@taiga-ui/core';
 import {Observable} from 'rxjs';
-
 import {TuiSheet} from '../../sheet';
 import {TuiSheetService} from '../../sheet.service';
+import {takeUntil} from 'rxjs/operators';
 
 @Component({
     selector: 'tui-sheets-host',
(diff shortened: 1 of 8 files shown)


projects/addon-mobile/components/sheet/components/sheets-host/sheets-host.component.ts
```diff
@@ -1,11 +1,17 @@
...
+    Inject,
+    Self,
+} from '@angular/core';
+import {TUI_WINDOW_HEIGHT, TuiDestroyService} from '@taiga-ui/cdk';
 import {TUI_ANIMATION_OPTIONS, tuiFadeIn, tuiSlideInTop} from '@taiga-ui/core';
 import {Observable} from 'rxjs';
-
 import {TuiSheet} from '../../sheet';
 import {TuiSheetService} from '../../sheet.service';
+import {takeUntil} from 'rxjs/operators';
```
waterplea, line 14: Looks like you need to run lint fix to order imports. It should have been ran automatically in precommit hook. Don't skip them :)

projects/cdk/components/alert-host/alert-host.component.ts
```diff
@@ -23,10 +27,23 @@ import {Observable} from 'rxjs';
     encapsulation: ViewEncapsulation.None,
 })
 export class TuiAlertHostComponent<T extends TuiDialog<unknown, unknown>> {
+    public alerts: readonly (readonly T[])[] = [];
```
waterplea, line 30: I think there's a lint rule for such types to be written as `ReadonlyArray<readonly T[]>`. We only use `readonly` modifier for simple array types, otherwise it becomes hard to read.

projects/cdk/components/dialog-host/dialog-host.component.ts
```diff
@@ -33,22 +43,42 @@ const isFakeHistoryState = (
...
 export class TuiDialogHostComponent<T extends TuiDialog<unknown, unknown>> {
-    readonly dialogs$ = combineLatest(this.dialogsByType).pipe(
-        map(allTypesDialogs =>
-            new Array<T>()
-                .concat(...allTypesDialogs)
-                .sort((a, b) => a.createdAt - b.createdAt),
-        ),
-    );
+    public dialogs: readonly T[] = [];
+    public isDialogClosesOnBack = defaultClosesOnBack;
```
waterplea, line 47: You can keep initial implementation of this. It is never changing amid change detection cycle, unlike dialogs list, so it will not present that issue.

projects/cdk/components/dialog-host/dialog-host.component.ts
```diff
@@ -33,22 +43,42 @@ const isFakeHistoryState = (
...
-    ) {}
+        @Self() @Inject(TuiDestroyService) destroy$: Observable<void>,
+        @Inject(ChangeDetectorRef) cdr: ChangeDetectorRef,
+    ) {
+        // Due to this view being parallel to app content, `markForCheck` from `async` pipe
+        // can happen after view was checked, so calling `detectChanges` instead
+        combineLatest(dialogsByType)
+            .pipe(
+                map(arr =>
+                    new Array<T>()
```
waterplea, line 64: What's the point of `new Array<T>().concat`? Why not just `[...arr]`?

projects/addon-mobile/components/sheet/components/sheets-host/sheets-host.component.ts
evantrimboli: I didn't skip them. Seems like whatever is setup didn't work for me. Maybe because I'm running Windows.

projects/cdk/components/dialog-host/dialog-host.component.ts
evantrimboli: `[...arr]` wouldn't flatten the array properly. I had originally used `arr.flat()` however at build time it threw an error because it's only targeting ES2019.

projects/addon-mobile/components/sheet/components/sheets-host/sheets-host.template.html
```diff
@@ -1,6 +1,6 @@
-<ng-container *ngIf="service.sheets$ | async as items">
+<ng-container *ngIf="sheets.length > 0">
```
waterplea, line 1: We do not need this `ng-container` and `*ngIf`

projects/cdk/components/alert-host/alert-host.component.ts
```diff
@@ -1,16 +1,20 @@
...
+    ChangeDetectorRef,
     Component,
     Inject,
     INJECTOR,
     Injector,
+    Self,
     ViewEncapsulation,
 } from '@angular/core';
 import {TUI_PARENT_ANIMATION} from '@taiga-ui/cdk/constants';
+import {TuiDestroyService} from '@taiga-ui/cdk/services/destroy.service';
```
waterplea, line 12: To fix build please change that entry point to `@taiga-ui/cdk/services`

projects/cdk/components/dialog-host/dialog-host.component.ts
```diff
@@ -1,11 +1,19 @@
...
+    ChangeDetectorRef,
+    Component,
+    Inject,
+    InjectionToken,
+    Self,
+} from '@angular/core';
 import {Title} from '@angular/platform-browser';
 import {HISTORY} from '@ng-web-apis/common';
 import {TUI_PARENT_ANIMATION} from '@taiga-ui/cdk/constants';
+import {TuiDestroyService} from '@taiga-ui/cdk/services/destroy.service';
```
waterplea, line 12: Same here

projects/addon-mobile/components/sheet/components/sheets-host/sheets-host.template.html
evantrimboli: What about the overlay? Without the container it would always render that and it has a reference to `sheets[0]`.

waterplea: It is hidden with CSS unless there is a Sheet with overlay visible.

waterplea:
Looks like you forgot to add TuiDestroyService to providers for all the hosts.

waterplea:
Looks like you cannot run `detectChanges` inside constructor and all those streams emit synchronously:
https://stackoverflow.com/questions/65712096/how-to-fix-error-assertion-error-should-be-run-in-update-mode-expected-fals
Moving to `ngOnInit` should solve this issue.

waterplea (approved):
Thank you for the contribution 👍

nsbarsukov (approved):

PR #4632, opened 2023-06-10
fix(cdk): using trackBy for re-order animations

### LINKED ISSUE #4610: 🐞 - Alert animation is repeated for all alerts instead of new one
### Playground Link

https://taiga-ui.dev/services/alert-service

### Description

When opening a second alert, the alert animation is repeated for all alerts, instead of only the second one. Same when alerts are closed.

The behavior changed somewhere between v3.24.0 and v3.30.0. Not sure which exact version.

Reproduction: 
- open https://taiga-ui.dev/services/alert-service
- click "Show" button several times

Current behavior (taiga-ui v3.30.0):

https://github.com/Tinkoff/taiga-ui/assets/24232962/a4834bd4-1874-4a88-939e-1b9ccf31a1c7

Expected behavior (taiga-ui v3.24.0):

https://github.com/Tinkoff/taiga-ui/assets/24232962/07cdb001-1d90-4488-9618-039327f55093

### Angular version

16

### Taiga UI version

3.30.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #4610

### DIFF
--- a/projects/cdk/components/alert-host/alert-host.component.ts
+++ b/projects/cdk/components/alert-host/alert-host.component.ts
@@ -7,6 +7,7 @@ import {
     Injector,
     OnInit,
     Self,
+    TrackByFunction,
     ViewEncapsulation,
 } from '@angular/core';
 import {TUI_PARENT_ANIMATION} from '@taiga-ui/cdk/constants';
@@ -39,6 +40,8 @@ export class TuiAlertHostComponent<T extends TuiDialog<unknown, unknown>>
         @Self() @Inject(TuiDestroyService) private readonly destroy$: Observable<void>,
         @Inject(ChangeDetectorRef) private readonly cdr: ChangeDetectorRef,
     ) {}
+
+    readonly trackBy: TrackByFunction<readonly T[]> = (index: number) => index;
 
     ngOnInit(): void {
         // Due to this view being parallel to app content, `markForCheck` from `async` pipe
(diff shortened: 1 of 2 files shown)


waterplea (commented):
I wonder why it stopped working just by changing the async pipe to assigning an array. It should still work without track by because items are Object.is equal.

splincode:
the problem is only in one line

image

new array reference, re-creates all div elements

waterplea:
> the problem is only in one line, we don't have async pipe here
> 
> image
> 
> but the new array reference notify Angular to recreate all div elements

Oh, I see, it's because alerts are two dimensional arrays. Others are flat. We only need track by for alerts then.

splincode:
`We only need track by for alerts then.` but we can improve the performance a bit for all lists, what do you think?

waterplea:
> `We only need track by for alerts then.` but we can improve the performance a bit for all lists, what do you think?

No, it will not improve a thing but will only introduce potential errors. Track by doesn't matter if items in arrays are ===

splincode:
@waterplea fixed

pwespi:
Thank you for the quick fix and your great work on this project in general! 🚀

PR #4716, opened 2023-06-20
fix(addon-editor): ignore delay for autofocus in cypress mode

splincode:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

### DIFF
--- a/projects/cdk/directives/auto-focus/autofocus.options.ts
+++ b/projects/cdk/directives/auto-focus/autofocus.options.ts
@@ -1,16 +1,20 @@
 import {
     ElementRef,
+    FactoryProvider,
     InjectionToken,
     NgZone,
     Optional,
     Renderer2,
     Self,
-    ValueProvider,
 } from '@angular/core';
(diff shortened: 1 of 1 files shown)


projects/cdk/directives/auto-focus/autofocus.options.ts
```diff
@@ -38,10 +42,17 @@ export const TUI_AUTOFOCUS_OPTIONS = new InjectionToken<TuiAutofocusOptions>(
 
 export function tuiAutoFocusOptionsProvider(
     options: Partial<TuiAutofocusOptions>,
-): ValueProvider {
+): FactoryProvider {
     return {
         provide: TUI_AUTOFOCUS_OPTIONS,
-        useValue: {...TUI_AUTOFOCUS_DEFAULT_OPTIONS, ...options},
+        useFactory: (isCypress: boolean) => {
```
waterplea, line 48: I don't think we should hook into `TUI_IS_CYPRESS` in a real code. Can we circumvent it inside our test environment?

PR #4848, opened 2023-07-05
fix(kit): `InputDateTime`'s min/max inputs should receive null (#4847)

### LINKED ISSUE #4847: 🚀 - `InputDateTime`'s `min`/`max` inputs must receive `null`
### Playground Link

-

### Description

Extending type of min/max inputs will allow to drop the extra code, e.g while using streams

```ts
import { TUI_FIRST_DAY } from '@taiga-ui/cdk';

@Component({...})
export class Component {
   readonly tuiFirstDay = TUI_FIRST_DAY;
   readonly max$ = this.someService.max$;
}
```

```html
  <tui-input-date-time [max]="(max$ | async) ?? tuiFirstDay"></tui-input-date-time>
```

### Angular version

12

### Taiga UI version

3.35.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

constantant:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #4847


projects/demo/src/modules/components/input-date-time/input-date-time.component.ts
```diff
@@ -62,25 +62,27 @@ export class ExampleTuiInputDateTimeComponent extends AbstractExampleTuiControl
...
         new TuiDay(new Date().getFullYear() + 3, 1, 1),
         [this.today.append({day: -1}), new TuiTime(12, 20)],
     ];
 
-    min: TuiDay | [TuiDay, TuiTime] = this.minVariants[0];
+    min: TuiDay | [TuiDay, TuiTime] | null = this.minVariants[0];
 
-    readonly maxVariants: ReadonlyArray<TuiDay | [TuiDay, TuiTime]> = [
+    readonly maxVariants: ReadonlyArray<TuiDay | [TuiDay, TuiTime] | null> = [
+        null,
```
waterplea, line 77: null should not be in the list, it is lack of value, not a particular value. Demo portal adds cleaner to select if null is in the type automatically.

waterplea (commented):
Please do the same changes to other similar date/time controls.

projects/kit/components/range/range.component.ts
```diff
@@ -184,13 +192,13 @@ export class TuiRangeComponent
     @tuiPure
     private computePureKeySteps(
         keySteps: TuiKeySteps | null,
-        min: number,
-        max: number,
+        min: number | null,
+        max: number | null,
```
nsbarsukov, line 196: Why is this change required ?

Just pass non-nullable arguments inside this function (actually, you already do it now).

```ts
this.computePureKeySteps(this.keySteps, this.min ?? 0, this.max ?? 100)
```

projects/kit/utils/math/key-steps.ts
```diff
@@ -33,11 +33,19 @@ export function tuiKeyStepValueToPercentage(
...
     keySteps: TuiKeySteps,
 ): number {
+    const stepsBoundariesByFn = tuiFindKeyStepsBoundariesByFn(
+        keySteps,
+        ([_, keyStepValue]) => value <= keyStepValue,
+    );
+
+    const [lower, upper] = stepsBoundariesByFn;
+
+    if (!(lower || upper)) {
```
nsbarsukov, line 43: I am not sure that we need to change anything inside this file.

projects/kit/components/range/range.component.ts
```diff
@@ -104,17 +104,37 @@ export class TuiRangeComponent
     }
 
     get fractionStep(): number {
-        return this.step / (this.max - this.min);
+        return this.step / ((this.max ?? 100) - (this.min ?? 0));
```
nsbarsukov, line 107:
```suggestion
        return this.step / (this.computedMax - this.computedMin);
```

projects/kit/components/input-date-time/input-date-time.component.ts
```diff
@@ -189,21 +197,28 @@ export class TuiInputDateTimeComponent
...
     get calendarMaxDay(): TuiDay {
-        return Array.isArray(this.max) ? this.max[0] : this.max;
+        const max = this.computedMax;
+
+        return Array.isArray(max) ? max[0] : max;
     }
 
     get computedActiveYearMonth(): TuiMonth {
+        const min = this.computedMin;
+        const max = this.computedMax;
```
nsbarsukov, line 213:
```suggestion
        const {computedMin, computedMax} = this;
```

projects/kit/utils/math/key-steps.ts
nsbarsukov: I got it, thanks!

Let's change `tuiFindKeyStepsBoundariesByFn` function instead:


```ts
function tuiFindKeyStepsBoundariesByFn(
    keySteps: TuiKeySteps,
    fn: ([keyStepPercent, keyStepValue]: [number, number]) => boolean,
): [[number, number], [number, number]] {
    const keyStepUpperIndex = keySteps.findIndex((keyStep, i) => i && fn(keyStep));

    const lowerStep = keySteps[keyStepUpperIndex - 1] || keySteps[0];
    const upperStep = keySteps[keyStepUpperIndex] || keySteps[keySteps.length - 1];

    return [lowerStep, upperStep];
}
```
And revert changes inside this function.

projects/kit/components/range/range.component.ts
constantant: no need due to https://github.com/Tinkoff/taiga-ui/pull/4848#discussion_r1258190049

projects/kit/utils/math/key-steps.ts
```diff
@@ -33,11 +33,19 @@ export function tuiKeyStepValueToPercentage(
     value: number,
     keySteps: TuiKeySteps,
 ): number {
+    const stepsBoundariesByFn = tuiFindKeyStepsBoundariesByFn(
```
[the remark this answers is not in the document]
constantant: no need due to https://github.com/Tinkoff/taiga-ui/pull/4848#discussion_r1258190049

projects/kit/components/input-date/input-date.directive.ts
```diff
@@ -14,11 +14,11 @@ export class TuiInputDateDirective extends AbstractTuiTextfieldHost<TuiInputDate
     }
 
     get max(): TuiDay {
-        return this.host.max;
+        return this.host.max ?? TUI_LAST_DAY;
```
waterplea, line 17: We use `computed` getters here.

waterplea (commented):
I don't think nullish min/max make sense for sliders and ranges — there always **has** to be a limit since it's a slider from min to max :)

projects/kit/components/input-date/input-date.directive.ts
constantant: Sorry, not getting what you mean. It is already a getter

waterplea: I mean you can use `this.host.computedMax`, no need to import fallback in that case.

constantant:
> I don't think nullish min/max make sense for sliders and ranges — there always **has** to be a limit since it's a slider from min to max :)

Any limits can depend on streams (`async` pipe returns `null` by default)

simple
```html
<tui-... [min]="min$ | async" [max]="max$ | async"></tui-...>
```

complex
```html
<ng-container *ngIf="min$ | async as min">
     <tui-...  *ngIf="max$ | async as max" [min]="min" [max]="max"></tui-...>
</ng-container>
```

waterplea:
> > I don't think nullish min/max make sense for sliders and ranges — there always **has** to be a limit since it's a slider from min to max :)
> 
> Any limits can depend on streams (`async` pipe returns `null` by default)
> 
> simple
> 
> ```
> <tui-... [min]="min$ | async" [max]="max$ | async"></tui-...>
> ```
> 
> complex
> 
> ```
> <ng-container *ngIf="min$ | async as min">
>      <tui-...  *ngIf="max$ | async as max" [min]="min" [max]="max"></tui-...>
> </ng-container>
> ```

Then it is your duty as a developer to provide a suitable fallback, component should not do the guesswork.

waterplea:
> If we follow the non-guess logic the inputs must be required

They are required, that's why they are non-nullable :D If you question if they should have default values — we use the same values as native sliders.

projects/kit/components/input-date-time/input-date-time.component.ts
```diff
@@ -84,10 +84,10 @@ export class TuiInputDateTimeComponent
     private month: TuiMonth | null = null;
 
     @Input()
-    min: TuiDay | [TuiDay, TuiTime] = this.options.min;
+    min: TuiDay | [TuiDay | null, TuiTime | null] | null = this.options.min;
```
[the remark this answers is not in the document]
waterplea: There is already `TuiDay` to cover a case when time is not yet entered. Maybe we can instead make a pipe:
```ts
transform(value: [TuiDay | null, TuiTime | null] | nulll): TuiDay | [TuiDay, TuiTime] {
  if (!value) {
    return null;
  }

  return value[1] ? value : value[0];
}
```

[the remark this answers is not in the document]
constantant: @waterplea I am not getting what you wanna achieve via this pipe. Could you show an example how to use it?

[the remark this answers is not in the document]
waterplea: > @waterplea I am not getting what you wanna achieve via this pipe. Could you show an example how to use it?

```ts
@Pipe({ name: 'tuiSafeDateTime' })
export class TuiSafeDateTimePipe implements PipeTransform {
  transform(value: [TuiDay | null, TuiTime | null] | null): TuiDay | [TuiDay, TuiTime] | null {
    if (!value) {
      return null;
    }

    return value[1] ? value : value[0];
  }
}
```

```html
<tui-input-date-time [(ngModel)]="min">...</tui-input-date-time>
<tui-input-date-time [min]="min | tuiSafeDateTime" [(ngModel)]="value">...</tui-input-date-time>
```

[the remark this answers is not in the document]
constantant: Looks like an extra complexity for such simple thing. As a developer I expect that limits receive the same type an access value provides

waterplea (approved):

projects/kit/components/input-date-time/input-date-time.component.ts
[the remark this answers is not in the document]
waterplea: Ok, maybe it just makes sense to change the control type to ` [TuiDay, TuiTime] | TuiDay | null` in 4.0. We'll think about it. Approved this PR.

splincode (approved):

vladimirpotekhin (approved):

PR #4856, opened 2023-07-07
fix(core): reuse `tuiOption` styles inside `tuiDataList`

### LINKED ISSUE #4803: 🐞 - EmptyContent in TuiDatalist gets wrong size
### Playground Link

https://taiga-ui.dev/components/combo-box/API?tuiTextfieldSize=s

### Description

When search queries in datalist controls return empty value, it always has the same size (doesn't change depends on datalist size by itself)

### Angular version

12.2.17

### Taiga UI version

2.34.0

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

AdrianKokot:
Extracted `tuiOption` styles into a separate `mixin.style.less` file. Imported into `tuiOption` and `tuiDataList` style files respectively.

Added appropriate text and margin/padding/size styles to the `.t-empty` selector to respect the size of the `tuiDataList`. 
Adding `tuiOption` and replacing `<div>` with either `<a>` or `<button>` would also work, but then custom content provided with polymorpheus could be broken, so decided to reuse styles instead.

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #4803

Size of `.t-empty` inside `data-list.template.html` was not respecting `tuiDataList`'s `size`.

## What is the new behavior?

Size of `tuiDataList` is used to determine text size of `.t-empty` selector.

### DIFF
--- a/projects/core/components/data-list/data-list.style.less
+++ b/projects/core/components/data-list/data-list.style.less
@@ -1,4 +1,5 @@
 @import '../../styles/taiga-ui-local.less';
+@import './mixin.style.less';
 
 tui-data-list {
     --tui-data-list-padding: 0.25rem;
@@ -22,15 +23,23 @@ tui-data-list {
     &[data-list-size='s'] {
         /* stylelint-disable-next-line */
         --tui-data-list-margin: 0rem;
+
+        & > .t-empty {
+            .data-list-item('s');
+        }
+    }
+
+    &[data-list-size='m'] > .t-empty {
+        .data-list-item('m');
     }
 
     &[data-list-size='l'] {
         --tui-data-list-padding: 0.375rem;
         --tui-data-list-margin: 0.125rem;
-    }
 
-    & > .t-empty {
-        margin: 0.75rem 1rem;
+        & > .t-empty {
+            .data-list-item('l');
+        }
     }
 }
 
--- a/projects/core/components/data-list/option/option.style.less
+++ b/projects/core/components/data-list/option/option.style.less
@@ -1,23 +1,17 @@
 @import '../../../styles/taiga-ui-local.less';
+@import '../mixin.style.less';
 
 :host {
     .clearbtn();
     .transition(background);
-    display: flex;
-    align-items: center;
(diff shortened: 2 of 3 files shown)


projects/core/components/data-list/data-list.style.less
```diff
@@ -1,4 +1,5 @@
 @import '../../styles/taiga-ui-local.less';
+@import 'shared.style.less';
```
splincode, line 2: @import './shared.style.less';

splincode (approved):

projects/core/components/data-list/shared.style.less
```diff
@@ -0,0 +1,20 @@
+.data-list-item(@size: 'm') when (@size = 's'), (@size = 'm'), (@size = 'l') {
```
splincode, line 1: `mixin.style.less`

projects/core/components/data-list/data-list.style.less
```diff
@@ -22,15 +23,29 @@ tui-data-list {
...
         --tui-data-list-margin: 0.125rem;
+
+        & > .t-empty {
+            .data-list-item('l');
+        }
     }
 
     & > .t-empty {
-        margin: 0.75rem 1rem;
+        display: flex;
```
waterplea, line 46: I think we can move these 3 rules under the mixin as well, since they are also used in `tuiOption`.

projects/core/components/data-list/option/option.style.less
waterplea, line 1:
```css
    min-height: 2.5rem;
    padding: 0.375rem 0.5rem;
    margin: 0.125rem 0;
```

under `:host` are always overridden by mixin, let's remove those lines. Or am I missing something? I wonder why `margin: var(--tui-data-list-margin) 0;` was repeated for each size 🤔

waterplea (approved):

PR #4892, opened 2023-07-11
feat(core): allow hierarchical DI options for components

waterplea:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #


projects/cdk/utils/miscellaneous/create-options.ts
```diff
@@ -0,0 +1,7 @@
+import {InjectionToken} from '@angular/core';
+
+export function tuiCreateOptions<T>(defaults: T): InjectionToken<T> {
+    return new InjectionToken<T>(``, {
```
splincode, line 4: `` - is it good?

projects/core/components/primitive-checkbox/checkbox.options.ts
```diff
@@ -1,8 +1,9 @@
-import {InjectionToken, ValueProvider} from '@angular/core';
-import {TuiContextWithImplicit} from '@taiga-ui/cdk';
+import {Provider} from '@angular/core';
+import {tuiCreateOptions, tuiProvideOptions, TuiContextWithImplicit} from '@taiga-ui/cdk';
 import {TuiAppearance} from '@taiga-ui/core/enums';
 import {TuiSizeL} from '@taiga-ui/core/types';
 import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
+import {} from '@taiga-ui/core/utils';
```
splincode, line 6: `{}`?

projects/cdk/utils/miscellaneous/create-options.ts
waterplea: I'm not sure it is of any help since it is mostly shown when there is no provider with this name, but all those tokens have default values so they never cause this situation really.

projects/kit/components/arrow/arrow.providers.ts
```diff
@@ -0,0 +1,29 @@
+import {Provider} from '@angular/core';
```
splincode, line 1: What do you think?
`arrow.providers.ts` -> `arrow-mode.options.ts`

waterplea: Let's keep it as is, it is not really an option, it's a regular DI token for particular behavior. I just think it makes sense to keep it here, rather than in `tokens` to simplify secondary entry points.

projects/kit/components/arrow/arrow.providers.ts
```diff
@@ -0,0 +1,29 @@
...
+    interactive: TUI_ARROW,
+    disabled: TUI_ARROW,
+};
+
+/**
+ * Type of icon in dropdowns for interactive or disable mode
+ */
+export const TUI_ARROW_MODE = tuiCreateOptions(TUI_ARROW_DEFAULT_MODE);
+
+export function tuiArrowModeProvider(options: Partial<TuiArrowMode>): Provider {
```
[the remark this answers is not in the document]
waterplea: See my other comment. Also all our helpers like that end with `Provider` word.

vladimirpotekhin (approved):

PR #4898, opened 2023-07-12
fix(core): `Textfield` straighten up icons spacings

waterplea:
Don't merge yet, let's see how hard it breaks screenshots.

### DIFF
--- a/projects/core/components/primitive-textfield/primitive-textfield.component.ts
+++ b/projects/core/components/primitive-textfield/primitive-textfield.component.ts
@@ -36,8 +36,9 @@ import {map} from 'rxjs/operators';
 
 import {TuiPrimitiveTextfield} from './primitive-textfield-types';
 
-const ICON_PADDING_L = 1.75;
-const ICON_PADDING = 1.625;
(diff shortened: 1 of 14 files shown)


projects/core/components/primitive-textfield/primitive-textfield.component.ts
```diff
@@ -36,8 +36,11 @@ import {map} from 'rxjs/operators';
 
 import {TuiPrimitiveTextfield} from './primitive-textfield-types';
 
-const ICON_PADDING_L = 1.75;
-const ICON_PADDING = 1.625;
+export const TUI_ICON_PADDINGS: Record<TuiSizeS | TuiSizeL, number> = {
```
splincode: Do we need use DI for this case in the future?

splincode (approved):

projects/core/components/primitive-textfield/primitive-textfield.component.ts
waterplea: No, we do not want this to be configurable.

vladimirpotekhin (approved):

PR #4911, opened 2023-07-13
chore(demo): fix flickering

splincode:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring
- [ ] Tests related changes

### DIFF
--- a/projects/demo/src/modules/components/island/examples/3/index.less
+++ b/projects/demo/src/modules/components/island/examples/3/index.less
@@ -39,15 +39,18 @@
     }
 
     [tuiaction] {
+        cursor: default;
         box-shadow: none;
         border: 1px solid var(--tui-base-03);
         border-bottom: none;
+        will-change: tranform;
         border-radius-right-bottom: 0;
         border-bottom-right-radius: 0;
         border-bottom-left-radius: 0;
+    }
 
-        &:hover {
-            transform: translateY(-1.6rem);
-        }
+    &:hover [tuiaction] {
+        transition-delay: 250ms;
+        transform: translateY(-1.6rem);
     }
 }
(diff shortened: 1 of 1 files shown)


projects/demo/src/modules/components/island/examples/3/index.less
```diff
@@ -39,15 +39,18 @@
     }
 
     [tuiaction] {
+        cursor: default;
         box-shadow: none;
         border: 1px solid var(--tui-base-03);
         border-bottom: none;
+        will-change: tranform;
```
vladimirpotekhin, line 46: is it necessary? 

`Warning: will-change is intended to be used as a last resort, in order to try to deal with existing performance problems. It should not be used to anticipate performance problems.` (mdn)

splincode: fixed

projects/demo/src/modules/components/island/examples/3/index.less
```diff
@@ -45,9 +45,10 @@
         border-radius-right-bottom: 0;
         border-bottom-right-radius: 0;
         border-bottom-left-radius: 0;
+    }
 
-        &:hover {
-            transform: translateY(-1.6rem);
-        }
+    &:hover [tuiaction] {
+        transition-delay: 250ms;
```
waterplea, line 51: This doesn't fix the issue, just delays it. It should either react to hover of the whole container, or have some :after element with height of 1.6rem to still be under the pointer.

splincode: `react to hover of the whole container` -> & - is a container

splincode: remove delay anyway

waterplea (approved):

Issue #5058, opened 2023-08-08
🐞 - `InputNumber` ignores native attribute `maxlength`


nsbarsukov:
### Playground Link

https://stackblitz.com/edit/tui-input-number-maxlength-bug

### Description

```html
<tui-input-number>
    <input tuiTextfield maxlength="7" />
</tui-input-number>
```

Why?
https://github.com/Tinkoff/taiga-ui/blob/fe99d1377203508e01834f6acf3785885ecd20c6/projects/kit/components/input-number/input-number.directive.ts#L29

### Angular version

any

### Taiga UI version

3.39.0

### Which browsers have you used?

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

40oleg:
May I take this one?

waterplea:
> May I take this one?

Yes, go ahead.

40oleg:
There is a problem.
Does native maxlength attribute have to count non-digit symbols like thousandSeparator?
e.g. "123 456" has length 6 or 7?
![image](https://github.com/taiga-family/taiga-ui/assets/60510512/cef64352-1705-41d0-ba0a-719c47c20ad0)
On picture above I set maxlength=7 but because of the space-symbol I can't type the seventh digit.
Is this behavior normal or not?

40oleg:
@waterplea @nsbarsukov  What do you think?

nsbarsukov:
> Does native maxlength attribute have to count non-digit symbols like thousandSeparator?

I think that `maxlength attribute` should count any character (digits and non-digits).
For example, `maxlength=7` should allow to insert 6 digits and 1 space between them.

splincode:
also `-123 456` is `maxlength=8`

PR #5119, opened 2023-08-15
InputNumber ignores native attribute maxlength #5058

40oleg:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

native maxlength just ignored

Closes # [](https://github.com/taiga-family/taiga-ui/issues/5058)

## What is the new behavior?

native maxlength works

### DIFF
--- a/projects/kit/components/input-number/input-number.component.ts
+++ b/projects/kit/components/input-number/input-number.component.ts
@@ -165,7 +165,7 @@ export class TuiInputNumberComponent
     get calculatedMaxLength(): number {
         const nativeMaxLength = Number(this.textfield?.nativeFocusableElement?.maxLength);
 
-        if (nativeMaxLength !== -1) {
+        if (nativeMaxLength !== -1 && !Number.isNaN(nativeMaxLength)) {
             return nativeMaxLength;
         }
 
--- a/projects/kit/components/input-number/input-number.component.ts
+++ b/projects/kit/components/input-number/input-number.component.ts
@@ -163,9 +163,9 @@ export class TuiInputNumberComponent
     }
 
     get calculatedMaxLength(): number {
-        const nativeMaxLength = Number(this.textfield?.nativeFocusableElement?.maxLength);
+        const nativeMaxLength = this.textfield?.nativeFocusableElement?.maxLength;
 
-        if (nativeMaxLength !== -1 && !Number.isNaN(nativeMaxLength)) {
+        if (nativeMaxLength !== -1 && nativeMaxLength !== undefined) {
             return nativeMaxLength;
         }
 
--- a/projects/kit/components/input-number/input-number.component.ts
+++ b/projects/kit/components/input-number/input-number.component.ts
@@ -163,6 +163,12 @@ export class TuiInputNumberComponent
     }
 
     get calculatedMaxLength(): number {
+        const nativeMaxLength = Number(this.textfield?.nativeFocusableElement?.maxLength);
+
+        if (nativeMaxLength !== -1) {
+            return nativeMaxLength;
+        }
+
         const decimalPart =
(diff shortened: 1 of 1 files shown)


projects/kit/components/input-number/input-number.component.ts
```diff
@@ -163,6 +163,12 @@ export class TuiInputNumberComponent
     }
 
     get calculatedMaxLength(): number {
+        const nativeMaxLength = this.textfield?.nativeFocusableElement?.maxLength;
+
+        if (nativeMaxLength !== -1 && nativeMaxLength !== undefined) {
+            return nativeMaxLength;
+        }
```
splincode, line 170: could you add unit tests?

40oleg:
I need to clarify something. One of the test failed. It just checks default "maxlength" value of input. But for some reason, the default "maxlength" value when running tests is different from the default "maxlength" value when running the application in the browser.
```ts
const el = document.createElement(`input`);
const out = el.maxLength;
console.log(out)
```
Another words, if you run code above in console of chrome you will get -1, but if you run it in .spec.ts file you will get 524228 (this is kind of max hard limit)
Apparently, something is wrong with runtime that was used to run the tests.
What should I do with this test? I can just hardcode this number in condition so test completes successfully, but maybe there is a better solution.
What do you think? @splincode

(found only one link for the kind of similar issue but it was dated 2011 https://bugs.webkit.org/show_bug.cgi?id=44883)

splincode:
@40oleg what if we mock input reference as `const input = jest.fn();`?

projects/kit/components/input-number/input-number.component.ts
```diff
@@ -163,6 +163,12 @@ export class TuiInputNumberComponent
     }
 
     get calculatedMaxLength(): number {
+        const nativeMaxLength = this.textfield?.nativeFocusableElement?.maxLength;
+
+        if (nativeMaxLength !== -1 && nativeMaxLength !== undefined) {
```
nsbarsukov, line 168: [Unit tests / kit](https://github.com/taiga-family/taiga-ui/actions/runs/5871927108/job/15922430479?pr=5119) job has failed with the following error:

```console
 FAIL  projects/kit/components/input-number/test/input-number.component.spec.ts
  ● InputNumber › maxlength is set to 23 by default (18 digits + 5 default separators)

    expect(received).toBe(expected) // Object.is equality

    Expected: "23"
    Received: "524288"

      335 |         const nativeInput = getNativeInput()!.nativeElement;
      336 |
    > 337 |         expect(nativeInput.getAttribute(`maxlength`)).toBe(`23`);
          |                                                       ^
      338 |     });
      339 |
      340 |     describe(`When decimal === always`, () => {

      at projects/kit/components/input-number/test/input-number.component.spec.ts:337:55
      at _ZoneDelegate.Object.<anonymous>._ZoneDelegate.invoke (node_modules/zone.js/bundles/zone-testing-bundle.umd.js:409:30)
      at ProxyZoneSpec.Object.<anonymous>.ProxyZoneSpec.onInvoke (node_modules/zone.js/bundles/zone-testing-bundle.umd.js:3830:43)
      at _ZoneDelegate.Object.<anonymous>._ZoneDelegate.invoke (node_modules/zone.js/bundles/zone-testing-bundle.umd.js:408:56)
      at Zone.Object.<anonymous>.Zone.run (node_modules/zone.js/bundles/zone-testing-bundle.umd.js:169:47)
      at Object.wrappedFunc (node_modules/zone.js/bundles/zone-testing-bundle.umd.js:4330:34)
```

nsbarsukov:
The same issue will be solved in this PR:

* https://github.com/taiga-family/taiga-ui/pull/5307

PR #5128, opened 2023-08-17
chore(demo): improve typings

splincode:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring
- [ ] Tests related changes

### DIFF
--- a/projects/demo/src/modules/app/pages.ts
+++ b/projects/demo/src/modules/app/pages.ts
@@ -1,6 +1,4 @@
-import {TuiDocPages} from '@taiga-ui/addon-doc';
-
-export const pages: TuiDocPages = [
+export const pages = [
     // Documentation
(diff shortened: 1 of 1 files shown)


nsbarsukov (approved):

projects/demo/src/modules/app/pages.ts
```diff
@@ -1326,4 +1324,4 @@ export const pages: TuiDocPages = [
             },
         ],
     },
-];
```
vladimirpotekhin, line 1329: So we can make a mistake in this config and the error will not be highlighted, why is this better?

splincode: for proprietary repo
image

vladimirpotekhin (approved):

PR #5198, opened 2023-08-25
feat(addon-doc): don't respect case-sensitive for tab link

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

image

## What is the new behavior?

image

### DIFF
--- a/projects/addon-doc/components/page/page.template.html
+++ b/projects/addon-doc/components/page/page.template.html
@@ -26,7 +26,7 @@
                     *tuiItem
                     routerLinkActive
                     tuiTab
-                    [routerLink]="first ? './' : (tabName | tuiReplace: from : to)"
+                    [routerLink]="first ? './' : (tabName | tuiReplace: from : to | lowercase)"
                     [routerLinkActiveOptions]="{exact: first}"
                 >
                     {{ tabName }}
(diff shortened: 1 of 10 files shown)


splincode:
@vladimirpotekhin @nsbarsukov hello, please review me

nsbarsukov (approved):

projects/addon-doc/tokens/url-lower-case.ts
```diff
@@ -0,0 +1,3 @@
+import {tuiCreateToken} from '@taiga-ui/cdk';
+
+export const TUI_DOC_URL_LOWER_CASE = tuiCreateToken(false);
```
waterplea, line 3: Looks overengineered to me. How about we just provide serializer in demo without all that configurations?

splincode: fixed

waterplea: I mean let's do it in demo app, because people might not want this behavior in addon-doc.

splincode: But I wanna use it in tui-editor demo. Will I need to duplicate it?

PR #5221, opened 2023-08-28
feat(experimental): add pin

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

## What is the new behavior?

image
image
image
image
image
image
image
image
image

https://github.com/taiga-family/taiga-ui/assets/12021443/e123ebde-3d21-4aa0-a6cf-acc3658d4b3f


splincode:
на ревью

```
1. Для чего нужны тайтл и сабтайтл? Может просто сделать :first-line? Не может же быть сабтайтла без тайтла.
2. tui-pin-dot может сделать через color, а не background?
3. tui-pin-selected может быть без точки? Может её зашить внутрь?
4. tui-pin-icon точно нужен? Нельзя как-то без него обойтись? Что там может быть, кроме него, почему этот внешний вид не зашит сразу?
```

waterplea (commented):
1) There is crazy amount of CSS variables, something needs to be done about it. I'll think about it and discuss with designers then propose my idea of styles to satisfy all customization requirements.
2) I don't see a reason for pointer version to be in the same module as the other pin, seems like an unrelated component which we should add separately.
3) Same goes for the label version. It is basically a wrapper around pin and some text, it should be a separate component.
4) Open state should be positioned around the bottom with top content not affecting it in terms of layout. Basically, upon toggling it, the center of the bottom dot must be at the same place the center of non-selected pin was. I think we can also come up with some transition, I'll think about how we can do that.
5) I don't think we should bundle interactivity into pin itself, especially with focus/blur - it would definitely not satisfy most of the cases. You would click a pin - it opens some info about it with buttons/inputs - you click there and blur closes pin. 

How about we make some "pin controller" instead with default implementation being completely passive, relying on `open` input, but with ability to provide your own service to control a bunch of pins at the same time. We can then add an example on the demo with say a map with a list of locations on the side with full address. When you click pin - it closes other pins and highlights the selected address on the side and vice-versa. Kinda like you would select a pick point in e-commerce delivery.

splincode:
@waterplea 

1. I'll wait for your suggestions
2. Do you mean that the component name will be different?  `<tui-cursor-map />` instead of `<tui-cursor pointer />` 
3. Do you mean that the component name will be different? 

before

```html
<tui-pin>
   Title <br>
   Subtitle
</tui-pin>
```

after?

```html
<div tuiPinText>
   <tui-pin src="image" />
   Title <br>
   Subtitle
</tui-pin>
```

4. `Open state should be positioned around the bottom with top content not affecting it in terms of layout. ` 
does it have to be absolute positioning?

5. You mean

```html
<div tuiPinGroup>
    <tui-pin src="img" tuiPinInteractive></tui-pin>
    <tui-pin tuiPinInteractive>123</tui-pin>
    <tui-pin tuiPinInteractive>5</tui-pin>
    <tui-pin tuiPinInteractive>23</tui-pin>
</div>
```

or just

```html
    <tui-pin src="img" tuiPinInteractive></tui-pin>
    <tui-pin tuiPinInteractive>123</tui-pin>
    <tui-pin tuiPinInteractive>5</tui-pin>
    <tui-pin tuiPinInteractive>23</tui-pin>
```

projects/experimental/components/pin/pointer/pin-pointer.component.ts
```diff
@@ -0,0 +1,18 @@
...
+    templateUrl: './pin-pointer.component.html',
+    styleUrls: ['../pin.shared.less', './pin-pointer.style.less'],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {
+        '[style.transform]': '"rotate(" + degrees + "deg)"',
+    },
+})
+export class TuiPinPointerComponent {
+    @Input()
+    pointer = '';
```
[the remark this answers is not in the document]
splincode: fixed

projects/experimental/components/pin/pin.module.ts
```diff
@@ -0,0 +1,13 @@
+import {CommonModule} from '@angular/common';
+import {NgModule} from '@angular/core';
+import {TuiSvgModule} from '@taiga-ui/core';
+
+import {TuiPinInteractiveComponent} from './interactive/pin-interactive.component';
+import {TuiPinPointerComponent} from './pointer/pin-pointer.component';
+
+@NgModule({
+    imports: [TuiSvgModule, CommonModule],
+    declarations: [TuiPinInteractiveComponent, TuiPinPointerComponent],
```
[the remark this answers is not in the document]
splincode: fixed

projects/demo/src/modules/experimental/pin/examples/1/index.html
```diff
@@ -0,0 +1,26 @@
+<!-- eslint-disable @html-eslint/no-inline-styles -->
+<div tuiPinGroup>
+    <tui-pin
+        src="assets/images/avatar.svg"
```
waterplea, line 4: Why is avatar an SVG? It should support raster images too.

waterplea (commented):
Let's move compass to a separate module and a separate PR so this is not such a big chunk of stuff to review.

projects/demo/src/modules/experimental/pin/examples/1/index.html
splincode: But tui-svg doesn't support raster image

waterplea: That means we cannot just use `tui-svg`. Perhaps it's better to externalize it so people can write both:
```html
<tui-pin><tui-svg src='...'></tui-svg></tui-pin>
<tui-pin><img src='...' /></tui-pin>
```

projects/demo/src/modules/experimental/pin/examples/1/index.html
```diff
@@ -0,0 +1,26 @@
+<!-- eslint-disable @html-eslint/no-inline-styles -->
```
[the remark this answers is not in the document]
splincode: fixed

projects/demo/src/modules/experimental/pin/examples/1/index.html
```diff
@@ -0,0 +1,26 @@
...
+<div tuiPinGroup>
+    <tui-pin
+        src="assets/images/avatar.svg"
+        tuiPinController
+    ></tui-pin>
+
+    <tui-pin
```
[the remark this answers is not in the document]
splincode: fixed

projects/demo/src/modules/experimental/pin/examples/2/index.html
```diff
@@ -0,0 +1,27 @@
+<tui-pin-text
+    src="assets/images/avatar.svg"
```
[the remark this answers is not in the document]
splincode: fixed

projects/demo/src/modules/experimental/pin/examples/3/index.less
```diff
@@ -0,0 +1,30 @@
...
+    gap: 2rem;
+    align-items: center;
+    flex-wrap: wrap;
+}
+
+.blue,
+.yellow,
+.green,
+.red {
+    height: 1rem;
```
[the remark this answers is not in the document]
splincode: fixed

projects/demo/src/modules/experimental/pin/examples/1/index.html
splincode: fixed

splincode:
@waterplea hello, please review me

projects/demo/src/modules/experimental/pin/pin.module.ts
```diff
@@ -0,0 +1,32 @@
...
+import {TuiPinExample3} from './examples/3';
+import {TuiPinExample4} from './examples/4';
+import {ExampleTuiPinComponent} from './pin.component';
+
+@NgModule({
+    imports: [
+        CommonModule,
+        TuiSvgModule,
+        TuiAddonDocModule,
+        RouterModule.forChild(tuiGenerateRoutes(ExampleTuiPinComponent)),
```
waterplea, line 19: New helper :)

projects/demo/src/modules/experimental/pin/pin.component.ts
```diff
@@ -0,0 +1,55 @@
...
+        TypeScript: import('./examples/4/index.ts?raw'),
+        HTML: import('./examples/4/index.html?raw'),
+        LESS: import('./examples/4/index.less?raw'),
+    };
+
+    readonly iconVariants = ['tuiIconClockLarge', 'tuiIconHeart'];
+
+    icon = this.iconVariants[0];
+
+    opened = false;
```
waterplea, line 48: nit: `open`

projects/demo/src/modules/experimental/pin/pin.component.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    degrees = 0;
```
waterplea, line 50: Not used anymore, right?

projects/demo/src/modules/experimental/pin/examples/import/insert-template.md
```diff
@@ -0,0 +1,20 @@
+```html
+<tui-pin src="assets/images/avatar.svg"></tui-pin>
```
waterplea, line 2: Seems like old API is used

projects/demo/src/modules/experimental/pin/pin.module.ts
splincode: fixed

projects/demo/src/modules/experimental/pin/pin.component.ts
splincode: fixed

waterplea:
Height is a bit off:
image
Maybe you can use the new `tuiTitle` component here? Perhaps style it with `::ng-deep` or disable style encapsulation for pin.

Image is missing the white border (the is one if Figma):
image


Small pin has the blue border around image and not in Figma:
image

Small pin in Figma has 2 modes of icon, could you add this example too? Note that image with white background has a smaller icon:
image

I think that is because the blue border is actually a border so the content inside gets smaller.

Here how I think this could work. Border would be `0 solid` by default, so it inherits color and people can just modify border-width if they need:
image

Here's the open state:
image
::before and ::after will be made of just the border, which inherits the border color. ::after will inherit `box-shadow` too. ::before will use SVG mask to have the desired shape, and it will also cover the box-shadow on the host so it is only visible on the small circle below. The blue border can be extended using `border-width`. Minimum of 2px which is currently there is part of the SVG mask and cannot be made smaller, I check with designers. In case of image we can make it's height a bit smaller so that box-shadow is visible and those 2px of white are there.

projects/experimental/directives/pin/pin.directive.ts
```diff
@@ -0,0 +1,54 @@
...
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    encapsulation: ViewEncapsulation.None,
+})
+export class TuiPinStyleComponent {}
+
+@Directive({
+    selector: '[tuiPin]',
+    host: {
+        tuiPin: '',
+        '[attr.tabindex]': '0',
```
waterplea, line 27: Let's not make it focusable by default. Since it's an attribute directive, users can put it on buttons or links if they want them focusable. Let's add `clearbtn` and `text-decoration: none` for such cases.

projects/experimental/directives/pin/pin-group.directive.ts
```diff
@@ -0,0 +1,54 @@
...
+    Self,
+} from '@angular/core';
+import {TuiDestroyService} from '@taiga-ui/cdk';
+import {Observable, Subject} from 'rxjs';
+import {filter, takeUntil} from 'rxjs/operators';
+
+import {TuiPinDirective} from './pin.directive';
+
+@Directive({
+    selector: '[tuiPinGroup]',
```
waterplea, line 17: Let's drop this directive for now and openChange so that for the first iteration pins are only manually open-able.

projects/experimental/directives/pin/pin.directive.ts
```diff
@@ -0,0 +1,54 @@
...
+})
+export class TuiPinDirective {
+    @Input('tuiPin')
+    open: boolean | '' = false;
+
+    @Output('tuiPinChange')
+    openChange = new EventEmitter<boolean>();
+
+    @Output()
+    focusChange = new EventEmitter<boolean>();
```
waterplea, line 40: If we drop pin group directive from first iteration, let's drop focus related stuff too. It's not needed without it and we can always add it later.

projects/experimental/directives/pin/pin.directive.ts
```diff
@@ -0,0 +1,54 @@
...
+import {TUI_PLATFORM, TuiDirectiveStylesService, TuiPlatform} from '@taiga-ui/cdk';
+
+@Component({
+    host: {class: 'tui-pin'},
+    template: '',
+    styleUrls: ['./pin.styles.less'],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    encapsulation: ViewEncapsulation.None,
+})
+export class TuiPinStyleComponent {}
```
waterplea, line 21: nit: since we do not have a pin component, only styles – there's no need to name it PinStyle, let's just move it to pin.component.ts and name it TuiPinComponent. That's how we did it everywhere, except for the buttons, because there's an actual button component there for icons and loader.

projects/experimental/directives/pin/pin.directive.ts
splincode: fixed

projects/experimental/directives/pin/pin.directive.ts
splincode: fixed

projects/experimental/directives/pin/pin-group.directive.ts
splincode: fixed

projects/experimental/directives/pin/pin.directive.ts
splincode: fixed

splincode:
@waterplea hello, please review me

projects/experimental/directives/pin/pin.directive.ts
```diff
@@ -0,0 +1,43 @@
...
+    styleUrls: ['./pin.styles.less'],
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {
+        class: 'tui-pin',
+    },
+})
+export class TuiPinComponent {}
+
+@Directive({
```
waterplea, line 24: I think there's no need for it to be a directive, a component with attribute selector is enough, because it is reliant on its content – it will not mix with other components anyway. But attribute selector would allow us to put it on `button`, for example. BTW, in such cases, what should be our convention, use `tui-pin,[tuiPin]` as selector, or always use `[tuiPin]`, what do you think, @vladimirpotekhin @nsbarsukov. Previously we aloways had 2 selectors, I believe.

projects/demo/src/modules/experimental/pin/pin.template.html
```diff
@@ -0,0 +1,96 @@
...
+                <div [(tuiPin)]="open">
+                    <tui-svg [src]="icon"></tui-svg>
+                </div>
+            </div>
+        </tui-doc-demo>
+
+        <tui-doc-documentation>
+            <ng-template
+                documentationPropertyMode="input"
+                documentationPropertyName="src"
```
waterplea, line 53: You ended up removing this as input. I guess the whole API tab is rather unnecessary in this case, it's just one input and all other cases are covered by examples.

projects/experimental/directives/pin/pin.directive.ts
```diff
@@ -0,0 +1,43 @@
...
+        tuiPin: '',
+        '[class._open]': 'open',
+    },
+})
+export class TuiPinDirective {
+    @Input('tuiPin')
+    open: boolean | '' = false;
+
+    @Output()
+    readonly tuiPinChange = new EventEmitter<boolean>();
```
waterplea, line 36: Let's remove the output for now, we can always add it later.

projects/demo/src/modules/experimental/pin/pin.template.html
splincode: fixed

projects/experimental/directives/pin/pin.directive.ts
splincode: fixed

projects/experimental/components/pin/pin.styles.less
```diff
@@ -0,0 +1,110 @@
...
+    text-align: center;
+    border-radius: 100%;
+    font: var(--tui-font-text-s);
+    font-weight: bold;
+    box-sizing: border-box;
+    color: var(--tui-accent-text);
+    background: var(--tui-accent);
+    box-shadow: 0 0 0 0.125rem var(--tui-elevation-01);
+    border: 0 solid var(--tui-accent);
+    transform: translate(-50%, -50%);
```
nsbarsukov, line 22: Why we need this ? 🤔

projects/demo/src/modules/experimental/pin/examples/1/index.less
```diff
@@ -0,0 +1,14 @@
+:host {
+    display: flex;
+    gap: 2rem;
+    background: var(--tui-base-02);
+    box-shadow: 0 0 0 2rem var(--tui-base-02);
+    padding: 1rem 0 0 1rem;
+    margin-bottom: -1rem;
```
nsbarsukov, line 7: It looks strange 🤔 

Example should not have any extra space below content.
Maybe something goes wrong ?

nsbarsukov: Is it connected to `transform: translate(-50%, -50%)` (on `<tui-pin />`) ?

projects/demo/src/modules/experimental/pin/examples/2/index.less
```diff
@@ -0,0 +1,24 @@
+:host {
+    display: flex;
+    gap: 2rem;
+    background: var(--tui-base-02);
+    box-shadow: 0 0 0 2rem var(--tui-base-02);
+    padding: 0.375rem 0 0 0.375rem;
+    margin-bottom: -0.375rem;
```
nsbarsukov, line 7: The same question

projects/demo/src/modules/experimental/pin/examples/3/index.html
```diff
@@ -0,0 +1,39 @@
...
+    (click)="b = !b"
+>
+    <tui-icon icon="assets/images/atm.svg"></tui-icon>
+</button>
+
+<button
+    tuiPin
+    class="link"
+    [open]="c"
+    [style.left.rem]="10"
```
nsbarsukov, line 25: Why not `display: flex` + `gap` ?

projects/demo/src/modules/experimental/pin/examples/3/index.ts
```diff
@@ -0,0 +1,17 @@
...
+@Component({
+    selector: 'tui-pin-example-3',
+    templateUrl: './index.html',
+    styleUrls: ['./index.less'],
+    encapsulation,
+    changeDetection,
+})
+export class TuiPinExample3 {
+    a = false;
+    b = false;
```
nsbarsukov, line 14: Let's make any item be initially open

projects/experimental/components/pin/pin.styles.less
```diff
@@ -0,0 +1,110 @@
+@import '@taiga-ui/core/styles/taiga-ui-local';
+
+:host {
+    .transition(~'transform, width, height, font, border');
+    .clearbtn();
```
nsbarsukov, line 5: Should we add `cursor: pointer` for `<button tuiPin />` ?

projects/demo/src/modules/experimental/pin/pin.component.ts
```diff
@@ -0,0 +1,36 @@
...
+        './examples/import/import-module.md?raw'
+    );
+
+    readonly exampleHtml: TuiRawLoaderContent = import(
+        './examples/import/insert-template.md?raw'
+    );
+
+    readonly example1: TuiDocExample = {
+        TypeScript: import('./examples/1/index.ts?raw'),
+        HTML: import('./examples/1/index.html?raw'),
```
nsbarsukov, line 21: I think we could move all `HTML` tabs on the first place for all examples on this documentation page.

They look like the most important part.

projects/demo/src/modules/experimental/pin/pin.module.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import {TuiPinExample1} from './examples/1';
+import {TuiPinExample2} from './examples/2';
+import {TuiPinExample3} from './examples/3';
+import {ExampleTuiPinComponent} from './pin.component';
+
+@NgModule({
+    imports: [
+        CommonModule,
+        TuiSvgModule,
```
nsbarsukov, line 15: Is it used anywhere ?

projects/demo/src/modules/experimental/pin/pin.module.ts
```diff
@@ -0,0 +1,31 @@
...
+        TuiPinModule,
+        TuiAddonDocModule,
```
nsbarsukov, line 17: `tuiGetDocModules` (which is imported in this module) already includes `TuiAddonDocModule`

projects/demo/src/modules/experimental/pin/pin.module.ts
```diff
@@ -0,0 +1,31 @@
...
+        TuiNotificationModule,
+        TuiIconModule,
+        tuiGetDocModules(ExampleTuiPinComponent),
+        TuiAppearanceModule,
```
nsbarsukov, line 21: Is it used anywhere ?

projects/demo/src/modules/experimental/pin/pin.template.html
```diff
@@ -0,0 +1,83 @@
...
+            heading="Dot"
+            [content]="example2"
+            [fullsize]="true"
+        >
+            <tui-pin-example-2></tui-pin-example-2>
+        </tui-doc-example>
+
+        <tui-doc-example
+            id="selected"
+            heading="Selected"
```
nsbarsukov, line 50: Input property (which is described in this example) is called `open`. Should we rename `id` and `heading` ? 🤔

projects/demo/src/modules/experimental/pin/examples/1/index.less
waterplea: Yes, it is. Pins would most likely be positioned on a map of sorts so we want to position their center at top/left.

projects/experimental/components/pin/pin.styles.less
waterplea: It is done this way so that when we position pin on a map with top/left – that place is at the center of the pin.

projects/demo/src/modules/experimental/pin/examples/2/index.less
waterplea: This is just a compensation for centered positioning of the pins. Alternatively we can absolutely position them and not have this weird compensation.

projects/demo/src/modules/experimental/pin/examples/3/index.html
waterplea: Because selected pins are wider so it would jump around on clicks.

vladimirpotekhin:
there's something wrong with the animation in safari

https://github.com/taiga-family/taiga-ui/assets/46284632/8454d5c9-d530-488f-92b8-f3ec6d1f65b1

nsbarsukov (approved):

vladimirpotekhin (approved):

PR #5280, opened 2023-09-04
feat(core): support appearance in dialog

### LINKED ISSUE #5243: 🚀 - Add configuration for dialog position
### Description

In some cases we need to fix position of dialog. A good example is dialog with content which height is changing (some kind of expand). So we need to be able to configure and fix position of dialog to prevent flickering.
[TuiAlertService](https://taiga-ui.dev/services/alert-service) has token `TUI_ALERT_POSITION` that solves similar problem.

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #5243

## What is the new behavior?

image


projects/core/interfaces/dialog-options.ts
```diff
@@ -15,6 +15,7 @@ import {Observable} from 'rxjs';
  * data - arbitrary data for dialog (undefined by default)
  */
 export interface TuiDialogOptions<I> {
+    readonly appearance?: string;
```
waterplea, line 18: Add TODO to make it required in 4.0 and default to empty string.

projects/demo/src/modules/components/dialog/examples/10/dialog-example/dialog-example.component.less
```diff
@@ -0,0 +1,60 @@
+.pulldown-menu {
+    background: var(--tui-elevation-01);
```
waterplea, line 2: We should use something less radical in terms of styles for this feature. For something this much different to default dialog people should create their own dialogs, not alter built-in appearance with CSS.

projects/demo/src/modules/components/dialog/examples/10/index.less
```diff
@@ -0,0 +1,17 @@
+[data-appearance='ios-style'][data-size='m'] {
+    width: 100% !important;
```
waterplea, line 2: Would be great to come up with an example without necessary `!important`.

nsbarsukov (approved):

projects/demo/src/modules/components/dialog/dialog.module.ts
```diff
@@ -42,6 +42,7 @@ import {SearchDialogExampleModule} from './examples/7/search-example/search-dial
 import {TuiDialogExampleComponent8} from './examples/8';
 import {TuiDialogExampleComponent9} from './examples/9';
 import {PayExampleModalModule} from './examples/9/pay-modal/pay-modal.module';
+import {TuiDialogExampleComponent10} from './examples/10';
```
waterplea, line 45: Not sorted

projects/demo/src/modules/components/dialog/examples/10/index.less
```diff
@@ -0,0 +1,51 @@
+tui-dialog[data-appearance='ios-style'][data-size] {
+    margin: 1.25rem auto;
+    border: none;
+    border-radius: 0;
+
+    .t-content {
```
waterplea, line 6: Please add an example that does not rely on internal structure.

projects/demo/src/modules/components/dialog/examples/10/index.ts
```diff
@@ -0,0 +1,32 @@
...
+    selector: 'tui-dialog-example-10',
+    templateUrl: './index.html',
+    styleUrls: ['./index.less'],
+    encapsulation: ViewEncapsulation.None,
+    changeDetection,
+})
+export class TuiDialogExampleComponent10 {
+    constructor(
+        @Inject(TuiDialogService)
+        readonly dialogs: TuiDialogService,
```
waterplea, line 15: nit: private + single line

projects/demo/src/modules/components/dialog/dialog.module.ts
splincode: lint succeeded because alphabetical sorting works

```
./examples/8/...
./examples/9/...
./examples/9/...
./examples/10/..
```

projects/demo/src/modules/components/dialog/examples/10/index.less
splincode: fix

projects/demo/src/modules/components/dialog/examples/10/index.ts
splincode: fix

waterplea (approved):

PR #5300, opened 2023-09-06
feat(kit): improve debug in dev mode for input component

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

## What is the new behavior?

image

### DIFF
--- a/projects/kit/components/input/input.component.ts
+++ b/projects/kit/components/input/input.component.ts
@@ -19,6 +19,8 @@ import {
     TuiContextWithImplicit,
     TuiFocusableElementAccessor,
     tuiIsNativeFocused,
+    tuiIsPresent,
+    tuiIsString,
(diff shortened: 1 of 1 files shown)


projects/kit/components/input/input.component.ts
```diff
@@ -117,6 +119,16 @@ export class TuiInputComponent
         this.open = false;
     }
 
+    override writeValue(value: string | null): void {
```
nsbarsukov, line 122: Why `null` ?

Empty text field - empty string, is it ?

splincode: fixed

waterplea:
I'm not sure we need this in a particular case. Maybe we can solve it systematically for all controls and inputs somehow?

PR #5303, opened 2023-09-06
fix(cdk): schematic error `Attempted to get information from a node that was removed or forgotten`

nsbarsukov:
## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

```
🚀 Your packages will be updated to @taiga-ui/*@^3.42.1

[...]

  ⚡️ replacing services...
    > replacing TuiPortalService...
    > replacing TuiNotificationsService...
Attempted to get information from a node that was removed or forgotten.

Node text: TuiNotificationsService
/Users/n.barsukov/WebstormProjects/sme-deposits-front/node_modules/@nrwl/workspace/node_modules/yargs/build/lib/yargs.js:1132
                throw err;
```

### DIFF
--- a/projects/cdk/schematics/ng-update/steps/remove-module.ts
+++ b/projects/cdk/schematics/ng-update/steps/remove-module.ts
@@ -29,6 +29,10 @@ export function removeModule(name: string, moduleSpecifier: string): void {
     const references = getNamedImportReferences(name, moduleSpecifier);
 
     references.forEach(ref => {
+        if (ref.wasForgotten()) {
+            return;
+        }
+
         const parent = ref.getParent();
 
         if (Node.isImportSpecifier(parent)) {
--- a/projects/cdk/schematics/ng-update/steps/rename-types.ts
+++ b/projects/cdk/schematics/ng-update/steps/rename-types.ts
@@ -34,6 +34,10 @@ function renameType(
     const references = getNamedImportReferences(from, moduleSpecifier);
 
     references.forEach(ref => {
+        if (ref.wasForgotten()) {
+            return;
+        }
+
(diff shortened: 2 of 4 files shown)


projects/cdk/schematics/ng-update/steps/remove-module.ts
```diff
@@ -29,6 +29,10 @@ export function removeModule(name: string, moduleSpecifier: string): void {
     const references = getNamedImportReferences(name, moduleSpecifier);
 
     references.forEach(ref => {
```
splincode, line 31: what about
```ts
references.filter(ref => !ref.wasForgotten()).forEach(ref => {
```

splincode (approved):

projects/cdk/schematics/ng-update/steps/remove-module.ts
splincode: Or

```ts
    return arrayFlat(
        namedImports.map(specifier => specifier?.findReferencesAsNodes() || []),
    ).filter(ref => !ref.wasForgotten());
```

image

splincode (approved):

projects/cdk/schematics/ng-update/steps/remove-module.ts
nsbarsukov: 1. It will iterate the same array twice.
2. We should check condition `wasForgotten()` before EVERY beginning of new iteration (because previous one can change everything).

splincode: sorry

vladimirpotekhin (approved):

Issue #5371, opened 2023-09-12
🐞 - `InputSlider` & `InputRange` should not validate `quantum` if textfield is focused


nsbarsukov:
### Playground Link

https://taiga-ui.dev/components/input-slider/API?max=1000&quantum=100

### Description

1. Open https://taiga-ui.dev/components/input-slider/API?max=1000&quantum=100
2. Type `150`

**Expected**: `150`
**Actual**: `200`

### Angular version

any

### Taiga UI version

3.44.1

### Which browsers have you used?

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

PR #5389, opened 2023-09-14
fix(kit): `InputSlider` should not validate `quantum` if textfield is focused

nsbarsukov:
## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Partially solves #5371


waterplea (approved):
Would we need some changes for InputRange too?

nsbarsukov:
> Would we need some changes for InputRange too?

Yep, i will do it in separate PR

splincode (approved):

PR #5407, opened 2023-09-17
fix(core): don't use hyphens in notification

splincode:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

image

### DIFF
--- a/projects/core/components/notification/notification.style.less
+++ b/projects/core/components/notification/notification.style.less
@@ -81,5 +81,5 @@
     flex: 1;
     word-wrap: break-word;
     color: var(--tui-text-01);
-    hyphens: auto;
+    overflow: hidden;
 }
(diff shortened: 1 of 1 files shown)


projects/core/components/notification/notification.style.less
```diff
@@ -81,5 +81,5 @@
     flex: 1;
     word-wrap: break-word;
     color: var(--tui-text-01);
-    hyphens: auto;
+    overflow: hidden;
```
waterplea, line 84: We do not need it here, otherwise it chops off extended parts of the letters on the last line. Why do you think we need it? Seems to work just fine without it.

waterplea (changes requested):

nsbarsukov (approved):

waterplea (approved):

PR #5414, opened 2023-09-18
feat(core): memoize getPosition method in dropdown position service

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [x] Refactoring
- [ ] Tests related changes

Called every time when dropdown opened

image


waterplea:
It is not that simple. These data can change from 2 sides:
1) Host position can change
2) Dropdown position can change due to content changes

PR #5421, opened 2023-09-18
feat(addon-doc): support customization top level navigation menu

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

image

## What is the new behavior?

image


waterplea:
Do we need this somewhere? I believe we will rework accordion component completely soon so I wouldn't want more things tie us up to the old API. Besides some options do not really make sense in the context of navigation, like rounded or borders.

PR #5442, opened 2023-09-20
fix(cdk): mask value may be different instead of real input value

### LINKED ISSUE #5098: 🐞 - InputDateTime - placeholder overlapping with value
### Playground Link

-

### Description

Problem in `basic` example at proprietary demo portal.

1. Choose a date
2. Enter first digit of timestamp
3. See that placeholder `H` is still visible and overlaps value

### Angular version

any

### Taiga UI version

3.40.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #5098

Screenshot 2023-09-20 at 21 36 13

`this.valueIdenticalComparator(this.value, value)` returned `true` when if the value does not match the mask

### DIFF
--- a/projects/cdk/abstract/control.ts
+++ b/projects/cdk/abstract/control.ts
@@ -215,6 +215,8 @@ export abstract class AbstractTuiControl<T>
      */
     protected updateValue(value: T): void {
         if (this.disabled || this.valueIdenticalComparator(this.value, value)) {
+            this.checkControlUpdate();
+
             return;
         }
 


projects/cdk/abstract/control.ts
```diff
@@ -215,6 +215,8 @@ export abstract class AbstractTuiControl<T>
      */
     protected updateValue(value: T): void {
         if (this.disabled || this.valueIdenticalComparator(this.value, value)) {
+            this.checkControlUpdate();
```
waterplea, line 218: Why do we need it? 🤔

splincode: `this.valueIdenticalComparator(this.value, value)` returned `true` when if the value does not match the mask
and therefore we need to update view

waterplea: This would sync native input value to control value with applied mask?

PR #5445, opened 2023-09-21
feat(core): reuse baseUrl in icons path factory

### LINKED ISSUE #4026: 🐞 - Broken icons via CDN
### Playground Link

impossible to show it on playground

### Description

If you deploy an app to CDN by [the docs](https://angular.io/guide/deployment#the-deploy-url) a browser breaks icons due to security reason (` Unsafe attempt to load URL <URL> from frame with URL <URL>. Domains, protocols and ports must match.`)

Note that `<base>` and [APP_BASE_HREF](https://angular.io/api/common/APP_BASE_HREF) are different in that case

### Angular version

15.2

### Taiga UI version

3.22.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #4026

## What is the new behavior?

```ts
providers: [
    tuiSvgOptionsProvider({
        path: (name, baseUrl) => `${baseUrl}assets/taiga-ui/icons/${name}`,
    }),
],
```


projects/core/components/svg/svg-options.ts
```diff
@@ -31,7 +31,7 @@ export interface TuiSvgOptions {
     readonly contentProcessor: TuiHandler<TuiSafeHtml, TuiSafeHtml>;
     readonly deprecated: TuiStringHandler<string>;
     readonly iconsPlace: string;
-    readonly path: TuiStringHandler<string>;
+    readonly path: (name: string, baseHref: string) => string;
```
vladimirpotekhin, line 34: breaking change? should we support old variant?

splincode: fixed

projects/demo/src/modules/icons/icons-mapping/examples/4/app.module.md
```diff
@@ -6,9 +6,12 @@ import {tuiSvgOptionsProvider, tuiSvgSrcInterceptors} from '@taiga-ui/core';
...
     tuiSvgOptionsProvider({
       /**
-       * The path to the folder with icons.
+       * @description:
+       * Resolve path to your icon by factory
+       * @name -> <tui-svg src="{name}" />
+       * @baseHref(optional) -> <base href="{baseHref}" />
        */
-      path: name => `assets/icons/${MAPPER[name]}.svg#${MAPPER[name]}`,
+      path: (name: string, baseHref: string) => `${baseHref}/assets/icons/${MAPPER[name]}.svg#${MAPPER[name]}`,
```
waterplea, line 14: If `baseHref` is empty, wouldn't that cause extra `/`?

splincode: fixed

waterplea (commented):
Super worried about all this base href stuff. What if there is no trailing slash at the end of it? Is it still valid? That would break current code. Also regarding `TUI_ENSURE_BASE_HREF`:
1) I don't see a point in the word `ENSURE` in its name and what would be the
2) What could be the scenario where `link.pathname` is not empty string, when base tag is missing?
3) `link.href = ''` seems unnecessary as it is the default value of newly created `a` tag

splincode:
`That would break current code` no, because you need overwrite path handler manually

waterplea (approved):

splincode:
@nsbarsukov hello, please review me

nsbarsukov (approved):

Issue #5436, opened 2023-09-20
🐞 - `Table` controls do not work in heading


waterplea:
### Playground Link

https://stackblitz.com/edit/angular-wszavt

### Description

We need to add relevant styles from TD component to TH as well:
```css
    // Create new viewport for fixed positioning
    transform: translate3d(0, 0, 0);
```
and also padding changes

### Which browsers have you used?

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

PR #5450, opened 2023-09-21
fix(addon-table): `table` controls do not work in heading

splincode:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Part of #5436

### DIFF
--- a/projects/addon-table/components/table/th/th.style.less
+++ b/projects/addon-table/components/table/th/th.style.less
@@ -15,6 +15,8 @@
     box-sizing: border-box;
     box-shadow: 0 0.3125rem rgba(237, 237, 237, 0);
     border: 1px solid var(--tui-base-04);
+    // Create new viewport for fixed positioning
+    transform: translate3d(0, 0, 0);
 
     &:not(:first-child) {
         border-left: none;


projects/addon-table/components/table/th/th.style.less
```diff
@@ -15,6 +15,8 @@
     box-sizing: border-box;
     box-shadow: 0 0.3125rem rgba(237, 237, 237, 0);
     border: 1px solid var(--tui-base-04);
+    // Create new viewport for fixed positioning
+    transform: translate3d(0, 0, 0);
```
waterplea, line 19: There's also padding reset in TD if there's a control inside.

splincode: fixed

waterplea: Where does the editable class come from?

splincode: fixed

waterplea (approved):

waterplea (commented):
Come to think of it, if somebody has some unrelated NgControl in their headers it would now break their padding. We probably should remove the padding part and control query.

waterplea (approved):

vladimirpotekhin (approved):

PR #5451, opened 2023-09-21
feat(addon-doc): support `Promise<unknown>`

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

image

### DIFF
--- a/projects/addon-doc/interfaces/page.ts
+++ b/projects/addon-doc/interfaces/page.ts
@@ -16,7 +16,7 @@ export interface TuiDocPageGroup extends TuiDocPageBase {
     readonly subPages: readonly TuiDocPage[];
 }
 
-export type RawLoaderContent = Promise<{default: string}> | string;
+export type RawLoaderContent = Promise<{default: string}> | Promise<unknown> | string;
 
 export const TUI_EXAMPLE_PRIMARY_FILE_NAME = {
     TS: `TypeScript`,
--- a/projects/addon-doc/utils/raw-load.ts
+++ b/projects/addon-doc/utils/raw-load.ts
@@ -1,5 +1,5 @@
 import {RawLoaderContent} from '@taiga-ui/addon-doc/interfaces';
 
 export async function tuiRawLoad(content: RawLoaderContent): Promise<string> {
-    return content instanceof Promise ? (await content).default : content;
+    return content instanceof Promise ? ((await content) as any)?.default ?? `` : content;
 }


projects/addon-doc/interfaces/page.ts
```diff
@@ -16,7 +16,7 @@ export interface TuiDocPageGroup extends TuiDocPageBase {
     readonly subPages: readonly TuiDocPage[];
 }
 
-export type RawLoaderContent = Promise<{default: string}> | string;
+export type RawLoaderContent = Promise<{default: string}> | Promise<unknown> | string;
```
waterplea, line 19: Why do we need this?

splincode: IDEA bugs

image

projects/addon-doc/utils/raw-load.ts
```diff
@@ -1,5 +1,5 @@
 import {RawLoaderContent} from '@taiga-ui/addon-doc/interfaces';
 
 export async function tuiRawLoad(content: RawLoaderContent): Promise<string> {
-    return content instanceof Promise ? (await content).default : content;
+    return content instanceof Promise ? (await (content as any))?.default ?? `` : content;
```
waterplea, line 4: Maybe `String((await content).default)` instead?

splincode: image

waterplea (approved):

splincode:
@nsbarsukov hello, please review me)

projects/addon-doc/interfaces/page.ts
nsbarsukov, line 19: To solve this "bug" we can just change a little these lines:
https://github.com/taiga-family/taiga-ui/blob/728f07ec7e3fed0a846b540b36ac7286083ab90d/projects/demo/src/modules/components/data-list-wrapper/data-list-wrapper.component.ts#L14-L16

```ts
readonly example1: TuiDocExample = {
        TypeScript: import('./examples/1/index.ts?raw'),
        HTML: import('./examples/1/index.html?raw'),
    };
```

I am not sure that we need to expand this type with so non-strict typing.

nsbarsukov: @vladimirpotekhin what do you think about it ?

waterplea: It would be nice to not require typing such objects. Why does this happen, is it just a webstorm getting the types wrong?

splincode: `just a webstorm`

PR #5477, opened 2023-09-25
fix(core): `PrimitiveTextfield` with prefix/postfix has CD problem after delayed fonts loading

### LINKED ISSUE #5472: 🐞 - `PrimitiveTextfield` with prefix/postfix has change detection problem after delayed fonts loading
### Playground Link

https://stackblitz.com/edit/primitive-textfield-delayed-font-loading?file=src%2Fapp%2Fapp.component.html,src%2Findex.html

### Description

1. Throttle network request for fonts
2. Open https://stackblitz.com/edit/primitive-textfield-delayed-font-loading?file=src%2Fapp%2Fapp.component.html,src%2Findex.html

https://github.com/taiga-family/taiga-ui/assets/35179038/ca6378b9-3e4f-4e66-b90a-00142a18c999

___

Related problem:

___

https://github.com/taiga-family/taiga-ui/blob/3d0ca4e9d70ab04582cbd111b13ac42f5db1afae/projects/core/components/primitive-textfield/value-decoration/value-decoration.component.ts#L72-L75

### Angular version

12

### Taiga UI version

3.46.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

nsbarsukov:
## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #5472

### DIFF
--- a/projects/cdk/tokens/fonts-ready.ts
+++ b/projects/cdk/tokens/fonts-ready.ts
@@ -1,14 +1,14 @@
 import {DOCUMENT} from '@angular/common';
 import {inject, InjectionToken} from '@angular/core';
-import {from, NEVER, Observable} from 'rxjs';
+import {EMPTY, from, Observable} from 'rxjs';
 import {shareReplay} from 'rxjs/operators';
 
 export const TUI_FONTS_READY = new InjectionToken<Observable<unknown>>(
     `[TUI_FONTS_READY]`,
     {
         factory: () =>
-            from((inject(DOCUMENT) as any).fonts?.ready || NEVER).pipe(
-                shareReplay({bufferSize: 1, refCount: true}),
+            from((inject(DOCUMENT) as any).fonts?.ready || EMPTY).pipe(
+                shareReplay({bufferSize: 1, refCount: false}),
             ),
     },
 );
--- a/projects/cdk/tokens/fonts-ready.ts
+++ b/projects/cdk/tokens/fonts-ready.ts
@@ -1,0 +1,14 @@
+import {DOCUMENT} from '@angular/common';
+import {inject, InjectionToken} from '@angular/core';
+import {from, NEVER, Observable} from 'rxjs';
+import {shareReplay} from 'rxjs/operators';
+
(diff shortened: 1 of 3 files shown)


projects/cdk/tokens/fonts-ready.ts
```diff
@@ -0,0 +1,14 @@
+import {DOCUMENT} from '@angular/common';
+import {inject, InjectionToken} from '@angular/core';
+import {from, NEVER, Observable} from 'rxjs';
+import {shareReplay} from 'rxjs/operators';
+
+export const TUI_FONTS_READY = new InjectionToken<Observable<unknown>>(
+    `[TUI_FONTS_READY]`,
+    {
+        factory: () =>
+            from((inject(DOCUMENT) as any).fonts?.ready || NEVER).pipe(
```
nsbarsukov, line 10: Guys, `NEVER` or `EMPTY` ? 🤔

splincode: I think `EMPTY` prefer here

NEVER well… It never fires anything. It doesn’t emit anything nor does it complete the observable. This may be useful in very limited cases where you need to return a value, but you don’t want subscribers to act on it *or* for anyone to act on the observable completion. (c)

splincode (approved):

vladimirpotekhin (approved):

projects/cdk/tokens/fonts-ready.ts
vladimirpotekhin: I think `EMPTY`, because it terminates, and `NEVER` does not

waterplea (approved):
I don't think we need `refCount`. If everybody unsubscribes it would reset the stream and this event will never happen again. I think it's ok to keep it in the memory forever. Also let's go with `EMPTY` and then we can merge.

projects/cdk/tokens/fonts-ready.ts
nsbarsukov: fixed

PR #5507, opened 2023-10-01
feat(core): introduces `TypedTuiMapper`

### LINKED ISSUE #5506: 🚀 - More type safe MapperPipe
### Description

The mapper pipe works perfectly, but currently, it's not type safe for more than one arguments, because of any[] varargs.

### Proposed Solution

Since typescript 4.0 we can use variadic tuple types to solve this problem.

(This unfortunately also includes a breaking change in the `TuiMapper` type)

```ts
// helper type
type Arr = unknown[];

export type TuiMapper<T extends Arr, G> = (...args: T) => G;

@Pipe({
  name: "mapper",
})
export class MapperPipe implements PipeTransform {
  transform<T extends Arr, V, G>(
    value: V,
    mapper: TuiMapper<[V, ...T], G>,
    ...args: T
  ): G {
    return mapper(value, ...args);
  }
}
```

This is breaking, because if someone used the type `TuiMapper` in component file, then it must be changed like the following.

```diff
- export type TuiMapper<T, G> = (item: T, ...args: any[]) => G;
+ export type TuiMapper<T extends Arr, G> = (...args: T) => G;

- const mapper: TuiMapper<number, string> = (a,, b) => `${a + b}`;
+ const mapper: TuiMapper<[number, number], string> = (a, b) => `${a + b}`; 
```

Frequency21:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #5506

### DIFF
--- a/projects/cdk/types/mapper.ts
+++ b/projects/cdk/types/mapper.ts
@@ -1,4 +1,11 @@
 /**
  * Typed mapping function.
+ *
+ * @deprecated Will be removed in v4.0. Use `TypedTuiMapper` instead.
  */
-export type TuiMapper<T extends unknown[], G> = (...args: T) => G;
+export type TuiMapper<T, G> = (item: T, ...args: any[]) => G;
+
+/**
+ * Typed mapping function.
+ */
+export type TypedTuiMapper<T extends unknown[], G> = (...args: T) => G;
--- a/projects/cdk/types/mapper.ts
+++ b/projects/cdk/types/mapper.ts
@@ -1,11 +1,11 @@
 /**
  * Typed mapping function.
  *
- * @deprecated Will be removed in v4.0. Use `TypedTuiMapper` instead.
+ * @deprecated Will be removed in v4.0. Use {@link TuiTypedMapper} instead.
  */
 export type TuiMapper<T, G> = (item: T, ...args: any[]) => G;
 
 /**
  * Typed mapping function.
  */
-export type TypedTuiMapper<T extends unknown[], G> = (...args: T) => G;
+export type TuiTypedMapper<T extends unknown[], G> = (...args: T) => G;
--- a/projects/cdk/types/mapper.ts
+++ b/projects/cdk/types/mapper.ts
@@ -1,4 +1,4 @@
 /**
  * Typed mapping function.
  */
-export type TuiMapper<T, G> = (item: T, ...args: any[]) => G;
+export type TuiMapper<T extends unknown[], G> = (...args: T) => G;
--- a/projects/cdk/types/matcher.ts
+++ b/projects/cdk/types/matcher.ts
@@ -1,17 +1,17 @@
 import {TuiStringHandler} from './handler';
-import {TuiMapper, TypedTuiMapper} from './mapper';
+import {TuiMapper, TuiTypedMapper} from './mapper';
 
 /**
  * A matcher function to test items against with extra arguments.
  *
- * @deprecated Will be removed in v4.0. Use `TypedTuiMatcher` instead.
+ * @deprecated Will be removed in v4.0. Use {@link TuiTypedMatcher} instead.
  */
 export type TuiMatcher<I> = TuiMapper<I, boolean>;
 
 /**
  * A matcher function to test items against with extra arguments.
  */
-export type TypedTuiMatcher<I extends unknown[]> = TypedTuiMapper<I, boolean>;
+export type TuiTypedMatcher<I extends unknown[]> = TuiTypedMapper<I, boolean>;
 
 export type TuiStringMatcher<I> = (
     item: I,
--- a/projects/cdk/types/matcher.ts
+++ b/projects/cdk/types/matcher.ts
@@ -4,7 +4,7 @@ import {TuiMapper} from './mapper';
 /**
  * A matcher function to test items against with extra arguments.
  */
-export type TuiMatcher<I> = TuiMapper<I, boolean>;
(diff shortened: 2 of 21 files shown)


waterplea:
Is it a breaking change? Can we avoid breaking by defaulting to any until 4.0?

Frequency21:
The change in `TuiMapper` is definitely breaking and it can't be avoided (at least by default generic parameters or something) I think. But the PR can be made non-breaking with the following changes.

We declare a new `TypedTuiMapper` type next to `TuiMapper` like the following.

```ts
/**
 * Typed mapping function.
 * 
 * @deprecated Will be removed in v4.0. Use `TypedTuiMapper` instead.
 */
export type TuiMapper<T, G> = (item: T, ...args: any[]) => G;

export type TypedTuiMapper<T extends unknown[], G> = (...args: T) => G;
```

We add an overload with the `TypedTuiMapper` to the `MapperPipe.transform` method as proposed.

```ts
@Pipe({name: `tuiMapper`})
export class TuiMapperPipe implements PipeTransform {
    /**
     * Maps object to an arbitrary result through a mapper function
     *
     * @param value an item to transform
     * @param mapper a mapping function
     * @param args arbitrary number of additional arguments
     */
    transform<T, G>(value: T, mapper: TuiMapper<T, G>, ...args: any[]): G;
    transform<T extends unknown[], U, G>(
        value: U,
        mapper: TypedTuiMapper<[U, ...T], G>,
        ...args: T
    ): G {
        return mapper(value, ...args);
    }
}
```

I will make a commit to this PR with these modifications to see the overall picture.

projects/cdk/types/mapper.ts
```diff
@@ -1,4 +1,11 @@
...
  * Typed mapping function.
+ *
+ * @deprecated Will be removed in v4.0. Use `TypedTuiMapper` instead.
  */
 export type TuiMapper<T, G> = (item: T, ...args: any[]) => G;
+
+/**
+ * Typed mapping function.
+ */
+export type TypedTuiMapper<T extends unknown[], G> = (...args: T) => G;
```
waterplea, line 11: Please change to `TuiTypedMapper` because our public imports should start with tui prefix.

projects/cdk/types/matcher.ts
```diff
@@ -1,11 +1,18 @@
...
  * A matcher function to test items against with extra arguments.
+ *
+ * @deprecated Will be removed in v4.0. Use `TypedTuiMatcher` instead.
  */
 export type TuiMatcher<I> = TuiMapper<I, boolean>;
 
+/**
+ * A matcher function to test items against with extra arguments.
+ */
+export type TypedTuiMatcher<I extends unknown[]> = TypedTuiMapper<I, boolean>;
```
waterplea, line 14: Same here.

projects/core/components/calendar/calendar.component.ts
```diff
@@ -81,17 +81,17 @@ export class TuiCalendarComponent implements TuiWithOptionalMinMax<TuiDay> {
...
-        TuiBooleanHandler<TuiDay>,
+    readonly disabledItemHandlerMapper: TypedTuiMapper<
+        [TuiBooleanHandler<TuiDay>, TuiDay, TuiDay],
         TuiBooleanHandler<TuiDay>
-    > = (disabledItemHandler, min: TuiDay, max: TuiDay) => item =>
+    > = (disabledItemHandler, min, max) => item =>
         item.dayBefore(min) || item.dayAfter(max) || disabledItemHandler(item);
 
-    get computedMin(): TuiMonth {
+    get computedMin(): TuiDay {
```
waterplea, line 90: New typed mapper already found a mistake? 👍

projects/kit/components/calendar-range/calendar-range.template.html
```diff
@@ -29,7 +29,7 @@
             class="t-menu"
         >
             <button
-                *ngFor="let item of items | tuiMapper: mapper : min : max : minLength : (otherDateText$ | async)"
+                *ngFor="let item of items | tuiMapper: mapper : min : max : minLength : (otherDateText$ | async) || ''"
```
waterplea, line 32: Let's add `()` so it's clear at a glance what this fallback is applied to

waterplea (commented):
Looks great! Migration would be very appreciated 👍

projects/kit/components/calendar-range/calendar-range.template.html
Frequency21: Prettier removes it, looks like its obvious for it 😂.

I remove the fallback and use default args in the template instead.

Frequency21:
Hi @waterplea! Are the pipes broken or I did something bad?

splincode:
@Frequency21 I think it's my fault about CI, please wait for a while

splincode:
@Frequency21 Could you change `taiga-family/ci/actions/setup/checkout@v...` to `taiga-family/ci/actions/setup/checkout@v1.29.5` in your branch yml files for test?

Frequency21:
Of course.

splincode:
so, apparently it doesn't work yet

what about `taiga-family/ci/actions/setup/checkout@v1.29.5` -> `actions/checkout@v4.1.0`?

Frequency21:
Let's try

Frequency21:
Looks good to me :) thanks for the quick fix

splincode:
@Frequency21 please, rebase into main branch, you can revert `action/checkout`

waterplea (approved):

splincode (approved):

PR #5540, opened 2023-10-05
docs(demo): add an example for wrapping tooltip

Drogenix:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes
- [x] Documentation content changes

https://github.com/taiga-family/taiga-ui/blob/main/projects/demo/src/modules/components/tooltip/examples/1/index.html

### DIFF
--- a/projects/demo/src/modules/components/tooltip/examples/1/index.less
+++ b/projects/demo/src/modules/components/tooltip/examples/1/index.less
@@ -4,3 +4,15 @@
     height: 1.25rem;
     min-width: 6.25rem;
 }
+
+.wrapping-tooltip {
+    width: fit-content;
+    background: var(--tui-base-02);
+    border-radius: 0.5rem;
+    line-height: 1.5rem;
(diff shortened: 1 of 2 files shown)


splincode (approved):

projects/demo/src/modules/components/tooltip/examples/1/index.less
```diff
@@ -4,3 +4,15 @@
     height: 1.25rem;
     min-width: 6.25rem;
 }
+
+.wrapping-tooltip {
+    width: fit-content;
```
waterplea, line 9: How about we add:
```css
resize: horizontal;
overflow: auto;
```
so we can play around with it without resizing the whole window?

Drogenix: Great idea! I'll do it :)

waterplea (approved):

PR #5580, opened 2023-10-09
feat(core): support show native scrollbar

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

## What is the new behavior?

```ts
    providers: [
        tuiScrollbarOptionsProvider({
            native: true,
        }),
    ],
```


waterplea:
What would be the purpose, though? 🤔 We need to be able to toggle native scrollbars on root, I guess, but for the rest you can just not use `tui-scrollbar` if you want native scrollbars. Am I missing some use case?

PR #5621, opened 2023-10-12
refactor: use nullish coalescing

splincode:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring
- [ ] Tests related changes

### DIFF
--- a/projects/kit/components/input-date-range/input-date-range.component.ts
+++ b/projects/kit/components/input-date-range/input-date-range.component.ts
@@ -250,7 +250,7 @@ export class TuiInputDateRangeComponent
     }
 
     get nativeValue(): string {
-        return this.nativeFocusableElement ? this.nativeFocusableElement.value : '';
+        return this.nativeFocusableElement?.value || '';
     }
 
     set nativeValue(value: string) {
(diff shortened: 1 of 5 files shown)


projects/kit/components/input-date-range/input-date-range.component.ts
```diff
@@ -250,7 +250,7 @@ export class TuiInputDateRangeComponent
     }
 
     get nativeValue(): string {
-        return this.nativeFocusableElement ? this.nativeFocusableElement.value : '';
+        return this.nativeFocusableElement?.value ?? '';
```
nsbarsukov, line 253: Should we use `||` for the final fallback in all changed lines ? 🤔 

```suggestion
        return this.nativeFocusableElement?.value || '';
```

nsbarsukov (approved):

projects/kit/components/input-date-range/input-date-range.component.ts
waterplea: I think so too, there's no need to concretize the fallback case to nullish only, it's just a fallback.

waterplea (approved):

projects/kit/components/input-date-range/input-date-range.component.ts
splincode: fixed

PR #5631, opened 2023-10-13
feat(addon-table): table column direction reset on third click

Drogenix:
## PR Type

What kind of change does this PR introduce?

- [x] Feature
- [ ] Bugfix
- [ ] Refactoring
- [ ] Build or CI related changes
- [ ] Tests related changes
- [ ] Documentation content changes

## What is the current behaviour?

Closes [#5064](https://github.com/taiga-family/taiga-ui/issues/5064)

PR #5640, opened 2023-10-14
feat(addon-table): reset column direction on third click

### LINKED ISSUE #5064: 🚀 - TuiTable - reset column direction on third click
### Description

Now we can only change column direction in ascending or descending order but we cant reset column direction to default.
My suggestion: 
1) column first click - ascending sort
2) column second click - descending sort
3) column third click - reset sort

Example: https://material.angular.io/components/sort/overview#sort-overview

JoelLau:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #5064 

click table headers currently toggles through asc and desc states (2 states)

## What is the new behaviour?

adding a 3rd state to column sort behaviour when clicking table headers - rotate through:
1. asc
2. desc
3. off

### DIFF
--- a/projects/addon-table/components/table/directives/table.directive.ts
+++ b/projects/addon-table/components/table/directives/table.directive.ts
@@ -16,7 +16,7 @@ import {Observable} from 'rxjs';
 
 import {TUI_STUCK} from '../providers/stuck.provider';
 import {TUI_TABLE_PROVIDERS} from '../providers/table.providers';
-import {TUI_TABLE_OPTIONS, TuiTableOptions} from '../table.options';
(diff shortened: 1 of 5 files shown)


waterplea:
Hi @JoelLau. Thank you for this PR. However there's already a fix for it which is simpler:
https://github.com/taiga-family/taiga-ui/pull/5631
I'm not sure adding 0 mode to the direction makes sense as it basically means not to sort and instead of introducing a "not sort" mode for sorter, removing sorter altogether (setting it to `null`) seems cleaner to me.

PR #5641, opened 2023-10-14
feat(kit): prevent combobox dropdown opening on value reset

### LINKED ISSUE #5598: 🐞 - Разные поведения при закрытии выпадающего списка с использованием tuiTextfieldCleaner
### Playground Link

https://stackblitz.com/edit/angular-grj9sv?file=src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.component.html&hideExplorer=1

### Description

При сбросе значения через крест, в select компоненте происходит закрытие списка и сброс значения.

При сбросе значения в комбобоксе, происходит автофокус на инпут и список снова выпадает.

### Angular version

12.2.17

### Taiga UI version

3.48.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

ddubrava:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #5598

## What is the new behaviour?

A combox dropdown is not opened on value reset

### DIFF
--- a/projects/kit/components/combo-box/combo-box.component.ts
+++ b/projects/kit/components/combo-box/combo-box.component.ts
@@ -207,7 +207,10 @@ export class TuiComboBoxComponent<T>
             this.value = null;
         }
 
-        this.hostedDropdown?.updateOpen(true);
+        // Clearing sets the empty value, the dropdown should not be opened on clear.
+        if (this.search !== '') {
+            this.hostedDropdown?.updateOpen(true);
+        }
     }
 
     /** @deprecated use 'value' setter */
--- a/projects/kit/components/combo-box/test/combo-box.component.spec.ts
+++ b/projects/kit/components/combo-box/test/combo-box.component.spec.ts
@@ -159,7 +159,9 @@ describe(`ComboBox`, () => {
         expect(testComponent.control.value).toBeNull();
 
         testComponent.items = ITEMS;
+        testComponent.component.toggle();
         fixture.detectChanges();
+        inputPO.sendText(`Sly cat`);
 
         await fixture.whenStable();
         expect(testComponent.control.value).toBe(ITEMS[1]);
@@ -209,8 +211,14 @@ describe(`ComboBox`, () => {
         });
 
         describe(`dropdown`, () => {
-            it(`empty value opens dropdown`, () => {
+            it(`empty value does not open dropdown`, () => {
                 testComponent.component.onValueChange(``);
+                fixture.detectChanges();
+                expect(testComponent.component.open).toBe(false);
+            });
+
+            it(`not empty value opens dropdown`, () => {
+                testComponent.component.onValueChange(`raccoon`);
                 fixture.detectChanges();
                 expect(testComponent.component.open).toBe(true);
             });
--- a/projects/kit/components/input-tag/input-tag.component.ts
+++ b/projects/kit/components/input-tag/input-tag.component.ts
@@ -194,9 +194,6 @@ export class TuiInputTagComponent
         readonly controller: TuiTextfieldController,
         @Inject(TUI_INPUT_TAG_OPTIONS)
         private readonly options: TuiInputTagOptions,
-        @Optional()
-        @Inject(TuiHostedDropdownComponent)
-        private readonly parentHostedDropdown: TuiHostedDropdownComponent | null,
         @Inject(TUI_COMMON_ICONS) readonly icons: TuiCommonIcons,
     ) {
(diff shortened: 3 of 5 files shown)


ddubrava:
Hi! I hope this change is safe. I've checjed that when you type something or make a selection, the dropdown always remains open. So, it's unnecessary to set it as open again. The only scenario where the value is updated and the dropdown is closed is when you reset the value

ddubrava:
> Hi! I hope this change is safe. I've checjed that when you type something or make a selection, the dropdown always remains open. So, it's unnecessary to set it as open again. The only scenario where the value is updated and the dropdown is closed is when you reset the value

well, it's not true if tests are failed. i don't think the fix is trivial. i'll check it today

projects/kit/components/combo-box/combo-box.component.ts
```diff
@@ -206,8 +206,6 @@ export class TuiComboBoxComponent<T>
         if (this.strict || this.search === '') {
             this.value = null;
         }
-
-        this.hostedDropdown?.updateOpen(true);
```
waterplea, line 210: Please check input-tage/multi-select as well

ddubrava:
image

Isn't it a test on the feature that we are fixing here? Do not open a dropdown if a value becomes empty?

image

Regarding this one, sorry, but I do not understand why it doesn't work. How `open` affects the stored value in a contol

projects/kit/components/combo-box/combo-box.component.ts
ddubrava: Removed there as well

splincode:
@waterplea what we do next?

waterplea:
> @waterplea what we do next?

Decide together with Sergey, our designer, what behavior we want to have. I think clicking on X to not change the current state of the drop-down is a good idea, it's less annoying to click a second time if you intended to clear and select a new value than to click away to close the drop-down if you just wanted to clear the control.

ddubrava:
So can we close it, ye?

splincode (approved):

splincode:
@ddubrava could you fix unit test please?

ddubrava:
> @ddubrava could you fix unit test please?

not sure, it's the tricky one. One test probably can be removed, but I don't know why the second test fails

https://github.com/taiga-family/taiga-ui/pull/5641#issuecomment-1763346772

splincode:
let's just change it to

image

waterplea (approved):

vladimirpotekhin:
@ddubrava the dropdown should open when you input a value


https://github.com/taiga-family/taiga-ui/assets/46284632/49873bc1-a795-49be-b5d4-fa8c749b007d

ddubrava:
Hi! I fixed this bug, but assuming that the empty value always represents clearing is very unreliable. I would create a 'clear' event from the `TuiPrimitiveTextfieldComponent` and allow each consumer to decide how they would like to handle this. 

Currently, there is a 'collision' between a true empty value (e.g., when the user deletes everything) and the 'clear' event. We cannot distinguish between them since the 'clear' event simply sets an empty value without triggering any events.

waterplea (approved):

splincode (approved):

splincode:
@vladimirpotekhin fyi

waterplea:
> Hi! I fixed this bug, but assuming that the empty value always represents clearing is very unreliable. I would create a 'clear' event from the `TuiPrimitiveTextfieldComponent` and allow each consumer to decide how they would like to handle this.
> 
> Currently, there is a 'collision' between a true empty value (e.g., when the user deletes everything) and the 'clear' event. We cannot distinguish between them since the 'clear' event simply sets an empty value without triggering any events.

I think it's good enough for now, could you please do the same in multiselect?

ddubrava:
> > Hi! I fixed this bug, but assuming that the empty value always represents clearing is very unreliable. I would create a 'clear' event from the `TuiPrimitiveTextfieldComponent` and allow each consumer to decide how they would like to handle this.
> > Currently, there is a 'collision' between a true empty value (e.g., when the user deletes everything) and the 'clear' event. We cannot distinguish between them since the 'clear' event simply sets an empty value without triggering any events.
> 
> I think it's good enough for now, could you please do the same in multiselect?

sure, I will. I didn't add it because the dropdown is always opened in multiselect

vladimirpotekhin (approved):

PR #5646, opened 2023-10-16
feat(core): support override tui-svg width and height by css variable

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

### DIFF
--- a/projects/core/components/svg/svg.style.less
+++ b/projects/core/components/svg/svg.style.less
@@ -1,5 +1,3 @@
-@default-size: 1.5rem;
-
 :host {
     display: inline-flex;
     vertical-align: middle;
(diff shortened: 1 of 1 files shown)


projects/core/components/svg/svg.style.less
```diff
@@ -1,14 +1,12 @@
...
 :host {
     display: inline-flex;
     vertical-align: middle;
     flex-shrink: 0;
     align-items: center;
     justify-content: center;
     line-height: 0;
-    height: @default-size;
-    width: @default-size;
+    height: var(--tui-svg-size, 1.5rem);
```
waterplea, line 8: But that's on the host, what's the point of it?

splincode: but if your icons

```html
<svg height="inherit" width="inherit" />
```

you can immediately set the overall size for all icons

image

Issue #5645, opened 2023-10-16
🐞 - `Table` disappearing borders


waterplea:
### Playground Link

https://taiga-ui.dev/components/table

### Description

Sometimes when resizing the table or reordering rows/columns, borders might disappear. Retriggering `isolation: isolate` or any other paint related CSS seems to bring them back:

![image](https://github.com/taiga-family/taiga-ui/assets/11832552/dbc5be6e-202f-4765-8207-c9ab40bc4dff)

### Which browsers have you used?

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

Drogenix:
@waterplea I figured that removing the `transform: translate3d(0,0,0)` in th and td elements fixes this bug: borders always appear. You can see this in the video:

https://github.com/taiga-family/taiga-ui/assets/101005283/a930ce2b-5b67-4fb3-baab-95fc88aa7666

waterplea:
> @waterplea I have found that removing the `transform: translate3d(0,0,0)` in th and td elements fixes this bug: borders always appear. You can see this in the video:
> 
>  fix.mp4

That's great! We use it to establish new containing block for fixed positioned border on controls inside table. Please see if any of the following options (4th bullet) are a suitable replacement:
https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block

Drogenix:
So i checked them all, `filter: opacity(1)`, for example, works well, but tables with sticky header are still broken... I don't know, how we can fix this :)

Drogenix:
No, they are just broke, even with `transform: translate(0)`. `filter` doens't have any affect on them

Drogenix:
@waterplea I didn't find any better solution then using `filter` and animation for th element (which we discussed earlier in telegram) to force the browser to repaint our table. Animation look like this:
```css
@keyframes repaint{
    100%{
        tab-size: 1;
    }
}
```
What do you think?

waterplea:
> @waterplea I didn't find any better solution then using `filter` and animation for th element (which we discussed earlier in telegram) to force the browser to repaint our table. Animation look like this:
> ```css
> @keyframes repaint{
>     100%{
>         tab-size: 1;
>     }
> }
> ```
> What do you think?

Ok, let's go for it for the time being.

[...]

PR #5650, opened 2023-10-16
fix(addon-table): table disappearing borders

Drogenix:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes  [#5645](https://github.com/taiga-family/taiga-ui/issues/5645)

## What is the new behaviour?

I added a simple animation for "th" and "td" which updates the border-color. The same animation but with setting "isolation: isolate" doesn't work as needed

It's fixes a bug, but there's a problem: if we want to customize the border-color, we also should change the animation


projects/addon-table/components/table/td/td.style.less
```diff
@@ -31,6 +35,12 @@
     }
 }
 
+@keyframes repaint {
+    to {
+        border-right-color: var(--tui-base-04);
```
waterplea, line 40: People might want to override this. Can we do something like `isolation: isolate` here? Or some other harmless rule. We need the same in `th` too.

Drogenix: I tried this, but it's doesn't work. I'm gonna try to find another solution, but if i won't be able to, we can just add a css variable for table border-color and use it in animation. What do you think?

PR #5662, opened 2023-10-17
fix(core): use correct host portal viewport for dropdown position

splincode:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes https://github.com/taiga-family/tui-editor/issues/201

image

## What is the new behaviour?

image

### DIFF
--- a/projects/core/directives/dropdown/dropdown-position.directive.ts
+++ b/projects/core/directives/dropdown/dropdown-position.directive.ts
@@ -36,23 +36,30 @@ export class TuiDropdownPositionDirective extends TuiPositionAccessor {
         }
 
         const hostRect = this.accessor?.getClientRect() ?? EMPTY_CLIENT_RECT;
-        const viewport = this.viewport.getClientRect();
+        const viewportRect = this.viewport.getClientRect();
         const {minHeight, align, direction, offset} = this.options;
+        const viewport = {
+            top: viewportRect.top - offset,
+            bottom: viewportRect.bottom - offset,
+            right: viewportRect.right - offset,
+            left: viewportRect.left - offset,
+        } as const;
         const previous = this.previous || direction || 'bottom';
-        const right = Math.max(hostRect.right - width, offset);
         const available = {
             top: hostRect.top - 2 * offset - viewport.top,
             bottom: viewport.bottom - hostRect.bottom - 2 * offset,
         } as const;
+        const right = Math.max(hostRect.right - width, offset);
+        const left = hostRect.left + width < viewport.right ? hostRect.left : right;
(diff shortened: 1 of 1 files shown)


projects/core/directives/dropdown/dropdown-position.directive.ts
```diff
@@ -44,15 +44,17 @@ export class TuiDropdownPositionDirective extends TuiPositionAccessor {
             top: hostRect.top - 2 * offset - viewport.top,
             bottom: viewport.bottom - hostRect.bottom - 2 * offset,
         } as const;
+        const left =
+            hostRect.left + width < viewport.right - offset ? hostRect.left : right;
         const position = {
             top: hostRect.top - offset - height,
             bottom: hostRect.bottom + offset,
-            right,
+            right: viewport.left > right ? viewport.left : right,
```
waterplea, line 52: Should it be `viewport.left + offset`? Also `Math.max(viewport.left + offset, right)`

projects/core/directives/dropdown/dropdown-position.directive.ts
```diff
@@ -44,15 +44,17 @@ export class TuiDropdownPositionDirective extends TuiPositionAccessor {
...
             center:
                 hostRect.left + hostRect.width / 2 + width / 2 < viewport.right - offset
                     ? hostRect.left + hostRect.width / 2 - width / 2
                     : right,
-            left: hostRect.left + width < viewport.right - offset ? hostRect.left : right,
+            left: viewport.left > left ? viewport.left : left,
```
waterplea, line 57: Same here

projects/core/directives/dropdown/dropdown-position.directive.ts
```diff
@@ -44,15 +44,17 @@ export class TuiDropdownPositionDirective extends TuiPositionAccessor {
             top: hostRect.top - 2 * offset - viewport.top,
             bottom: viewport.bottom - hostRect.bottom - 2 * offset,
         } as const;
+        const left =
+            hostRect.left + width < viewport.right - offset ? hostRect.left : right;
```
waterplea, line 48: I wonder if we can somehow simplify all the offset subtractions by just subtracting it from the viewport rect initially 🤔

waterplea (commented):
Do we need something like that for hints too?

splincode:
`Do we need something like that for hints too?` maybe separate PR

splincode:
@vladimirpotekhin hello, please review me)

vladimirpotekhin (approved):

PR #5669, opened 2023-10-17
fix(core): correct `TuiHostedDropdownComponent` with `tuiHostedDropdownHost`

olegnowikow:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

In the [docs](https://taiga-ui.dev/components/hosted-dropdown): 
> If you want another element to be host, use `tuiHostedDropdownHost` directive.

However, this [doesn't work](https://stackblitz.com/edit/angular-wugh4z?file=src%2Fapp%2Fapp.component.html) when using `tuiDropdownHover`.

## What is the new behaviour?

This fix solves problem of displaying a dropdown list on hover for selected item


projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
```diff
@@ -103,8 +103,21 @@ export class TuiHostedDropdownComponent implements TuiFocusableElementAccessor {
...
     canOpen = true;
 
+    private readonly onHoverTarget$ = new Subject<HTMLElement>();
+
+    private readonly _hover$ = combineLatest([
+        this.onHoverTarget$,
+        this.hover$ || EMPTY,
+    ]).pipe(
+        map(
+            ([target, hovered]: [HTMLElement, boolean]) =>
```
waterplea, line 113: I don't think you need to types these explicitly

projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
```diff
@@ -103,8 +103,21 @@ export class TuiHostedDropdownComponent implements TuiFocusableElementAccessor {
...
+                hovered && this.computedHost.contains(target),
+        ),
+        share(),
```
waterplea, line 116: Do we need a `share` here? Seems like it is only used in one place. Maybe we should add `share` to `open$` instead?

projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
```diff
@@ -184,6 +197,11 @@ export class TuiHostedDropdownComponent implements TuiFocusableElementAccessor {
         }
     }
 
+    @HostListener('mouseover', ['$event.target'])
```
waterplea, line 200: This will spam change detection as mouse moves across nested tags. How about we use `tuiTypedFromEvent(this.el.nativeElement, 'mouseover')` instead of `onHoverTarget$`?

projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
```diff
@@ -103,8 +103,21 @@ export class TuiHostedDropdownComponent implements TuiFocusableElementAccessor {
     @Input()
     canOpen = true;
 
+    private readonly onHoverTarget$ = new Subject<HTMLElement>();
+
+    private readonly _hover$ = combineLatest([
```
waterplea, line 108: I don't like underscore names, let's name it `hostHover$` instead so we know it's about host being hovered.

projects/demo/src/modules/directives/dropdown-hover/examples/4/index.html
```diff
@@ -0,0 +1,30 @@
...
+            tuiHostedDropdownHost
+            type="button"
+        >
+            it will open here
+        </button>
+    </div>
+</tui-hosted-dropdown>
+<ng-template
+    #dropdown
+    let-activeZone
```
waterplea, line 27: This one is unused, let's remove it to simplify the example.

projects/demo/src/modules/directives/dropdown-hover/examples/4/index.html
```diff
@@ -0,0 +1,30 @@
+<tui-hosted-dropdown
+    tuiDropdownHover
+    tuiDropdownLimitWidth="fixed"
+    [content]="dropdown"
+    [(open)]="open"
+>
+    <div tuiGroup>
+        <button
+            size="l"
```
waterplea, line 9: I think we can remove size and type for both buttons to simplify the example

projects/demo/src/modules/directives/dropdown-hover/examples/4/index.html
```diff
@@ -0,0 +1,30 @@
+<tui-hosted-dropdown
+    tuiDropdownHover
+    tuiDropdownLimitWidth="fixed"
+    [content]="dropdown"
+    [(open)]="open"
```
waterplea, line 5: Do we need to track `open`? We can probably drop it from here.

projects/demo/src/modules/directives/dropdown-hover/dropdown-hover.component.ts
```diff
@@ -36,6 +36,11 @@ export class ExampleTuiDropdownHoverComponent extends AbstractExampleTuiDropdown
         HTML: import('./examples/3/index.html?raw'),
     };
 
+    readonly example4: TuiDocExample = {
+        TypeScript: import('./examples/4/index.ts?raw'),
```
waterplea, line 40: If we drop `open` we can probably keep just `HTML` here.

projects/core/components/hosted-dropdown/hosted-dropdown.component.ts
```diff
@@ -103,10 +104,21 @@ export class TuiHostedDropdownComponent implements TuiFocusableElementAccessor {
     @Input()
     canOpen = true;
 
+    private readonly hostHover$ = combineLatest([
+        tuiTypedFromEvent(this.el.nativeElement, 'mouseover'),
+        this.hover$ || EMPTY,
+    ]).pipe(
+        map(
+            ([{target}, hovered]) =>
+                hovered && this.computedHost.contains(target as HTMLElement),
```
waterplea, line 113: To take delays into account you can rewrite the first part of this stream as follows:
```ts
tuiTypedFromEvent(this.el.nativeElement, 'mouseover').pipe(
    map(
        e => this.computedHost.contains(tuiGetActualTarget(e)),
        switchMap(visible =>
            of(visible).pipe(
                delay(
                    (visible ? this.hover$?.showDelay : this.hover$?.hideDelay) || 0,
                ),
            ),
        ),
    ),
);
```

olegnowikow: fixed

waterplea (approved):

splincode (approved):

PR #5676, opened 2023-10-18
refactor: restrict underscore prefix in class properties

splincode:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring
- [ ] Tests related changes

### DIFF
--- a/projects/addon-charts/components/line-chart/line-chart.component.ts
+++ b/projects/addon-charts/components/line-chart/line-chart.component.ts
@@ -40,7 +40,7 @@ import {TuiLineChartHintDirective} from './line-chart-hint.directive';
     viewProviders: [tuiHintOptionsProvider({direction: 'top', hideDelay: 0})],
 })
 export class TuiLineChartComponent {
-    private readonly _hovered$ = new Subject<number>();
+    private readonly hover$ = new Subject<number>();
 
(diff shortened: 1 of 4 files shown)


projects/addon-charts/components/line-chart/line-chart.component.ts
```diff
@@ -40,6 +40,7 @@ import {TuiLineChartHintDirective} from './line-chart-hint.directive';
     viewProviders: [tuiHintOptionsProvider({direction: 'top', hideDelay: 0})],
 })
 export class TuiLineChartComponent {
+    // eslint-disable-next-line @typescript-eslint/naming-convention
     private readonly _hovered$ = new Subject<number>();
```
waterplea, line 44: It's a private property, you can just rename it to something like `hover$`

waterplea (approved):

projects/addon-charts/components/line-chart/line-chart.component.ts
splincode: But this class already has `get hover$`

splincode: fixed

nsbarsukov (approved):

Issue #2900, opened 2022-10-14
🚀 - Allow dynamic changes to `TUI_NUMBER_FORMAT`


theorlovsky:
### Which @taiga-ui/* package(s) are relevant/releated to the feature request?

core

### Description

Different locales have different decimal separators, so it would be good to be able to change it dynamically on language change. Maybe change it to `InjectionToken<TuiNumberFormatSettings | Observable<TuiNumberFormatSettings>>`, this way it will be backward compatible with the previous versions.

waterplea:
That's a good suggestion. Will you be willing to make a PR for it?

theorlovsky:
sure, will take a look

ddubrava:
Hi! Any ideas for the implementation? It seems that `TUI_NUMBER_FORMAT` is used in many places, so it's not trivial.

Issue #2578, opened 2022-09-02
🐞 - Providing TUI_NUMBER_FORMAT with useClass causes incorrect parsing of input


pwespi:
### Which @taiga-ui/* package(s) are the source of the bug?

core

### Please provide a link to a minimal reproduction of the bug

https://github.com/pwespi/tui-input-number

### Is this issue blocking you?

Blocking

### Description

We are providing `TUI_NUMBER_FORMAT` with `useClass` (to be able to change the format dynamically). This causes incorrect parsing of the input when the input loses focus.

![Screen Recording 2022-09-02 at 16 46 04(1)](https://user-images.githubusercontent.com/24232962/188175783-8c137bf8-4c39-4c8f-aa70-ff5451e39e60.gif)

This used to work in Taiga UI 2.x.

### Angular version

^14.1.0

### Taiga UI version

^3.0.1

### Which browsers have you used?

- [X] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Edge

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

ddubrava:
Hey, guys! @splincode

The main reason of this problem is that we use spread operator to pass settings. Since spread operator doesn't include accessors (see https://github.com/Microsoft/TypeScript/issues/26547), we lost all the settings. 

![image](https://user-images.githubusercontent.com/22116465/190873843-6e598458-621a-4c29-80e5-d92801f1cf37.png)

The workaround is to copy everything (including accessors) from the settings to an object but I'm not sure it's a great idea to hardcode this.

waterplea:
Well, this kind of spreading is used everywhere, we cannot take such cases into account in all situations. Can you show the code of your class? Let's try to think of the best workaround together.

ddubrava:
@waterplea https://github.com/pwespi/tui-input-number/blob/main/src/app/app.component.ts

<img width="558" alt="Screenshot 2022-09-27 at 16 18 39" src="https://user-images.githubusercontent.com/22116465/192537199-291dab8a-527e-470c-963a-9e9d77e7c5bc.png">

waterplea:
> @waterplea https://github.com/pwespi/tui-input-number/blob/main/src/app/app.component.ts
> 
[...]

PR #5735, opened 2023-10-24
feat(experimental): `TuiAmount` add new pipe

vladimirpotekhin:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #


splincode (approved):

waterplea (commented):
We will eventually need to support dynamic number formatting:
https://github.com/taiga-family/taiga-ui/issues/2900
https://github.com/taiga-family/taiga-ui/issues/2578

So how about this pipe to return `Observable<string>` so the API is ready for it?

Also, setting number format through just the token is pretty inconvenient. How about we add a new `Observable` token for number formatting and make a directive for it that extends `BehaviorSubject` with `TUI_NUMBER_FORMAT` as initial value and a partial `TuiNumberFormatSettings` as input?

projects/experimental/pipes/amount/amount.options.ts
```diff
@@ -0,0 +1,26 @@
...
+    readonly sign: TuiMoneySign;
+}
+
+export const TUI_AMOUNT_DEFAULT_OPTIONS: TuiAmountOptions = {
+    decimal: `not-zero`,
+    currency: null,
+    currencyAlign: `right`,
+    sign: `negative-only`,
+    precision: 2,
+};
```
waterplea, line 20: Should we move `decimal`, `sign` and `precision` to that new directive I proposed?

projects/experimental/pipes/amount/amount.module.ts
```diff
@@ -0,0 +1,11 @@
+import {NgModule} from '@angular/core';
+import {TuiFormatNumberPipe} from '@taiga-ui/core';
+
+import {TuiAmountPipePipe} from './amount.pipe';
+
+@NgModule({
+    declarations: [TuiAmountPipePipe],
+    exports: [TuiAmountPipePipe],
+    providers: [TuiFormatNumberPipe],
```
waterplea, line 9: Looks like an unnecessary provider.

projects/experimental/pipes/amount/amount.pipe.ts
```diff
@@ -0,0 +1,45 @@
...
+        currency = this.options.currency,
+        currencyAlign = this.options.currencyAlign,
+    ): string {
+        const sign = tuiFormatSignSymbol(value, this.options.sign);
+        const currencySymbol = tuiFormatCurrency(currency);
+        const formatted = tuiFormatNumber(Math.abs(value), {
+            ...this.format,
+            decimalLimit: this.getDecimalLimit(value),
+        });
+        // eslint-disable-next-line no-irregular-whitespace
```
waterplea, line 31: It should be non breaking space, otherwise it could wrap to the next line.

vladimirpotekhin: Yes, this exception is just for use non breaking space. Linter doesn't allow nbsp. Maybe disable this rule?

waterplea: Don't we have a const for it?

projects/core/interfaces/number-format-settings.ts
```diff
@@ -28,4 +28,8 @@ export interface TuiNumberFormatSettings {
      * Enable zeros at the end of decimal part.
      */
     readonly zeroPadding: boolean;
+    /**
+     * Decimal part display mode. ('not-zero' by default)
+     */
+    readonly decimal: TuiDecimal;
```
waterplea, line 34: We need to make it optional until 4.0, otherwise it's a breaking change

waterplea (approved):

PR #5841, opened 2023-11-03
feat(kit): highlight improvements

### LINKED ISSUE #5838: 🚀 - Добавить поддержку searchWords, regex и сaseSensitive в TuiHighlight
### Description

Для TuiHighlight просьба добавить  поддержку:
- Нескольких слов одновременно
- Поиск по регулярным выражениям
- Флаг CaseSensitive

MillerSvt:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

1. Not support for multiple occurrences
2. Not support for multiple highlights
3. Not support for case-sensitive mode
4. Not support for regexp

Closes #5838 

## What is the new behaviour?

1. Adding support for multiple occurrences
2. Adding support for multiple highlights
3. Adding support for case-sensitive mode
4. Adding support for regexp


splincode:
What about if `search` will be `string[]`?

projects/kit/directives/highlight/highlight.component.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    @HostBinding('style.top.px')
+    top = NaN;
+
+    @HostBinding('style.width.px')
+    width = NaN;
+
+    @HostBinding('style.height.px')
+    height = NaN;
+}
```
splincode, line 21: Can we use native `mark` (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark) instead of angular component?

MillerSvt:
> What about if `search` will be `string[]`?

This will work the same way as with regular expressions. Or do you mean that I didn't add a demo?

splincode:
Yes, you didn't add a demo

projects/kit/directives/highlight/highlight.component.ts
MillerSvt: This is interesting, but using mark implies modifying an existing DOM tree, which can lead to unexpected consequences. Mark is good for static text selection, but not for dynamic. In my opinion.

splincode: okey

MillerSvt:
> Yes, you didn't add a demo

added

projects/kit/directives/highlight/highlight.directive.ts
```diff
@@ -12,90 +43,162 @@ import {Observable} from 'rxjs';
...
     );
 
+    private readonly clearHighlights$ = new Subject<void>();
+
+    private readonly addHighlight$ = new Subject<Observable<never>>();
+
+    private readonly cf: ComponentFactory<TuiHighlightComponent>;
+
+    @Input()
+    tuiHighlight: TuiArrayOrValue<RegExp | string> = '';
```
splincode, line 59:
```
 @Input(`tuiHighlight`)
  values: TuiArrayOrValue<RegExp | string> = '';
```

projects/kit/directives/highlight/highlight.directive.ts
```diff
@@ -1,7 +1,38 @@
...
+type TuiArrayOrValue<T> = T | readonly T[];
+
+interface TuiHighlightOccurrence {
+    index: number;
+    length: number;
+}
+
+function tuiToArray<T>(value: T | readonly T[]): readonly T[] {
+    return Array.isArray(value) ? value : [value];
+}
```
splincode, line 35: you can move it to `@taiga-ui/cdk/utils`

projects/kit/directives/highlight/highlight.directive.ts
```diff
@@ -1,7 +1,38 @@
...
+    tuiIsNumber,
+    tuiIsString,
+    TuiResizeService,
+} from '@taiga-ui/cdk';
+import {Observable, Subject} from 'rxjs';
+import {mergeAll, switchMap, takeUntil} from 'rxjs/operators';
+
+import {TuiHighlightComponent} from './highlight.component';
+
+type TuiArrayOrValue<T> = T | readonly T[];
```
splincode, line 26: you can move it to `@taiga-ui/cdk/types`

projects/kit/directives/highlight/test/highlight.directive.spec.ts
```diff
@@ -3,7 +3,7 @@ import {TestBed} from '@angular/core/testing';
 import {TuiHighlightModule} from '@taiga-ui/kit';
 import {configureTestSuite} from '@taiga-ui/testing';
 
-describe(`TuiHighlight directive`, () => {
+describe(`TuiHighlight directive in single occurrence mode`, () => {
```
splincode, line 6: Please, use one top level describe

ref: https://github.com/taiga-family/taiga-ui/pull/5843

splincode (approved):

projects/kit/directives/highlight/highlight.directive.ts
```diff
@@ -12,90 +39,162 @@ import {Observable} from 'rxjs';
...
+                        length: itemValue.length,
+                    };
+                }
+            } else {
+                for (const match of source.matchAll(item)) {
+                    if (tuiIsNumber(match.index) && match.length) {
+                        yield {
+                            index: match.index,
+                            length: match[0].length,
+                        };
```
splincode, line 173: What are the benefits of using generators?

projects/kit/directives/highlight/highlight.directive.ts
```diff
@@ -12,90 +39,162 @@ import {Observable} from 'rxjs';
...
+                    }
+                }
+            }
+        }
     }
 
-    private setUpHighlight(): HTMLElement {
-        const highlight = this.renderer.createElement('div');
-        const {style} = highlight;
+    private *getRanges(): Generator<Range> {
```
splincode, line 180: What are the benefits of using generators?

projects/kit/directives/highlight/highlight.directive.ts
MillerSvt: 1. We can control how many occurrences we want to calculate. In `match` getter, for example, we need only one occurrence for say "yes". In second, if we pass tuiHighlightMultiOccurrences: false, then also we need to calculate only one occurrence.
2. In generators we can split iteration into smaller parts of logic for cleaner code, without iterate same array several times. We can execute all calculations on the fly.

splincode:
@waterplea @nsbarsukov let's wait for our jedi to review

projects/kit/directives/highlight/highlight.directive.ts
```diff
@@ -12,90 +39,162 @@ import {Observable} from 'rxjs';
...
+            .subscribe();
+
+        this.cf = cfr.resolveComponentFactory(TuiHighlightComponent);
     }
 
     get match(): boolean {
-        return this.indexOf(this.el.nativeElement.textContent) !== -1;
+        const [occurrence] = this.getOccurrences(this.el.nativeElement.textContent);
+
+        return Boolean(occurrence);
```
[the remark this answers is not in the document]
MillerSvt: No, it's not possible. Method 'getOccurrences' returns a Generator, not an Array.

MillerSvt:
I found some problems here:
1. We can't highlight multiline occurrences correctly in this solution.
2. We may get performance degradation when resize with large DOM tree.

I'll think about what we can do with it.

MillerSvt:
> I found some problems here

fixed, ready to review

projects/cdk/utils/miscellaneous/to-array.ts
```diff
@@ -0,0 +1,5 @@
+import type {TuiArrayOrValue} from '@taiga-ui/cdk/types';
+
+export function tuiToArray<T>(value: TuiArrayOrValue<T>): readonly T[] {
```
splincode, line 3: use `coerceArray` from angular/cdk

splincode:
@waterplea Could we close it? I'm not sure it makes sense to keep an open PR for three years

waterplea:
I guess so, sorry for wasting your time :(

PR #5857, opened 2023-11-06
ci: add `functional` eslint plugin

splincode:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

### DIFF
--- a/projects/addon-table/directives/table-filters/generic-filter.directive.ts
+++ b/projects/addon-table/directives/table-filters/generic-filter.directive.ts
@@ -13,10 +13,6 @@ import {AbstractTuiTableFilter} from './abstract-table-filter';
     ],
 })
 export class TuiGenericFilterDirective<T, G> extends AbstractTuiTableFilter<T, G> {
-    @Input()
-    tuiGenericFilter: (item: T, value: G) => boolean = ALWAYS_TRUE_HANDLER;
-
-    filter(item: T, value: G): boolean {
-        return this.tuiGenericFilter(item, value);
-    }
+    @Input('tuiGenericFilter')
+    filter: (item: T, value: G) => boolean = ALWAYS_TRUE_HANDLER;
 }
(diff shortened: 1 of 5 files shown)


projects/addon-table/directives/table-filters/generic-filter.directive.ts
```diff
@@ -16,7 +16,5 @@ export class TuiGenericFilterDirective<T, G> extends AbstractTuiTableFilter<T, G
     @Input()
     tuiGenericFilter: (item: T, value: G) => boolean = ALWAYS_TRUE_HANDLER;
 
-    filter(item: T, value: G): boolean {
-        return this.tuiGenericFilter(item, value);
-    }
+    readonly filter = this.tuiGenericFilter;
```
waterplea, line 19: What's the point of this method?

projects/demo/src/modules/tables/table/examples/4/index.ts
waterplea, line 1: Not sure about these changes. Why?

projects/demo/src/modules/tables/table/examples/5/index.ts
```diff
@@ -55,18 +55,15 @@ export class TuiTableExample5 {
 
     readonly columns = ['name', 'dob', 'age'];
 
-    readonly ageSorter: TuiComparator<User> = (a: User, b: User) => getAge(a) - getAge(b);
+    readonly ageSorter: TuiComparator<User> = (a: User, b: User) =>
+        this.getAge(a) - this.getAge(b);
```
waterplea, line 59: Same here. The idea of moving pure functions away from this is:
- We are sure they are pure
- Their names can be minified
- They do not pollute class API (not applicable here since they are still publicly exposed)

projects/addon-table/directives/table-filters/generic-filter.directive.ts
splincode: fixed

projects/demo/src/modules/tables/table/examples/4/index.ts
splincode: fixed

nsbarsukov (approved):

waterplea (approved):

PR #5872, opened 2023-11-07
feat(experimental): `Icon` add new component

waterplea:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #

### DIFF
--- a/projects/demo/src/modules/experimental/icon/examples/import/insert-template.md
+++ b/projects/demo/src/modules/experimental/icon/examples/import/insert-template.md
@@ -1,3 +1,3 @@
 ```html
-<tui-icon>"Batman"</tui-icon>
+<tui-icon [icon]="'tuiIconCheck' | tuiIcon"></tui-icon>
 ```
--- a/projects/demo/src/modules/experimental/icon/examples/import/insert-template.md
+++ b/projects/demo/src/modules/experimental/icon/examples/import/insert-template.md
@@ -1,0 +1,3 @@
+```html
+<tui-icon>"Batman"</tui-icon>
+```
--- a/projects/demo/src/modules/experimental/icon/icon.template.html
+++ b/projects/demo/src/modules/experimental/icon/icon.template.html
@@ -1,0 +1,62 @@
+<tui-doc-page
+    header="Icon"
+    package="EXPERIMENTAL"
+    type="components"
+>
+    <ng-template pageTab>
+        <tui-notification status="warning">
+            This code is
+            <strong>experimental</strong>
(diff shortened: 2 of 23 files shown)


splincode (approved):

projects/demo/src/modules/experimental/icon/examples/import/insert-template.md
```diff
@@ -0,0 +1,3 @@
+```html
+<tui-icon>"Batman"</tui-icon>
```
nsbarsukov: Why as `ng-content` ?
Should it be `<tui-icon icon="Batman"></tui-icon>` ?

projects/demo/src/modules/experimental/icon/icon.template.html
```diff
@@ -0,0 +1,62 @@
...
+        >
+            <tui-icon-example-2></tui-icon-example-2>
+        </tui-doc-example>
+
+        <tui-doc-example
+            id="duo"
+            heading="Two colors"
+            [content]="example3"
+        >
+            <tui-notification class="tui-space_bottom-4">This example only works in modern browsers</tui-notification>
```
nsbarsukov, line 36: Should we add link to `Can i use`?

```suggestion
            <tui-notification class="tui-space_bottom-4">This example only works in <a tuiLink href="https://caniuse.com/?search=mask-mode">modern browsers</a></tui-notification>
```

I think this example can cause many questions from our users (chrome is modern browser but it does not support this feature)?

nsbarsukov: Should we hide this example if it does not support this feature ?

Something like the following:
```less
@supports not (mask-mode: luminance) {
    .all-example {
        display: none;
    }
}
```

nsbarsukov (approved):

projects/demo/src/modules/experimental/icon/examples/import/insert-template.md
waterplea: Fixed

projects/demo/src/modules/experimental/icon/icon.template.html
waterplea: Good idea. I believe Chrome 120 actually should support it, even though it is not mentioned on caniuse. It comes out in a month, we'll see :)

waterplea: Yes, tested in Canary, it will work starting next major version.

Issue #5808, opened 2023-10-31
🤖 - 4.0 migrations


vladimirpotekhin:
### Description

In this discussion we will be listing breaking changes we plan for the next major version. No clear forecast on when it will happen yet. Besides obviously dropping all deprecated stuff, these things will also take place:

- [x] Update Angular to 15/16 (https://github.com/taiga-family/taiga-ui/issues/6158)
- [x] Refactor all CSS variables according to new internal design system
- [x] Move all dependencies to peerDeps to avoid issues with new versions (read ng-polymorpheus issues we constantly face)
- [x] Bump browser support (https://github.com/Tinkoff/taiga-ui/discussions/3232)
- [x] Refactor dialogs/alerts so you won't need to add modules to global module
- [x] Refactor input-files so that there is no built-in rejection mechanism, provide utils for external file processing
- [x] Make all i18n dynamic (number format, date format, etc.)
- [x] Use `--tui-radius-l` on large controls

This list will be updated as we collect more necessary migrations.

We need to cover as many breaking changes with migrations as possible. 

# Migration list

## Experimental

- [x] `Badge`
- [x] `BadgedContent`
- [x] `ThumbnailCard`
- [x] `Avatar`
- [x] `Checkbox`
- [x] `Radio`
- [x] `Radio` identityMatcher? 
- [x] `Toggle`
- [x] `Button`

## CDK

- [x] `TuiMapper / TuiMatcher` 
- [x] `TuiPreventDefault` -> `(event.prevent)="0"`

## Core

- [x] `PrimitiveTextfield` - prefix, postfix, filler
- [x] `TuiAlertOptions` - autocloseTimer, hasIcon
- [x] `HostedDropdown` -> `DropdownOpen`

## Kit
- [x] `InputCount`  --> `InputNumber`
- [x] `InputNumber, InputSlider` -  prefix, postfix
- [x] `InputTag`, `Multiselect`, `InputDateMulti` -  [expandable]=false --> [rows]=1
- [ ] [nativeId] --> id. (toggle, checkbox, etc)
- [ ] InputNumber. [precision][decimal] --> [tuiNumberFormat]={decimalMode, precision}
- [x] progress-segmented  
- [ ] date format
- [x] rating
- [x] thumbnail-card
- [x] Prompt renamed to Confirm
- [x] tuiMobileTabs --> segmented
- [x] `Action` -> `CardLarge`
- [x] marker icon -> avatar
[...]

PR #5881, opened 2023-11-08
chore(cdk): `schematics` add `Avatar` migration

vladimirpotekhin:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

part of #5808


splincode (approved):

projects/cdk/schematics/ng-update/v4/tests/schematic-migrate-avatar.spec.ts
```diff
@@ -0,0 +1,140 @@
+/* eslint-disable rxjs/no-topromise */
```
nsbarsukov, line 1: I think that we should disable this rule for all `**/schematics/**/*.spec.ts` files.

Now every test file for schematics contains this line.

nsbarsukov (approved):

projects/cdk/schematics/ng-update/v4/tests/schematic-migrate-avatar.spec.ts
nsbarsukov: @splincode what do you think ?

Could we update this file ?
https://github.com/taiga-family/linters/blob/main/projects/eslint-plugin-experience/configs/angular/rxjs/index.js

nsbarsukov: Every test file for schematics contains:

```ts
const tree = await runner
    .runSchematicAsync(...)
    .toPromise();
```

splincode: I will do it right now

PR #5910, opened 2023-11-10
chore: small refactor

splincode:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring
- [ ] Tests related changes

### DIFF
--- a/projects/core/components/root/root.style.less
+++ b/projects/core/components/root/root.style.less
@@ -1,5 +1,4 @@
-@import '../../styles/mixins/mixins.less';
-@import '../../styles/mixins/text.less';
+@import '@taiga-ui/core/styles/taiga-ui-local';
 
 // Required by some Taiga UI components
 @keyframes tuiPresent {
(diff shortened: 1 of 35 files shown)


vladimirpotekhin (approved):

projects/core/components/root/root.style.less
```diff
@@ -1,5 +1,5 @@
-@import '../../styles/mixins/mixins.less';
-@import '../../styles/mixins/text.less';
+@import '@taiga-ui/core/styles/mixins/mixins';
+@import '@taiga-ui/core/styles/mixins/text';
```
waterplea, line 2: Why not entire local file?

waterplea (approved):

projects/core/components/root/root.style.less
splincode: what do you mean?

waterplea: I mean single import of `@taiga-ui/core/styles/taiga-ui-local';`

splincode: aaa

PR #5919, opened 2023-11-11
ci: enable `@angular-eslint/component-max-inline-declarations`

splincode:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

### DIFF
--- a/projects/core/directives/hint/hint.component.ts
+++ b/projects/core/directives/hint/hint.component.ts
@@ -42,12 +42,7 @@ const GAP = 4;
 
 @Component({
     selector: 'tui-hint',
-    template: `
-        <span
-            *polymorpheusOutlet="content as text; context: context"
-            [innerHTML]="text"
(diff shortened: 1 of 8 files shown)


projects/core/directives/hint/hint.component.ts
```diff
@@ -42,12 +42,7 @@ const GAP = 4;
 
 @Component({
     selector: 'tui-hint',
-    template: `
-        <span
-            *polymorpheusOutlet="content as text; context: context"
-            [innerHTML]="text"
-        ></span>
-    `,
```
vladimirpotekhin, line 50: Dubious rule. I think it's ok to use template like this in decorator

waterplea: I agree, I think we can just rely on our own judgement whether or not we should move it to a separate file.

PR #5970, opened 2023-11-15
feat(addon-doc): improve code block

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes # 

## What is the new behaviour?

good example

image

----

image

image

image

### DIFF
--- a/projects/addon-doc/components/code/code.style.less
+++ b/projects/addon-doc/components/code/code.style.less
@@ -9,10 +9,19 @@
     font-weight: bold;
 }
 
+.t-header + .t-code {
+    border-radius: 0.25rem;
+}
+
 .t-code {
     position: relative;
     margin: 0;
     white-space: normal;
+    outline: 1px solid var(--tui-base-03);
+
+    ::ng-deep .hljs-ln {
+        width: max-content;
+    }
 }
 
 .t-code + .t-code {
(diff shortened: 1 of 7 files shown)


projects/addon-doc/components/code/code.style.less
```diff
@@ -24,26 +33,41 @@
     top: 0.75rem;
     right: 0.75rem;
     display: flex;
-    gap: 0.375rem;
     justify-content: center;
     align-items: center;
+    flex-direction: row-reverse;
+
+    @media @tui-mobile {
```
vladimirpotekhin, line 40: nit: :host-context(tui-root._mobile) &

vladimirpotekhin (approved):

projects/addon-doc/components/code/code.style.less
splincode: fixed

waterplea (commented):
Maybe it can be `button tuiLink`?
image

projects/demo/src/modules/app/stackblitz/starter/stackblitz-edit-button.component.ts
```diff
@@ -4,13 +4,15 @@ import {ChangeDetectionStrategy, Component} from '@angular/core';
...
             size="s"
             tuiButton
             type="button"
         >
-            Edit
+            <a tuiLink>
+                <tui-svg src="assets/icons/stackblitz.svg"></tui-svg>
+                Edit
+            </a>
         </button>
```
nsbarsukov, line 16: anchor inside button ? 🤔 

Should we just use the following code?
```html
<a
    icon="assets/icons/stackblitz.svg"
    iconAlign="left"
    tuiLink
>
    Edit
</a>
```

splincode: fixed

waterplea: It shouldn't be `a`, it should be `<button tuiLink`, the link component can be applied to buttons too. `a` is a link while `button tuiLink` is a button that looks like a link. You cannot focus `a` without `href` and it is semantically wrong — what's why we made it apply to buttons too :)

splincode: fixed

waterplea (approved):

PR #5981, opened 2023-11-16
chore(demo): add `Dynamic color` example for `ProgressCircle`

nsbarsukov:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes
- [X] Documentation content changes

## What is the new behaviour?
Screenshot 2023-11-16 at 16 02 02

### DIFF
--- a/projects/demo/src/modules/components/progress-circle/examples/5/index.less
+++ b/projects/demo/src/modules/components/progress-circle/examples/5/index.less
@@ -1,4 +1,3 @@
 tui-progress-circle {
-    transform: scaleX(-1) // reversed direction
-        rotate(-90deg); // top side starting point
+    transition: color 2s;
 }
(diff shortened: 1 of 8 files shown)


splincode (approved):

nsbarsukov:
> Let's add a Restart button, otherwise people might miss it because it's down the page.

I can add rxjs operator `repeat()` to `value$`.
It will not overcomplicate example by extra heavy logic.
What do you think ?

waterplea:
> > Let's add a Restart button, otherwise people might miss it because it's down the page.
> 
> @waterplea 
> I can add rxjs operator `repeat()` to `value$`.
> It will not overcomplicate example by extra heavy logic.
> What do you think ?

Sure. My goal is so that people who spent time in examples at the top are not met by just a green circle when they scroll down 🙂

waterplea:
> > Let's add a Restart button, otherwise people might miss it because it's down the page.
> 
> @waterplea 
> I can add rxjs operator `repeat()` to `value$`.
> It will not overcomplicate example by extra heavy logic.
> What do you think ?

Sure. My goal is so that people who spent time in examples at the top are not met by just a green circle when they scroll down 🙂

PR #5999, opened 2023-11-17
fix(addon-table): `Table` disappearing borders

Drogenix:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Part of #5645

## What is the new behaviour?
`filter: opacity(1)` doesn't work as expected in Safari, so we have to use `transform: translate3d(0, 0, 0)`

### DIFF
--- a/projects/addon-table/components/table/th/th.style.less
+++ b/projects/addon-table/components/table/th/th.style.less
@@ -17,6 +17,11 @@
     border: 1px solid var(--tui-base-04);
     // Create new viewport for fixed positioning
     filter: opacity(1);
+    .safari-only({
+        & {
+            transform: translate3d(0, 0, 0);
+        }
+    });
 
     &:not(:first-child) {
         border-left: none;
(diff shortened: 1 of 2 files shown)


splincode (approved):

projects/addon-table/components/table/th/th.style.less
```diff
@@ -17,6 +17,11 @@
     border: 1px solid var(--tui-base-04);
     // Create new viewport for fixed positioning
     filter: opacity(1);
+    .safari-only({
+        & {
```
waterplea, line 21: I don't think `&` is necessary

waterplea (approved):

projects/addon-table/components/table/th/th.style.less
Drogenix: I hope it would :) 
Fixed!

PR #6050, opened 2023-11-21
chore(demo): improve home page

splincode:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring
- [ ] Tests related changes

## What is the new behaviour?

image

image


projects/demo/src/modules/app/home/home.module.ts
```diff
@@ -13,6 +14,9 @@ import {HomeComponent} from './home.component';
         TuiNotificationModule,
         TuiLinkModule,
         TuiAccordionModule,
+        TuiTabsModule,
+        CommonModule,
+        TuiSvgModule,
```
waterplea, line 19: Let's not use SVG going forward. If you have a colored SVG - just use IMG, if you have a single color icon - let's use `tui-icon` instead.

waterplea (commented):
1. For some reason there's 1px instead of 2px 
image
2. Intstead of Angular < 15 and >= 16 let's write modules vs standalone
3. Now the headings of steps look weak, because everything else looks cool :) Maybe spice them up a little?

projects/demo/src/modules/app/home/home.module.ts
splincode: fixed

waterplea (approved):

vladimirpotekhin (approved):

PR #6058, opened 2023-11-22
feat(schematics): add experimental package for proprietary-core

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

### DIFF
--- a/projects/cdk/schematics/ng-update/v3-55/index.ts
+++ b/projects/cdk/schematics/ng-update/v3-55/index.ts
@@ -1,0 +1,37 @@
+import {Rule, SchematicContext, Tree} from '@angular-devkit/schematics';
+import {NodePackageInstallTask} from '@angular-devkit/schematics/tasks';
+import {
+    addPackageJsonDependency,
+    getPackageJsonDependency,
+    saveActiveProject,
+} from 'ng-morph';
+
+import {TAIGA_VERSION} from '../../ng-add/constants/versions';
+import {TuiSchema} from '../../ng-add/schema';
+import {FINISH_SYMBOL, titleLog} from '../../utils/colored-log';
+import {getFileSystem} from '../utils/get-file-system';
+
+export function updateToV3_55(options: TuiSchema): Rule {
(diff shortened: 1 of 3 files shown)


projects/cdk/schematics/ng-update/v3-55/index.ts
```diff
@@ -0,0 +1,37 @@
...
+export function updateToV3_55(options: TuiSchema): Rule {
+    return (tree: Tree, context: SchematicContext) => {
+        if (!getPackageJsonDependency(tree, `@taiga-ui/proprietary-core`)) {
+            !options[`skip-logs`] &&
+                titleLog(`${FINISH_SYMBOL} No migrations required\n`);
+
+            return;
+        }
+
+        const fileSystem = getFileSystem(tree);
```
vladimirpotekhin, line 23: there is no need to use a file system unless you mutate templates

just 
```
setActiveProject(<file pattern>)
... add dependency ...
saveActiveProject()
```
also, probably there is no need to use ALL FILES pattern, may be .json is enough

projects/cdk/schematics/ng-update/v3-55/index.ts
```diff
@@ -0,0 +1,37 @@
...
+} from 'ng-morph';
+
+import {TAIGA_VERSION} from '../../ng-add/constants/versions';
+import {TuiSchema} from '../../ng-add/schema';
+import {FINISH_SYMBOL, titleLog} from '../../utils/colored-log';
+import {getFileSystem} from '../utils/get-file-system';
+
+export function updateToV3_55(options: TuiSchema): Rule {
+    return (tree: Tree, context: SchematicContext) => {
+        if (!getPackageJsonDependency(tree, `@taiga-ui/proprietary-core`)) {
```
nsbarsukov, line 16: `@taiga-ui/experimental` is peer dependency of `@taiga-ui/proprietary-core`.
It should be install automatically by modern npm (7+) (if `--legacy-peer-deps` flag is not enabled).

Do we actually need to install this package manually ?

PR #6091, opened 2023-11-24
feat(addon-doc): use `Source Code Pro` font for code

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

## What is the new behaviour?

image


waterplea:
I don't think it's beneficial — the difference is not that big and this way we introduce another Google Fonts dependency for addon doc.

PR #6060, opened 2023-11-22
fix(kit): prevent clear country code prefix

splincode:
## PR Type

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Build or CI related changes
- [ ] Tests related changes
- [ ] Documentation content changes

## What is the current behaviour?

Closes #5787

PR #6214, opened 2023-12-11
fix(kit): `InputPhone` should erase value on blur if it's just country code

### LINKED ISSUE #6199: 🐞 - `InputPhone` should erase value on blur if it's just country code
### Playground Link

_No response_

### Description

We should clear the input on blur when there is no value other than forced country code.

We also need to have empty string in form control while value is only the country code to allow optional form control to pass validation as empty.

### Angular version

_No response_

### Taiga UI version

_No response_

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

nsbarsukov:
## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #6199
Relates #6060

### DIFF
--- a/projects/kit/components/input-phone/input-phone.component.ts
+++ b/projects/kit/components/input-phone/input-phone.component.ts
@@ -181,6 +181,8 @@ export class TuiInputPhoneComponent
         if (this.nativeValue === this.nonRemovablePrefix || this.isTextValue) {
             this.updateSearch('');
             this.nativeValue = '';
+
+            return;
         }
 
         if (!active && !this.allowText && this.nativeFocusableElement) {
@@ -194,11 +196,12 @@ export class TuiInputPhoneComponent
             : value.replace(TUI_MASK_SYMBOLS_REGEXP, '').slice(0, this.maxPhoneLength);
 
         this.updateSearch(parsed);
-
-        const prepared = parsed === this.countryCode || isText(parsed) ? '' : parsed;
-
-        this.value = prepared === '' && !this.allowText ? this.countryCode : prepared;
+        this.value = parsed === this.countryCode || isText(parsed) ? '' : parsed;
         this.open = true;
+
+        if (!this.value && !this.allowText) {
+            this.nativeValue = this.nonRemovablePrefix;
+        }
     }
 
     handleOption(item: string): void {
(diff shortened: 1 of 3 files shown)


projects/kit/components/input-phone/input-phone.component.ts
```diff
@@ -181,6 +181,8 @@ export class TuiInputPhoneComponent
         if (this.nativeValue === this.nonRemovablePrefix || this.isTextValue) {
```
waterplea, line 181: Should it be `this.nativeValue.trim() === this.nonRemovablePrefix.trim()`? Is it possible one of them contains space at the end and the other does not?

waterplea (approved):

projects/kit/components/input-phone/input-phone.component.ts
```diff
@@ -181,6 +181,8 @@ export class TuiInputPhoneComponent
...
             this.updateSearch('');
             this.nativeValue = '';
+
+            return;
         }
 
         if (!active && !this.allowText && this.nativeFocusableElement) {
```
waterplea, line 188: What does this branch do?

nsbarsukov:
```
"+1 (123) 456-" => blur => "+1 (123) 456"
```

projects/kit/components/input-phone/input-phone.component.ts
nsbarsukov: >  Is it possible one of them contains space at the end and the other does not?

* `nonRemovablePrefix` always includes space at the end:
https://github.com/taiga-family/taiga-ui/blob/a67423fe82616925ebaeb912b627b9e66a6e3984/projects/kit/components/input-phone/input-phone.component.ts#L227-L229

* User can't delete space in `nativeValue`
https://github.com/taiga-family/taiga-ui/blob/a67423fe82616925ebaeb912b627b9e66a6e3984/projects/kit/components/input-phone/input-phone.component.ts#L269

vladimirpotekhin (approved):

PR #6319, opened 2023-12-23
feat(core): use match media inside `TuiBreakpointService`

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes


waterplea (changes requested):
We specifically switched from matchMedia to resize, because matchMedia is buggy as hell.

PR #6322, opened 2023-12-24
ci: refactor e2e

splincode:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring
- [ ] Tests related changes

### DIFF
--- a/projects/demo-playwright/utils/page-objects/input-card-grouped.po.ts
+++ b/projects/demo-playwright/utils/page-objects/input-card-grouped.po.ts
@@ -1,21 +1,10 @@
 import {Locator} from '@playwright/test';
 
 export class TuiInputCardGroupedPO {
-    readonly numberTextfield = this.host.locator(
-        '[automation-id="tui-input-card-grouped__card"]',
-    );
-
-    readonly cleanerIcon = this.host.locator(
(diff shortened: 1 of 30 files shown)


projects/demo-playwright/utils/page-objects/input-card-grouped.po.ts
```diff
@@ -1,21 +1,10 @@
...
-        '[automation-id="tui-input-card-grouped__expire"]',
-    );
-
-    readonly cvcTextfield = this.host.locator(
-        '[automation-id="tui-input-card-grouped__cvc"]',
-    );
+    readonly numberTextfield = this.host.getByTestId('tui-input-card-grouped__card');
+    readonly cleanerIcon = this.host.getByTestId('tui-input-card-grouped__cleaner');
+    readonly expiryTextfield = this.host.getByTestId('tui-input-card-grouped__expire');
+    readonly cvcTextfield = this.host.getByTestId('tui-input-card-grouped__cvc');
```
nsbarsukov, line 7: I think it should be reverted for all page objects.

These POs can be potentially published later.
They will not work without correct global configuration.

nsbarsukov (changes requested):

PR #6483, opened 2024-01-18
feat(core): applies styles only inside taiga context

### LINKED ISSUE #6473: 🐞 - Web Components - `TuiRootComponent` applies a style that affects the host application
### Playground Link

_No response_

### Description

To use Taiga UI in a web component (Angular Elements) that will be used on Angular and non-Angular applications, we need to put the `TuiRootComponent` at the top level of the web component as it's necessary for the other Taiga UI components.

The issue is that `TuiRootComponent` applies CSS styles to the body element ([root.style.less#L21](https://github.com/taiga-family/taiga-ui/blob/v3.62.0/projects/core/components/root/root.style.less#L21)) which breaks the style of the host application.

My suggestion is to add the `data-tui-theme` attribute to the `body {}` selector: `body[data-tui-theme] {}`. 

### Angular version

17.0.9

### Taiga UI version

3.62.0

- [X] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #6473

### DIFF
--- a/projects/core/components/root/root.component.ts
+++ b/projects/core/components/root/root.component.ts
@@ -60,9 +60,9 @@ export class TuiRootComponent {
         @Inject(TuiBreakpointService) private readonly breakpoint: TuiBreakpointService,
         @Inject(TUI_IS_IOS) readonly isIOS: boolean,
         @Inject(TUI_IS_ANDROID) readonly isAndroid: boolean,
-        @Inject(DOCUMENT) {body}: Document,
+        @Inject(DOCUMENT) document: Document,
         @Inject(TUI_THEME) theme: string,
     ) {
-        body.setAttribute('data-tui-theme', theme.toLowerCase());
+        document.documentElement.setAttribute('data-tui-theme', theme.toLowerCase());
     }
 }
--- a/projects/core/components/root/root.style.less
+++ b/projects/core/components/root/root.style.less
@@ -13,12 +13,12 @@
     }
 }
 
-html,
+html[data-tui-theme],
 .tui-zero-scrollbar {
     .scrollbar-hidden();
 }
 
-body {
+html[data-tui-theme] body {
     .text-basic();
     background: var(--tui-base-01);
     margin: 0;


projects/core/components/root/root.component.ts
```diff
@@ -63,6 +63,6 @@ export class TuiRootComponent {
         @Inject(DOCUMENT) {body}: Document,
         @Inject(TUI_THEME) theme: string,
     ) {
-        body.setAttribute('data-tui-theme', theme.toLowerCase());
+        body.parentElement?.setAttribute('data-tui-theme', theme.toLowerCase());
```
waterplea, line 66: Instead of `body` use `documentElement`.

projects/core/components/root/root.style.less
```diff
@@ -13,15 +13,18 @@
     }
 }
 
-html,
-.tui-zero-scrollbar {
+html[data-tui-theme] {
```
waterplea, line 16: I wonder if we should also hide global `tui-scroll-controls` if there is no `[data-tui-theme]` attribute.

projects/demo/src/styles.less
```diff
@@ -1,12 +1,14 @@
 @import '@taiga-ui/core/styles/taiga-ui-local';
 
-html,
-body {
-    .scroll-behavior();
-    font: var(--tui-font-text-m);
+html[data-tui-theme] {
```
waterplea, line 3: These do not matter, it is just our demo app.

projects/styles/basic/main.less
```diff
@@ -1,22 +1,37 @@
-html {
+html[data-tui-theme] {
```
waterplea, line 1: I don't think we should add this change here, it is a conscious decision to use our global styles, makes no sense to make them disableable like that.

splincode: fixed

projects/demo/src/styles.less
splincode: fixed

projects/core/components/root/root.style.less
splincode: fixed

projects/core/components/root/root.component.ts
splincode: fixed

projects/core/components/root/root.style.less
waterplea: Where? I don't see related changes.

projects/addon-doc/components/main/main.style.less
waterplea, line 1: There's quite a lot of changes in this file, why so?

splincode: image

do we need to use global styles outside of the taiga context?

waterplea: All of these styles are for addon doc, everything is in Taiga context.

projects/core/components/root/root.style.less
```diff
@@ -13,15 +13,17 @@
     }
 }
 
-html,
-.tui-zero-scrollbar {
-    .scrollbar-hidden();
-}
+html[data-tui-theme] {
+    &,
+    .tui-zero-scrollbar {
```
waterplea, line 18: `.tui-zero-scrollbar` should probably work regardless of the attribute.

waterplea (approved):

projects/core/components/root/root.style.less
splincode: fixed

vladimirpotekhin (approved):

PR #6491, opened 2024-01-19
feat(kit): `Sensitive` increase number of rows for larger sizes

vladimirpotekhin:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes #


splincode (approved):

projects/experimental/directives/sensitive/sensitive.directive.ts
```diff
@@ -15,9 +25,18 @@ export class TuiSensitiveDirective {
     tuiSensitive: boolean | null = false;
 
     readonly offset = Math.round(Math.random() * 10) * 10;
+    readonly height$ = this.resize$.pipe(
+        map(([entry]) => entry.contentRect.height),
```
waterplea, line 29: map(([{contentRect}]) => Math.max(2, Math.floor(contentRect.height / 16) + 1)),
map(rows => height * rowsInSvg / rows),

projects/experimental/directives/sensitive/sensitive.directive.ts
```diff
@@ -15,9 +25,18 @@ export class TuiSensitiveDirective {
...
+        map(height => {
+            const rows = Math.max(2, Math.floor(height / 16) + 1);
+
+            return height * (rowsInSvg / rows);
+        }),
+    );
 
     constructor(
         @Inject(TuiDirectiveStylesService) directiveStyles: TuiDirectiveStylesService,
+        @Inject(TuiResizeService) private readonly resize$: TuiResizeService,
```
waterplea, line 39: In 4.0 we can use resize observer from ng-web-apis directly, we will drop this service and there will be no need for destroy service. If it is intended to go into 3.0 too then let's keep it like this for now.

waterplea (approved):

projects/experimental/directives/sensitive/sensitive.directive.ts
vladimirpotekhin: Yes, ok I'll refactor this after cherry-picking in 3.0

PR #6538, opened 2024-01-24
feat(core): `Loader` has new animation

nsbarsukov:
spec

### DIFF
--- a/projects/core/components/loader/loader.style.less
+++ b/projects/core/components/loader/loader.style.less
@@ -13,7 +13,7 @@
 @width-xl: 14;
 @width-xxl: 10;
 
-@radius: 50;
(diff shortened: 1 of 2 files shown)


projects/core/components/loader/loader.style.less
nsbarsukov, line 1: Guys, should I cherry-pick these changes to `v3.x` ?

splincode (approved):

projects/core/components/loader/loader.template.html
nsbarsukov, line 1: DONT MERGE

I've found some mobile safari problems, it needs some investigations

waterplea (approved):

PR #6550, opened 2024-01-25
feat(core)!: `Loader` has new algorithm to calculate `stroke-width`

nsbarsukov:
Screenshot 2024-01-25 at 16 06 58

### DIFF
--- a/projects/core/components/loader/loader.style.less
+++ b/projects/core/components/loader/loader.style.less
@@ -6,12 +6,6 @@
 @size-l: 2.5rem;
 @size-xl: 3.5rem;
 @size-xxl: 5rem;
-@width-xs: 38;
-@width-s: 25;
-@width-m: 17;
-@width-l: 15;
-@width-xl: 14;
-@width-xxl: 10;
 
 :host {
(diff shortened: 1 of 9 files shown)


projects/core/components/loader/loader.style.less
nsbarsukov, line 1: Dont merge, i've made a mistake

projects/core/components/loader/loader.style.less
```diff
@@ -54,6 +48,33 @@
     &._loading {
         overflow: hidden;
     }
+
+    // Use native CSS max function instead LESS one
+    stroke-width: ~'max(calc(var(--t-size) / 12), 1.5px)';
```
nsbarsukov, line 53: It is ok to use in v4.0
https://caniuse.com/css-math-functions

Screenshot 2024-01-25 at 16 49 10

projects/core/components/loader/loader.style.less
nsbarsukov: fixed ✅

waterplea (approved):

PR #6554, opened 2024-01-25
fix(cdk): when the consumer unsubscribes, clean up data ready for next subscription

### LINKED ISSUE #6553: 🐞 - tuiControlValue memory leak
### Playground Link

_No response_

### Description

You subscribe to a control in `tuiControlValue` function, but do not unsubscribe when the source observable completes or is unsubscribed.

Here is the source code:

```
export function tuiControlValue<T>(
    control: AbstractControl | AbstractControlDirective,
): Observable<T> {
    return new Observable(subscriber => {
        if (!control.valueChanges) {
            throw new TuiValueChangesException();
        }

        control.valueChanges.pipe(startWith(control.value)).subscribe(subscriber);
    });
}
```

### Angular version

_No response_

### Taiga UI version

_No response_

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
Closes #6553

### DIFF
--- a/projects/cdk/observables/control-value.ts
+++ b/projects/cdk/observables/control-value.ts
@@ -13,6 +13,6 @@ export function tuiControlValue<T>(
             throw new TuiValueChangesException();
         }
 
-        control.valueChanges.pipe(startWith(control.value)).subscribe(subscriber);
+        return control.valueChanges.pipe(startWith(control.value)).subscribe(subscriber);
     });
 }


vladimirpotekhin (approved):

projects/cdk/observables/control-value.ts
```diff
@@ -13,6 +13,10 @@ export function tuiControlValue<T>(
             throw new TuiValueChangesException();
         }
 
-        control.valueChanges.pipe(startWith(control.value)).subscribe(subscriber);
+        const subscription = control.valueChanges
+            .pipe(startWith(control.value))
+            .subscribe(subscriber);
+
+        return () => subscription.unsubscribe();
```
waterplea, line 20: You can just add `return`, the subscription is a valid teardown logic.

waterplea (approved):

projects/cdk/observables/control-value.ts
splincode: fixed

PR #6564, opened 2024-01-26
refactor: change `TuiRangeState` from enum to type

splincode:


vladimirpotekhin (approved):
don't forget plz to add instruction for migration there https://github.com/taiga-family/taiga-ui/issues/5808 (or add migration)

waterplea (approved):

waterplea:
> don't forget plz to add instruction for migration there #5808 (or add migration)

I don't think we need it, it's only used internally

PR #6573, opened 2024-01-26
chore(demo): set disable-animation page as standalone

splincode:


projects/demo/src/modules/info/testing/disable-animation/disable-animation.component.ts
```diff
@@ -1,12 +1,32 @@
...
+    standalone: true,
     selector: 'disable-animation',
+    imports: [
+        NgForOf,
+        TuiDocDocumentationModule,
+        TuiDocPageModule,
+        TuiLinkModule,
+        TuiTabsModule,
+        TuiDocTabModule,
+        TuiDocCodeModule,
```
nsbarsukov, line 22: `TuiAddonDocModule` ?

nsbarsukov (approved):

projects/demo/src/modules/info/testing/disable-animation/disable-animation.component.ts
splincode: fixed

PR #6607, opened 2024-01-30
feat(core): disable zoom on Safari when focus textfield

### LINKED ISSUE #6582: 🐞 - DialogService
### Playground Link

_No response_

### Description

1. В мобильном браузере на IOS в диалоге криво работает выделение текста. Можно выделить только один раз, потом ползунки выделения не перемещаются

2. В ComponentDialog на IOS также нелзя двигать ползунки, если в Dialog присутствуют поля зполнения (input, select, etc)

### Angular version

_No response_

### Taiga UI version

_No response_

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X] iOS
- [ ] Android

splincode:
Closes #6582

### DIFF
--- a/projects/core/styles/mixins/textfield.less
+++ b/projects/core/styles/mixins/textfield.less
@@ -1,3 +1,5 @@
+@import '../taiga-ui-local.less';
+
 .textfield-host() {
     .text-basic();
     position: relative;
@@ -288,6 +290,11 @@
 
     .t-input {
         .textfield-input();
+        .ios-only({
+            &:focus { // disable zoom
+              font-size: 16px;
+            }
+        });
     }
 
     .t-content {


projects/core/styles/mixins/textfield.less
```diff
@@ -288,6 +290,11 @@
 
     .t-input {
         .textfield-input();
+        .ios-only({
+            &:focus { // disable zoom
+              font-size: 16px;
```
waterplea, line 295: We cannot do that, it would break the design. Large inputs are already 16px, small ones would look bad this way.

waterplea:
How is that relevant to the mentioned issue?

MRustem:
the problem is still not solved

PR #6624, opened 2024-01-31
feat(cdk): scroll controls should be visible only inside tui-root area

splincode:


projects/cdk/components/dialog-host/dialog-host.component.ts
```diff
@@ -84,7 +84,7 @@ export class TuiDialogHostComponent<T extends TuiDialog<unknown, unknown>>
                 this.dialogs = dialogs;
                 this.cdr.markForCheck();
 
-                this.doc.documentElement.classList.toggle(
+                this.doc.defaultView?.document.documentElement.classList.toggle(
```
waterplea, line 87: What difference does it make?

waterplea (approved):

waterplea (approved):

projects/cdk/components/dialog-host/dialog-host.component.ts
splincode: for FM

PR #6640, opened 2024-02-01
chore: add `swipe.cy.ts`

splincode:


projects/demo-cypress/src/tests/swipe.cy.ts
```diff
@@ -0,0 +1,95 @@
...
+    const touchEvent = new TouchEvent(eventType, {
+        cancelable: true,
+        bubbles: true,
+        touches: [touchObj],
+        targetTouches: [],
+        changedTouches: [touchObj],
+        shiftKey: true,
+    });
+
+    element.dispatchEvent(touchEvent);
```
nsbarsukov, line 94: The same https://github.com/taiga-family/taiga-ui/pull/6642#discussion_r1476027194

nsbarsukov (approved):

PR #6642, opened 2024-02-01
chore: add `zoom.cy.ts`

splincode:


projects/demo-cypress/src/tests/zoom.cy.ts
```diff
@@ -0,0 +1,95 @@
...
+    const touchEvent = new TouchEvent(eventType, {
+        cancelable: true,
+        bubbles: true,
+        touches: [touchObj1, touchObj2],
+        targetTouches: [],
+        changedTouches: [touchObj1, touchObj2],
+        shiftKey: true,
+    });
+
+    element.dispatchEvent(touchEvent);
```
nsbarsukov, line 94: Could we use Cypress API?
https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/testing-dom__drag-drop

splincode: fixed

nsbarsukov (approved):

waterplea (approved):

PR #6652, opened 2024-02-01
feat(addon-table): add optional first/last page buttons to the TablePagination component

pubiqq:
New property for the TablePagination component - `showFirstLastPageButtons`:

<table>
  <tr>
    <th><code>showFirstLastPageButtons = false</code> (default)</th>
    <th><code>showFirstLastPageButtons = true</code></th>
   </tr> 
   <tr>
    <td></td>
    <td></td>
    </tr>
</table>


splincode (approved):

projects/addon-table/components/table-pagination/table-pagination.component.ts
```diff
@@ -108,15 +139,83 @@ export class TuiTablePaginationComponent {
...
+
+    protected selectFirstPage(): void {
+        this.selectPage(this.firstPage);
+    }
+
+    protected selectPreviousPage(): void {
+        const previousPage = this.previousPage;
+
+        if (previousPage === null) {
+            return;
```
waterplea, line 164: Early returns make no sense if it's just one line. This looks like it would be better to just support `null` in `selectPage` and do the early return there.

projects/addon-table/components/table-pagination/table-pagination.component.ts
```diff
@@ -38,6 +38,9 @@ export class TuiTablePaginationComponent {
     @Input()
     public size = this.options.size;
 
+    @Input()
+    public showFirstLastPageButtons = false;
```
waterplea, line 42: Sorry for taking so long to review this. We need shorter names and we do not need to deprecate and rename things, let's do this with minimum intervention to how it works right now. If we only want it in 4.0, we can just rename everything to better suited names.

projects/addon-table/components/table-pagination/table-pagination.component.ts
```diff
@@ -93,12 +102,34 @@ export class TuiTablePaginationComponent {
...
     protected get rightDisabled(): boolean {
-        return this.end === this.total;
+        return this.nextPageDisabled;
+    }
+
+    protected get firstPageDisabled(): boolean {
+        return this.previousPage === null;
+    }
+
+    protected get previousPageDisabled(): boolean {
```
waterplea, line 123: Let's drop `Page` from all these getters. This is a pagination component, it's implicit that we are talking about pages. And class prop names are not minified by compiler so it's better to keep them sort bundlesize-wise.

projects/addon-table/components/table-pagination/table-pagination.component.ts
```diff
@@ -93,12 +102,34 @@ export class TuiTablePaginationComponent {
...
+        return this.previousPage === null;
+    }
+
+    protected get nextPageDisabled(): boolean {
+        return this.nextPage === null;
+    }
+
+    protected get lastPageDisabled(): boolean {
+        return this.nextPage === null;
```
waterplea, line 132: Why do we have 2 getters that do the same?

projects/addon-table/components/table-pagination/table-pagination.component.ts
```diff
@@ -108,15 +139,83 @@ export class TuiTablePaginationComponent {
...
+    }
+
+    protected selectLastPage(): void {
+        this.selectPage(this.lastPage);
+    }
+
+    /**
+     * Returns the index of the first page.
+     */
+    private get firstPage(): number {
```
waterplea, line 187: This getter makes no sense. It's obvious that first index is 0, there's no need to move into a semantic getter.

projects/addon-table/components/table-pagination/table-pagination.component.ts
```diff
@@ -108,15 +139,83 @@ export class TuiTablePaginationComponent {
...
+        return 0;
+    }
+
+    /**
+     * Returns the index of the last page.
+     */
+    private get lastPage(): number {
```
waterplea, line 194: It is only used once, let's not bloat class with a getter for `lastPage`, just inline `this.pageCount - 1` where you use it, it's self explanatory.

PR #6660, opened 2024-02-02
feat(cdk)!: refactor the way new custom dialogs are created

waterplea:
Closes #


projects/cdk/services/popover.service.ts
```diff
@@ -20,32 +20,36 @@ export type TuiPopover<T, O> = T &
 @Injectable()
 // eslint-disable-next-line @typescript-eslint/naming-convention
```
splincode, line 21: you can remove this comment

splincode (approved):

PR #6685, opened 2024-02-05
chore: fix flaky

splincode:


waterplea (approved):

projects/demo-cypress/src/tests/elastic-sticky.cy.ts
```diff
@@ -51,14 +51,14 @@ describe('ElasticSticky', () => {
     it('callback is triggered with 0.5 when half of sticky would be hidden', () => {
         cy.get('#scroll')
             .then(query => query.get(0).scrollTo({top: 75}))
-            .then(() => cy.wait(100))
+            .then(() => cy.wait(300))
```
nsbarsukov, line 54: Should we use Cypress API ?
https://docs.cypress.io/api/commands/scrollTo

```ts
cy.get('#scroll').scrollTo().then(...);
```

nsbarsukov (approved):

projects/demo-cypress/src/tests/elastic-sticky.cy.ts
splincode: didn't work

PR #6704, opened 2024-02-07
fix(kit): add `line-height` for `[tuiTab]`

splincode:


projects/kit/components/tabs/tabs-with-more/tabs-with-more.style.less
```diff
@@ -63,6 +63,7 @@
         justify-content: flex-start;
         margin: 0.125rem 0.375rem;
         padding: 0 0.625rem;
+        line-height: 1.5rem;
```
waterplea, line 66: Mobile tabs and bootstrap are irrelevant, only this one matters. And also you need to add it to the actual `tab.styles.less`.

splincode: fixed

waterplea (approved):

vladimirpotekhin (approved):

PR #6721, opened 2024-02-08
feat(core): add iOS support to TuiDropdownContextDirective

MillerSvt:


projects/core/directives/dropdown/dropdown-context.directive.ts
```diff
@@ -44,6 +57,44 @@ export class TuiDropdownContextDirective extends TuiRectAccessor {
...
+    ])
+    onTouchStart(x: number, y: number): void {
+        if (!this.isIOS) {
+            return;
+        }
+
+        this.currentRect = tuiPointToClientRect(x, y);
+        this.longTapTimeout = setTimeout(() => {
+            this.driver.next(true);
+        }, tuiLongTapDelay);
```
splincode, line 84: `runOutsideAngular`?

projects/core/directives/dropdown/dropdown-context.directive.ts
```diff
@@ -22,14 +25,24 @@ function activeZoneFilter(this: TuiDropdownContextDirective, target: Element): b
...
+    host: {
+        '[style.user-select]': '"none"',
+        '[style.-webkit-touch-callout]': '"none"',
+        '[style.-webkit-user-select]': '"none"',
+    },
 })
 export class TuiDropdownContextDirective extends TuiRectAccessor {
     private readonly driver = inject(TuiDropdownDriver);
     private currentRect = EMPTY_CLIENT_RECT;
+    private longTapTimeout?: any;
```
splincode, line 37: why `any`?

MillerSvt: it conflicts with Node.js setTimeout, that returns `Timeout` object insted of `number`

projects/core/directives/dropdown/dropdown-context.directive.ts
MillerSvt: fixed

projects/core/directives/dropdown/dropdown-context.directive.ts
MillerSvt: fixed

splincode (approved):

projects/core/directives/dropdown/dropdown-context.directive.ts
```diff
@@ -44,6 +62,48 @@ export class TuiDropdownContextDirective extends TuiRectAccessor {
...
+        '$event.touches[0].clientX',
+        '$event.touches[0].clientY',
+    ])
+    onTouchStart(x: number, y: number): void {
+        if (!this.isIOS) {
+            return;
+        }
+
+        this.currentRect = tuiPointToClientRect(x, y);
+        this.zone.runOutsideAngular(() => {
```
waterplea, line 87: What's the point of running it outside if we immediately re-enter?

projects/core/directives/dropdown/dropdown-context.directive.ts
```diff
@@ -44,6 +62,48 @@ export class TuiDropdownContextDirective extends TuiRectAccessor {
...
+            this.longTapTimeout = this.win.setTimeout(() => {
```
waterplea, line 88: No need for `win` here, this event will not happen in the context where `setTimeout` does not exist ever.

projects/core/directives/dropdown/dropdown-context.directive.ts
```diff
@@ -44,6 +62,48 @@ export class TuiDropdownContextDirective extends TuiRectAccessor {
...
+
+    @HostListener('touchend.silent')
+    @HostListener('touchcancel.silent')
+    onTouchEnd(): void {
+        if (!this.isIOS) {
+            return;
+        }
+
+        clearTimeout(this.longTapTimeout);
+        this.driver.next(false);
```
waterplea, line 104: Wouldn't that close the dropdown when we release the finger?

projects/core/directives/dropdown/dropdown-context.directive.ts
```diff
@@ -44,6 +62,48 @@ export class TuiDropdownContextDirective extends TuiRectAccessor {
...
+    @HostListener('touchmove.silent', [
+        '$event.touches[0].clientX',
+        '$event.touches[0].clientY',
+    ])
+    onTouchMove(x: number, y: number): void {
+        if (!this.isIOS) {
+            return;
+        }
+
+        this.currentRect = tuiPointToClientRect(x, y);
```
waterplea, line 74: Why do we need to move it on touchmove? Shouldn't touchmove reset timer as well, meaning we pan rather than longtap?

projects/core/directives/dropdown/dropdown-context.directive.ts
MillerSvt: https://github.com/taiga-family/taiga-ui/pull/6721#discussion_r1483218811

projects/core/directives/dropdown/dropdown-context.directive.ts
MillerSvt: It looks like a bug in TuiPortalService.remove.
Because DropdownContextDirective emits false, but dropwown still present until next interaction.

image

It fixes the issue.

projects/core/directives/dropdown/dropdown-context.directive.ts
MillerSvt: I think it's very difficult not to move your finger at all during a long tap. In android, the browser allows you to move your finger slightly to the side, and the menu will still open. I just simplified this logic a bit.

projects/core/directives/dropdown/dropdown-context.directive.ts
MillerSvt: fixed

projects/core/directives/dropdown/dropdown-context.directive.ts
MillerSvt: fixed

projects/core/directives/dropdown/dropdown-context.directive.ts
splincode: why not just `runOutsideAngular`?

```ts
  this.zone.runOutsideAngular(() => {
    this.longTapTimeout = 
         setTimeout(() => this.driver.next(true)) as unknown as number;
  });
```

projects/core/directives/dropdown/dropdown-context.directive.ts
splincode: or you can

```ts
setTimeout(() => ...) as unknown as number;
```

projects/cdk/abstract/portals.ts
```diff
@@ -62,9 +62,10 @@ export abstract class TuiPortalService {
         return this.safeHost.addComponentChild(component);
     }
 
-    remove<C>({hostView}: ComponentRef<C>): void {
+    remove<C>({hostView, changeDetectorRef}: ComponentRef<C>): void {
         if (!hostView.destroyed) {
             hostView.destroy();
+            changeDetectorRef.detectChanges();
         }
```
splincode, line 69: could you it add in separate PR?

projects/core/directives/dropdown/dropdown-context.directive.ts
MillerSvt: Do you think is more beautiful than `any`?

projects/cdk/abstract/portals.ts
MillerSvt: ok

projects/core/directives/dropdown/dropdown-context.directive.ts
MillerSvt: fixed

projects/core/directives/dropdown/dropdown-context.directive.ts
MillerSvt: fixed

projects/core/directives/dropdown/dropdown-context.directive.ts
splincode: You agree, I'm sorry, you can use `any` if you want

projects/core/directives/dropdown/dropdown-context.directive.ts
splincode: `as unknown as number;` instead of `as unknown as any;`

projects/core/directives/dropdown/dropdown-context.directive.ts
MillerSvt: fixed

projects/core/directives/dropdown/dropdown-context.directive.ts
waterplea: It is already outside of angular because of `.silent`

waterplea: Should `touchstart` also be `.passive`?

projects/core/directives/dropdown/dropdown-context.directive.ts
```diff
@@ -22,14 +26,28 @@ function activeZoneFilter(this: TuiDropdownContextDirective, target: Element): b
...
     private readonly driver = inject(TuiDropdownDriver);
     private currentRect = EMPTY_CLIENT_RECT;
+    private touchStartRect = EMPTY_CLIENT_RECT;
+    private longTapTimeout: any = NaN;
 
     readonly activeZone = inject(TuiActiveZoneDirective);
     readonly type = 'dropdown';
 
+    constructor(
+        @Inject(TUI_IS_IOS) private readonly isIOS: boolean,
```
waterplea, line 45: Use `inject` so we will not need constructor and `super()`.

projects/core/directives/dropdown/dropdown-context.directive.ts
```diff
@@ -13,6 +14,9 @@ function activeZoneFilter(this: TuiDropdownContextDirective, target: Element): b
     return !this.activeZone.contains(target);
 }
 
+const tuiLongTapDelay = 700;
```
waterplea, line 17: nitpicking: Prefix is not necessary since it is not exported. We also write such constants in caps.

projects/core/directives/dropdown/dropdown-context.directive.ts
```diff
@@ -44,6 +62,53 @@ export class TuiDropdownContextDirective extends TuiRectAccessor {
...
+        }
+
+        if (
+            Math.hypot(x - this.touchStartRect.x, y - this.touchStartRect.y) >
+            tuiLongTapMoveThreshold
+        ) {
+            this.onTouchEnd();
+        }
+
+        this.currentRect = tuiPointToClientRect(x, y);
```
waterplea, line 81: How about we will only keep `currentRect` set at the initial touchstart and not update it on move, then we will not need to keep touchStartRect, so we will only use touchmove listener to cancel if it's moved too far.

PR #6728, opened 2024-02-09
refactor: prefer regexp exec

splincode:
String#match is defined to work the same as RegExp#exec when the regular expression does not include the g flag. Keeping to consistently using one of the two can help improve code readability.

https://typescript-eslint.io/rules/prefer-regexp-exec/


waterplea (approved):

vladimirpotekhin:
I see no reason for this
We already used only one of two - match, why change it?

PR #6730, opened 2024-02-09
fix(core): prevent body scroll by keydown outside dialog

splincode:


waterplea:
I think this is kind of a bigger issue that we need to fix more generally. Like when you added overflow: hidden to body, but it got broken in iframe. We need something like that. Ideally `overscroll-behavior` should fix that, but it's a longstanding bug and even native `dialog` tag does not address that. Let's not introduce this, as it fixes only part of the issue and only for our concrete dialog implementation.

PR #6731, opened 2024-02-09
fix(kit): support `tuiUnfinishedValidator` for `input-date-time`

### LINKED ISSUE #6513: 🐞 - tuiUnfinishedValidator in tuiInputDateTime
### Playground Link

_No response_

### Description

I want to replicate the behavior of tuiUnfinishedValidator from tuiInputDate to tuiInputDateTime, but this is impossible

### Angular version

17.0.0

### Taiga UI version

3.56.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
Closes #6513

### DIFF
--- a/projects/core/components/primitive-textfield/primitive-textfield.component.ts
+++ b/projects/core/components/primitive-textfield/primitive-textfield.component.ts
@@ -60,7 +60,7 @@ export class TuiPrimitiveTextfieldComponent
     extends AbstractTuiInteractive
     implements TuiPrimitiveTextfield
 {
-    @ViewChild('focusableElement')
+    @ViewChild('focusableElement', {static: true})
     private readonly focusableElement?: ElementRef<HTMLInputElement>;
 
     @Input()
--- a/projects/demo/src/modules/components/input-date-time/examples/1/index.html
+++ b/projects/demo/src/modules/components/input-date-time/examples/1/index.html
@@ -2,13 +2,22 @@
     class="b-form"
     [formGroup]="testForm"
 >
-    <tui-input-date-time formControlName="testValue">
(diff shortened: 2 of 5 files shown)


splincode:
@waterplea I don't think that a good idea change `[TuiDay, null]` to `null`

waterplea:
> @waterplea I don't think that a good idea for transform `[TuiDay, null]` to `null`

I meant `[null, null]` the case when you write date properly then delete part of it:

image

It should reset value to `null` in that case and unfinished validator will work as is. However there's also a case where the date is there but the time is not fully entered:

image

One way to handle it would be to reset the whole value to `null` if at least one part of it is partly entered, which is probably a good idea. After all, treating `08.02.2022, 2` as `[TuiDay, null]` is guesswork, we are not sure if user meant to enter the day and does not care about time, more like user forgot to finish his input.

splincode:
@waterplea `One way to handle it would be to reset the whole value to null if at least one part of it is partly entered`

image

in this case, do we still have to force a reset to null?

waterplea:
I think it's ok if this `08.02.2022` meant `[TuiDay, null]` and `08.02.2022, 2` meant `null`. Then if user deletes `2` it becomes `[TuiDay, null]` again.

splincode:
`
I think it's ok if this 08.02.2022 meant [TuiDay, null] and 08.02.2022, 2 meant null. Then if user deletes 2 it becomes [TuiDay, null] again.
`

It's so difficult to my mind

@vladimirpotekhin what do you think?

PR #6732, opened 2024-02-09
fix(kit): set focus after change input files

### LINKED ISSUE #6437: 🐞 - Focus Shift Issue in tui-radio-block After Uploading Document with tui-input-files in Form
### Playground Link

https://stackblitz.com/edit/angular-bkyhfx-p92ydw

### Description

Precondition: A form containing both tui-radio-block and tui-input-files components.
After uploading a document, the focus is shifted to the tui-radio-block
Can be reproduced - https://stackblitz.com/edit/angular-bkyhfx-p92ydw

### Angular version

12.2.17

### Taiga UI version

3.61.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
Closes #6437

### DIFF
--- a/projects/kit/components/input-files/input-files.template.html
+++ b/projects/kit/components/input-files/input-files.template.html
@@ -11,7 +11,6 @@
     <label
         automation-id="tui-input-file__label"
         class="t-label"
-        (mousedown.prevent.silent)="(0)"
         (tuiDroppableDragOverChange)="onDragOver($event)"
         (tuiDroppableDropped)="onDropped($event)"
     >


projects/kit/components/input-files/input-files.component.ts
```diff
@@ -178,6 +178,7 @@ export class TuiInputFilesComponent
             return;
         }
 
+        this.input?.nativeElement.focus();
```
waterplea, line 181: The issue is `(mousedown.prevent)` in the template. I don't know what it was there for, seems to have been there from the beginning. We probably can drop it.

splincode: fixed

PR #6733, opened 2024-02-09
refactor: inject `EnvironmentInjector` only when necessary

splincode:


waterplea:
That's optimizing nanoseconds for longer code. Let's not do that.

PR #6735, opened 2024-02-09
chore: remove `wait` because flaky

splincode:


vladimirpotekhin (approved):

projects/demo-cypress/src/support/get-spy-value.ts
```diff
@@ -0,0 +1,3 @@
+export function getSpyValue<T>(args: unknown): T {
+    return (args as any)?.getCalls?.()?.[0]?.args?.[0] as T;
```
nsbarsukov, line 2: Use auto-retrying assertion instead.
See this [example](https://github.com/taiga-family/maskito/blob/b54385870ce52967e2ade1148954da23ef45dfab/projects/demo-integrations/src/tests/component-testing/placeholder/placeholder-dispatch-input-events.cy.ts#L52):

```ts
cy.get('@inputEvent')
    .should('have.callCount', 3)
    .should('have.been.calledWith', ['arg1, 'arg2']);

```

Learn more: https://docs.cypress.io/guides/references/assertions

splincode: fixed

nsbarsukov (approved):

PR #6769, opened 2024-02-13
fix(addon-charts): cancel macrotask in `tui-arc-chart` when component destroyed

splincode:


projects/addon-charts/components/arc-chart/arc-chart.component.ts
```diff
@@ -94,13 +105,10 @@ export class TuiArcChartComponent {
...
-        @Inject(ChangeDetectorRef) cdr: ChangeDetectorRef,
+        @Inject(ChangeDetectorRef) private readonly cd: ChangeDetectorRef,
+        @Self() @Inject(TuiDestroyService) private readonly destroy$: Observable<void>,
     ) {
-        // So initial animation works
-        setTimeout(() => {
-            this.initialized = true;
-            cdr.markForCheck();
-        });
+        this.initializeAnimation();
```
waterplea, line 111: If we do not move it to a method we can keep `destroy$` and `cdr` from storing as class props. Also let's use `tuiWatch(cdr)`

waterplea (approved):

vladimirpotekhin (approved):

projects/addon-charts/components/arc-chart/arc-chart.component.ts
splincode: fixed

PR #6778, opened 2024-02-13
chore(demo): cancel macrotask when element page destroyed

splincode:


projects/demo/src/modules/components/accordion/accordion.component.ts
```diff
@@ -61,19 +64,23 @@ export class ExampleTuiAccordionComponent {
 
     size: TuiSizeS = this.sizeVariants[1];
 
+    constructor(
+        @Self() @Inject(TuiDestroyService) private readonly destroy$: Observable<void>,
```
waterplea, line 68: Please use `inject` everywhere

waterplea (approved):

vladimirpotekhin (approved):

projects/demo/src/modules/components/accordion/accordion.component.ts
splincode: Sure, next PR

because I want to make cherry picks in 3.x

PR #6801, opened 2024-02-15
feat(experimental): add autoClose when another element is focused, hide actions on non-touch devices

MillerSvt:


splincode (approved):

projects/experimental/components/swipe-action/swipe-actions.style.less
```diff
@@ -37,6 +37,10 @@
     ::ng-deep & > * {
         .loop(6);
     }
+
+    @media (hover: hover) and (pointer: fine) {
```
vladimirpotekhin, line 41: seems like these lines - [one](https://github.com/taiga-family/taiga-ui/blob/v3.x/projects/demo/src/modules/experimental/swipe-action/examples/5/index.less#L8), [two](https://github.com/taiga-family/taiga-ui/blob/v3.x/projects/demo/src/modules/experimental/swipe-action/examples/5/index.less#L30) is redundant now. Let's remove them

MillerSvt: fixed

PR #6806, opened 2024-02-16
feat(kit): add auto lineHeight for TuiLineClamp

MillerSvt:


projects/kit/components/line-clamp/line-clamp.component.ts
```diff
@@ -103,6 +105,19 @@ export class TuiLineClampComponent implements DoCheck, AfterViewInit {
         return this.options.showHint && this.overflown ? this.content : '';
     }
 
+    private get computedLineHeight(): number {
+        if (this.lineHeight === 'auto') {
+            return (
+                parseInt(
+                    this.window.getComputedStyle(this.el.nativeElement).lineHeight,
```
waterplea, line 112: We did this initially and it turned out to be very slow when there are multiple instances of it. I think we better not add this option. One day we will be able to use `lh` CSS units :harold_hides_the_pain:

PR #6809, opened 2024-02-16
feat(experimental): add autoClose when another element is focused

MillerSvt:


projects/experimental/components/swipe-action/swipe-actions-auto-close.directive.ts
vladimirpotekhin, line 1: redundant file

waterplea (approved):

splincode (approved):

PR #6827, opened 2024-02-19
feat(experimental): `Navigation` add new module

waterplea:
Closes #

### DIFF
--- a/projects/experimental/components/navigation/index.ts
+++ b/projects/experimental/components/navigation/index.ts
@@ -2,4 +2,5 @@ export * from './aside.component';
 export * from './head.component';
 export * from './logo.component';
 export * from './main.component';
+export * from './nav.component';
 export * from './navigation.module';
--- a/projects/experimental/directives/header/header.styles.less
+++ b/projects/experimental/directives/header/header.styles.less
@@ -14,7 +14,6 @@
     [tuiAccessories] {
         display: flex;
         align-items: center;
-        margin-inline-start: 1rem;
         gap: 0.75rem;
 
         &:before {


splincode (approved):

projects/demo/src/modules/experimental/navigation/navigation.template.html
```diff
@@ -0,0 +1,67 @@
...
+            This code is
+            <strong>experimental</strong>
+            and is a subject to change. Expect final solution to be shipped in the next major version
+        </tui-notification>
+
+        <p>A set of basic layout components for global navigation</p>
+
+        <tui-doc-example
+            id="basic"
+            heading="Basic"
```
vladimirpotekhin, line 17: it seems like the first example is expansive, and the second is basic, isn't?

projects/experimental/components/navigation/main.style.less
```diff
@@ -0,0 +1,95 @@
...
+        margin: 0 -1.25rem -3rem;
+        border-radius: 1.25rem 1.25rem 0 0;
+        box-shadow: 0 -1.25rem var(--tui-theme-color, #000);
+    }
+
+    > [tuiHeader] {
+        margin: 1rem 0 0.5rem;
+
+        &[tuiCardLarge] {
+            flex-direction: column;
```
vladimirpotekhin, line 27: these styles don't seem to affect the card

Снимок экрана 2024-02-20 в 10 43 29

waterplea: They are for this card:
image

projects/demo/src/modules/experimental/navigation/navigation.template.html
waterplea: Sure, we can rename them to `Full` and `Expansive header`

vladimirpotekhin (approved):

hakimio:
@waterplea some minor issues with the demo:
- there is an issue with the demo where content overflows when you scroll
- in the first demo there is some visual artifact (happens every time)
- shouldn't the navigation panel change color when switching between dark and light mode?

image

waterplea:
1. That is just how demo components behave, this module is meant for the entire page so those will not be visible. We can probably just hack-hide padding in these particular examples if this bothers people :)
2. I think that is some Chrome rendering bug, I started seeing it in some other places a couple of weeks ago. There's no reason for it to be there, as far as I can tell, nor is it visible in Firefox. Not even in Chrome on MacOS for me.
3. That is how our design team drew it to be, and since there's nothing proprietary about this design and Taiga UI lacked global layout components and we plan to keep expanding in that area — we decided this design can be open sourced. You can control color of the navigation panel by `--tui-theme-color` variable since you would probably want it to fit within your corporate styles. By default it does not react to dark/light mode by design.

waterplea:
> @waterplea some minor issues with the demo:
> 
> * there is an issue with the demo where content overflows when you scroll
> * in the first demo there is some visual artifact (happens every time)
> * shouldn't the navigation panel change color when switching between dark and light mode?

1. That is just how demo components behave, this module is meant for the entire page so those will not be visible. We can probably just hack-hide padding in these particular examples if this bothers people :)
2. I think that is some Chrome rendering bug, I started seeing it in some other places a couple of weeks ago. There's no reason for it to be there, as far as I can tell, nor is it visible in Firefox. Not even in Chrome on MacOS for me.
3. That is how our design team drew it to be, and since there's nothing proprietary about this design and Taiga UI lacked global layout components and we plan to keep expanding in that area — we decided this design can be open sourced. You can control color of the navigation panel by `--tui-theme-color` variable since you would probably want it to fit within your corporate styles. By default it does not react to dark/light mode by design.

PR #6832, opened 2024-02-19
chore: `Money` add migration

vladimirpotekhin:
part of #5808

### DIFF
--- a/projects/cdk/schematics/utils/templates/inputs.ts
+++ b/projects/cdk/schematics/utils/templates/inputs.ts
@@ -3,3 +3,7 @@ import {Attribute} from 'parse5';
 export function findAttr(attrs: Attribute[], name: string): Attribute | undefined {
     return attrs.find(attr => attr.name === name || attr.name === `[${name}]`);
 }
+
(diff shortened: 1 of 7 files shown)


projects/cdk/schematics/utils/templates/inputs.ts
```diff
@@ -3,3 +3,7 @@ import {Attribute} from 'parse5';
 export function findAttr(attrs: Attribute[], name: string): Attribute | undefined {
     return attrs.find(attr => attr.name === name || attr.name === `[${name}]`);
 }
+
+export function isInput(attr: Attribute): boolean {
+    return attr.name.startsWith('[');
```
splincode, line 8: what if `[(ngModel)]` or `[(data)]`?

splincode (approved):

projects/cdk/schematics/utils/templates/inputs.ts
vladimirpotekhin: I can rename it to `isBinding` 👍

PR #6841, opened 2024-02-21
chore(demo): fix example

splincode:
# before

image

# after

image


vladimirpotekhin (approved):

waterplea (commented):
Let's change `background` to `background-color` in elevated surface instead. And in flat as well.

waterplea (approved):

PR #6856, opened 2024-02-22
feat(kit): allow dynamic changes to TUI_NUMBER_FORMAT

### LINKED ISSUE #2900: 🚀 - Allow dynamic changes to `TUI_NUMBER_FORMAT`

vladimirpotekhin:
Closes #2900, #2578

BREAKING CHANGES: remove `TUI_NUMBER_FORMAT_OBSERVABLE` in favour of `TUI_NUMBER_FORMAT`, `tuiFormatNumber` pipe requires async pipe now


projects/kit/components/input-range/test/input-range.component.spec.ts
```diff
@@ -149,7 +149,7 @@ describe('InputRange', () => {
         });
 
         it('does not rounds to nearest multiple of [quantum] until text field losses focus', () => {
-            inputPORight.sendText('8');
+            inputPORight.sendText('8 лет');
```
vladimirpotekhin, line 152: perhaps this is due to the removal of the initial calibration plugin from `input-number`, everything seems to be ok in the demo

splincode (approved):

projects/addon-commerce/pipes/amount/amount.pipe.ts
```diff
@@ -20,14 +20,16 @@ const DEFAULT_DECIMAL_LIMIT = 2;
...
-    private readonly format = inject(TUI_NUMBER_FORMAT_OBSERVABLE);
+    private readonly format = inject(TUI_NUMBER_FORMAT);
 
     transform(
         value: number,
         currency: TuiCurrencyVariants = this.options.currency,
         currencyAlign: TuiHorizontalDirection = this.options.currencyAlign,
     ): Observable<string> {
-        return this.format.pipe(
+        const format = isObservable(this.format) ? this.format : of(this.format);
```
waterplea, line 30: Let's not branch this and use observable all the time

projects/core/directives/number-format/number-format.directive.ts
```diff
@@ -1,26 +1,28 @@
...
     set tuiNumberFormat(format: Partial<TuiNumberFormatSettings>) {
-        this.next({...this.settings, decimalLimit: NaN, ...format});
+        this.next({...this.settings, ...format});
     }
 
     constructor(
-        @Inject(TUI_NUMBER_FORMAT) private readonly settings: TuiNumberFormatSettings,
+        @SkipSelf()
+        @Inject(TUI_NUMBER_FORMAT)
+        private readonly settings: TuiNumberFormatSettings,
```
waterplea, line 24: That will be tricky once `TUI_NUMBER_FORMAT` is always and `Observable` but we can do it. Basically, it will no longer be a `BehaviorSubject` and instead there will be `combineLatest` of upper number format and an internal `ReplaySubject(1)` that we will `next` to in setter.

waterplea (commented):
I'll review `InputNumber` changes once other comments are addressed as it seems like it will change quite a bit.

projects/core/directives/number-format/number-format.directive.ts
vladimirpotekhin: rewritten :+1: I hope I got you right

projects/kit/components/input-number/input-number.component.ts
```diff
@@ -82,6 +86,10 @@ export class TuiInputNumberComponent
 
     protected readonly options = inject(TUI_INPUT_NUMBER_OPTIONS);
 
+    protected get numberFormat(): TuiNumberFormatSettings {
```
waterplea, line 89: I think we should just add subscription, instead of trying to turn it into `BehaviorSubject` with `.value`. With `tuiWatch`.

projects/core/tokens/number-format.ts
```diff
@@ -1,15 +1,26 @@
...
     options: Partial<TuiNumberFormatSettings>,
 ): Provider {
-    return tuiProvideOptions(TUI_NUMBER_FORMAT, options, TUI_DEFAULT_NUMBER_FORMAT);
+    return {
+        provide: TUI_NUMBER_FORMAT,
+        deps: [[new Optional(), new SkipSelf(), TUI_NUMBER_FORMAT]],
+        useFactory: (
+            parent: Observable<TuiNumberFormatSettings> | null,
+        ): Observable<TuiNumberFormatSettings> =>
+            (parent || of(TUI_DEFAULT_NUMBER_FORMAT)).pipe(
```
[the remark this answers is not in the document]
vladimirpotekhin: without `Optional` unit tests using `tuiNumberFormatProvider` fail with a `NullInjectorError`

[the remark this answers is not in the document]
waterplea: Right, if we provide it at the top level there will be no parent, because token factory has the same level. Ok then.

projects/demo/src/modules/components/input-number/examples/7/index.ts
```diff
@@ -0,0 +1,31 @@
...
+
+@Component({
+    selector: 'tui-input-number-example-7',
+    templateUrl: './index.html',
+    encapsulation,
+    changeDetection,
+    providers: [
+        {
+            provide: TUI_NUMBER_FORMAT,
+            deps: [TUI_LANGUAGE],
```
waterplea, line 16: What do you guys think if we ban `deps` in factories in favor of `inject`? It is better typed. Maybe not ban, but at least prefer.

waterplea (approved):
I hope it's all well without the initial calibration plugin? :)

projects/demo/src/modules/components/input-number/examples/7/index.ts
vladimirpotekhin: Yes, I also prefer `inject` instead of `deps`. But real type of `TUI_LANGUAGE` in our demo is different than type of token due to [LanguageSwitcher](https://github.com/taiga-family/taiga-ui/blob/main/projects/i18n/switch/language-switcher.providers.ts#L14). So it is just to avoid ugly typecasting like `as unknown as Observable<Observable<TuiLanguage>>>`

PR #6869, opened 2024-02-23
refactor: add some utils for stabilize flaky

splincode:


projects/demo-playwright/tests/demo/landing.spec.ts
```diff
@@ -7,7 +7,11 @@ test.describe('Landing', () => {
     });
 
     test('take snapshot', async ({page}) => {
-        await tuiGoto(page, '/', {hideHeader: false});
+        await tuiGoto(page, '/', {
+            hideHeader: false,
+            checkStable: false,
```
nsbarsukov, line 12: Why shouldn't we wait for stable state here ?

nsbarsukov (approved):

vladimirpotekhin (approved):

projects/demo-playwright/tests/demo/landing.spec.ts
splincode: fixed

PR #6876, opened 2024-02-26
feat(addon-doc): support target property for links

splincode:


vladimirpotekhin (approved):

waterplea (approved):

waterplea:
I think that's excessive. Can we just add `rel="noreferrer noopener"` there? What's the potential downside?

PR #6906, opened 2024-02-28
chore: fix radius for `tuiCardMedium`, `tuiCardLarge`

splincode:


waterplea:
> @waterplea https://taiga-ui.dev/next/experimental/card-medium
> 
> image
> 

I think we should remove border radius from surface instead. We decided that padding and radius are responsibility of card, while surface is more like «appearance on steroids», responsible for colors and interaction animations.

waterplea (approved):

vladimirpotekhin (approved):

Issue #6914, opened 2024-02-29
🚀 - Add to tui-reorder possibility to transfer a template


said4505:
### Description

Currently, it is impossible to add translate pipe to values. so, i think it would be perfect give users possibility to add custom template, similarly as other components

40oleg:
May I take this one?

PR #6967, opened 2024-03-10
feat(addon-table): reorder add possibility to transfer a template

40oleg:
Closes https://github.com/taiga-family/taiga-ui/issues/6914


40oleg:
I see that the text moved a little. I will try to fix it a bit later

projects/addon-table/components/reorder/reorder.component.ts
```diff
@@ -29,6 +30,9 @@ export class TuiReorderComponent<T> {
     @Output()
     public readonly enabledChange = new EventEmitter<T[]>();
 
+    @Input()
+    public container?: PolymorpheusContent;
```
waterplea, line 34: We call these things `content`. And let's type context: `PolymorpheusContent<TuiContext<T> & { index: number }>`

projects/addon-table/components/reorder/reorder.template.html
```diff
@@ -15,7 +15,15 @@
                     class="t-icon"
                     [src]="options.icons.drag"
                 />
-                {{ item }}
+                <ng-container
+                    *polymorpheusOutlet="
+                        container ? container : primitiveValue as template;
```
waterplea, line 20: No need to fallback, you can just add default content as a function that stringifies $implicit: `({$implicit}) => String($implicit)`

projects/addon-table/components/reorder/reorder.component.ts
```diff
@@ -44,6 +46,11 @@ export class TuiReorderComponent<T> {
         }
     }
 
+    @Input()
+    public content?: PolymorpheusContent<TuiContext<T> & {index: number}> = ({
```
waterplea, line 50: We need to drop `?:`, otherwise LGTM

waterplea (approved):

splincode (approved):

40oleg:
I restarted pipelines because e2e-test for calendar failed for some reasons. I think this is because branch was outdated

Issue #6957, opened 2024-03-06
🚀 - Migrate demo application to standalone components


nsbarsukov:
# Why is it important ?
We have many examples when it is not obvious which modules are required to fully reproduce the example.

For instance, open this example:
https://taiga-ui.dev/next/components/input#sizes

If developer tries to reproduce it, it will not completely work.
Why ? Because, `TuiInputModule` is not the only required module.
It also requires `TuiTextfieldControllerModule` and `TuiHintModule`.
If we migrate the whole demo application to standalone, we can solve this problem (developer can see all required modules inside `TypeScript` tab).

# Algorithm
1. Take any folder from https://github.com/taiga-family/taiga-ui/tree/main/projects/demo/src/modules/components
2. Check if it does not already contains standalone components
3. Migrate all examples to standalone (double check that standalone example imports ONLY REQUIERED modules, no any redundant imports)
4. Migrate host component of this documentation page (add `export default`)
5. Update routing file 
https://github.com/taiga-family/taiga-ui/blob/12b2a00c53fa443b584c7e631da3f9789f2ae722/projects/demo/src/modules/app/app.routes.ts#L5
6. Commit and push you branch (**don't include many documentation pages in single PR; 1 PR per documentation page!**)

**Hints:**
* You can use https://angular.io/guide/standalone-migration script. It has `--path` flag, if you find it useful. My own experience showed that it does not save so much time.

PR #6969, opened 2024-03-11
chore(demo): migrate documentation page `Badge` to standalone components

KrollikRoddzer:
Partially solves https://github.com/taiga-family/taiga-ui/issues/6957
Done: Migrated `Badge` example components to standalone.


projects/demo/src/modules/components/badge/examples/4/index.ts
```diff
@@ -1,9 +1,13 @@
 import {Component} from '@angular/core';
 import {changeDetection} from '@demo/emulate/change-detection';
 import {encapsulation} from '@demo/emulate/encapsulation';
+import {TuiIconModule} from '@taiga-ui/experimental';
+import {TuiBadgeDirective, TuiFadeDirective} from '@taiga-ui/kit';
 
 @Component({
+    standalone: true,
     selector: 'tui-badge-example-4',
+    imports: [TuiBadgeDirective, TuiIconModule, TuiFadeDirective],
```
nsbarsukov, line 10: Do we actually need `TuiIconModule` here ?

KrollikRoddzer: Yes. It is responsible for this icon: 
Снимок экрана 2024-03-11 в 11 38 54

nsbarsukov (approved):

projects/demo/src/modules/components/badge/examples/4/index.ts
KrollikRoddzer: I was wrong, about this case. We don't need it in this example

PR #6975, opened 2024-03-11
refactor: self closing

splincode:


projects/demo/src/index.html
```diff
@@ -167,6 +167,8 @@
         <script src="https://cdn.polyfill.io/v3/polyfill.min.js?features=Reflect"></script>
     </head>
     <body>
+        <!-- TODO: self closing doesn't work with SSR, I don't know why -->
+        <!-- eslint-disable-next-line -->
         <app></app>
```
nsbarsukov, line 172:
```suggestion
        <!-- TODO: self closing doesn't work with SSR -->
        <!-- eslint-disable-next-line -->
        <app></app>
```

nsbarsukov (approved):

vladimirpotekhin (approved):

projects/demo/src/index.html
waterplea: Could it be because it needs to be something-app with hyphen to be treated as a custom element?

PR #6988, opened 2024-03-12
ci: add workflow to check `dry-run` of the latest migration schematic

nsbarsukov:


projects/cdk/schematics/ng-update/v3/tests/schematic-migrate-hints.spec.ts
```diff
@@ -98,7 +98,7 @@ describe('ng-update', () => {
         saveActiveProject();
     });
 
-    it('should migrate polymorpheus', async () => {
+    it('should migrate hints', async () => {
```
splincode, line 101: revert

splincode (approved):

vladimirpotekhin (approved):

projects/cdk/schematics/ng-update/v3/tests/schematic-migrate-hints.spec.ts
nsbarsukov: Actually, it is right correction 😄

splincode: sorry

PR #7009, opened 2024-03-15
feat(addon-doc): support external link inside `[routerLink]`

splincode:


projects/addon-doc/components/navigation/navigation.template.html
```diff
@@ -19,6 +19,7 @@
                     *ngFor="let item of group"
                     rel="noopener noreferrer"
                     tuiOption
+                    tuiRouterLink
```
waterplea, line 22: I'm not sure I like this directive. Maybe we can just check if path contains :// and use regular a href instead?

splincode: fixed

nsbarsukov (approved):

vladimirpotekhin (approved):

PR #7022, opened 2024-03-18
fix(kit): `TasbWithMore` fix redundant left margin for single element

### LINKED ISSUE #6989: 🐞 - Unwanted margin for tui-tabs-with-more component
### Playground Link

_No response_

### Description

image
If there is only 1 element left in the component, a margin appears on the left. It seems that for a single element it should not exist. The bug appears on small screens with a long element.

### Angular version

12.2.13

### Taiga UI version

3.70.0

- [X] Chrome
- [X] Firefox
- [X] Safari
- [ ] Edge

- [X] macOS
- [X] Windows
- [X] Linux
- [X] iOS
- [X] Android

vladimirpotekhin:
Closes #6989

### DIFF
--- a/projects/kit/components/tabs/tab/tab.style.less
+++ b/projects/kit/components/tabs/tab/tab.style.less
@@ -6,9 +6,12 @@
 
 :host-context(tui-tabs >):first-child,
 :host-context([tuiTabs] >):first-child,
-:host-context(tui-tabs > :first-child),
-:host-context([tuiTabs] > :first-child) {
+:host-context(tui-tabs > :not(.t-overflown) >):first-child {
     margin-left: 0;
+}
+
+:host-context(tui-tabs > [tuiTab] ~ :not(.t-overflown) >):first-child {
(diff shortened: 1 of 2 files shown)


splincode (approved):

projects/kit/components/tabs/tab/tab.style.less
```diff
@@ -6,11 +6,14 @@
...
 :host-context(tui-tabs >):first-child,
 :host-context([tuiTabs] >):first-child,
-:host-context(tui-tabs > :first-child),
-:host-context([tuiTabs] > :first-child) {
+:host-context(tui-tabs > :not(.t-overflown) >):first-child {
     margin-left: 0;
 }
 
+:host-context(tui-tabs > [tuiTab] ~ :not(.t-overflown) >):first-child {
+    margin-left: var(--tui-tab-margin, 1.5rem);
```
waterplea, line 14: What is this case?

waterplea (approved):
Shouldn't we also take care of `[tuiTabs]` selector?

PR #7048, opened 2024-03-20
feat(addon-doc): support lazy component for preview

splincode:


projects/addon-doc/components/example/example.component.ts
```diff
@@ -78,13 +84,27 @@ export class TuiDocExampleComponent {
             map(value => this.processContent(value)),
         );
 
+    protected readonly lazyComponent$ = this.lazyLoader$$.pipe(
+        switchMap(async lazyImport =>
```
waterplea, line 88: Maybe `switchAll` + `map` so we don't mix RxJS and async.then? If I remember correctly what switchAll does :)

projects/demo/src/modules/components/avatar/index.html
```diff
@@ -7,62 +7,59 @@
         <tui-doc-example
             id="content"
             heading="Content types"
-            [content]="example1"
+            [component]="example1"
```
waterplea, line 10: Maybe a pipe like `tuiExample`?

projects/addon-doc/components/example/example.component.ts

projects/demo/src/modules/components/avatar/index.html
splincode: fixed

nsbarsukov (approved):

PR #7060, opened 2024-03-21
refactor!: move `Title` / `Header` to core / layout packages, make it standalone

vladimirpotekhin:
Closes #


splincode (approved):

projects/demo/src/modules/directives/header/index.html
```diff
@@ -0,0 +1,36 @@
...
+        <tui-doc-example
+            id="interactive"
+            heading="Interactive"
+            [component]="3 | tuiComponent"
+            [content]="3 | tuiExample: 'html,less'"
+        />
+    </ng-template>
+
+    <ng-template pageTab="Setup">
+        <tui-setup
```
waterplea, line 31: We can probably simplify it even further by dropping `ng-template` and writing:
```html
<tui-setup
  *pageTab="'Setup'"
  ...
```

projects/demo/src/modules/directives/header/index.ts
```diff
@@ -0,0 +1,24 @@
...
+    standalone: true,
+    imports: [TuiExamplePipe, TuiComponentPipe, TuiAddonDocModule, TuiSetupComponent],
+    templateUrl: './index.html',
+    styleUrls: ['./index.less'],
+    encapsulation: ViewEncapsulation.None,
+    changeDetection,
+    providers: [tuiDocExampleOptionsProvider({fullsize: true})],
+})
+export default class ExampleComponent {
+    protected readonly exampleModule: TuiRawLoaderContent = import(
```
waterplea, line 17: nit: I try to name those `import` and `template` so they are named the same across pages. We can rename the files as well. Makes it easier to copy-paste pages.

projects/demo/src/modules/directives/title/examples/1/index.ts
```diff
@@ -1,12 +1,14 @@
...
-    selector: 'tui-header-example-1',
+    standalone: true,
+    imports: [TuiTitleDirective],
     templateUrl: './index.html',
     styleUrls: ['./index.less'],
     encapsulation,
     changeDetection,
 })
-export class TuiHeaderExample1 {}
+export default class ExampleComponent {}
```
waterplea, line 14: Do we need the name for the class?

waterplea (approved):

projects/demo/src/modules/directives/header/index.ts
splincode: what about

```html
[code]="'insert-module' | tuiSetup"
```

```html
[code]="'insert-template' | tuiSetup"
```

projects/demo/src/modules/directives/title/examples/1/index.ts
vladimirpotekhin: yes, we need, otherwise the error `ERROR Error: The component 'default_1' needs to be compiled using the JIT compiler` will appear

projects/demo/src/modules/directives/header/index.ts
vladimirpotekhin: I've renamed it for now, we can think about the pipe in a separate PR

projects/demo/src/modules/directives/header/index.html
vladimirpotekhin: great point 👍

Issue #7052, opened 2024-03-21
🚀 - cvc code should be hidden for InputCardGrouped


tinkoffspectrum:
### Description

We need to add an option to hide cvc code input for the component.
Some clients request this feature

Thank you

waterplea:
I suppose we can add a flag to hide value using this CSS rule:
https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-security
It is now finally supported by Firefox.

PR #7068, opened 2024-03-22
feat(addon-commerce): add flag hide cvc code in InputCard, InputCardGrouped

mdlufy:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring
- [ ] Tests related changes

Closes https://github.com/taiga-family/taiga-ui/issues/7052

## What is the new behaviour?

Add flag to hide cvc code in InputCard, InputCardGrouped

### DIFF
--- a/projects/addon-commerce/components/input-card-grouped/input-card-grouped.component.ts
+++ b/projects/addon-commerce/components/input-card-grouped/input-card-grouped.component.ts
@@ -111,6 +111,8 @@ export class TuiInputCardGroupedComponent
 
     protected exampleTextCVC = this.options.exampleTextCVC;
 
+    protected cvcHidden = this.options.cvcHidden;
+
     protected maskCVC: MaskitoOptions = {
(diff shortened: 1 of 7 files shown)


splincode (approved):

waterplea (approved):
Let's default to `false` so we do not change the existing behavior

splincode:
Could you add these changes to v3.x?

mdlufy:
> Let's default to `false` so we do not change the existing behavior

`false` is okay for current behaviour?

waterplea:
> > Let's default to `false` so we do not change the existing behavior
> 
> `false` is okay for current behaviour?

It currently does not hide CVC. We want to introduce the ability to hide it, not change the default value, that would be breaking change, even if it's a small one. Actually, in 4.0 we maybe want `true` to be default, but in v.3.x let's have it false.

mdlufy:
Set `false` as default and make changes to v3.x https://github.com/taiga-family/taiga-ui/pull/7070

PR #7102, opened 2024-03-27
fix(kit): remove pure getNativeElements

### LINKED ISSUE #7017: 🐞 - Stepper: autoscroll to selected element not working after updating stepper state
### Playground Link

https://stackblitz.com/edit/angular-fasgvr?file=src%2Fapp%2Fapp.component.html

### Description

https://github.com/taiga-family/taiga-ui/assets/1676900/7d41e751-b7fa-41fa-8792-da71f935cb0e

In the example it looks at steps$ observable, and the autoscroll works perfectly, both when switching the stepper itself and when navigating through pages. But as soon as I update the steps$ value (in my example it happens after buttons click), the autoscroll doesn't move anymore.

### Angular version

12.2

### Taiga UI version

3.70.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

AlexeyPortnyagin:
Closes #7017

### DIFF
--- a/projects/kit/components/stepper/stepper.component.ts
+++ b/projects/kit/components/stepper/stepper.component.ts
@@ -17,7 +17,6 @@ import {ResizeObserverService} from '@ng-web-apis/resize-observer';
 import {
     EMPTY_QUERY,
     TuiDestroyService,
-    tuiGetOriginalArrayFromQueryList,
     tuiIsElement,
     tuiMoveFocus,
     tuiPure,
@@ -71,9 +70,9 @@ export class TuiStepperComponent {
     }
 
     public indexOf(step: HTMLElement): number {
-        return tuiGetOriginalArrayFromQueryList(this.steps).findIndex(
-            ({nativeElement}) => nativeElement === step,
-        );
+        return this.steps
+            .toArray()
+            .findIndex(({nativeElement}) => nativeElement === step);
     }
 
     public isActive(index: number): boolean {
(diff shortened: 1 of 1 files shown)


projects/kit/components/stepper/stepper.component.ts
```diff
@@ -71,9 +70,9 @@ export class TuiStepperComponent {
     }
 
     public indexOf(step: HTMLElement): number {
-        return tuiGetOriginalArrayFromQueryList(this.steps).findIndex(
```
AlexeyPortnyagin, line 74: It looked like it could be replaced, correct if wrong

waterplea: It is used by `StepComponent` to determine if it is the active one and to activate it. `toArray` creates a new array which is a bit slower than `tuiGetOriginalArrayFromQueryList`. We can get rid of `indexOf` in favor of 
```ts
isActive(step: HTMLElement): boolean {
  return this.steps.get(this.activeItemIndex)?.nativeElement === step;
}
```
and we also need to change `activate` method to accept `HTMLElement` instead of index then.

AlexeyPortnyagin: Changing `activate` method looks like a breaking change. This is a public method, perhaps someone uses it in their projects. Is this really necessary? Maybe just return `tuiGetOriginalArrayFromQueryList`

waterplea: Sure, let's just return it.

waterplea (approved):

waterplea:
Same thing will be needed for v3.x branch

AlexeyPortnyagin:
> Same thing will be needed for v3.x branch

Can you tell me how to pull up the 3.x branch in my fork?

vladimirpotekhin (approved):

splincode (approved):

PR #7106, opened 2024-03-27
refactor(core)!: replace `TUI_DATE_FORMAT`, `TUI_DATE_SEPARATOR` with single observable token

vladimirpotekhin:
Closes #


splincode (approved):

projects/demo/src/modules/components/input-date/index.html
```diff
@@ -9,20 +9,24 @@
             – input with a calendar.
         </div>
 
-        <h3>DI-tokens for date localization:</h3>
+        <h3>
```
waterplea, line 12: Looks like this piece repeats a lot, should we make a component for it?

projects/kit/components/input-date-multi/input-date-multi.component.ts
```diff
@@ -132,16 +133,24 @@ export class TuiInputDateMultiComponent
...
-            changeDateSeparator(dateTexts[this.dateFormat], this.dateSeparator),
+            changeDateSeparator(
+                dateTexts[this.dateFormat.mode],
+                this.dateFormat.separator,
+            ),
         ),
     );
 
+    protected readonly dateFormat$ = inject(TUI_DATE_FORMAT)
+        .pipe(tuiWatch(this.cdr), takeUntil(this.destroy$))
```
waterplea, line 149: Let's migrate to `takeUntilDestroyed` while we're at it here and in other places where you use destroy service.

projects/kit/components/input-date-multi/input-date-multi.component.ts
```diff
@@ -132,16 +133,24 @@ export class TuiInputDateMultiComponent
...
         map(dateTexts =>
-            changeDateSeparator(dateTexts[this.dateFormat], this.dateSeparator),
+            changeDateSeparator(
+                dateTexts[this.dateFormat.mode],
+                this.dateFormat.separator,
+            ),
         ),
     );
 
+    protected readonly dateFormat$ = inject(TUI_DATE_FORMAT)
```
waterplea, line 148: Now that we have DI in decorators we can actually make a decorator for this and reuse it everywhere. I'll try to implement it after we merge, should be pretty cool.
```ts
@tuiFromObservable(TUI_DATE_FORMAT)
dateFormat = TUI_DEFAULT_DATE_FORMAT
```

projects/kit/components/input-date/input-date.directive.ts
```diff
@@ -21,6 +22,10 @@ export class TuiInputDateDirective extends AbstractTuiTextfieldHost<TuiInputDate
         return this.host.computedMin;
     }
 
+    public get format(): TuiDateFormatSettings {
```
waterplea, line 25: With the decorator I spoke of it would be clean to reuse it inside `TuiNativeDateDirective` instead of introducing more coupling.

waterplea (approved):

projects/demo/src/modules/components/input-date/index.html
vladimirpotekhin: made a link to the token page like NumberFormat

projects/kit/components/input-date-multi/input-date-multi.component.ts
vladimirpotekhin: cool idea

projects/kit/components/input-date/input-date.directive.ts
vladimirpotekhin: agree

PR #7110, opened 2024-03-28
feat(kit): add constant for default country masks

nguyenphutrong:
This PR introduces a new provider `tuiCountriesMasksProvider` that allows overriding or extending the default country masks for phone number formatting.

Usage example:
```js
providers: [
        tuiCountriesMasksProvider({
            [TuiCountryIsoCode.VN]: '+84#########' // Overriding the mask for Vietnam
        })
],
```

### DIFF
--- a/projects/kit/tokens/countries-masks.ts
+++ b/projects/kit/tokens/countries-masks.ts
@@ -1,4 +1,5 @@
-import {tuiCreateToken} from '@taiga-ui/cdk';
+import type {Provider} from '@angular/core';
+import {tuiCreateToken, tuiProvideOptions} from '@taiga-ui/cdk';
 import {TuiCountryIsoCode} from '@taiga-ui/i18n';
 
 export const TUI_COUNTRIES_DEFAULT_MASKS: Record<TuiCountryIsoCode, string> = {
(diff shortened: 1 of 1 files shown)


projects/kit/tokens/countries-masks.ts
```diff
@@ -217,4 +217,8 @@ export const TUI_COUNTRIES_MASKS = tuiCreateToken<Record<TuiCountryIsoCode, stri
     [TuiCountryIsoCode.ZA]: '+27##-###-####',
     [TuiCountryIsoCode.ZM]: '+260##-###-####',
     [TuiCountryIsoCode.ZW]: '+263#-######',
-});
+};
+
+export const TUI_COUNTRIES_MASKS = tuiCreateToken<Record<TuiCountryIsoCode, string>>(
```
waterplea, line 222: Let's add a helper like we do with other options (for example [here](https://github.com/taiga-family/taiga-ui/blob/main/projects/kit/components/checkbox/checkbox.options.ts#L26))

projects/kit/tokens/countries-masks.ts
```diff
@@ -217,4 +218,12 @@ export const TUI_COUNTRIES_MASKS = tuiCreateToken<Record<TuiCountryIsoCode, stri
     [TuiCountryIsoCode.ZA]: '+27##-###-####',
     [TuiCountryIsoCode.ZM]: '+260##-###-####',
     [TuiCountryIsoCode.ZW]: '+263#-######',
-});
+};
+
+export const TUI_COUNTRIES_MASKS = tuiCreateToken(TUI_COUNTRIES_DEFAULT_MASKS);
```
nsbarsukov, line 223: @waterplea @vladimirpotekhin 

We have plans to use `@maskito/phone` in `v4.0`.
Do we actually require this new token in `main`-branch ? (probably, it will be never released 🤔 )

vladimirpotekhin: yes, in fact, we need these changes in the 3.x branch only

waterplea: @nguyenphutrong could you please make these changes into `v3.x` branch instead of `main` so it will be available for everybody in the current major version. Upcoming 4.0 will have this mechanism reworked.

PR #7148, opened 2024-04-01
chore(demo): cleanup

splincode:


projects/demo/src/modules/components/action/index.ts
```diff
@@ -27,11 +26,11 @@ import {TuiActionExample5} from './examples/5';
...
         TuiActionModule,
     ],
-    templateUrl: './action.template.html',
+    templateUrl: './index.html',
     encapsulation,
     changeDetection,
 })
-export default class ExampleTuiActionComponent {
+export default class PageComponent {
     protected readonly exampleModule = import('./examples/import/import-module.md?raw');
```
waterplea, line 34: Would be nice to drop all those imports as well in favor of default names

waterplea (approved):

projects/demo/src/modules/components/action/index.ts
splincode: image

vladimirpotekhin (approved):

PR #7149, opened 2024-04-02
chore(demo): migrate documentation page `BadgedContent` to standalone components

mdlufy:
Partially solves #6957 
Done: Migrate BadgedContent components to standalone


projects/demo/src/modules/components/badged-content/badged-content.component.ts
```diff
@@ -1,13 +1,40 @@
...
+        TuiFallbackSrcPipe,
+        TuiBadgedContentExample1,
+        TuiBadgedContentExample2,
+        TuiBadgedContentExample3,
+    ],
     templateUrl: './badged-content.template.html',
     changeDetection,
 })
-export class ExampleTuiBadgedContentComponent {
+export default class ExampleTuiBadgedContentComponent {
```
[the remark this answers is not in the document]
mdlufy: good idea

projects/demo/src/modules/components/badged-content/index.html
```diff
@@ -0,0 +1,53 @@
...
+    type="components"
+>
+    <ng-template pageTab>
+        <p>BadgedContent is a wrapper for other components to add badges and notifications to them.</p>
+
+        <tui-doc-example
+            *ngFor="let example of examples; let index = index"
+            [component]="index + 1 | tuiComponent"
+            [content]="index + 1 | tuiExample"
+            [description]="index === 1 && description"
```
mdlufy, line 13: it can become cumbersome when there are several descriptions

can we leave it like this or think about how we can make it more beautiful?

mdlufy: this is ok for one description, but if several maybe some interface to use in cycle `examples`

`
interface TuiExample {
    name: string;
    description: string | PolymorpheusContent;
}
`

projects/demo/src/modules/components/badged-content/badged-content.component.ts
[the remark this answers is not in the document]
mdlufy: done

projects/demo/src/modules/components/badged-content/index.html
vladimirpotekhin: yes, let's leave it like this 

for more complex cases we can simply write examples without the loop

projects/demo/src/modules/components/badged-content/examples/1/index.html
```diff
@@ -7,16 +7,16 @@
     </tui-badge-notification>
     <tui-avatar
         size="s"
-        src="tuiIconUserLarge"
-        [round]="false"
+        [round]="round"
```
vladimirpotekhin, line 10: let's return strings and bools without binding and extra fields in class

vladimirpotekhin (approved):

splincode (approved):

PR #7164, opened 2024-04-03
refactor!: move `Surface` to core package

vladimirpotekhin:
Closes #


splincode (approved):

projects/demo/src/modules/app/pages.ts
```diff
@@ -915,10 +915,10 @@ export const pages: TuiDocPages = [
         route: '/experimental/segmented',
     },
     {
-        section: 'Experimental',
+        section: 'Components',
         title: 'Surface',
         keywords: 'card, container, wrapper, image, blur, overlay',
-        route: '/experimental/surface',
+        route: '/components/surface',
```
waterplea, line 921: Should it be directive? Should button and link be directives, since they are actually directives? We need to think about it :)

projects/demo/src/modules/components/surface/index.html
```diff
@@ -24,21 +18,22 @@
...
             </tui-notification>
-            <tui-surface-example-1 />
         </tui-doc-example>
 
         <tui-doc-example
             id="presets"
             heading="Presets"
-            [content]="example2"
+            [component]="2 | tuiComponent"
+            [content]="2 | tuiExample: 'html'"
```
waterplea, line 36: Why not go over them with ngFor? We need typescript all the time anyway, since these are standalone components that include imports.

projects/demo/src/modules/app/pages.ts
vladimirpotekhin: moved the discussion to the team chat

waterplea (approved):

PR #7165, opened 2024-04-03
chore(demo): migrate documentation page `Block` to standalone components

mdlufy:
Partially solves #6957 
Done: Migrate Block component to standalone


projects/demo/src/modules/components/block/index.html
```diff
@@ -0,0 +1,22 @@
...
+    <ng-template pageTab>
+        <p>
+            <code>Block</code>
+            is a special presentation of a checkbox/radiobutton which can display actual control or be a control itself
+        </p>
+
+        <tui-doc-example
+            *ngFor="let example of examples; let index = index"
+            [component]="index + 1 | tuiComponent"
+            [content]="index + 1 | tuiExample: 'html'"
```
mdlufy, line 15: why only html, not ts and less in examples?

vladimirpotekhin: empty `ts` files were useless until they became `standalone`

but now I think it's better to add ts, let's do it

(`less` are still useless because they don't affect components, just for layout purposes)

mdlufy: done

mdlufy: `ts` were not completely empty, there are form controls values, so I asked)

vladimirpotekhin (approved):

splincode (approved):

PR #7167, opened 2024-04-04
refactor: drop enum for `TuiCurrency`

splincode:


projects/addon-commerce/enums/currency.ts
```diff
@@ -397,5 +395,7 @@ export enum TuiCurrency {
      * @description Countries and territories: China
      * @description Currency symbol: CN¥
      */
-    YuanRenminbi = 'CNY',
-}
+    YuanRenminbi: 'CNY',
+} as const;
+
+export type TuiCurrencyT = (typeof TuiCurrency)[keyof typeof TuiCurrency];
```
waterplea, line 401: We only used `xxxT` naming to distinguish between enum and joint type while we have both for backwards compatibility. Surely we shouldn't keep this name in the next major release.

splincode: `TuiCurrencyType` is okey?

splincode: fixed

waterplea (approved):
We should keep in mind migration of the type

PR #7176, opened 2024-04-05
refactor(kit)!: drop `Action` and `MarkerIcon`

waterplea:
Closes #


splincode (approved):

projects/demo/src/modules/components/select/examples/9/account/my-account.component.html
```diff
@@ -1,7 +1,6 @@
 <span class="icon">
-    <tui-marker-icon
+    <tui-svg
```
nsbarsukov, line 2: Why not new `<tui-icon />` ?

nsbarsukov (approved):

projects/demo/src/modules/components/select/examples/9/account/my-account.component.html
waterplea: We need colored SVGs here. Ideally it would be `img` but I don't have time to fix this in a demo example. We can just open source tuiAccount later.

vladimirpotekhin (approved):

PR #7193, opened 2024-04-09
feat(experimental): `DropdownMobile` add new dropdown behavior

waterplea:
Closes #


vladimirpotekhin (approved):

projects/addon-mobile/directives/dropdown-mobile/dropdown-mobile.component.ts
```diff
@@ -0,0 +1,106 @@
...
+        '(mousedown.silent.prevent)': '0',
+        '(document:click.silent.capture)': 'onClick($event)',
+        '(window>scroll.silent.capture)': 'refresh($event.currentTarget.visualViewport)',
+        '(visualViewport>resize.silent)': 'refresh($event.target)',
+        '(visualViewport>scroll.silent)': 'refresh($event.target)',
+    },
+})
+export class TuiDropdownMobileComponent implements OnDestroy {
+    private readonly scrollTop = this.doc.documentElement.scrollTop;
+    private readonly observer = new ResizeObserver(() =>
```
nsbarsukov, line 40: SSR ?

Why not `@ng-web-apis` ?

```
import {ResizeObserverService} from '@ng-web-apis/resize-observer';
```

projects/demo/src/modules/directives/dropdown/dropdown.component.ts
```diff
@@ -42,6 +42,11 @@ export class ExampleTuiDropdownComponent extends AbstractExampleTuiDropdown {
         LESS: import('./examples/4/index.less?raw'),
     };
 
+    readonly example5: TuiDocExample = {
+        TypeScript: import('./examples/4/index.ts?raw'),
+        HTML: import('./examples/4/index.html?raw'),
```
nsbarsukov, line 47:
```suggestion
    readonly example5: TuiDocExample = {
        TypeScript: import('./examples/5/index.ts?raw'),
        HTML: import('./examples/5/index.html?raw'),
```

nsbarsukov (approved):

projects/addon-mobile/directives/dropdown-mobile/dropdown-mobile.component.ts
waterplea: We do not need RxJS here. I think SSR is ok as it is used only on mobile when dropdown is opened, which is not going to happen under SSR.

PR #7196, opened 2024-04-10
refactor: use `TuiFlagPipe` as standalone

splincode:


projects/demo/src/modules/pipes/flag/index.ts
```diff
@@ -0,0 +1,15 @@
...
+
+@Component({
+    standalone: true,
+    imports: [TuiAddonDocModule, TuiComponentPipe, TuiExamplePipe],
+    templateUrl: './index.html',
+    changeDetection,
+})
+export default class PageComponent {
+    protected readonly exampleModule = import('./examples/import/import-module.md?raw');
+    protected readonly exampleHtml = import('./examples/import/insert-template.md?raw');
```
vladimirpotekhin, line 14: forgot to use `tui-setup` ?

vladimirpotekhin (approved):

projects/demo/src/modules/pipes/flag/index.ts
splincode: fixed

projects/demo/src/modules/pipes/flag/index.html
```diff
@@ -0,0 +1,22 @@
...
+            id="base"
+            heading="Basic"
+            [component]="1 | tuiComponent"
+            [content]="1 | tuiExample: 'html,ts'"
+        />
+    </ng-template>
+
+    <tui-setup
+        *pageTab
+        [import]="import"
```
waterplea, line 19: We can rename files so it will use default values

waterplea (approved):

PR #7206, opened 2024-04-11
refactor: set `tui-svg` as standalone

splincode:


waterplea (approved):
I think we have some deprecated svg related options/tokens such as interceptors, processors etc. Let's refactor those next.

vladimirpotekhin (approved):

PR #7209, opened 2024-04-11
refactor: set `tui-root` as standalone

splincode:


projects/cdk/schematics/ng-update/v4/steps/constants/identifiers-to-replace.ts
```diff
@@ -293,4 +293,8 @@ export const IDENTIFIERS_TO_REPLACE: ReplacementIdentifier[] = [
         from: {name: 'TuiFlagPipeModule', moduleSpecifier: '@taiga-ui/core'},
         to: {name: 'TuiFlagPipe', moduleSpecifier: '@taiga-ui/core'},
     },
+    {
+        from: {name: 'TuiRootModule', moduleSpecifier: '@taiga-ui/core'},
```
waterplea, line 297: It would break `importProvidersFrom`

projects/core/components/hosted-dropdown/test/hosted-dropdown.component.spec.ts
```diff
@@ -126,9 +127,10 @@ describe('TuiHostedDropdown', () => {
                 TuiTextfieldControllerModule,
                 TuiDataListModule,
                 TuiElementModule,
-                TuiRootModule,
+                TuiRootComponent,
             ],
             declarations: [TestComponent],
+            providers: [NG_EVENT_PLUGINS],
```
waterplea, line 133: Ideally we would want to have this in every test, since you never know what component relies on plugins. Can we somehow do that?

projects/demo-cypress/src/tests/elastic-sticky.cy.ts
```diff
@@ -36,7 +37,8 @@ describe('ElasticSticky', () => {
 
     beforeEach(() =>
         cy.mount(TestComponent, {
-            imports: [TuiRootModule, TuiScrollbarComponent, TuiElasticStickyModule],
+            imports: [TuiRootComponent, TuiScrollbarComponent, TuiElasticStickyModule],
```
waterplea, line 40: Like here, I don't think we need component, but we need plugins. Some other tests are also like that.

waterplea (approved):

projects/cdk/schematics/ng-update/v4/steps/constants/identifiers-to-replace.ts
splincode: I added comment to https://github.com/taiga-family/taiga-ui/issues/5808

projects/demo-cypress/src/tests/elastic-sticky.cy.ts
splincode: What do you mean? We have `<tui-root>` inside test component

splincode: without TuiRootComponent 
image

projects/core/components/hosted-dropdown/test/hosted-dropdown.component.spec.ts
splincode: fixed

projects/demo-cypress/src/tests/elastic-sticky.cy.ts
waterplea: I don't know what is searching for `tui-root` here, but we do not need it here. We only had it there before to add event plugins. Mostly we do not need `tui-root` for anything, other than portal based components (dropdowns, dialogs, alerts, hints).

splincode: I can create an issue about `tui-root` in e2e tests and deal with the error

waterplea: Sounds good 👍

vladimirpotekhin (approved):

PR #7220, opened 2024-04-13
fix(core): remove overflow clip from animation

splincode:


waterplea:
Why? It's there for a reason. Otherwise scrollbar on Windows was appearing and disappearing during fullscreen dialog if I remember correctly.

Issue #7223, opened 2024-04-14
Zoneless components for Angular 18


splincode:
### Discussed in https://github.com/orgs/taiga-family/discussions/7221

<div type='discussions-op-text'>

<sup>Originally posted by **hakimio** April 13, 2024</sup>
Angular 18 will introduce zoneless signal components and "Angular Material" component library has already done the required migrations to support zoneless. Do you have any plans to support "zoneless Angular"?

Some of the PRs merged to migrate Angular Material:
https://github.com/angular/components/pull/28332
https://github.com/angular/components/pull/28782
https://github.com/angular/components/pull/28783

PS: right now "Taiga UI" doesn't work correctly with the private `ɵprovideZonelessChangeDetection()` API but that might be quite different from the real zoneless experience which will be delivered in Angular 18.</div>

https://github.com/angular/angular/blob/main/adev/src/content/guide/zoneless.md#requirements-for-zoneless-compatibility

PR #7224, opened 2024-04-14
feat(addon-mobile): reduce zone.js in `tuiElasticSticky`

splincode:
Part of #7223

### DIFF
--- a/projects/addon-mobile/directives/elastic-sticky/elastic-sticky.service.ts
+++ b/projects/addon-mobile/directives/elastic-sticky/elastic-sticky.service.ts
@@ -1,41 +1,41 @@
-import {ElementRef, inject, Injectable, NgZone} from '@angular/core';
+import {
+    afterNextRender,
+    ChangeDetectorRef,
+    ElementRef,
+    inject,
+    Injectable,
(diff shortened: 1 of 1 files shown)


projects/addon-mobile/directives/elastic-sticky/elastic-sticky.service.ts
```diff
@@ -1,41 +1,41 @@
...
-                        const {offsetTop} = tuiGetElementOffset(host, this.el);
+        super(subscriber => {
+            afterNextRender(
+                () => {
+                    const host = this.el.closest(SCROLL_REF_SELECTOR) || this.scrollRef;
+                    const {offsetHeight} = this.el;
+                    const {offsetTop} = tuiGetElementOffset(host, this.el);
 
-                        return tuiScrollFrom(host).pipe(
+                    return tuiScrollFrom(host)
```
waterplea, line 36: I don't think this return matters. We are supposed to return teardown logic but there is void returned from the super callback.

nsbarsukov (approved):

vladimirpotekhin (approved):

PR #7241, opened 2024-04-16
feat(cdk): `Keyboard` add new service

waterplea:
Closes #


projects/cdk/services/keyboard.service.ts
```diff
@@ -0,0 +1,59 @@
+import {DOCUMENT} from '@angular/common';
+import {Inject, Injectable, OnDestroy} from '@angular/core';
+import {tuiGetNativeFocused} from '@taiga-ui/cdk';
+import {fromEvent, merge} from 'rxjs';
+import {filter} from 'rxjs/operators';
+
+@Injectable({
+    providedIn: 'root',
+})
+export class TuiKeyboardService implements OnDestroy {
```
splincode, line 10: why from `cdk` and not from `addon-mobile`?

splincode (approved):

PR #7249, opened 2024-04-17
feat(addon-mobile): `Keyboard` add new service

waterplea:
Closes #


projects/demo/src/modules/services/keyboard/examples/inject.md
```diff
@@ -0,0 +1,16 @@
+```ts
+import {Inject, Self} from '@angular/core';
+import {TuiKeyboardService} from '@taiga-ui/cdk';
+
+// ...
+export class MyComponent {
+  constructor(
+    @Inject(TuiKeyboardService)
+    private readonly keyboard: TuiKeyboardService,
```
splincode, line 9: `inject(TuiKeyboardService)`

splincode (approved):

vladimirpotekhin (approved):

PR #7276, opened 2024-04-19
refactor!: `Chip` move to kit

vladimirpotekhin:
Closes #


splincode (approved):

projects/demo/src/modules/components/tag/tag.template.html
```diff
@@ -1,4 +1,5 @@
 <tui-doc-page
+    deprecated
```
nsbarsukov, line 2: Will it be deleted in separate PR ?

nsbarsukov (approved):

projects/demo/src/modules/components/tag/tag.template.html
vladimirpotekhin: yes, after replacing tags to chips in `multiselect` / `input-tag`. After new textfields, probably

PR #7287, opened 2024-04-22
fix(cdk): `Dialog` fix jumping to top on closing

splincode:
Cherry pick

### DIFF
--- a/projects/core/animations/animations.ts
+++ b/projects/core/animations/animations.ts
@@ -34,10 +34,7 @@ export const tuiHost = trigger('tuiHost', [
         style({overflow: 'clip'}),
         query(':scope > *', [animateChild()], {optional: true}),
     ]),
-    transition(':leave', [
-        style({overflow: 'clip'}),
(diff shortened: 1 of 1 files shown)


projects/core/animations/animations.ts
```diff
@@ -30,14 +30,8 @@ export const TUI_PARENT_ANIMATION = trigger('tuiParentAnimation', [
 export const TUI_PARENT_STOP = trigger('tuiParentStop', [transition(':enter', [])]);
 
 export const tuiHost = trigger('tuiHost', [
-    transition(':enter', [
-        style({overflow: 'clip'}),
```
waterplea, line 34: I thought it jumps on iOS if we remove clip on :enter

splincode: image

I just remove `style({overflow: 'clip'}),`

splincode: fixed

vladimirpotekhin (approved):

nsbarsukov (approved):

PR #7294, opened 2024-04-23
refactor: rename `textfield-placeholder` to `label`

splincode:


projects/cdk/schematics/ng-update/v4/steps/migrate-styles.ts
```diff
@@ -10,6 +11,7 @@ export function migrateStyles(): void {
             let fullText = sourceFile.getFullText();
 
             fullText = fullText
+                .replaceAll('textfield-placeholder()', 'label()')
```
vladimirpotekhin, line 14: there is no `tui` prefix, unfortunately, seems like it is better to not migrate this

splincode: fixed

vladimirpotekhin (approved):

projects/core/styles/mixins/textfield.less
```diff
@@ -166,8 +166,7 @@
     padding-inline-end: 0.25rem;
 }
 
-// TODO: Rename to label in 4.0
-.textfield-placeholder() {
+.label() {
```
nsbarsukov, line 169: `textfield-label` ? 🤔 

(all 4 mixins above still has `textfield-` prefix)

projects/core/styles/mixins/textfield.scss
```diff
@@ -157,8 +157,7 @@ $line-height-l: 1.25rem;
     padding-inline-end: 0.25rem;
 }
 
-// TODO: Rename to label in 4.0
-@mixin textfield-placeholder() {
+@mixin label() {
```
nsbarsukov, line 160: Same question

splincode: fixed

nsbarsukov (approved):

PR #7343, opened 2024-05-01
fix(core): fix caret in InputPassword while clicking show/hide button

### LINKED ISSUE #7281: 🐞 - Caret in InputPassoword works instable while clicking show/hide password button
### Playground Link

_No response_

### Description

When starting randomly hitting "show" or "hide" buttons in InputPassword component, the caret can change its position to the beginning of the field though initially it was set to other place.

The behavior can be reproduced on https://taiga-ui.dev/.

See attachment
Kapture 2024-04-22 at 11 42 01

### Angular version

17.1.0

### Taiga UI version

3.68.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

al-march:
Closes #7281

NgDoCheck was invoking type change of the input too frequently, which, I believe, was causing the cursor to lag. In the current implementation, the type changes only when necessary, and the issue is no longer observed.

### DIFF
--- a/projects/kit/components/input-password/input-password.directive.ts
+++ b/projects/kit/components/input-password/input-password.directive.ts
@@ -1,7 +1,9 @@
-import type {DoCheck} from '@angular/core';
+import type {AfterViewInit} from '@angular/core';
 import {Directive} from '@angular/core';
+import {takeUntilDestroyed} from '@angular/core/rxjs-interop';
 import {tuiIsInput} from '@taiga-ui/cdk';
 import {AbstractTuiTextfieldHost, tuiAsTextfieldHost} from '@taiga-ui/core';
+import {startWith} from 'rxjs';
 
 import type {TuiInputPasswordComponent} from './input-password.component';
 
@@ -11,9 +13,14 @@ import type {TuiInputPasswordComponent} from './input-password.component';
 })
 export class TuiInputPasswordDirective
     extends AbstractTuiTextfieldHost<TuiInputPasswordComponent>
-    implements DoCheck
+    implements AfterViewInit
 {
     protected input?: HTMLInputElement;
+
+    protected readonly isPasswordHidden$ = this.host.passwordHidden.pipe(
+        startWith(this.host.inputType),
+        takeUntilDestroyed(),
+    );
 
     public onValueChange(value: string): void {
         this.host.onValueChange(value);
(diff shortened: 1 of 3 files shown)


projects/kit/components/input-password/input-password.directive.ts
```diff
@@ -23,12 +30,14 @@ export class TuiInputPasswordDirective
...
-            this.host.nativeFocusableElement &&
-            tuiIsInput(this.host.nativeFocusableElement)
-        ) {
-            this.host.nativeFocusableElement.type = this.host.inputType;
-        }
+    public ngAfterViewInit(): void {
+        this.isPasswordHidden$.subscribe(() => {
+            if (
+                this.host.nativeFocusableElement &&
+                tuiIsInput(this.host.nativeFocusableElement)
```
splincode, line 37:
```ts
this.isPasswordHidden$.pipe(
  filter(() => this.host.nativeFocusableElement && tuiIsInput(this.host.nativeFocusableElement))
).subscribe(() => {});
```

projects/kit/components/input-password/input-password.directive.ts
```diff
@@ -23,12 +30,14 @@ export class TuiInputPasswordDirective
...
     }
 
-    public ngDoCheck(): void {
-        if (
-            this.host.nativeFocusableElement &&
-            tuiIsInput(this.host.nativeFocusableElement)
-        ) {
-            this.host.nativeFocusableElement.type = this.host.inputType;
-        }
+    public ngAfterViewInit(): void {
```
splincode, line 33: can we use `afterNextRender(() => {})` in constructor?

al-march: From the documentation:

> Components are not guaranteed to be hydrated before the callback runs.

Are we sure we want to use afterNextRender? I believe using the ngAfterViewInit hook is sufficient, as it guarantees that the necessary input will be initialized.

splincode: sure

splincode (approved):

projects/kit/components/input-password/input-password.directive.ts
al-march: unfortunately the filter pipe won't give the expected result since Typescript won't consider `this.host.nativeFocusableElement` as an `HTMLInputElement` and I'll have to use `if` inside the callback anyway

waterplea (commented):
While I appreciate trying to optimize the code, I don't like the idea of adding subscription and output.

1) Would it work if we move it to `ngAfterContentCheck` rather than `ngDoCheck`?
2) What if we just add check `this.host.nativeFocusableElement.type !== this.host.inputType`?
3) Please do it in `v3.x` branch, all inputs in `main` will be completely reworked

PR #7347, opened 2024-05-02
refactor: zoneless for `tui-doc-main`

splincode:
image


waterplea:
We don't want to enforce `OnPush`/`zoneless` to addon-doc users at this point.

PR #7351, opened 2024-05-02
feat(core): add possibility to control alerts queue

MillerSvt:
I added the ability to manage the queue. I implemented this through the operator because we need to modify not only the number of alerts but also their order. By default, `mergeAll` displays the first `N ?? Infinity` alerts, but we need to show the last `N` alerts. To achieve this, we need to pass not only a number but also an operator who will control the process.

Generally speaking, if you need to restrict the first N elements, it is sufficient to provide:
```
{
  provide: TUI_ALERT_QUEUE_OPERATOR,
  useValue: () => mergeAll(N),
}
```

Alternatively, you can use your own operator.

### DIFF
--- a/projects/demo/src/modules/services/alerts/examples/7/app.module.ts
+++ b/projects/demo/src/modules/services/alerts/examples/7/app.module.ts
@@ -1,0 +1,14 @@
+import {NgModule} from '@angular/core';
+import {TUI_ALERT_QUEUE_OPERATOR, TuiAlertModule} from '@taiga-ui/core';
+import {mergeAll} from 'rxjs/operators';
+
+@NgModule({
+    providers: [
+        TuiAlertModule,
+        {
+            provide: TUI_ALERT_QUEUE_OPERATOR,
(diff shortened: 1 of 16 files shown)


splincode (approved):

MillerSvt:
@splincode Added a demo for the new functionality. It's not working very well yet, but it couldn't do any better.
After the weekend, I suggest we meet and discuss it.

projects/demo/src/modules/services/alerts/examples/7/app.module.ts
```diff
@@ -0,0 +1,14 @@
+import {NgModule} from '@angular/core';
+import {TUI_ALERT_QUEUE_OPERATOR, TuiAlertModule} from '@taiga-ui/core';
+import {mergeAll} from 'rxjs/operators';
+
+@NgModule({
+    providers: [
+        TuiAlertModule,
```
splincode, line 7: ?

PR #7355, opened 2024-05-02
fix(core): active tab does not have class _active

al-march:
Before:

https://github.com/taiga-family/taiga-ui/assets/42899726/df7c4d8f-1602-4ea7-bb45-4a84dd548042

After:

https://github.com/taiga-family/taiga-ui/assets/42899726/26623e9e-4e42-40bf-9a20-72b427e1aef8

### DIFF
--- a/projects/kit/components/tabs/tabs.directive.ts
+++ b/projects/kit/components/tabs/tabs.directive.ts
@@ -1,10 +1,12 @@
 import type {AfterViewChecked} from '@angular/core';
 import {
+    afterNextRender,
     Directive,
     ElementRef,
     EventEmitter,
     HostListener,
     inject,
+    INJECTOR,
     Input,
     Output,
 } from '@angular/core';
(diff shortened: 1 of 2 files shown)


splincode (approved):

projects/kit/components/tabs/tabs.directive.ts
```diff
@@ -51,14 +54,12 @@ export class TuiTabsDirective implements AfterViewChecked {
     }
 
     public ngAfterViewChecked(): void {
```
splincode, line 56: can we move this to another lifecycle hook? for example `ngOnInit`, will this work?

al-march: I think not, because we have to update the state after every rendering, and not just after initialization. But if anyone has any ideas on how to implement it without `afterNextRender`, I’d be glad for a hint

waterplea (approved):

PR #7364, opened 2024-05-06
fix(core): fix compatibility tuiGroup & tuiLabel

MillerSvt:


projects/core/components/label/label.style.less
```diff
@@ -35,4 +35,5 @@
     overflow: inherit;
     text-overflow: ellipsis;
     pointer-events: auto;
+    border-radius: inherit;
```
waterplea, line 38: Needs the same in new label under new textfield

waterplea (approved):

splincode (approved):

PR #7367, opened 2024-05-06
chore(demo): animation modules should be standalone

splincode:


projects/demo/src/modules/components/animations/state.ts
```diff
@@ -0,0 +1,9 @@
+import {Injectable} from '@angular/core';
+import {BehaviorSubject} from 'rxjs';
+
+@Injectable()
```
waterplea, line 4: Can be root, so no provide in `PageComponent`

waterplea (approved):

vladimirpotekhin (approved):

PR #7390, opened 2024-05-08
refactor: drop TuiBarModule

splincode:


projects/demo/src/modules/components/bar/index.ts
```diff
@@ -0,0 +1,42 @@
...
+import {TuiBarComponent} from '@taiga-ui/addon-charts';
+import {TuiAddonDocModule} from '@taiga-ui/addon-doc';
+import type {TuiSizeL, TuiSizeS} from '@taiga-ui/core';
+
+@Component({
+    standalone: true,
+    imports: [
+        TuiAddonDocModule,
+        TuiBarComponent,
+        TuiSetupComponent,
```
waterplea, line 19: We have `TuiDemoModule` for all the demo related components in one import

nsbarsukov (approved):

PR #7418, opened 2024-05-14
fix(core): same overflow behaviour inside `tui-svg`

splincode:


waterplea:
Let's not do this with svg, too risky

PR #7424, opened 2024-05-14
[WIP] chore(demo-playwright): use Firefox as temporary workaround for unstable Chromium tests

nsbarsukov:


nsbarsukov:
Font problem persists((

01-character-with-descenders-inside-placeholder diff
01-native-select-value diff

PR #7445, opened 2024-05-15
refactor: rename TuiTabBarModule to TuiTabBar

splincode:


waterplea (approved):

projects/demo/src/modules/components/tab-bar/index.html
```diff
@@ -44,58 +45,28 @@
...
-                    [code]="exampleModule"
-                />
-            </li>
-
-            <li>
-                <p>
-                    Add to the template inside
-                    <code>Root</code>
-                    component at the required layer and position with CSS:
-                </p>
```
nsbarsukov, line 92: I'm not sure that basic `<tui-setup />` is enough here 🤔

splincode: I add that comment inside template

image

nsbarsukov (approved):

PR #7458, opened 2024-05-16
feat(schematics): add manual migration

splincode:
```
npx schematics ./dist/cdk:updateToV3_78 --allow-private --skip-deep-imports --dry-run false --experimental
```


projects/cdk/schematics/collection.json
```diff
@@ -17,6 +17,12 @@
             "description": "Manually run migration to update Taiga to v3.x.x",
             "factory": "./ng-update/v3/index#updateToV3",
             "schema": "./ng-update/v3/schema.json"
+        },
+        "updateToV3_78": {
+            "private": true,
+            "description": "Manually run migration to update proprietary icons",
+            "factory": "./ng-update/v3-78/index#updateToV3_78",
+            "schema": "./ng-update/v3/schema.json"
```
nsbarsukov, line 25: v3 `schema.json` does not contains `experimental` property

https://github.com/taiga-family/taiga-ui/blob/0955f9928354fe8478cc4ba864ef166b3485c672/projects/cdk/schematics/ng-update/v3/schema.json#L6-L12

Should we create the separate `./ng-update/v3-78/schema.json` and use it here ?

nsbarsukov (approved):

PR #7482, opened 2024-05-20
refactor: change signature for `tuiGetElementObscures`

splincode:


projects/cdk/utils/dom/get-element-obscurers.ts
```diff
@@ -5,14 +5,10 @@ import {tuiGetDocumentOrShadowRoot} from './get-document-or-shadow-root';
 
 /**
  * Returns array of Elements covering edges of given element or null if at least one edge middle point is visible
- *
- * CAUTION: Empty array means element if offscreen i.e. covered by no elements, rather than not covered
```
nsbarsukov, line 9: Why was this line deleted ?

projects/cdk/utils/dom/get-element-obscurers.ts
```diff
@@ -50,10 +46,12 @@ export function tuiGetElementObscures(element: Element): readonly Element[] | nu
     const nonNull = elements.filter(tuiIsPresent);
 
     if (!nonNull.length) {
-        return nonNull;
+        return nonNull as [Element, Element, Element, Element];
```
nsbarsukov, line 49:
```suggestion
        return nonNull as [];
```

splincode: fixed

projects/cdk/utils/dom/get-element-obscurers.ts
```diff
@@ -50,10 +51,12 @@ export function tuiGetElementObscures(element: Element): readonly Element[] | nu
     const nonNull = elements.filter(tuiIsPresent);
 
     if (!nonNull.length) {
-        return nonNull;
+        return nonNull as [];
```
waterplea, line 54: No need to typecast, we can just return `[]`

waterplea (approved):

projects/cdk/utils/dom/get-element-obscurers.ts
splincode: fixed

nsbarsukov (approved):

PR #7485, opened 2024-05-20
refactor: remove $any in 4.0

splincode:


projects/core/components/data-list/data-list.component.ts
```diff
@@ -63,7 +69,7 @@ export class TuiDataListComponent<T> implements TuiDataListAccessor<T> {
 
     @Input()
     @HostBinding('attr.data-list-size')
-    public size = this.controller?.size || 'm';
+    public size: TuiSizeL | TuiSizeS | TuiSizeXS = this.controller?.size || 'm';
```
waterplea, line 72: TuiSizeS is unnecessary, it is included in TuiSizeXS

waterplea (approved):

projects/core/components/data-list/data-list.component.ts
splincode: fixed

nsbarsukov (approved):

PR #7503, opened 2024-05-20
refactor: drop TuiSidebarModule

splincode:


projects/addon-mobile/directives/sidebar/sidebar.module.ts
```diff
@@ -1,13 +1,10 @@
...
 import {TuiSidebarDirective} from './sidebar.directive';
 
 @NgModule({
-    imports: [TuiActiveZoneDirective, PolymorpheusModule],
-    declarations: [TuiSidebarDirective, TuiSidebarComponent],
+    imports: [TuiSidebarDirective, TuiSidebarComponent],
     exports: [TuiSidebarDirective],
 })
-export class TuiSidebarModule {}
+export class TuiSidebar {}
```
waterplea, line 10: There's no need for module at all, component is created dynamically, no need to import it.

projects/addon-doc/components/internal/header/header.module.ts
```diff
@@ -13,7 +12,6 @@ import {TuiDocHeaderComponent} from './header.component';
         CommonModule,
         PolymorpheusModule,
         TuiButtonDirective,
-        TuiSidebarModule,
```
nsbarsukov, line 16: Why ?
https://github.com/taiga-family/taiga-ui/blob/c55dfbf2f361c8a397fea6cce74edf22ff26ad83/projects/addon-doc/components/internal/header/header.template.html#L12

https://github.com/taiga-family/taiga-ui/assets/35179038/cddc4400-d460-48d5-b646-cc806c76f89b

projects/cdk/schematics/ng-update/v4/steps/constants/identifiers-to-replace.ts
```diff
@@ -967,7 +967,7 @@ export const IDENTIFIERS_TO_REPLACE: ReplacementIdentifierMulti[] = [
             moduleSpecifier: '@taiga-ui/addon-mobile',
         },
         to: {
-            name: 'TuiTabBar',
+            name: 'TuiSidebarDirective',
```
nsbarsukov, line 970: image

projects/addon-doc/components/internal/header/header.module.ts
splincode: fixed

projects/cdk/schematics/ng-update/v4/steps/constants/identifiers-to-replace.ts
splincode: fixed

mdlufy (approved):

PR #7510, opened 2024-05-21
feat(addon-commerce): update currencies

### LINKED ISSUE #6896: 🚀 - Add the missing currency codes
### Description

ISO 4217 contains many currency codes that are not listed in the enum: https://github.com/taiga-family/taiga-ui/blob/main/projects/addon-commerce/enums/currency-code.ts

Already, we are lacking the following currencies: 048, 203, 348, 498, 792, 951, 985

AndreiBelokopytov:
Closes #6896 

Added the following currency codes mentioned in the issue: 048, 203, 348, 498, 951, 985.

Code 742 in PR is ommited because it belongs to the old Turkish Lira which is now obsolete and its name can be confused with the new Turkish Lira, which has been officially called "Turkish lira" since 2009.


splincode (approved):

splincode:
@AndreiBelokopytov Could you add also cherry-pick to v3.x, please

waterplea (approved):

PR #7538, opened 2024-05-24
refactor(kit): dump legacy controls to legacy package

waterplea:
Closes #


splincode (approved):

projects/demo/src/modules/components/input/examples/import/import.md
```diff
@@ -1,6 +1,6 @@
 ```ts
 import {ReactiveFormsModule} from '@angular/forms';
-import {TuiInputModule} from '@taiga-ui/kit';
```
vladimirpotekhin, line 3: nit: we need to change the chip in the demos as well 
Снимок экрана 2024-05-24 в 21 35 15

vladimirpotekhin (approved):

projects/demo/src/modules/components/input/examples/import/import.md
waterplea: I don't want to do it because we will eventually replace all those controls in kit with updated versions, presumably before 4.0. I only updated import instructions so I don't miss any imports :)

PR #7556, opened 2024-05-27
refactor!: `TableBars` move to `kit` package, refactor

vladimirpotekhin:
Closes #


splincode (approved):

mdlufy (approved):

projects/demo/src/modules/app/app.template.html
```diff
@@ -62,6 +62,6 @@
         </div>
         <tui-sheets-host ngProjectAs="tuiOverContent" />
         <custom-host ngProjectAs="tuiOverContent" />
-        <tui-table-bars-host ngProjectAs="tuiOverAlerts" />
+        <tui-actions-bars-host ngProjectAs="tuiOverAlerts" />
```
waterplea, line 65: Why do we need a host? Why not just use portals?

waterplea (approved):

PR #7573, opened 2024-05-29
refactor(kit): final iteration

waterplea:
Closes #


splincode (approved):

vladimirpotekhin (approved):

projects/legacy/utils/date-mode-maskito-adapter.ts
nsbarsukov, line 1: Why is it deprecated ?

Will next generation of `InputDate` use the date format similar to `Maskito`'s one ?
https://github.com/taiga-family/maskito/blob/a7682bfad40b6aaefc3a6a396b0af6d22eb01ff1/projects/kit/src/lib/types/date-mode.ts#L1-L8

nsbarsukov (approved):

projects/legacy/utils/date-mode-maskito-adapter.ts
waterplea: Probably. I just wanted to get everything cleaned up so that there is no entities in kit package that we do not use at the moment.

PR #7580, opened 2024-05-30
refactor(kit): `InputPhoneInternational` uses `@maskito/phone`

### LINKED ISSUE #917: 🚀 - InputPhoneInternational component with phone number & type validation

core, i18n

## Problem 1. Validation and trust of data used in Taiga
It's not specified what source is used for internalization and phone masks in Taiga and if it can be trusted. According to my tests I faced an issue with incorrect phone mask #863. I tried to make a little PR #868 to fix that. But later I found much more inconsistencies with other countries, too.  Besides, there are issues with invalid ISO codes,  #915, #916.

## Problem 2. Phone mask and validation
It's not the best approach to use the mask constants as the are now. The reason is that in some countries the phone number length is different for mobiles and landline phones. Also, some countries have strict rules about the phone numbers, for example, mobile phones can start only with digit 5, and landlines with digit 4. All other prefixes are invalid. Therefore, even if the input is correct in term of length, it's still invalid phone number.

## How to solve that?
I think Taiga should use a trusted source of data for phone number masks & validation, and internalization, i.e. include and use some other package that keeps updated when ISO codes change, etc.

## My solution
I recommend to use `libphonenumber-js`:
- https://www.npmjs.com/package/libphonenumber-js
- https://gitlab.com/catamphetamine/libphonenumber-js

Because:
- it covers all the officially assigned country codes and phone numbers
- regularly updated 
- based on official Google lib, which adds more trust that the data is valid
- perfectly documented
- provides phone masks
- provides sample phone numbers for each country and type (mobile, landline)
- detects phone number type (mobile, landline, toll free, etc.)
- provides a limited phone number validation (by length) and full validation (that this number is really confirms the phone number rules of the specified country)

## Demo
I modified the InputPhoneInternational component to work with `libphonenumber-js`. The modification includes:
- removed deprecated `country-iso-codes.ts`, instead used  `import {TuiCountryIsoCode} from '@taiga-ui/i18n'`
- removed `countries.ts` file with mask constants, used masks generated by `libphonenumber-js`
- displays additional info as the user types in. The following is displayed in `tui-form__field-note` below the input field:
    - check if phone number is valid
    - check if phone number length is valid and how to fix that
    - show phone type.
    - show sample phone number

GIF demo:

phoneinput

I think there's a plenty room for discussion here. I'm eager to make a PR.

nsbarsukov:
Closes #917


projects/kit/components/input-phone-international/input-phone-international.style.less
nsbarsukov, line 1: The content is identical to already existing file

[`projects/legacy/components/input-phone-international/input-phone-international.style.less`](https://github.com/taiga-family/taiga-ui/blob/main/projects/legacy/components/input-phone-international/input-phone-international.style.less)

Nothing is changed here!

projects/kit/components/input-phone-international/input-phone-international.options.ts
```diff
@@ -0,0 +1,38 @@
...
+    readonly countryIsoCode: TuiCountryIsoCode;
+    readonly metadata: Observable<MetadataJson>;
+}
+
+export const TUI_INPUT_PHONE_INTERNATIONAL_DEFAULT_OPTIONS: TuiInputPhoneInternationalOptions =
+    {
+        countries: [],
+        countryIsoCode: 'RU',
+        metadata: defer(async () =>
+            import('libphonenumber-js/max/metadata').then(m => m.default),
```
nsbarsukov, line 19: New options are almost identical to the previous [one](https://github.com/taiga-family/taiga-ui/blob/main/projects/legacy/components/input-phone-international/input-phone-international.options.ts).

`metadata` is the only addition to it.

projects/kit/components/input-phone-international/input-phone-international.style.less
nsbarsukov, line 1: This file is copy-pasted from already existing file

[`projects/legacy/components/input-phone-international/input-phone-international.style.less`](https://github.com/taiga-family/taiga-ui/blob/main/projects/legacy/components/input-phone-international/input-phone-international.style.less)

Only `t-country-select` and `t-flag` classes was modified a little bit.

splincode (approved):

projects/kit/components/input-phone-international/input-phone-international.component.ts
```diff
@@ -0,0 +1,252 @@
...
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    providers: [
+        tuiAsFocusableItemAccessor(TuiInputPhoneInternationalComponent),
+        tuiAsControl(TuiInputPhoneInternationalComponent),
+        tuiTextfieldOptionsProvider({cleaner: false}),
+    ],
+    host: {
+        '[attr.data-size]': 'textfieldSize.size',
+    },
+    viewProviders: [
```
waterplea, line 89: Do we need to isolate it? Sounds like we can just move it to providers?

projects/kit/components/input-phone-international/input-phone-international.template.html
```diff
@@ -0,0 +1,74 @@
+<ng-container *tuiLet="options.metadata | async as phoneMetadata">
+    <div
+        *ngIf="countriesNames$ | async as countriesNames"
+        tuiGroup
```
waterplea, line 4: Do we need this wrapper? Maybe we can move all this to host directives to minimize unnecessary DOM nesting? We probably need to come up with some infrastructure for binding values to host directives, we can probably achieve it with signals and effects, I'll think about it, you can keep it as is for now.

projects/kit/components/input-phone-international/input-phone-international.template.html
```diff
@@ -0,0 +1,74 @@
...
+                <img
+                    *ngIf="countryIsoCode$ | async as countryIsoCode"
+                    class="t-flag"
+                    [alt]="countriesNames[countryIsoCode]"
+                    [src]="countryIsoCode | tuiFlag"
+                />
+            </ng-template>
+        </tui-textfield>
+
+        <tui-primitive-textfield
```
waterplea, line 31: Why not new textfield, since you already got familiar enough with it to use it for country selection? :D

projects/kit/components/input-phone-international/input-phone-international.style.less
```diff
@@ -0,0 +1,60 @@
...
+
+    &._disabled {
+        pointer-events: none;
+    }
+}
+
+.t-country-select {
+    width: 3.625rem;
+
+    & select:not(:disabled):not(._readonly) {
```
waterplea, line 17: Wouldn't disabled be covered by pointer-events: none on the host? Also sounds like point should be built in from the tui-textfield side here, every control based on select would require that.

projects/kit/components/input-phone-international/input-phone-international.style.less
```diff
@@ -0,0 +1,60 @@
...
+
+    &[data-size='m'] {
+        width: 5.5rem;
+    }
+
+    &[data-size='s'] {
+        width: 2rem;
+
+        .t-flag {
+            margin-left: -@space * 4;
```
waterplea, line 29: Let's ditch this variable

projects/kit/components/input-phone-international/input-phone-international.style.less
```diff
@@ -0,0 +1,60 @@
+@import '@taiga-ui/core/styles/taiga-ui-local';
+
+@flag-width: 1.75rem;
+@flag-height: 1.25rem;
```
waterplea, line 4: Not a fan of those variables either, this is not something we will change

projects/kit/components/input-phone-international/input-phone-international.style.less
```diff
@@ -0,0 +1,60 @@
...
+    .fullsize(absolute, inset);
+
+    left: 0.5rem;
+    width: @flag-width;
+    height: @flag-height;
+    margin: auto;
+}
+
+.t-country-item-flag {
+    width: @flag-width;
```
waterplea, line 48: We can get rid of duplication if we call it .t-flag as well and current .t-flag with some postfix like .t-flag_select

projects/kit/components/input-phone-international/input-phone-international.style.less
```diff
@@ -0,0 +1,60 @@
...
+    height: @flag-height;
+}
+
+.t-country-item-name {
+    margin-left: @space * 3;
```
waterplea, line 53: Let's switch to margin shorthand instead of 2 lines

projects/kit/components/input-phone-international/input-phone-international.style.less
```diff
@@ -0,0 +1,60 @@
...
+    margin-right: auto;
+}
+
+.t-country-item-code {
+    color: var(--tui-text-02);
+    margin-right: @space;
```
waterplea, line 59: As a general rule of thumb we should try avoiding -left and -right for margin, padding and border and use -inline-start and -inline-end to be better prepared for RTL languages. It is now supported by all our browsers.

projects/kit/components/input-phone-international/input-phone-international.component.ts
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    public get nativeFocusableElement(): HTMLElement | null {
+        return this.inputPhone && !this.computedDisabled
+            ? this.inputPhone.nativeFocusableElement
+            : null;
+    }
+
+    public get focused(): boolean {
+        return Boolean(this.countrySelect?.focused || this.inputPhone?.focused);
```
waterplea, line 137: You should also include this.open, focus can be in the dropdown

projects/kit/components/input-phone-international/input-phone-international.component.ts
```diff
@@ -0,0 +1,252 @@
...
+            .pipe(
+                map(args => this.computeMask(...args)),
+                takeUntilDestroyed(),
+            )
+            .subscribe(this.mask$);
+    }
+
+    public override setDisabledState(): void {
+        super.setDisabledState();
+        this.close();
```
waterplea, line 183: I think we can get rid of this by adding `!computedDisabled` to the [tuiDropdownOpen], or maybe come up with some universal for this later, as we will need it in every control with a dropdown.

projects/kit/components/input-phone-international/input-phone-international.component.ts
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    protected onActiveZone(active: boolean): void {
+        this.updateFocused(active);
+    }
+
+    protected onFocus(phonesMetadata: MetadataJson): void {
+        const countryCallingCode = `${CHAR_PLUS + getCountryCallingCode(this.countryIsoCode$.value, phonesMetadata)} `;
+
+        if (!this.textfieldValue) {
```
waterplea, line 204: Why only when its empty?

projects/kit/components/input-phone-international/input-phone-international.component.ts
nsbarsukov: Otherwise, any `tuiDropdown*`-controller will override `limitWidth` to `auto`.

For example:

```html
<tui-input-phone-international
    tuiDropdownDirection="top"
/>
```

projects/legacy/index.ts
```diff
@@ -2,4 +2,5 @@ export * from '@taiga-ui/legacy/classes';
 export * from '@taiga-ui/legacy/components';
 export * from '@taiga-ui/legacy/directives';
 export * from '@taiga-ui/legacy/pipes';
+export * from '@taiga-ui/legacy/tokens';
```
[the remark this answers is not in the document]
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/7580/commits/7c8b8796dde969cc991e53ae8ed2f1a503627f55

projects/kit/components/input-phone-international/input-phone-international.style.less
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/7580/commits/18f5fea3632817d0f77b2408751721d12509218f

projects/kit/components/input-phone-international/input-phone-international.style.less
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/7580/commits/18f5fea3632817d0f77b2408751721d12509218f

projects/kit/components/input-phone-international/input-phone-international.component.ts
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/7580/commits/006c66da85e757dc13595046e600d6255e5873c1

projects/kit/components/input-phone-international/input-phone-international.style.less
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/7580/commits/0b2f448326e9c9cf5a33becb91f000f7fba50525

projects/kit/components/input-phone-international/input-phone-international.style.less
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/7580/commits/0b2f448326e9c9cf5a33becb91f000f7fba50525

projects/kit/components/input-phone-international/input-phone-international.style.less
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/7580/commits/0b2f448326e9c9cf5a33becb91f000f7fba50525

projects/kit/components/input-phone-international/input-phone-international.component.ts
nsbarsukov: In all other cases country calling code already exists (no need to add it).

projects/kit/components/input-phone-international/input-phone-international.style.less
nsbarsukov: Drop this line in https://github.com/taiga-family/taiga-ui/pull/7580/commits/fd43ede112379c0cd1e908839204fdd7912098ec + open the following PR:

* https://github.com/taiga-family/taiga-ui/pull/7603

projects/kit/components/input-phone-international/input-phone-international.template.html
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/7580/commits/314477590a0bd9455a5719cbd3db2ea2cb6ca916

projects/demo/src/modules/components/input-phone-international/examples/2/index.ts
```diff
@@ -4,11 +4,8 @@ import {FormControl, FormGroup, ReactiveFormsModule, Validators} from '@angular/
 import {changeDetection} from '@demo/emulate/change-detection';
 import {encapsulation} from '@demo/emulate/encapsulation';
 import type {TuiCountryIsoCode} from '@taiga-ui/i18n';
-import {TuiSortCountriesPipe} from '@taiga-ui/kit';
-import {
-    TUI_ALL_COUNTRIES_ISO_CODES,
-    TuiInputPhoneInternationalComponent,
-} from '@taiga-ui/legacy';
+import {TuiInputPhoneInternationalComponent, TuiSortCountriesPipe} from '@taiga-ui/kit';
+import {TUI_ALL_COUNTRIES_ISO_CODES} from '@taiga-ui/legacy';
```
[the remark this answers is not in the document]
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/7580/commits/4641e12031e73edd820a36357a52dce15667d6b8

projects/kit/components/input-phone-international/input-phone-international.options.ts
vladimirpotekhin: Maybe it is better to set default metadata `libphonenumber-js/min` ? I think it should cover 90% of use cases. 
(and show in "setup" page or in some example that we can provide our own metadata)

nsbarsukov: > Maybe it is better to set default metadata libphonenumber-js/min ?

Fixed in https://github.com/taiga-family/taiga-ui/pull/7580/commits/a81ff3547f6258624a0c5836b8942548ecdd2ee9


> and show in "setup" page or in some example that we can provide our own metadata

I will rebuild the whole documentation page of `PhoneInternational` in a separate PR later.

projects/kit/components/input-phone-international/input-phone-international.template.html
waterplea: I have made that infrastructure, let's put a TODO here to get rid of nesting as much as possible.

projects/kit/components/input-phone-international/input-phone-international.style.less
```diff
@@ -0,0 +1,50 @@
...
+        margin: auto;
+    }
+
+    :host(._disabled) & {
+        opacity: var(--tui-disabled-opacity);
+    }
+}
+
+.t-country-item-name {
+    margin-inline: 0.75rem auto;
```
waterplea, line 44: Not supported by Safari in a shorthand form:
https://caniuse.com/?search=margin-inline

projects/kit/components/input-phone-international/input-phone-international.template.html
```diff
@@ -0,0 +1,74 @@
...
+        *ngIf="countriesNames$ | async as countriesNames"
+        tuiGroup
+        [tuiDropdown]="dropdown"
+        [tuiDropdownEnabled]="!readOnly"
+        [(tuiDropdownOpen)]="!computedDisabled && open"
+        (tuiActiveZoneChange)="onActiveZone($event)"
+    >
+        <tui-textfield
+            tuiChevron
+            class="t-country-select tui-group__auto-width-item"
```
waterplea, line 12: I would ideally drop those group classes someday in the future. Would be nice if we didn't use them here if possible, since we have local classes on both of those controls.

projects/kit/components/input-phone-international/input-phone-international.component.ts
```diff
@@ -0,0 +1,246 @@
...
+export class TuiInputPhoneInternationalComponent
+    extends AbstractTuiControl<string>
+    implements TuiFocusableElementAccessor
+{
+    /* eslint-disable @typescript-eslint/member-ordering */
+    protected readonly options = inject(TUI_INPUT_PHONE_INTERNATIONAL_OPTIONS);
+    protected readonly countryIsoCode$ = new BehaviorSubject(this.options.countryIsoCode);
+    protected readonly mask$ = new BehaviorSubject<MaskitoOptions | null>(null);
+    protected readonly countriesNames$ = inject(TUI_COUNTRIES);
+    protected readonly textfieldOptions = inject(TuiTextfieldOptionsDirective, {
```
waterplea, line 98: I think there's a token `TUI_TEXTFIELD_OPTIONS` instead of the optional directive?

projects/kit/components/input-phone-international/input-phone-international.component.ts
```diff
@@ -0,0 +1,246 @@
...
+        super();
+        combineLatest([this.countryIsoCode$, this.options.metadata])
+            .pipe(
+                map(args => this.computeMask(...args)),
+                takeUntilDestroyed(),
+            )
+            .subscribe(this.mask$);
+    }
+
+    public override setDisabledState(): void {
```
waterplea, line 180: Unnecessary

projects/kit/components/input-phone-international/input-phone-international.component.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    public override writeValue(unmaskedValue: string): void {
+        super.writeValue(unmaskedValue);
+
+        const maskOptions = this.mask$.value;
+
+        this.textfieldValue = maskOptions
+            ? maskitoTransform(unmaskedValue, maskOptions)
+            : unmaskedValue; // it will be calibrated later when mask is ready (by maskitoInitialCalibrationPlugin)
+        this.cdr.detectChanges();
```
waterplea, line 192: Do we really need it and not markForCheck?

projects/kit/components/input-phone-international/input-phone-international.component.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    public onItemClick(isoCode: TuiCountryIsoCode): void {
+        this.open = false;
+        this.countryIsoCode$.next(isoCode);
+        this.nativeFocusableElement?.focus();
+    }
+
+    constructor() {
+        super();
+        combineLatest([this.countryIsoCode$, this.options.metadata])
```
waterplea, line 172: This looks like it can be optimized through signals. Are we sure there's no double load of metadata with 2 subscriptions? You can keep it like that for now, maybe I'll take a stab at it later when I will be moving directives to the host and refactoring it to use new signal control.

projects/kit/components/input-phone-international/input-phone-international.style.less
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/7580/commits/a6e889c5c2109ef32eb045d9f560acacc8c584aa

projects/kit/components/input-phone-international/input-phone-international.template.html
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/7580/commits/022347a56c0e1f04052fad24885b1a58e84b24d9

projects/kit/components/input-phone-international/input-phone-international.component.ts
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/7580/commits/814e19734dec70b4e86c6214d264dd9bbf580b79

projects/kit/components/input-phone-international/input-phone-international.component.ts
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/7580/commits/f7004ea0a626d6ab885468cb904fb8ccea8d14bb

projects/kit/components/input-phone-international/input-phone-international.component.ts
nsbarsukov: Does `formControl.patchValue` trigger tick of application ? 🤔

nsbarsukov: I've try to replace it by `markForCheck`.
This test fails:

https://github.com/taiga-family/taiga-ui/blob/c8a24a8d70ade7674f83222f1aa8239af1055a6f/projects/demo-cypress/src/tests/input-phone-international.cy.ts#L227-L231

waterplea: Ok then, let's keep it like this for now

waterplea (approved):

projects/kit/components/input-phone-international/input-phone-international.template.html
nsbarsukov: > let's put a TODO here to get rid of nesting as much as possible

Done in https://github.com/taiga-family/taiga-ui/pull/7580/commits/a4614994750d48b5ccf395520ac5f7a59838ce22

projects/kit/components/input-phone-international/input-phone-international.component.ts
nsbarsukov: I'll rewrite code (using signals) in next PR

PR #7596, opened 2024-05-31
feat(kit): emails pipe from proprietary

splincode:


projects/cdk/pipes/emails/emails.options.ts
```diff
@@ -0,0 +1,18 @@
+import {tuiCreateToken} from '@taiga-ui/cdk/utils/miscellaneous';
+
+export const TUI_EMAIL_PIPE_DEFAULT_OPTIONS = [
+    'mail.ru',
```
waterplea, line 4: Let's use international emails by default in public and override them in proprietary providers.

splincode: fixed

projects/cdk/pipes/emails/emails.options.ts
```diff
@@ -0,0 +1,22 @@
+import {tuiCreateToken} from '@taiga-ui/cdk/utils/miscellaneous';
+
+export const TUI_EMAIL_PIPE_OPTIONS = tuiCreateToken([
+    'me.com',
```
waterplea, line 4: nit: Would be great to sort them by popularity, pretty sure gmail should go first :)

waterplea (approved):

projects/cdk/pipes/emails/emails.options.ts
waterplea: Also let's keep the list a bit shorter, only showing most popular options, so that there is no scroll in the dropdown. Lists UX best practice is to keep it below 7 items.

PR #7606, opened 2024-06-03
refactor(core): `Textfield` should not shift all content if empty `label[tuiLabel]` is passed

nsbarsukov:
Imagine that you develop any custom input using new `Textfield`.

**your-component.template.html:**
```html
<tui-textfield>
    <input tuiTextfield />

    <label tuiLabel><ng-content /></label>
</tui-textfield>
```

Public API of your component:
```html
<your-custom-input>
    <strong>Awesome label</strong>
</your-custom-input>

<your-custom-input/>
```


splincode (approved):

waterplea (approved):
I wish we could just use `:has([tuiLabel]:not(:empty)` :)

PR #7615, opened 2024-06-04
chore: include inputs/outputs of host directive `ActiveZone` in `DropdownOpen`'s API

nsbarsukov:
```ts
@Component({
    standalone: true,
    imports: [TuiButtonDirective, TuiActiveZoneDirective, TuiDropdownModule],
    template: `
        <button
            tuiButton
            type="button"
            [tuiDropdown]="'ANY CONTENT DROPDOWN'"
            [(tuiDropdownOpen)]="open"
            (tuiActiveZoneChange)="(0)"
        >
            Button
        </button>
    `,
})
export default class ExampleComponent {
    protected open = false;
}
```

It will throw:
```shell
Error: NG0309:
Directive TuiActiveZoneDirective matches multiple times on the same element.
Directives can only match an element once.
```

### DIFF
--- a/projects/core/directives/dropdown/dropdown-open.directive.ts
+++ b/projects/core/directives/dropdown/dropdown-open.directive.ts
@@ -49,7 +49,14 @@ function shouldClose(
 @Directive({
     standalone: true,
     selector: '[tuiDropdownOpen],[tuiDropdownOpenChange]',
-    hostDirectives: [TuiObscuredDirective, TuiActiveZoneDirective],
+    hostDirectives: [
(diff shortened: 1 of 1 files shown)


splincode (approved):

waterplea (approved):
What a stupid error. Basically, it forces us to add all inputs/outputs to ANY host directive that might be applied externally. I'm really disappointed with how Angular team handled Directive Composition API. Almost as if they just heard we might need this thing and haven't put much thought into how it could actually be used and what it should be able to achieve.

PR #7631, opened 2024-06-04
chore: move to standalone

splincode:


projects/cdk/directives/control/control.directive.ts
```diff
@@ -10,7 +10,7 @@ import {NgControl} from '@angular/forms';
 export class TuiControlDirective {
     private readonly ngControl = inject(NgControl);
 
-    protected get control(): AbstractControl {
+    public get control(): AbstractControl {
```
vladimirpotekhin, line 13: why public ?

splincode: That is, most likely it was assumed that this directive could be injected and access the control

projects/kit/components/accordion/test/accordion.component.spec.ts
```diff
@@ -40,6 +39,14 @@ describe('Accordion', () => {
     let testComponent: TestComponent;
 
     @Component({
+        standalone: true,
+        imports: [
+            TuiAccordion,
+            ReactiveFormsModule,
+            TuiTextfieldControllerModule,
```
nsbarsukov, line 46: Could it be dropped ?

projects/kit/components/accordion/test/accordion.component.spec.ts
```diff
@@ -40,6 +39,14 @@ describe('Accordion', () => {
...
+            TuiDataList,
+            TuiDataListWrapper,
```
nsbarsukov, line 48: Could we drop them to ?

nsbarsukov (approved):

projects/kit/components/accordion/test/accordion.component.spec.ts
splincode: fixed

projects/kit/components/accordion/test/accordion.component.spec.ts
splincode: fixed

vladimirpotekhin (approved):

PR #7658, opened 2024-06-06
refactor!: use `tui-icon` instead `tui-svg`

mdlufy:
Closes #


projects/addon-table/components/table-pagination/table-pagination.style.less
```diff
@@ -22,7 +22,9 @@
 
 .t-checkmark {
     min-width: 1.5rem;
-    border-left: 5px solid transparent;
+    margin-inline-start: 5px;
```
waterplea, line 25: That's weird, should probably be 0.25rem. I think we should also have `margin-inline-end: -0.125rem` so the icon is equally far from item edges:
image

projects/addon-table/components/table/th/th.style.less
```diff
@@ -124,6 +124,10 @@
     &:hover {
         color: var(--tui-text-01);
     }
+
+    &-icon {
```
waterplea, line 128: Let's just rename it `.t-icon`, no need for it to be nested. Strange that we had that class in HTML with no styles applied to it before.

projects/core/components/tooltip/tooltip.component.ts
```diff
@@ -34,6 +34,11 @@ export class TuiTooltipComponent<C = any> extends TuiHintOptionsDirective {
         return this.appearance || '';
     }
 
+    @HostBinding('class._small')
+    protected get small(): boolean {
```
waterplea, line 38: Let's drop that, we should not depend on icon names, leave it always small for now.

waterplea (approved):

waterplea (approved):

projects/core/components/tooltip/tooltip.component.ts
mdlufy: droped

projects/addon-table/components/table/th/th.style.less
mdlufy: renamed

waterplea (approved):

vladimirpotekhin (approved):

PR #7660, opened 2024-06-06
refactor(core): `PrimitiveTextfield` and `HostedDropdown` move to legacy

waterplea:
Closes #


splincode (approved):

projects/addon-doc/components/documentation/documentation.template.html
```diff
@@ -168,22 +143,22 @@
...
-                                    [nativeId]="propertyConnector.attrName"
-                                    [tuiTextfieldCleaner]="true"
-                                    [tuiTextfieldLabelOutside]="true"
-                                    [value]="propertyConnector.documentationPropertyValue.toString()"
-                                    (valueChange)="propertyConnector.onValueChange($event)"
-                                >
-                                    Default content
-                                </tui-primitive-textfield>
-                            </ng-container>
+                            <!--                            <ng-container *ngSwitchCase="'PolymorpheusContent'">-->
```
vladimirpotekhin, line 146: should it be removed or uncommented?

waterplea: I don't know yet, probably removed. I didn't find any usage due to the nested ngIf never being true. I didn't look to closely though, but anyway, we can probably drop it and later improve docs at the spot if we find a need to do so. I'll drop it in my next PR.

vladimirpotekhin (approved):

Issue #7073, opened 2024-03-22
🐞 - InputCopy returns incorrect copy hint


SeregaZnam:
### Description

If you click on the copy button before the tooltip appears, the tooltip will show up after the click displaying the text "Copy", seemingly prompting the user to click it a second time.

https://github.com/taiga-family/taiga-ui/assets/18699867/7e75ba2a-ea62-4b38-b4bf-904db06a93ed

### Angular version

15.2.0

### Taiga UI version

3.71.0

### Which browsers have you used?

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

vladimirpotekhin:
@chipaza will take it

PR #7674, opened 2024-06-07
fix(kit): fix incorrect copy hint in inputCopy

chipaza:
Closes [#7073](https://github.com/taiga-family/taiga-ui/issues/7073)


projects/kit/components/input-copy/input-copy.component.ts
```diff
@@ -87,14 +87,19 @@ export class TuiInputCopyComponent
...
-                this.copy$.pipe(
-                    switchMap(() =>
-                        merge(
+                this.copied$.pipe(
+                    switchMap(copied => {
+                        if (!copied) {
+                            return of(texts[0]);
+                        }
+
+                        this.copied$.next(false);
```
waterplea, line 96: This feels like a pretzel. Ideally it would just be a single stream without shooting itself back :) We gotta refactor that in 4.0, as for 3.0 we can probably merge this.

waterplea (approved):

splincode (approved):

PR #7701, opened 2024-06-10
chore(demo): add `API` section for documentation page `Icon`

nsbarsukov:
### [Example](https://taiga-ui--pr7701-icons-api-page-iq7aplku.web.app/components/icon/API?icon=tuiIconInfoLarge&background=tuiIconInfoLarge&style.color=red&style.background=%233aa981&style.height.px=96)

### DIFF
--- a/projects/demo/src/modules/components/icon/index.ts
+++ b/projects/demo/src/modules/components/icon/index.ts
@@ -17,12 +17,12 @@ import {tuiInputNumberOptionsProvider} from '@taiga-ui/legacy';
     ],
 })
 export default class PageComponent {
-    protected readonly iconVariants = TUI_USED_ICONS;
-    protected readonly colorVariants = ['var(--tui-text-01)', 'red', '#3aa981'];
+    protected readonly iconVariants = ['', ...TUI_USED_ICONS];
+    protected readonly colorVariants = ['', 'var(--tui-text-01)', 'red', '#3aa981'];
 
     protected icon = 'tuiIconHeartLarge';
     protected backgroundIcon = '';
-    protected color = this.colorVariants[0];
-    protected backgroundColor = this.colorVariants[0];
+    protected color = this.colorVariants[1];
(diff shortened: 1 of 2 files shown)


splincode (approved):

projects/demo/src/modules/components/icon/index.ts
```diff
@@ -1,12 +1,28 @@
...
     changeDetection,
+    providers: [
+        tuiInputNumberOptionsProvider({
+            min: 0,
+        }),
+    ],
 })
-export default class PageComponent {}
+export default class PageComponent {
+    protected readonly iconVariants = TUI_USED_ICONS;
```
vladimirpotekhin, line 20: can you please add empty variant? like `''`

vladimirpotekhin (approved):

projects/demo/src/modules/components/icon/index.ts
```diff
@@ -1,12 +1,28 @@
...
+    protected readonly colorVariants = ['var(--tui-text-01)', 'red', '#3aa981'];
+
+    protected icon = 'tuiIconHeartLarge';
+    protected backgroundIcon = '';
+    protected color = this.colorVariants[0];
+    protected backgroundColor = this.colorVariants[0];
```
vladimirpotekhin, line 26: seems like it's better to use different color for background, otherwise there is a black box when you visit the page first time

nsbarsukov: My mistake, thanks)

projects/demo/src/modules/components/icon/index.ts
nsbarsukov: ✅

projects/demo/src/modules/components/icon/index.ts
nsbarsukov: ✅

PR #7731, opened 2024-06-13
refactor!: use new icons approach

vladimirpotekhin:
Closes #


splincode (approved):

projects/demo/src/modules/components/icon/examples/3/index.html
```diff
@@ -1,6 +1,7 @@
 <tui-icon
-    background="tuiIconInfoLarge"
-    icon="tuiIconInfoLarge"
+    background="tuiIconInfo
```
waterplea, line 2: Something's wrong here

projects/demo/src/modules/components/icons/icons.tokens.ts
```diff
@@ -2,9 +2,9 @@ import {inject} from '@angular/core';
 import type {TuiStringHandler} from '@taiga-ui/cdk';
 import {tuiCreateTokenFromFactory} from '@taiga-ui/cdk';
 import {TUI_SVG_OPTIONS} from '@taiga-ui/core';
-import * as allIcons from '@taiga-ui/icons';
+import {TUI_ALL_ICONS} from '@taiga-ui/icons';
```
waterplea, line 5: Should we still compile a typescript dictionary of all the icons? Doesn't seem like it makes sense now that you cannot pass source code to the icon component. Preloading icons in bundle would work differently now.

projects/icons/scripts/prepare-lucide-icons.ts
```diff
@@ -0,0 +1,24 @@
...
+        const content = fs.readFileSync(path.join(src, filename), 'utf-8');
+
+        fs.writeFileSync(
+            path.join(dest, processName(filename)),
+            content.replace('stroke-width="2"', 'stroke-width="3"'),
+        );
+    });
+})();
+
+function processName(name: string, postfix = ''): string {
```
waterplea, line 18: Just a thought for future: I think we should minimize icon processing, probably in the next PR — we can keep the names as they are. I also wonder if it's possible to created filled version of the icons from lucide, would setting stroke to transparent and fill to current color do the trick?

projects/demo/src/modules/components/icons/icons.tokens.ts
waterplea: I dropped it in my PR, I don't think if we maintain name equivalence we need a separate page to list all the 1500 icons.

waterplea: That's why I also want to drop file/name processing and just write it like `iconLeft="@tui.align-horizontal-distribute-start"` so you can just copy icon name from Lucide docs.

projects/icons/scripts/prepare-lucide-icons.ts
vladimirpotekhin: > I also wonder if it's possible to created filled version of the icons from lucide, would setting stroke to transparent and fill to current color do the trick?

Seems like it works (example with apple icon). But why do we need to create filled versions? This is a little contrary to minimizing preprocessing

Снимок экрана 2024-06-14 в 10 27 23
Снимок экрана 2024-06-14 в 10 27 48

waterplea (approved):

projects/icons/scripts/prepare-lucide-icons.ts
vladimirpotekhin: I will add filled icons in next PR

PR #7779, opened 2024-06-17
Sync pipes

MillerSvt:


projects/cdk/pipes/observable/observable.pipe.ts
```diff
@@ -0,0 +1,53 @@
...
+ *         map(i => value - i)
+ *         takeWhile(i => i >= 0, true),
+ *      );
+ *
+ *      return this.toSignal(countdown$, value);
+ *   }
+ * }
+ * ```
+ */
+// eslint-disable-next-line @taiga-ui/experience/no-private-esnext-fields
```
MillerSvt, line 35: I believe it is necessary to use ECMA private fields to prevent any conflicts with subclass definitions.

vladimirpotekhin: unfortunately we can't use private fields now because Safari 13.1 doesn't support it

MillerSvt: It doesn't matter because TypeScript automatically added the necessary polyfills.

MillerSvt: Source:
```
class Foo {
  #bar = 123;

  method() {
    console.log(this.#bar);
  }
}
```

Compiled:
```
"use strict";
var __classPrivateFieldGet = (this && this.__classPrivateFieldGet) || function (receiver, state, kind, f) {
    if (kind === "a" && !f) throw new TypeError("Private accessor was defined without a getter");
    if (typeof state === "function" ? receiver !== state || !f : !state.has(receiver)) throw new TypeError("Cannot read private member from an object whose class did not declare it");
    return kind === "m" ? f : kind === "a" ? f.call(receiver) : f ? f.value : state.get(receiver);
};
var _Foo_bar;
class Foo {
    constructor() {
        _Foo_bar.set(this, 123);
    }
    method() {
        console.log(__classPrivateFieldGet(this, _Foo_bar, "f"));
    }
}
_Foo_bar = new WeakMap();
```

waterplea: We are just 1 Safari major release away from real private fields. I think we can avoid them for now, we lived without them all this time.

waterplea:
Sorry, but let's drop this for now. Those pipes are like that because underlying data sources are observables, like translation tokens or format options — we should probably just switch those to signals and take it from there, maybe some of those pipes won't even make sense anymore and will turn into helper functions that inject things they need under the hood.

PR #7804, opened 2024-06-17
fix(kit): save current active period to distinguish periods with same…

mdlufy:
… date ranges

Closes #


mdlufy:
Problem: on item select, control value hold **only range**, not considering name/content

```
onItemSelect(item: TuiDayRangePeriod | string): void {
        if (typeof item !== 'string') {
            this.updateValue(item.range.dayLimit(this.min, this.max));

            return;
        }

        if (this.activePeriod !== null) {
            this.updateValue(null);
        }
}
```
And when it attemps to get current active period, in finds only by range  
```
private get activePeriod(): TuiDayRangePeriod | null {
        return (
            this.items.find(item =>
                tuiNullableSame<TuiDayRange>(
                    this.value,
                    item.range,
                    (a, b) =>
                        a.from.daySame(b.from.dayLimit(this.min, this.max)) &&
                        a.to.daySame(b.to.dayLimit(this.min, this.max)),
                ),
            ) || null
        );
}
```

Solutions:
1. To hold TuiDayRangePeriod separately
2. To hold TuiDayRangePeriod in control value (breaking change)

splincode (approved):

vladimirpotekhin (approved):

splincode (approved):

PR #7807, opened 2024-06-18
refactor: convert module to const for `TuiTable`

splincode:
Part of https://github.com/taiga-family/taiga-ui/pull/7753


projects/addon-table/components/table/directives/row.directive.ts
```diff
@@ -8,16 +8,17 @@ import type {TuiRowContext} from '@taiga-ui/addon-table/types';
  * TODO v4.0 delete it.
```
waterplea, line 8: Should we delete it now? Looks like we forgot some refactor for Table.

waterplea (approved):

vladimirpotekhin (approved):

PR #7810, opened 2024-06-18
feat(core): `Tooltip` add directive, move component to legacy

waterplea:
Closes #


splincode (approved):

projects/core/directives/tooltip/tooltip.directive.ts
```diff
@@ -0,0 +1,107 @@
...
+    private readonly textfield = inject(TuiTextfieldComponent, {optional: true});
+    private readonly isMobile = inject(TUI_IS_MOBILE);
+    private readonly describe = inject(TuiHintDescribeDirective);
+    private readonly driver = inject(TuiHintHoverDirective);
+
+    protected readonly nothing = tuiWithStyles(TuiTooltipStyles);
+    protected readonly state = bindAppearanceState();
+
+    constructor() {
+        inject(TuiIcon).icon = inject(TuiIcon).icon || inject(TUI_HINT_OPTIONS).icon;
```
demensky, line 75: Would such an operator be suitable here?

```suggestion
        inject(TuiIcon).icon ??= inject(TUI_HINT_OPTIONS).icon;
```

waterplea: It's empty line there by default, so that wouldn't work. I'll refactor it to default icon token so there's no such hack.

nsbarsukov (approved):

projects/core/directives/tooltip/tooltip.directive.ts
waterplea: Added the token

Issue #6723, opened 2024-02-08
🚀 - Use exact requests for internal imports


Markiewic:
### Description

Since when using Module Federation, webpack disables tree-shake, the entire packages are included in the output. For example, if in my code I import `@taiga-ui/core/components/alert`, which itself requests the `@taiga-ui/cdk` package, then both the former package and the entire latter package are included in the output.

I suggest using exact paths to packages everywhere instead of accessing their index files at the first level. This will greatly reduce the chance of the entire packet ending up in the output if it is not required.

I also suggest completely emptying index files so that editors always offer full paths to secondary entry points.

splincode:
breaking changes?

Markiewic:
Yes, but this is just a suggestion, just to think about it. Maybe there is some less radical method.

waterplea:
This is something we will consider for 4.0 only, since after it is established, those paths that previously were internal will become kind of public for Module Federation users. So after we commit to this, changing them will be considered breaking change.

mdlufy:
> I also suggest completely emptying index files so that editors always offer full paths to secondary entry points.

It's can be problem, because if now apps can share modules by `@taiga-ui/core` key module path, with exact imports instead root entry points, then will be forced to use, for example, `@taiga-ui/core/components/alert` path as shared key - and will have to rewrite the code, or caching will not work correctly, or disable sharing to enable tree-shaking

In fact, there may be several cases:
1. not to share modules - tree-shaking working, microfrontends bundle output tree-shakes and some package modules loads twice
2. share root entrypoints, such `@taiga-ui/core` - tree-shaking not working, but microfrontends load whole package from cache and each import from `@taiga-ui/core` not loading twice
3. share deep entrypoints, such `@taiga-ui/core/components/alert` - tree-shaking not working, to share modules correctly, you need to declare each deep entrypoint in shared section of module federation plugin

If I'm not mistaken, when library use exact paths internally, this can only be useful for the third case

PR #7857, opened 2024-06-19
refactor(addon-commerce): use deep entry points for imports of other Taiga packages

nsbarsukov:
Partially solves #6723


projects/addon-commerce/components/input-card-grouped/input-card-grouped.component.ts
```diff
@@ -186,7 +179,7 @@ export class TuiInputCardGroup
     public exampleText = this.options.exampleText;
 
     @Input()
-    public cardValidator = this.options.cardValidator;
+    public cardValidator: TuiBooleanHandler<string> = this.options.cardValidator;
```
nsbarsukov, line 182: Otherwise, it will produces build error `import(../../../../dist/cdk/types)`

splincode (approved):

PR #7878, opened 2024-06-20
refactor(kit): use deep entry points for imports of other Taiga packages

### LINKED ISSUE #6723: 🚀 - Use exact requests for internal imports

nsbarsukov:
Closes #6723


projects/kit/components/rating/rating.component.ts
```diff
@@ -46,7 +45,7 @@ export class TuiRatingComponent extends TuiControl<number> {
     protected active = 0;
 
     @Input()
-    public icon = this.options.icon;
+    public icon: TuiRatingOptions['icon'] = this.options.icon;
```
nsbarsukov, line 48: Otherwise, it will produce build error `import(../../../../dist/cdk/types)`

projects/kit/components/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -32,5 +35,5 @@ export class TuiBreadcrumbsComponent {
 
     @Input()
     @HostBinding('attr.data-size')
-    public size = this.options.size;
+    public size: TuiBreadcrumbsOptions['size'] = this.options.size;
```
nsbarsukov, line 38: Otherwise, it will produce build error `import(../../../../dist/core/types)`

splincode (approved):

mdlufy (approved):

PR #8003, opened 2024-07-01
fix(styles): emoji should be visible on safari

splincode:
## Before

image

## After

image


waterplea (approved):

waterplea:
New line height seems a bit different. Are we sure it's correct? Does it work well with growing height and red background for exceeding max length?

splincode:
@waterplea updated

mdlufy (approved):

PR #8005, opened 2024-07-01
fix(addon-mobile): prevent touchstart after close sheet

### LINKED ISSUE #8002: 🐞 - Elements are clickable under the overlay of TuiDropdownMobile
### Playground Link

_No response_

### Description

When open TuiDropdownMobile as a sheet (setting the "tuiDropdownMobile" param) I expect only dropdown closing after clicking on the overlay. But additionally it triggers click on an elem under the overlay.
Снимок экрана 2024-07-01 в 18 29 00

### Angular version

17

### Taiga UI version

3.86.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
Close #8002

### DIFF
--- a/projects/addon-mobile/directives/dropdown-mobile/dropdown-mobile.template.html
+++ b/projects/addon-mobile/directives/dropdown-mobile/dropdown-mobile.template.html
@@ -3,7 +3,7 @@
     waIntersectionObserver
     waIntersectionThreshold="1"
     class="t-filler"
-    (touchstart.silent)="close()"
+    (touchstart.silent.prevent)="close()"
     (waIntersectionObservee)="onIntersection($event)"
 ></div>
 <div


projects/addon-mobile/directives/dropdown-mobile/dropdown-mobile.template.html
```diff
@@ -3,7 +3,7 @@
     waIntersectionObserver
     waIntersectionThreshold="1"
     class="t-filler"
-    (touchstart.silent)="close()"
+    (touchstart.silent.prevent)="close()"
```
waterplea, line 6: Maybe if we remove `silent` it will fix the change detection issue on iOS?

waterplea (approved):

mdlufy (approved):

PR #8026, opened 2024-07-03
ci: add prepublish target

splincode:


package.json
```diff
@@ -34,7 +34,7 @@
         "start:host": "nx serve demo --open --host 0.0.0.0 --disable-host-check",
         "build:ssr": "nx build demo && nx run demo:server:production",
         "prerender": "nx run demo:prerender",
-        "test": "nx run-many --target test --all --output-style=stream --parallel=1",
```
vladimirpotekhin, line 37: let's keep without nxBail

vladimirpotekhin (approved):

package.json
splincode: fixed

mdlufy (approved):

PR #8069, opened 2024-07-09
chore: fix migrations

vladimirpotekhin:
Closes #

### DIFF
--- a/projects/cdk/schematics/ng-update/v4/steps/constants/identifiers-to-replace.ts
+++ b/projects/cdk/schematics/ng-update/v4/steps/constants/identifiers-to-replace.ts
@@ -734,7 +734,6 @@ export const IDENTIFIERS_TO_REPLACE: ReplacementIdentifierMulti[] = [
         to: {
             name: 'TuiDialog',
             moduleSpecifier: '@taiga-ui/core',
-            spreadInModule: true,
         },
     },
     {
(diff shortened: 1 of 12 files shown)


projects/cdk/schematics/ng-update/v4/steps/constants/identifiers-to-replace.ts
```diff
@@ -734,6 +734,7 @@ export const IDENTIFIERS_TO_REPLACE: ReplacementIdentifierMulti[] = [
         to: {
             name: 'TuiDialog',
             moduleSpecifier: '@taiga-ui/core',
+            spreadInModule: true,
```
waterplea, line 737: This is not going to be needed once we fix `TuiDialog`. It will not be an array.

projects/cdk/schematics/ng-update/v4/steps/constants/identifiers-to-replace.ts
```diff
@@ -1256,6 +1247,26 @@ export const IDENTIFIERS_TO_REPLACE: ReplacementIdentifierMulti[] = [
...
+            moduleSpecifier: '@taiga-ui/addon-preview',
+        },
+        to: {
+            name: 'TuiPreviewDialogService',
+            moduleSpecifier: '@taiga-ui/kit',
+        },
+    },
+    {
+        from: {
+            name: 'PolymorpheusModule',
```
waterplea, line 1262: We also need update imports for `PolymorpheusComponent`.

projects/cdk/schematics/ng-update/v4/steps/constants/identifiers-to-replace.ts
vladimirpotekhin: I removed from dialog. But we have many constants like this besides dialogue, I will add them later in separate pr

mdlufy (approved):

waterplea (approved):

PR #8074, opened 2024-07-10
fix(core): do not update dropdown position on every animation tick

dmlukichev:
Performance fix.

Dropdown component calls update on every emit TuiPositionService (which is every animation frame).


waterplea:
The position might not change but the content inside the dropdown itself might grow and force it to change position. Also I'm not sure how a case with disappearing from DOM would behave here (see isConnected check). I'm not sure this is a good change.

dmlukichev:
Another possible approach is to take into account dropdown element dimensions and `isConnected` value in distinctUntilChange comparator function.

waterplea:
Then we will move all the heaviest part into that check anyway :) I'm not seeing the issue you described, but the only thing we can safely improve here is not setting value on `style` object unless they change. Could you see if that helps?

waterplea:
You can try mapping values to an object of new styles, add `distinctUntilChanged` to test if keys values have changed and then do:
```ts
Object.assign(this.el.style, yourNewObject);
```

dmlukichev:
> You can try mapping values to an object of new styles...

@waterplea, did it like you proposed.

image

Performance boost is tiny in demo component, although already noticeable.
But DevTools is not crashing anymore because of frequent style updates and that is what was bothering me in the first place :)

waterplea:
Great! Seems like you didn't run pre-commit hooks that format the code or you need npm ci to update our linter configs. Also I see now that the width does not get properly applied as per screenshots. This needs to be addressed.

projects/core/directives/dropdown/dropdown.component.ts
```diff
@@ -64,9 +77,11 @@ export class TuiDropdownComponent implements OnInit {
         .closest('[tuiTheme]')
         ?.getAttribute('tuiTheme');
 
+    protected readonly styles$ = new BehaviorSubject<Partial<DynamicStyles>>({});
```
waterplea, line 80: There's no need to introduce another subject and subscription, just instead of `subscribe` with `update` in original subscription you can move `isConnected` check to `tap` (or `takeWhile` + close on complete) and `update` can be changed into `map` so it stays in the same stream and `Object.assign` would then be performed in `next` of it. I also wonder if `ngOnInit` can be changed to `startWith` inside `TuiPositionService` so it immediately reports size upon subscription instead of waiting for the next animation frame.

dmlukichev: Updated it accordingly.
Tried to add `startWith()` into TuiPositionService, but it makes hint behave weird, so kept `ngOnInit()` logic.

waterplea:
Closing in favor of #8113, thank you for your help

PR #8075, opened 2024-07-10
feat(core): rename `TuiAlertDirective` to `TuiAlert`

mdlufy:
Closes #


projects/cdk/schematics/ng-update/v4/steps/constants/identifiers-to-replace.ts
```diff
@@ -736,13 +736,6 @@ export const IDENTIFIERS_TO_REPLACE: ReplacementIdentifierMulti[] = [
             moduleSpecifier: '@taiga-ui/core',
         },
     },
-    {
-        from: {name: 'TuiAlertModule', moduleSpecifier: '@taiga-ui/core'},
```
waterplea, line 740: We should keep it as previously it was imported where the directive was used, not just in global component. We only should drop it from `importProvidersFrom`.

waterplea (approved):

PR #8076, opened 2024-07-10
feat(kit): add margin and limit properties for Range component

### LINKED ISSUE #3434: 🚀 - `Range`, `InputRange` add `[margin]` & `[limit]`

kit

## Range
We should add new `@Input()`-properties for [Range](https://taiga-ui.dev/components/range)-component:

* `margin` - the minimum distance between the thumbs.
See https://refreshless.com/nouislider/slider-options/#section-margin

* `limit` -  the maximum distance between two thumbs.
See https://refreshless.com/nouislider/slider-options/#section-limit

## InputRange
Should we just duplicates the same input properties ?

Maybe it is better to pass `Range`-component  inside `InputRange` ?
```html
<tui-input-range>
    <input tuiTextfield />

    <tui-range [margin]="5" [limit]="30"></tui-range>
</tui-input-range>
```
(we need to discuss it and probably create the separate issue for `InputRange`)

LarisaGl:
Closes #3434

### DIFF
--- a/projects/kit/components/range/range.component.ts
+++ b/projects/kit/components/range/range.component.ts
@@ -84,6 +84,12 @@ export class TuiRange extends TuiControl<[number, number]> implements OnChanges 
     @Input()
     public focusable = true;
 
+    @Input()
+    public margin = 0;
+
+    @Input()
+    public limit = 0;
+
     @ViewChildren(TuiSliderComponent, {read: ElementRef})
     public readonly slidersRefs: QueryList<ElementRef<HTMLInputElement>> = EMPTY_QUERY;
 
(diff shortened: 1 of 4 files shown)


projects/kit/components/range/range.component.ts
```diff
@@ -167,10 +173,31 @@ export class TuiRange extends TuiControl<[number, number]> implements OnChanges
...
+
+        if (this.isDistanceExceeded(distance)) {
+            return;
+        }
+
+        this.onChange([this.value()[0], newValue]);
+    }
+
+    private isDistanceExceeded(distance: number): boolean {
+        return (
```
waterplea, line 198: Can we write `tuiClamp(distance, this.limit, this.margin) === distance`? I think it looks easier to read.

projects/kit/components/range/range.component.ts
```diff
@@ -167,10 +173,31 @@ export class TuiRange extends TuiControl<[number, number]> implements OnChanges
...
+
+        this.onChange([newValue, this.value()[1]]);
     }
 
     private updateEnd(value: number): void {
-        this.onChange([this.value()[0], Math.max(value, this.value()[0])]);
+        const newValue = Math.max(value, this.value()[0]);
+        const distance = newValue - this.value()[0];
+
+        if (this.isDistanceExceeded(distance)) {
```
waterplea, line 190: I think `checkDistance` is a better name, because it checks both limits.

projects/kit/components/range/range.component.ts
```diff
@@ -84,6 +84,12 @@ export class TuiRange extends TuiControl<[number, number]> implements OnChanges
     @Input()
     public focusable = true;
 
+    @Input()
+    public margin = 0;
+
+    @Input()
+    public limit = 0;
```
waterplea, line 91: Should be `Infinity` by default.

waterplea (approved):

waterplea:
Thank you for this PR! A few small comments, otherwise LGTM 👍

projects/kit/components/range/range.component.ts
LarisaGl: Oh, sure! It looks better. I didn't know about tuiClamp function. Fixed

projects/kit/components/range/range.component.ts
LarisaGl: Fixed

projects/kit/components/range/range.component.ts
LarisaGl: Yes, my bad, Infinity is correct. Fixed

vladimirpotekhin (approved):

PR #8085, opened 2024-07-11
feat(core): dialog-button-handling

PawelTr:
Closes #

### DIFF
--- a/projects/core/components/dialog/dialog.template.html
+++ b/projects/core/components/dialog/dialog.template.html
@@ -22,7 +22,7 @@
                     tuiAutoFocus
                     tuiButton
                     type="button"
-                    (click)="context.$implicit.complete()"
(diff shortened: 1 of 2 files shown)


projects/core/components/dialog/dialog.template.html
```diff
@@ -22,7 +22,7 @@
                     tuiAutoFocus
                     tuiButton
                     type="button"
-                    (click)="context.$implicit.complete()"
+                    (click)="context.$implicit.next(undefined)"
```
waterplea, line 25: It has to be `completeWith(undefined)` and a typecast has to be here, because this is a little hack. This button can appear only in string dialogs so they cannot receive value any other way, no need to change type to `O | undefined`, it's a special case that is too hard to properly handle with types and there no need to really.

Issue #8115, opened 2024-07-16
🐞 - TuiLineChart too small line, when coordinates have 0 value by Y


Zheleznyak-F:
### Playground Link

https://stackblitz.com/edit/angular-zfirxq?file=src%2Fapp%2Fapp.component.ts

### Description

![image](https://github.com/user-attachments/assets/e2396869-dd1c-4ab9-9e38-9c99eb7acd9f)
If the points touch the zero along the axis Y, then the line becomes thinner then usual

### Angular version

17.3.0

### Taiga UI version

3.83.0

### Which browsers have you used?

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

Zheleznyak-F:
Also actual, when points touch maximum value by Y

PR #8117, opened 2024-07-16
chore(demo): add example for zero line

splincode:


nsbarsukov (approved):

projects/demo/src/modules/components/line-chart/examples/1/index.ts
```diff
@@ -21,7 +21,8 @@ export default class Example {
         [200, 150],
         [250, 155],
         [300, 190],
-        [350, 90],
+        [350, 0],
+        [400, 0],
```
vladimirpotekhin, line 25: we already have an issue about this https://github.com/taiga-family/taiga-ui/issues/8115

I think we need to add an example when solving this issue rather than just adding a broken example.

PR #8139, opened 2024-07-17
fix(addon-charts): handle divide by zero in pie-chart

Exiragor:
Closes # [issue](https://github.com/taiga-family/taiga-ui/issues/7225)


splincode (approved):

projects/addon-charts/components/pie-chart/pie-chart.component.ts
```diff
@@ -132,7 +132,7 @@ export class TuiPieChart {
     }
 
     private getDeg(value: number): number {
-        return 360 * (value / this.getSum(this.value));
+        return 360 * (value / this.getSum(this.value) ?? 0);
```
waterplea, line 135: I'm not sure I understand how that helps, the value on the left is never `null` or `undefined`, why does `??` change anything? 🤔

Exiragor: but it's NaN, when we try to divide by zero(getSum for array like this [0, 0, 0] will be zero)
So we handle NaN and use "0" instead

waterplea: When we divide by zero we get `Infinity`, not `NaN` and `??` doesn't work with `NaN`:
image

Exiragor: yeah, you are right
sorry for misunderstanding, i mean, if we divide zero by zero, it's only the case when it's `NaN`
but anyway i should use `||` instead of `??`
i'll fix it now

Exiragor: but i change it to ternary, i guess it's more clean way
but u can correct me

waterplea: No need for ternary, just add `|| 0` at the very end.

Exiragor: just thought about prevent even divide, but i guess it's unnecessary
fix it, so

waterplea (approved):

PR #8153, opened 2024-07-18
feat(core): externalize some components styles

waterplea:
Closes #


splincode (approved):

projects/core/components/button/button.directive.ts
```diff
@@ -18,7 +18,7 @@ import {TUI_BUTTON_OPTIONS} from './button.options';
 @Component({
     standalone: true,
     template: '',
-    styleUrls: ['./button.style.less'],
+    styles: ['@import "@taiga-ui/core/styles/components/button.less";'],
```
nsbarsukov, line 21: What do you think about this approach ?
```suggestion
    styleUrls: ['../../styles/components/button.less'],
```

It has smarter IDE integration


https://github.com/user-attachments/assets/a64112c1-a733-42f5-8608-5eb9923cb52f

nsbarsukov (approved):

projects/core/components/button/button.directive.ts
waterplea: Pinged WebStorm guys about this bug:
https://youtrack.jetbrains.com/issue/WEB-56888/imports-with-tilde-not-resolved-in-Angular-inline-styles-section

PR #8168, opened 2024-07-19
refactor(core): `Notification` simplify to directive

waterplea:
Closes #


splincode (approved):

projects/core/components/notification/notification.options.ts
```diff
@@ -1,30 +1,27 @@
...
-export interface TuiNotificationOptions {
-    readonly status: TuiNotificationStatus;
-    readonly icon: PolymorpheusContent<TuiContext<TuiNotificationStatus>>;
+export interface TuiNotificationOptions extends TuiAppearanceOptions {
+    readonly icon: TuiStringHandler<string> | string;
     readonly size: TuiSizeL | TuiSizeS;
 }
 
-const STATUS_ICON = {
+const STATUS_ICON: any = {
```
splincode, line 12: any?

waterplea: Oops, forgot temporary fix. Thanks!

projects/core/components/notification/notification.directive.ts
```diff
@@ -0,0 +1,80 @@
...
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {
+        class: 'tui-notification',
+    },
+})
+class TuiNotificationStyles {}
+
+@Directive({
+    standalone: true,
+    selector: 'tui-notification,a[tuiNotification],button[tuiNotification]',
```
nsbarsukov, line 37: Should we remove `tui-notification`  and simplify selector to the following way ?

```suggestion
    selector: '[tuiNotification]',
```

For example
```html
<div tuiNotification />
```

nsbarsukov (approved):

PR #8176, opened 2024-07-22
refactor(demo-playwright): drop deprecated method `type`

splincode:


vladimirpotekhin (approved):

projects/demo-playwright/tests/kit/input-date-range/input-date-range.spec.ts
```diff
@@ -46,23 +46,23 @@ test.describe('InputDateRange', () => {
                     `01-calendar-size-${size}-empty.png`,
                 );
 
-                await inputDateRange.textfield.type('01');
+                await inputDateRange.textfield.fill('01');
```
nsbarsukov, line 49: Let's use `pressSequentially` (for masked textfields) instead 

> Type into the field character by character, as if it was a user with a real keyboard with `locator.pressSequentially()`.
> This method will emit all the necessary keyboard events, with all the keydown, keyup, keypress events in place.

https://playwright.dev/docs/input#type-characters

splincode: fixed

waterplea (approved):

Issue #8166, opened 2024-07-19
🐞 - Unexpected scrollbar behavior in Safari


splincode:
### Playground Link

https://taiga-ui.dev/next/getting-started

### Description

https://github.com/user-attachments/assets/896433ac-970c-4f35-bb9a-468daf793c35

### Angular version

latest

### Taiga UI version

latest

### Which browsers have you used?

- [ ] Chrome
- [ ] Firefox
- [X] Safari
- [ ] Edge

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

PR #8183, opened 2024-07-24
fix(core): it's not possible to scroll container elements on safari

splincode:
Closes https://github.com/taiga-family/taiga-ui/issues/8166


waterplea (approved):
Fucking piece of shit Safari

PR #8184, opened 2024-07-24
fix(core): user selection on safari

splincode:
# Current behaviour

https://github.com/user-attachments/assets/cd9d11cb-0e7d-4666-a86d-9e517370af07


nsbarsukov (approved):

waterplea:
It was like that before, I believe this had some issues with iframes and letting go of mouse outside of the current frame, which would cause the drag to stick to cursor next time you hover back to the frame. I think it was visible on Stackblitz.

Issue #7552, opened 2024-05-27
🐞 - `InputInline` has broken scroll in Safari / Firefox


goodday21:
### Description

Браузер - Safari.

Компонент - https://taiga-ui.dev/components/input-inline#heading

Шаги:
1) нажать на иконку редактирования у элемента "Page heading"
2) ввести длинную строку, более 150 символов
3) сохранить изменения
4) снова нажать на иконку редактирования
5) попробовать поменять строку

Ожидаемый результат: фокус установлен в конец строки, пользователь может передвигать курсор по тексту и редактировать его
Фактический результат: фокус остается на начале строки. Если пользователь работает с концом строки, то он не видит, что вводит

### Taiga UI version

3.80.0

### Which browsers have you used?

- [ ] Chrome
- [ ] Firefox
- [X] Safari
- [ ] Edge

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

waterplea:
Safari does not dispatch scroll events for inputs. We need to use additional keypress events, selection change events and whatnot to be able to adjust text-indent to input scroll. Maybe it's even easier to just use silent animation frame for polling, but that might cause a lot of CSS reflow.

nsbarsukov:
## Just some intermediate investigations

We can replace the following logic 
https://github.com/taiga-family/taiga-ui/blob/432a4867f60c99fa266cd1bfaadf2e983d7c4984/projects/kit/components/input-inline/input-inline.component.ts#L32-L38

with 

```ts
protected indent$ = merge(
    fromEvent(this.element, 'scroll', {capture: true}), // Only Chrome emits this event for <input />
    // Chrome does not fire `selectionchange` on deletion (but do it on insertion) ¯\_(ツ)_/¯
    fromEvent(this.element, 'selectionchange', {capture: true}), // Firefox + Safari fallback
    // Chrome & Safari scrolls input to the beginning on blur
    // Firefox – keeps the same scrollLeft position on blur
    fromEvent(this.element, 'focusout').pipe(
        concatMap(
            (x) => of(x).pipe(delay(0)), // Without delay `scrollLeft` is not yet updated
        ),
    ),
).pipe(
    map(({target}) => target),
    filter((x): x is HTMLInputElement => tuiIsElement(x) && tuiIsInput(x)),
    map(
        ({scrollLeft}) => -scrollLeft - 1, // -1 for Safari (see styles)
[...]

PR #8186, opened 2024-07-24
fix(kit): show editing mode in safari for input-inline

splincode:
Closes https://github.com/taiga-family/taiga-ui/issues/7552

### DIFF
--- a/projects/kit/components/input-inline/input-inline.component.ts
+++ b/projects/kit/components/input-inline/input-inline.component.ts
@@ -79,6 +79,13 @@ export class TuiInputInlineComponent
 
     onFocused(focused: boolean): void {
         this.updateFocused(focused);
+
+        if (focused) {
+            this.nativeFocusableElement?.scrollTo(
+                this.nativeFocusableElement.scrollWidth,
+                0,
+            );
+        }
     }
 
     onScroll(): void {
--- a/projects/kit/components/input-inline/input-inline.style.less
+++ b/projects/kit/components/input-inline/input-inline.style.less
@@ -6,6 +6,14 @@
     overflow: hidden;
     white-space: nowrap;
     box-sizing: border-box;
+
+    &._focused .t-native {
+        -webkit-text-fill-color: inherit;
+    }
+
+    &._focused .t-before {
+        color: transparent;
+    }
 }
 
 .t-before {
--- a/projects/kit/components/input-inline/input-inline.template.html
+++ b/projects/kit/components/input-inline/input-inline.template.html
@@ -24,7 +24,6 @@
     [id]="id"
     [ngModel]="value"
     [readOnly]="readOnly"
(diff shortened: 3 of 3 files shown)


projects/kit/components/input-inline/input-inline.template.html
```diff
@@ -24,6 +24,7 @@
     [id]="id"
     [ngModel]="value"
     [readOnly]="readOnly"
+    [style.padding-left.px]="indent"
```
vladimirpotekhin, line 27: if I remember correctly padding can't be negative, it is mistake?

splincode: fixed

vladimirpotekhin (approved):

projects/kit/components/input-inline/input-inline.component.ts
```diff
@@ -79,6 +79,13 @@ export class TuiInputInlineComponent
 
     onFocused(focused: boolean): void {
         this.updateFocused(focused);
+
+        if (focused) {
```
waterplea, line 83: This is not enough, unfortunately. We need to update scroll position of the wrapper to be in sync with the input, and input scroll position can change whenever. Try listening to `selectionchange` event, if it doesn't help — we probably would have to use polling with rAF when input is focused. Fucking Safari piece of shit 🤬 We can probably isolate it into a service/host directive with early return when not in Safari.

projects/kit/components/input-inline/input-inline.style.less
```diff
@@ -6,6 +6,14 @@
     overflow: hidden;
     white-space: nowrap;
     box-sizing: border-box;
+
+    &._focused .t-native {
```
waterplea, line 10: What are these changes for?

waterplea (changes requested):

projects/kit/components/input-inline/input-inline.style.less
splincode: For show native input element instead

projects/kit/components/input-inline/input-inline.component.ts
splincode: No, that's need for another PR
This is reproduced in chrome when input has too long value

projects/kit/components/input-inline/input-inline.style.less
waterplea: But that was this way on purpose, so it doesn't just by 0.5px when you focus since it's impossible to align input and div 100% the same

projects/kit/components/input-inline/input-inline.component.ts
waterplea: Ok

waterplea: Then remove "closes" from this PR as it only addresses part of the issue

PR #8188, opened 2024-07-24
fix(kit): reuse cursor position after showing password

splincode:
Closes https://github.com/taiga-family/taiga-ui/issues/7281

### DIFF
--- a/projects/kit/components/input-password/input-password.component.ts
+++ b/projects/kit/components/input-password/input-password.component.ts
@@ -31,8 +31,8 @@ import {
 } from '@taiga-ui/core';
 import {TUI_PASSWORD_TEXTS} from '@taiga-ui/kit/tokens';
 import {PolymorpheusContent} from '@tinkoff/ng-polymorpheus';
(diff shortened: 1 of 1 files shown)


projects/kit/components/input-password/input-password.component.ts
```diff
@@ -128,9 +128,24 @@ export class TuiInputPasswordComponent
...
+
+    private reuseCursorPosition(): void {
+        const selectionStart =
+            this.textfield?.nativeFocusableElement?.selectionStart ?? 0;
+
+        timer(0)
+            .pipe(takeUntil(this.destroy$))
+            .subscribe(() => {
+                if (this.textfield?.nativeFocusableElement) {
+                    this.textfield.nativeFocusableElement.selectionStart = selectionStart;
```
nsbarsukov, line 147: What's about cases when user selects several characters ?
Let's save `selectionEnd` too + use `setSelectionRange`

splincode: fixed

waterplea (approved):

Issue #8182, opened 2024-07-23
📚 - Update documentation about `PdfViewer`


nsbarsukov:
### Description

Documentation https://taiga-ui.dev/components/pdf-viewer says:
<img width="701" alt="mobile-description" src="https://github.com/user-attachments/assets/00637ca7-0ce5-4e21-a6d4-8dd4e125ebaa">

Actually, it already works for some mobile.
For example, IOS 17.5.1 supports it.

Investigate new status about it and update documentation.

PR #8190, opened 2024-07-24
chore(demo): improve documentation about pdf viewer

splincode:
Closes https://github.com/taiga-family/taiga-ui/issues/8182


projects/demo/src/modules/components/pdf-viewer/index.html
```diff
@@ -6,14 +6,8 @@
     <ng-template pageTab>
         <p>Custom dialog to preview PDF</p>
 
-        <tui-notification status="warning">
-            Keep in mind mobile devices do not support displaying PDFs in iframe, so you need to rely on
```
nsbarsukov, line 10: I am not sure that we should remove this warning from the top of the documentation page.

This issue https://github.com/taiga-family/taiga-ui/issues/8182 is create to just update text of this warning – provide the full list (or link to this information) which mobile devices supports pdf inside iframe and which ones not.

PR #8224, opened 2024-07-26
refactor: move `TuiStaticRequestService` to legacy

splincode:


projects/demo-cypress/src/tests/static-request.cy.ts
```diff
@@ -1,5 +1,5 @@
 import {TestBed} from '@angular/core/testing';
-import {TuiStaticRequestService} from '@taiga-ui/cdk';
+import {TuiStaticRequestService} from '@taiga-ui/legacy';
```
nsbarsukov, line 2: Add schematic migration
https://github.com/taiga-family/taiga-ui/blob/main/projects/cdk/schematics/ng-update/v4/steps/constants/identifiers-to-replace.ts

waterplea (approved):

projects/demo-cypress/src/tests/static-request.cy.ts
splincode: fixed

vladimirpotekhin (approved):

PR #8235, opened 2024-07-26
chore: add layout package if needed

vladimirpotekhin:
Closes #


projects/cdk/schematics/ng-update/v4/steps/update-packages.ts
```diff
@@ -53,4 +55,11 @@ export function updatePackages({tree}: DevkitFileSystem, _: TuiSchema): void {
             version: '^4.0.1',
         });
     }
+
+    if (getImports(ALL_TS_FILES, {moduleSpecifier: '@taiga-ui/layout'}).length) {
```
nsbarsukov, line 59: Should we add the similar `if`-condition for `@taiga-ui/legacy` above ? 🤔

nsbarsukov (approved):

splincode (approved):

PR #8239, opened 2024-07-28
feat(cdk): add dialog name for override overlay styles

splincode:
image


waterplea:
I would like to not increase complexity until we really need something like that. Did somebody request it internally?

PR #8240, opened 2024-07-28
feat(cdk): add `TuiThemeColorService`

### LINKED ISSUE #5685: 🚀 - Change status-bar color when opening dialog
### Description

When dialog and sheetDialog are used in PWA applications, there is an issue that the status bar remains the same color as the main background of the application. It would be better to dynamically change the color of the status bar when the dialog is opened. 

`document.querySelector('meta[name="theme-color"]')?.setAttribute('content', content)`

i suggest using а `#404040` or `#333` color

splincode:
Closes #5685

### DIFF
--- a/projects/cdk/services/theme-color.service.ts
+++ b/projects/cdk/services/theme-color.service.ts
@@ -1,0 +1,31 @@
+import {DOCUMENT} from '@angular/common';
+import {inject, Injectable} from '@angular/core';
+import {Meta} from '@angular/platform-browser';
+import {tuiCreateTokenFromFactory} from '@taiga-ui/cdk/utils';
+
+export const TUI_THEME_COLOR = tuiCreateTokenFromFactory<string>(
+    () => inject(Meta).getTag('name="theme-color"')?.content ?? '',
+);
+
+interface TuiThemeColor {
+    get color(): string;
+    set color(value: string);
+}
+
+@Injectable({
+    providedIn: 'root',
+})
+export class TuiThemeColorService implements TuiThemeColor {
+    private readonly current = inject(TUI_THEME_COLOR);
+    private readonly doc = inject(DOCUMENT);
+    private readonly meta = inject(Meta);
+
+    public get color(): string {
+        return this.current;
+    }
+
+    public set color(content: string) {
+        this.meta.updateTag({name: 'theme-color', content});
+        this.doc.documentElement.style.setProperty('--tui-theme-color', content);
+    }
+}
--- a/projects/demo/src/modules/components/dialog/examples/11/index.ts
+++ b/projects/demo/src/modules/components/dialog/examples/11/index.ts
@@ -1,0 +1,31 @@
+import {Component, inject} from '@angular/core';
+import {changeDetection} from '@demo/emulate/change-detection';
+import {encapsulation} from '@demo/emulate/encapsulation';
+import {TuiThemeColorService} from '@taiga-ui/cdk';
+import {TuiButton, TuiDialogService} from '@taiga-ui/core';
+
+@Component({
+    standalone: true,
+    imports: [TuiButton],
+    templateUrl: './index.html',
+    encapsulation,
+    changeDetection,
+})
+export default class Example {
+    private readonly dialogs = inject(TuiDialogService);
+    private readonly theme = inject(TuiThemeColorService);
+
+    protected showDialog(): void {
+        this.theme.color = '#ffdd2d';
+        this.dialogs
+            .open(
+                'Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry`s standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum',
+                {label: 'What is Lorem Ipsum?'},
+            )
+            .subscribe({
+                complete: () => {
+                    this.theme.color = '#ff7043';
+                },
+            });
+    }
+}
--- a/projects/cdk/services/index.ts
+++ b/projects/cdk/services/index.ts
@@ -1,3 +1,4 @@
 export * from './id.service';
 export * from './popover.service';
 export * from './scroll.service';
+export * from './theme-color.service';
--- a/projects/demo/src/modules/components/dialog/examples/11/index.html
+++ b/projects/demo/src/modules/components/dialog/examples/11/index.html
@@ -1,0 +1,8 @@
+<button
+    size="m"
+    tuiButton
+    type="button"
+    (click)="showDialog()"
+>
+    Show
+</button>
--- a/projects/demo/src/modules/components/dialog/index.html
+++ b/projects/demo/src/modules/components/dialog/index.html
@@ -104,6 +104,21 @@
             [component]="10 | tuiComponent"
             [content]="10 | tuiExample"
         />
+
+        <tui-doc-example
+            id="status-bar"
+            description="Change status-bar color when opening dialog"
+            heading="Theme color"
+            [component]="11 | tuiComponent"
+            [content]="11 | tuiExample: 'html,ts'"
+        >
+            <tui-notification
+                appearance="warning"
+                class="tui-space_bottom-4"
+            >
+                Works only on mobile browsers
+            </tui-notification>
+        </tui-doc-example>
     </ng-template>
 
     <ng-template pageTab>


projects/cdk/services/meta-color.service.ts
```diff
@@ -0,0 +1,44 @@
...
+import {Subject} from 'rxjs';
+
+@Injectable({
+    providedIn: 'root',
+})
+export class TuiMetaColorService implements OnDestroy {
+    private readonly document = inject(WA_WINDOW).document;
+    private readonly value = new Subject<string>();
+
+    protected sub = this.value.subscribe({
```
waterplea, line 13: What is the point of this subscription? Why not directly in `setColor`?

projects/cdk/services/meta-color.service.ts
```diff
@@ -0,0 +1,44 @@
+import type {OnDestroy} from '@angular/core';
+import {inject, Injectable} from '@angular/core';
+import {WA_WINDOW} from '@ng-web-apis/common';
+import {Subject} from 'rxjs';
+
+@Injectable({
+    providedIn: 'root',
+})
+export class TuiMetaColorService implements OnDestroy {
+    private readonly document = inject(WA_WINDOW).document;
```
waterplea, line 10: `inject(DOCUMENT)`

projects/cdk/services/meta-color.service.ts
```diff
@@ -0,0 +1,44 @@
...
+    public getColor(): string {
+        return this.metaThemeColor.getAttribute('content') ?? '';
+    }
+
+    public setColor(value: string): [prev: string, next: string] {
+        const prev = this.getColor();
+
+        this.value.next(value);
+
+        return [prev, value];
```
waterplea, line 26: I don't think we should return anything. Wherever you call `setColor` you can just call `getColor` yourself and you already have the new value.

projects/cdk/services/meta-color.service.ts
```diff
@@ -0,0 +1,44 @@
...
+
+@Injectable({
+    providedIn: 'root',
+})
+export class TuiMetaColorService implements OnDestroy {
+    private readonly document = inject(WA_WINDOW).document;
+    private readonly value = new Subject<string>();
+
+    protected sub = this.value.subscribe({
+        next: (value) => this.metaThemeColor.setAttribute('content', value),
```
waterplea, line 14: We also have a CSS variable used in Navigation — `--tui-theme-color`. Let's also set it on `documentElement`

projects/cdk/services/meta-color.service.ts
```diff
@@ -0,0 +1,44 @@
+import type {OnDestroy} from '@angular/core';
+import {inject, Injectable} from '@angular/core';
+import {WA_WINDOW} from '@ng-web-apis/common';
+import {Subject} from 'rxjs';
+
+@Injectable({
+    providedIn: 'root',
+})
+export class TuiMetaColorService implements OnDestroy {
```
waterplea, line 9: Maybe better call it `TuiThemeColorService`?

projects/demo/src/modules/components/dialog/examples/11/index.ts
```diff
@@ -0,0 +1,29 @@
...
+    templateUrl: './index.html',
+    encapsulation: ViewEncapsulation.None,
+    changeDetection,
+})
+export default class Example {
+    private readonly dialogs = inject(TuiDialogService);
+    private readonly themeColor = inject(TuiMetaColorService);
+
+    protected showDialog(): void {
+        const [prev] = this.themeColor.setColor('#ffdd2d');
```
waterplea, line 18: I wonder if we should have a token for initial value and bundle it into dialogs 🤔

projects/cdk/services/meta-color.service.ts
splincode: fixed

projects/cdk/services/meta-color.service.ts
splincode: fixed

projects/cdk/services/meta-color.service.ts
splincode: fixed

projects/cdk/services/meta-color.service.ts
splincode: fixed

projects/cdk/services/meta-color.service.ts
splincode: fixed

projects/cdk/services/theme-color.service.ts
```diff
@@ -0,0 +1,49 @@
+import {DOCUMENT} from '@angular/common';
+import {inject, Injectable} from '@angular/core';
+import {tuiCreateTokenFromFactory} from '@taiga-ui/cdk/utils';
+
+export const TUI_THEME_COLOR = tuiCreateTokenFromFactory(() =>
+    getColor(inject(DOCUMENT)),
```
waterplea, line 6: I don't think we should get it from meta or CSS variable, we should instead default it too nothing and allow people to provide it through token, which would in turn set it in meta and in CSS variable.

splincode: updated

projects/cdk/services/theme-color.service.ts
```diff
@@ -0,0 +1,42 @@
...
+    private readonly doc = inject(DOCUMENT);
+    private readonly initial = inject(TUI_THEME_COLOR);
+
+    constructor() {
+        if (this.initial) {
+            this.setColor(this.initial);
+        }
+    }
+
+    public getColor(): string {
```
waterplea, line 20: We don't need `getColor` to query DOM. We can just store `color` as property, instead of a readonly `initial`. Maybe also using setter instead of method to set it.

waterplea (approved):

projects/cdk/services/theme-color.service.ts
splincode: fixed

projects/demo/src/modules/components/dialog/examples/11/index.ts
```diff
@@ -0,0 +1,30 @@
+import {Component, inject, ViewEncapsulation} from '@angular/core';
+import {changeDetection} from '@demo/emulate/change-detection';
+import {TuiThemeColorService} from '@taiga-ui/cdk';
+import {TuiButton, TuiDialogService} from '@taiga-ui/core';
+
+@Component({
+    standalone: true,
+    imports: [TuiButton],
+    templateUrl: './index.html',
+    encapsulation: ViewEncapsulation.None,
```
nsbarsukov, line 10: Why `None` ?

projects/cdk/services/theme-color.service.ts
```diff
@@ -0,0 +1,32 @@
...
+        }
+    }
+
+    public set color(color: string) {
+        let meta = this.doc.querySelector('meta[name="theme-color"]');
+
+        if (!meta) {
+            meta = this.doc.createElement('meta');
+            meta.setAttribute('name', 'theme-color');
+            this.doc.head.appendChild(meta);
```
nsbarsukov, line 26: Could we use just built-in `Meta`-service ?
https://angular.dev/api/platform-browser/Meta

Do we need actually need `TuiThemeColorService` or can just add example with `Meta` ?

splincode: fixed

projects/demo/src/modules/components/dialog/examples/11/index.ts
splincode: fixed

nsbarsukov (approved):

projects/cdk/services/theme-color.service.ts
```diff
@@ -0,0 +1,26 @@
+import {DOCUMENT} from '@angular/common';
+import {inject, Injectable} from '@angular/core';
+import {Meta} from '@angular/platform-browser';
+import {tuiCreateToken} from '@taiga-ui/cdk/utils';
+
+export const TUI_THEME_COLOR = tuiCreateToken<string | null>(null);
```
waterplea, line 6: Technically we can have it out of sync if somebody adds this meta tag to index.html and not provide the token. Maybe we should use `Meta` to try and get default value in case it is defined after all?

projects/cdk/services/theme-color.service.ts
```diff
@@ -0,0 +1,26 @@
...
+    private readonly doc = inject(DOCUMENT);
+    private readonly initial = inject(TUI_THEME_COLOR);
+
+    constructor() {
+        if (this.initial) {
+            this.color = this.initial;
+        }
+    }
+
+    public set color(color: string) {
```
waterplea, line 22: There's no way to get current value now. What I wanted was to rename `initial` to `current`, update it in the setter and return in the getter.

waterplea (approved):

projects/cdk/services/theme-color.service.ts
splincode: fixed

projects/cdk/services/theme-color.service.ts
splincode: fixed

PR #8246, opened 2024-07-29
feat(core): add `mode` for `tui-scrollbar`

splincode:
based idea

image

### DIFF
--- a/projects/core/components/scrollbar/scrollbar.options.ts
+++ b/projects/core/components/scrollbar/scrollbar.options.ts
@@ -1,0 +1,22 @@
+import type {Provider} from '@angular/core';
+import {tuiCreateToken, tuiProvideOptions} from '@taiga-ui/cdk/utils/miscellaneous';
+
+export interface TuiScrollbarOptions {
+    mode: 'always' | 'hover';
+}
+
+export const TUI_DEFAULT_SCROLLBAR_OPTIONS: TuiScrollbarOptions = {
+    mode: 'always',
+};
+
+export const TUI_SCROLLBAR_OPTIONS = tuiCreateToken(TUI_DEFAULT_SCROLLBAR_OPTIONS);
+
+export function tuiScrollbarOptionsProvider(
+    options: Partial<TuiScrollbarOptions>,
+): Provider {
+    return tuiProvideOptions(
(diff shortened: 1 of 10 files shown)


projects/core/components/scrollbar/scrollbar.style.less
```diff
@@ -26,3 +33,8 @@
 .t-bars {
     color: var(--tui-text-primary);
 }
+
+.t-auto-hide:not(:active) {
+    opacity: 0;
```
waterplea, line 38: `.transition(opacity)`

projects/core/components/scrollbar/scrollbar.options.ts
```diff
@@ -0,0 +1,22 @@
+import type {Provider} from '@angular/core';
+import {tuiCreateToken, tuiProvideOptions} from '@taiga-ui/cdk/utils/miscellaneous';
+
+export interface TuiScrollbarOptions {
+    showScrollbarWhenScrolling: boolean;
```
waterplea, line 5: This should be some kind of `mode` rather than a boolean. Also iOS shows bars when **scrolling**, not when hovering. I think showing it on hover is much better, just pointing out the naming inconsistency here.

projects/demo/src/modules/components/scrollbar/index.html
```diff
@@ -66,6 +66,13 @@
                 .
             </p>
         </tui-doc-example>
+
+        <tui-doc-example
+            id="auto-hide"
```
waterplea, line 71: Let's use ngFor to iterate over examples

projects/core/components/scrollbar/scrollbar.style.less
splincode: fixed

projects/core/components/scrollbar/scrollbar.options.ts
splincode: fixed

projects/demo/src/modules/components/scrollbar/index.html
splincode: fixed

waterplea (approved):

vladimirpotekhin (approved):

Issue #8198, opened 2024-07-24
🐞 - tui-input-time  sets the correct value but doesn't  Preselect Current value from the dropdown 


dhawaldenny:
### Playground Link

https://stackblitz.com/edit/angular-vj227c-wexvym?file=src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.module.ts

### Description

For the Combobox, it preselects values from the dropdown, but it doesn't preselect values from the input time. If a minor change in time is needed, you have to scroll again to the selected value.

![image](https://github.com/user-attachments/assets/844c5ffb-b947-4dfe-9c1d-36786808eb15)

![image](https://github.com/user-attachments/assets/2281fed9-3286-4ea8-83d5-bbe836d69a38)

![image](https://github.com/user-attachments/assets/5534d11e-81fe-42a0-9df5-87194e0eed19)

In the attached stackblitz demo, the Combobox is working fine, but there are issues with the input time.

### Angular version

16.1.4

### Taiga UI version

3.76.0

### Which browsers have you used?

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Which operating systems have you used?

- [X] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

ShawnDvn:
Hi,

I am experiencing the same issue with `tui-input-time` not preselecting the current value from the dropdown. 

**Observed Behavior**: The dropdown does not preselect the value from the input time, requiring users to manually scroll to the correct time.

![image](https://github.com/user-attachments/assets/a03f3228-563e-420c-88c9-017bdc237d50)

In this case, the user needs to scroll down from 00:00 to the current value 09:00. It should allow the user to make minor time adjustments such as changing from 09:00 to 08:00 or 10:00.

08:00  
09:00 <= preselected value  
10:00

PR #8249, opened 2024-07-29
fix(kit): correct present of control value from items list for input-time

splincode:
Closes https://github.com/taiga-family/taiga-ui/issues/8198

### DIFF
--- a/projects/core/components/data-list/option/option.component.ts
+++ b/projects/core/components/data-list/option/option.component.ts
@@ -97,6 +97,7 @@ export class TuiOptionComponent<T = unknown> implements OnDestroy {
     onClick(): void {
         if (this.host && this.value !== undefined) {
             this.host.handleOption(this.value);
+            this.el.nativeElement.focus();
         }
     }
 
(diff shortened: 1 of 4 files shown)


projects/core/interfaces/data-list-host.ts
```diff
@@ -2,8 +2,8 @@ import {TuiIdentityMatcher, TuiStringHandler} from '@taiga-ui/cdk';
 
 // TODO: Consider refactoring checkOption, it is only needed in ComboBox
 export interface TuiDataListHost<T> {
-    checkOption?(option: T): void;
-    handleOption(option: T): void;
+    checkOption?(option: T, element: HTMLElement): void;
```
waterplea, line 5: Let's not expand the public API and instead reuse select option component that does the scroll automatically.

splincode: fixed

projects/core/components/data-list/option/option.component.ts
```diff
@@ -97,6 +97,7 @@ export class TuiOptionComponent<T = unknown> implements OnDestroy {
     onClick(): void {
         if (this.host && this.value !== undefined) {
             this.host.handleOption(this.value);
+            this.el.nativeElement.focus();
```
waterplea, line 100: Why do we need it? It gets focused on hover, what is the case where it can be clicked while not focused?

waterplea (approved):

waterplea (approved):

PR #8253, opened 2024-07-29
fix(kit): placeholder in multiselect disappears while hovering options

### LINKED ISSUE #8233: 🐞 - Placeholder in multiselect disappears while hovering options if the search is empty
### Playground Link

https://stackblitz.com/edit/angular-lpncnx

### Description

Placeholder in multiselect disappears while hovering options if the search is empty. Currently, we're using v3.42.1 and the placeholder doesn't disappear. Is it intended? it looks strange a bit.

### Angular version

12.2.7

### Taiga UI version

3.87.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
Closes #8233


projects/kit/components/input-tag/input-tag.component.ts
```diff
@@ -269,6 +269,10 @@ export class TuiInputTagComponent
         );
     }
 
+    get hasNativePlaceholder(): boolean {
+        return !!this.nativeFocusableElement?.placeholder;
```
waterplea, line 273: There's a `placeholder` input, no need to reach to actual element with a getter

projects/kit/components/input-tag/input-tag.template.html
```diff
@@ -47,6 +47,7 @@
                 >
                     <div
                         class="t-tags"
+                        [class.t-has-native-placeholder]="hasNativePlaceholder"
```
waterplea, line 50: Can be host binding on placeholder input, `.t-with-placeholder` for brevity

waterplea (approved):

projects/kit/components/input-tag/input-tag.component.ts
splincode: fixed

projects/kit/components/input-tag/input-tag.template.html
splincode: fixed

PR #8264, opened 2024-07-30
feat(addon-mobile): possibility drop mobile calendar header

MillerSvt:
Closes #

### DIFF
--- a/projects/addon-mobile/components/mobile-calendar/mobile-calendar.template.html
+++ b/projects/addon-mobile/components/mobile-calendar/mobile-calendar.template.html
@@ -1,4 +1,7 @@
-<header class="t-header">
+<header
+    *ngIf="chooseDayOrRangeTexts$ | async as texts"
+    class="t-header"
+>
     <button
         appearance=""
         automation-id="tui-mobile-calendar__cancel"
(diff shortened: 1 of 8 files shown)


splincode (approved):

projects/addon-mobile/components/mobile-calendar/mobile-calendar.template.html
```diff
@@ -1,4 +1,7 @@
-<header class="t-header">
+<header
+    *ngIf="chooseDayOrRangeTexts$ | async as texts"
```
nsbarsukov, line 2: Why only `chooseDayOrRangeTexts` ? Should we check also `closeWord$` & `cancelWord$` ?

projects/i18n/interfaces/language.ts
```diff
@@ -62,7 +62,7 @@ export interface TuiLanguageKit {
     /**
      * [@string 'choose day', @param 'choose range', @param 'choose days']
      */
-    mobileCalendarTexts: [string, string, string];
+    mobileCalendarTexts: [string, string, string] | null;
```
nsbarsukov, line 65: Could we just use empty array to hide these labels ?

projects/addon-mobile/components/mobile-calendar/mobile-calendar.template.html
MillerSvt: I think not, because by typing there can not be null. And these are common variables that can be used in other places.

projects/i18n/interfaces/language.ts
MillerSvt: `[string, string, string] | []` ?

projects/demo/src/modules/components/mobile-calendar/examples/6/index.less
```diff
@@ -0,0 +1,3 @@
+.example {
+    height: 35rem;
+}
```
nsbarsukov, line 3: Let's drop redundant nesting

```suggestion
:host {
    display: block;
    height: 35rem;
}
```

\+ delete `<div class="example">` from HTML-file

nsbarsukov (approved):

PR #8296, opened 2024-08-01
fix(addon-mobile): fix mobile-calendar size without header

MillerSvt:
Closes #


projects/addon-mobile/components/mobile-calendar/mobile-calendar.component.ts
```diff
@@ -82,6 +95,10 @@ export class TuiMobileCalendarComponent implements AfterViewInit {
     @ViewChild('monthsScrollRef')
     private readonly monthsScrollRef?: CdkVirtualScrollViewport;
 
+    private readonly value$ = new BehaviorSubject<
```
waterplea, line 98: Looks like this can be an output already, no need for middle man

projects/demo/src/modules/components/mobile-calendar/mobile-calendar.template.html
```diff
@@ -144,6 +152,14 @@
             >
                 Single date or a range
             </ng-template>
+            <ng-template
+                documentationPropertyMode="input"
+                documentationPropertyName="multi"
+                documentationPropertyType="boolean"
+                [(documentationPropertyValue)]="multi"
+            >
+                Multi date or a range
```
waterplea, line 161: `Array of single dates` would be easier to understand. After refactor that should probably become a singular mode `singe | multi | range` instead of 2 booleans

projects/addon-mobile/components/mobile-calendar/mobile-calendar.component.ts
```diff
@@ -145,6 +165,23 @@ export class TuiMobileCalendarComponent implements AfterViewInit {
...
             this.value = value;
         });
+
+        this.value$
+            .pipe(
+                skip(1),
+                distinctUntilChanged((a, b) => a?.toString() === b?.toString()),
+                takeUntil(this.destroy$),
+            )
+            .subscribe(this.valueChange);
```
waterplea, line 175: Instead of subscribing, I think it's better to do `readonly valueChange = this.value$.pipe(...)`

waterplea (approved):

waterplea (approved):

PR #8326, opened 2024-08-03
feat(cdk): add `TuiReadonlyDate` type

splincode:
image


waterplea:
I don't think we need this. We barely ever use native dates

Issue #3708, opened 2023-02-17
🚀 - Dont allow to select disabled dates in calendar


sakuranasty:
### Which @taiga-ui/* package(s) are relevant/releated to the feature request?

core, addon-mobile

### Description

Currently there is an opportunity to pass disabledItemHandler, but it is possible to select disabled dates, if these dates are betwee two awailable dates.

Example:
![image](https://user-images.githubusercontent.com/34821835/219650969-7cb716f1-2f29-4ad8-933d-8d8500b75214.png)

Please add this feature to version 2.x

p-ivannikov:
I will take it

waterplea:
Had to rollback the change due to performance. Reopening

splincode:
@p-ivannikov fyi

waterplea:
I'm closing this as working as intended. Selection of dates should always be paired with a validator because in InputDateRange a user can manually type in a range with disabled dates inside and there's nothing we can do about it, other than invalidate the control. Since disabling dates is done through predicate to allow for rules/repeated dates, rather than an array of disabled dates – it is techincally very taxing to go over each day in an attempted range to check if it is not disabled. The benefits of adding this is minimal while technical and performance challenges are great, therefore closing this as not planned.

PR #8329, opened 2024-08-04
feat(kit): prevent disabled date selection for calendar-range

p-ivannikov:
Closes https://github.com/taiga-family/taiga-ui/issues/3708

### DIFF
--- a/projects/kit/components/calendar-range/calendar-range.component.ts
+++ b/projects/kit/components/calendar-range/calendar-range.component.ts
@@ -49,6 +49,7 @@ export class TuiCalendarRange implements OnInit, OnChanges {
     protected previousValue: TuiDayRange | null = null;
     protected hoveredItem: TuiDay | null = null;
     protected readonly capsMapper = TUI_DAY_CAPS_MAPPER;
+    protected availableRange: TuiDayRange | null = null;
 
     @Input()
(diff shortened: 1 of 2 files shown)


projects/kit/components/calendar-range/calendar-range.component.ts
```diff
@@ -208,7 +208,41 @@ export class TuiCalendarRange implements OnChanges {
...
+
+    private isExternalDaysDisabled(
+        disabledItemHandler: TuiBooleanHandler<TuiDay>,
+        startDay: TuiDay,
+        item: TuiDay,
+    ): boolean {
+        let temp = item;
+
+        while (temp.dayBefore(startDay) || temp.dayAfter(startDay)) {
+            if (disabledItemHandler(temp)) {
```
waterplea, line 237: Disabled item handler is called on each change detection on each available day, this function runs in creating classes in a loop for each day in between, I fear it might be detrimental to performance. Can we optimize it somehow? Can we measure how much of an impact this change brings?

p-ivannikov: changed search logic. After first day selection we search range of available dates. Than for each item we check its availability if item contains in selected range

waterplea (approved):

splincode (approved):

PR #8330, opened 2024-08-04
refactor: support `noUncheckedIndexedAccess`

splincode:
## Current behavior

image

This is because by default (even with `strict: true`), TypeScript will assume that any property in `obj` will be a string, even though it could be undefined at runtime.

## New behavior

image

But with `noUncheckedIndexedAccess`, we can get an error at compile time.

That's because TypeScript forces you to check if the property exists before accessing it.


projects/addon-charts/components/line-chart/line-chart-hint.directive.ts
```diff
@@ -63,14 +63,17 @@ export class TuiLineChartHint implements AfterViewInit {
     }
 
     // _chart is required by TuiLineDaysChartComponent that impersonates this directive
-    public getContext(index: number, _chart: TuiLineChart): readonly TuiPoint[] {
+    public getContext(
+        index: number,
+        _chart: TuiLineChart,
+    ): readonly TuiPoint[] | undefined {
```
waterplea, line 69: Why the return type has changed? Why not empty array?

projects/addon-charts/components/line-chart/line-chart-hint.directive.ts
```diff
@@ -63,14 +63,17 @@ export class TuiLineChartHint implements AfterViewInit {
...
         return this.computeContext(...this.charts.map(({value}) => value))[index];
     }
 
     // _chart is required by TuiLineDaysChartComponent that impersonates this directive
     public raise(index: number, _chart: TuiLineChart): void {
         const current = this.charts.map((chart) => chart.value[index]);
-        const sorted = [...current].sort((a, b) => a[1] - b[1]);
+        const sorted = [...current].sort((a, b) => (a?.[1] || 0) - (b?.[1] || 0));
```
waterplea, line 76: chart.value is an array of guaranteed tuples, we should be able to add [0, 0] fallback on the line above when we map, rather than using ?. and fallbacks here.

projects/addon-charts/components/pie-chart/pie-chart.directive.ts
```diff
@@ -48,12 +48,14 @@ export class TuiPieChartDirective {
...
             .subscribe(([start, end]) =>
-                el.setAttribute('d', tuiDescribeSector(start, end)),
+                el.setAttribute('d', tuiDescribeSector(start || 0, end || 0)),
             );
     }
 
     @Input()
-    public set tuiPieChart(sector: readonly [number, number]) {
-        this.sector$.next(sector);
+    public set tuiPieChart(sector: readonly [number, number] | undefined) {
```
waterplea, line 56: We shouldn't have to change the API because we make stricter typescript rules

projects/addon-charts/components/pie-chart/pie-chart.directive.ts
```diff
@@ -48,12 +48,14 @@ export class TuiPieChartDirective {
                 takeUntilDestroyed(),
             )
             .subscribe(([start, end]) =>
-                el.setAttribute('d', tuiDescribeSector(start, end)),
+                el.setAttribute('d', tuiDescribeSector(start || 0, end || 0)),
```
waterplea, line 51: Maybe just add default values to `tuiDescribeSector`?

projects/addon-charts/utils/control-point.ts
```diff
@@ -4,18 +4,18 @@ import {tuiLineAngle} from './line-angle';
 import {tuiLineLength} from './line-length';
 
 export function tuiControlPoint(
-    current: TuiPoint,
-    previous?: TuiPoint,
-    next?: TuiPoint,
+    current: TuiPoint | undefined,
```
waterplea, line 7: current should not be undefined. Also why change from ?: to explicit undefined?

PR #8336, opened 2024-08-05
chore(schematics): more identifiers migrations

nsbarsukov:
Relates:
* https://github.com/taiga-family/taiga-ui/issues/5808

### DIFF
--- a/projects/cdk/schematics/ng-update/v4/steps/constants/identifiers-to-replace.ts
+++ b/projects/cdk/schematics/ng-update/v4/steps/constants/identifiers-to-replace.ts
@@ -671,7 +671,7 @@ export const IDENTIFIERS_TO_REPLACE: ReplacementIdentifierMulti[] = [
     },
     {
         from: {name: 'TuiHostedDropdownComponent', moduleSpecifier: '@taiga-ui/core'},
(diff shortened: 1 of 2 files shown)


splincode (approved):

projects/cdk/schematics/ng-update/v4/steps/constants/identifiers-to-replace.ts
```diff
@@ -669,6 +669,10 @@ export const IDENTIFIERS_TO_REPLACE: ReplacementIdentifierMulti[] = [
             spreadInModule: true,
         },
     },
+    {
+        from: {name: 'TuiHostedDropdownComponent', moduleSpecifier: '@taiga-ui/core'},
+        to: {name: 'TuiDropdownDirective', moduleSpecifier: '@taiga-ui/core'},
```
waterplea, line 674: Shouldn't it be DropdownOpen?

waterplea (approved):

projects/cdk/schematics/ng-update/v4/steps/constants/identifiers-to-replace.ts
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/8336/commits/f18814b2a6abc7fef10e2574806b60f8368d5def

PR #8337, opened 2024-08-05
fix(legacy): correct present of control value from items list for input-time

splincode:
Close https://github.com/taiga-family/taiga-ui/issues/8198

### DIFF
--- a/projects/legacy/components/input-time/input-time.component.ts
+++ b/projects/legacy/components/input-time/input-time.component.ts
@@ -121,6 +121,12 @@ export class TuiInputTimeComponent
         controlValue instanceof TuiTime &&
         controlValue.valueOf() === dropdownValue.valueOf();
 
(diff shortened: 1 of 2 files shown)


waterplea (approved):

projects/legacy/components/input-time/input-time.component.ts
```diff
@@ -110,6 +110,12 @@ export class TuiInputTimeComponent
         this.nativeFocusableElement.value = value;
     }
 
+    public checkOption(option: TuiTime): void {
+        if (option.toAbsoluteMilliseconds() === this.value?.toAbsoluteMilliseconds()) {
```
waterplea, line 114: I wonder if we can use valueOf to make it shorter

splincode: fixed

mdlufy (approved):

PR #8344, opened 2024-08-05
chore: upgrade editor

splincode:


.npmrc
```diff
@@ -2,5 +2,5 @@ legacy-peer-deps=true # TODO: remove this line in v4.0
 strict-ssl=false # Fix problem when downloading cypress binary
 ca = ''
 unsafe-perm=true
-engine-strict=true
+engine-strict=false
```
nsbarsukov, line 5: Why do we need to change it ?

splincode: image

mdlufy (approved):

.npmrc
nsbarsukov: Should we just update CI Node version instead ?

waterplea (approved):

Issue #8339, opened 2024-08-05
🐞 - Paste function does not work in InputCardGrouped component in mobile devices


ayudich:
### Playground Link

https://taiga-ui.dev/components/input-card-grouped

### Description

1. Open https://taiga-ui.dev/components/input-card-grouped
2. Copy to buffer any number(card number for example)
3. Hold on in the InputCardGrouped element
Expected: Context menu with "Paste" function should be appeared
Actual: Context menu does not appear

### Angular version

latest on the site

### Taiga UI version

3.88.0

### Which browsers have you used?

- [X] Chrome
- [X] Firefox
- [X] Safari
- [ ] Edge

### Which operating systems have you used?

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [X] iOS
- [ ] Android

PR #8345, opened 2024-08-05
fix(core): ios menu is not triggered

splincode:
Closes https://github.com/taiga-family/taiga-ui/issues/8339


waterplea (approved):
Don't forget 4.0

PR #8348, opened 2024-08-05
fix(core): ios menu is not triggered

splincode:


projects/legacy/styles/mixins/mixins.less
```diff
@@ -33,12 +33,6 @@
     appearance: none;
     word-break: keep-all;
     -webkit-text-fill-color: currentColor; // for Safari
-
-    .ios-only({
```
waterplea, line 37: Something like this exists in modern textfields too, please remove it there as well

waterplea (approved):

vladimirpotekhin (approved):

PR #8356, opened 2024-08-05
ci: add eslint consistent-return

splincode:


projects/kit/components/pagination/pagination.component.ts
```diff
@@ -298,6 +298,7 @@ function tuiHorizontalDirectionToNumber(direction: TuiHorizontalDirection): -1 |
         case 'left':
             return -1;
         case 'right':
+        default:
```
waterplea, line 301: Let's just use ternary

waterplea (approved):

projects/kit/components/pagination/pagination.component.ts
splincode: fixed

waterplea (approved):

vladimirpotekhin (approved):

PR #8378, opened 2024-08-06
fix(core): background should be filled relatively `tui-root`

splincode:
I want add a custom content before or after tui-root

image


waterplea:
We should drop this ::before element completely and instead use `border-image` to just apply color.

PR #8395, opened 2024-08-07
chore(kit): `Stepper` vertical align set to stretch

mdlufy:
Closes #

### DIFF
--- a/projects/kit/components/stepper/step.style.less
+++ b/projects/kit/components/stepper/step.style.less
@@ -32,7 +32,7 @@
         cursor: default;
     }
 
-    &:focus-visible::before {
+    &:focus-visible:before {
         content: '';
         position: absolute;
         left: 2.75rem;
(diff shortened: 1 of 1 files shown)


projects/kit/components/stepper/step.style.less
```diff
@@ -47,13 +47,15 @@
...
 
+    position: relative;
     display: flex;
     width: 2rem;
     height: 2rem;
     border-radius: 100%;
     margin-right: 0.75rem;
     flex-shrink: 0;
     align-items: center;
+    align-self: stretch;
```
waterplea, line 58: Shouldn't it be flex-start?

waterplea (approved):

splincode (approved):

projects/kit/components/stepper/step.style.less
mdlufy: yes, rather flex-start

mdlufy: fixed

PR #8419, opened 2024-08-08
refactor: labeled tuple

splincode:


waterplea:
We cannot do that now, because it is not normalized across the codebase. In some places it is the other way around. I have in 5.0 RFC.

PR #8447, opened 2024-08-09
chore: some fixes

waterplea:
Closes #


splincode (approved):

projects/core/components/textfield/textfield.component.ts
```diff
@@ -97,6 +98,10 @@ export class TuiTextfieldComponent<T> implements TuiDataListHost<T> {
         return this.el?.nativeElement.id || '';
     }
 
+    public get size(): TuiSizeL | TuiSizeS {
```
waterplea, line 101: Needed for data list to properly size itself

vladimirpotekhin (approved):

projects/kit/components/segmented/segmented.directive.ts
```diff
@@ -1,4 +1,10 @@
-import type {AfterContentChecked, AfterContentInit, QueryList} from '@angular/core';
+import {isPlatformBrowser} from '@angular/common';
+import {
```
splincode, line 2: `import type`

PR #8454, opened 2024-08-10
feat: add `@taiga-ui/icons-material`

splincode:
image


waterplea:
Let's name it `@taiga-ui/icons-material`. So that all icon packages are alphabetically close. Not sure about 3 words, should we keep `-design`?

splincode:
@waterplea fixed

waterplea (approved):

vladimirpotekhin (approved):

PR #8456, opened 2024-08-11
feat(core): button has no default behavior and does nothing when pressed by default

splincode:
Related https://github.com/taiga-family/taiga-ui/pull/8455


waterplea:
We don't want to mess with browser defaults here

PR #8467, opened 2024-08-12
chore: `filterByInputWith` add missing migration

vladimirpotekhin:
Closes #

### DIFF
--- a/projects/cdk/schematics/ng-update/v4/tests/schematic-textfield-controller.spec.ts
+++ b/projects/cdk/schematics/ng-update/v4/tests/schematic-textfield-controller.spec.ts
@@ -84,9 +84,9 @@ const TEMPLATE_AFTER = `
     <tui-data-list-wrapper
         *tuiDataList
         tuiMultiSelectGroup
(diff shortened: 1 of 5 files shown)


splincode (approved):

projects/cdk/schematics/ng-update/v4/tests/schematic-textfield-controller.spec.ts
```diff
@@ -53,6 +53,12 @@ const TEMPLATE_BEFORE = `
         tuiTextfield
         type="email"
     />
+    <tui-data-list-wrapper
+        *tuiDataList
+        tuiMultiSelectGroup
+        [items]="items | tuiFilterByInputWith : matcher"
```
waterplea, line 59: `tuiFilterByInputWith` was used for stringify not for matcher. Now it gets stringify from host, so we should replace:
`xxx | tuiFilterByInputWith: yyy` with `xxx | tuiFilterByInput`

waterplea (approved):

PR #8478, opened 2024-08-12
fix(core): ignore pointer events for disabled button

splincode:
Close https://github.com/taiga-family/taiga-ui/issues/8479

### DIFF
--- a/projects/demo/src/modules/components/button/examples/1/index.less
+++ b/projects/demo/src/modules/components/button/examples/1/index.less
@@ -17,3 +17,8 @@
         animation: rotating 1.5s linear infinite;
     }
 }
+
+button:hover,
+a:hover {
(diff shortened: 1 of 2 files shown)


waterplea (approved):

projects/demo/src/modules/components/button/examples/1/index.less
```diff
@@ -17,3 +17,8 @@
         animation: rotating 1.5s linear infinite;
     }
 }
+
+button:hover,
+a:hover {
+    outline: 2px solid var(--tui-accent-hover);
```
waterplea, line 23: I think this might confuse people, let's not add it just for testing.

splincode: fixed

mdlufy (approved):

PR #8483, opened 2024-08-12
fix(core): ignore pointer events for disabled button

splincode:


projects/demo/src/modules/components/button/index.html
```diff
@@ -27,6 +28,22 @@
...
+                documentationPropertyType="boolean"
+                [(documentationPropertyValue)]="loading"
+            >
+                Loading indicator
+            </ng-template>
+            <ng-template
+                documentationPropertyMode="input"
+                documentationPropertyName="disabled"
+                documentationPropertyType="boolean"
+                [(documentationPropertyValue)]="disabled"
```
waterplea, line 43: This is a native property, I don't think we need to add it

projects/core/styles/components/button.less
```diff
@@ -35,6 +35,10 @@
     font: var(--tui-font-text-m);
     font-weight: bold;
 
+    &:disabled {
```
waterplea, line 38: Buttons have basic appearance applied to them, which already takes care of that:
https://github.com/taiga-family/taiga-ui/blob/main/projects/core/styles/components/appearance.less#L35

waterplea (commented):
Please check my comments. Should we just close it, as it seems no changes are necessary in main branch?

PR #8484, opened 2024-08-12
feat(legacy): added directive for activation expandable feature in textarea

denzo1993:
When there are a lot of expandable textareas, they take up a lot of space. This code allows to save space on the screen and use expandable feature.

If the textarea is in focus, then expandable feature is activated, else textarea has default height.


waterplea:
That's an interesting idea, however I would like to not expand public API, this can already be achieved with same amount of lines:
```html
<tui-textarea #textarea [expandable]="textarea.focused">
```
Maybe you can just add such example to the docs?

denzo1993:
> That's an interesting idea, however I would like to not expand public API, this can already be achieved with same amount of lines:
> 
> ```
> <tui-textarea #textarea [expandable]="textarea.focused">
> ```
> 
> Maybe you can just add such example to the docs?

Ok, commit reverted and added information into docs 👍. I use this feature in my project now like in example below above  🙃

The idea of ​​my modification is that it is possible to expand the allowed values of TuiTextareaExpandableTrigger ​​and leave this concern inside the library

splincode (approved):

PR #8574, opened 2024-08-16
fix(kit): segmented subscription to control value changes

Vadim2805:
Closes #


projects/kit/components/segmented/segmented.directive.ts
```diff
@@ -42,12 +43,13 @@ export class TuiSegmentedDirective implements AfterContentChecked, AfterContentI
...
-                switchMap((controls) => merge(controls.map((c) => c.valueChanges))),
+                switchMap((controls) => merge(controls.map((c) => tuiControlValue(c)))),
+                mergeAll(),
                 filter(() => this.isBrowser),
                 takeUntilDestroyed(this.destroyRef),
             )
-            .subscribe(() => {
-                this.update(this.el.querySelector(':checked'));
+            .subscribe((value) => {
+                this.update(this.el.querySelector(`[value=${value}]`));
```
waterplea, line 52: Why this change?

.eslintrc.json
waterplea, line 1: Is that to normalize MacOS vs Windows?

projects/kit/components/segmented/segmented.directive.ts
```diff
@@ -15,9 +15,10 @@ import {RouterLinkActive} from '@angular/router';
 import {EMPTY_QUERY} from '@taiga-ui/cdk/constants';
 import {tuiQueryListChanges} from '@taiga-ui/cdk/observables';
 import {tuiInjectElement} from '@taiga-ui/cdk/utils/dom';
-import {filter, merge, switchMap} from 'rxjs';
+import {filter, merge, mergeAll, switchMap} from 'rxjs';
 
 import {TuiSegmented} from './segmented.component';
+import {tuiControlValue} from '@taiga-ui/cdk/observables';
```
waterplea, line 21: I think precommit should have moved it up to other imports

projects/kit/components/segmented/segmented.directive.ts
Vadim2805: Yeh, you are right, we don't need this change.

.eslintrc.json
Vadim2805: yes

projects/kit/components/segmented/segmented.directive.ts
waterplea: And then `mergeAll()` is unnecessary too

Vadim2805: we still need mergeAll here, tested locally.

projects/kit/components/segmented/segmented.directive.ts
Vadim2805: thanks, corrected.

waterplea (approved):

package.json
```diff
@@ -80,7 +80,13 @@
...
         ],
         "rules": {
-            "jest/prefer-importing-jest-globals": "off"
+            "jest/prefer-importing-jest-globals": "off",
+            "prettier/prettier": [
+                "error",
+                {
+                    "endOfLine": "auto"
+                }
+            ]
```
nsbarsukov, line 89: Should it be reverted ? 🤔

nsbarsukov (approved):

package.json
splincode: I added this line to taiga linters, it will be updated later

PR #8581, opened 2024-08-16
feat(icons): support typings

splincode:
# If import `font-awesome`

`import {TuiIcon, TuiIconPipe} from '@taiga-ui/icons-fontawesome';`

image

# If import `material`

`import {TuiIcon, TuiIconPipe} from '@taiga-ui/icons-material';`

image

# If import `icons(lucide)`

`import {TuiIcon, TuiIconPipe} from '@taiga-ui/icons';`

image

# If just import from core

No typings by default

image


projects/kit/components/switch/switch.options.ts
```diff
@@ -3,14 +3,14 @@ import type {TuiStringHandler} from '@taiga-ui/cdk/types';
 import {tuiCreateToken, tuiProvideOptions} from '@taiga-ui/cdk/utils/miscellaneous';
 import type {TuiSizeS} from '@taiga-ui/core/types';
 
-export interface TuiSwitchOptions {
+export interface TuiSwitchOptions<Icon extends string> {
     readonly showIcons: boolean;
     readonly size: TuiSizeS;
-    readonly icon: TuiStringHandler<TuiSizeS> | string;
+    readonly icon: Icon | TuiStringHandler<TuiSizeS>;
```
waterplea, line 9: Should be `TuiHandler<TuiSizeS, Icon>`

waterplea (commented):
That's great! Do you think we can come up with something like that for `iconStart/iconEnd` directives though? 🤔

projects/kit/components/switch/switch.options.ts
splincode: fixed

PR #8584, opened 2024-08-17
fix(legacy): display view/hide password button with `readOnly` property

### LINKED ISSUE #6973: 🐞 - tui-input-password does not display the view/hide password button
### Playground Link

https://stackblitz.com/edit/angular-n6jrjp?file=src%2Fapp%2Fapp.component.html

### Description

Can you add [showButtons] directory in 3-th version? Because readOnly attribute can't work with dynamic form.

### Angular version

17

### Taiga UI version

3.69

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

olegnowikow:
Closes #6973

### DIFF
--- a/projects/legacy/components/input-password/input-password.component.ts
+++ b/projects/legacy/components/input-password/input-password.component.ts
@@ -67,7 +67,10 @@ export class TuiInputPasswordComponent
     }
 
     public get inputType(): string {
-        return this.isPasswordHidden || !this.interactive ? 'password' : 'text';
+        return this.isPasswordHidden ||
+            !(this.interactive || (this.readOnly && !this.computedDisabled))
+            ? 'password'
+            : 'text';
     }
 
     public onValueChange(textValue: string): void {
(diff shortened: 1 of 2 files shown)


splincode (approved):

projects/legacy/components/input-password/input-password.component.ts
```diff
@@ -67,7 +67,10 @@ export class TuiInputPasswordComponent
     }
 
     public get inputType(): string {
-        return this.isPasswordHidden || !this.interactive ? 'password' : 'text';
+        return this.isPasswordHidden ||
+            !(this.interactive || (this.readOnly && !this.computedDisabled))
```
waterplea, line 71: I believe interactive is a getter for not read-only && not disabled. Not computed disabled is enough to check. Same in template. Also please do this in both v3.x and main.

waterplea (approved):

PR #8589, opened 2024-08-18
refactor(core): minimizing reflows

splincode:


waterplea (approved):
Let's destructure the argument everywhere.

PR #8599, opened 2024-08-19
chore(demo): improve LCP

splincode:


waterplea:
Looks excessive

splincode:
@waterplea without inline image on SSR

image

waterplea:
> @waterplea without inline images on SSR
> 
> image
> 

But why is that? 🤔 Is it generating broken links?

splincode:
@waterplea SSR just gives out static

and we can inline images so that additional requests are not sent, and the picture is displayed immediately

PR #8601, opened 2024-08-19
refactor: post comments

splincode:


projects/core/directives/dropdown/dropdown-context.directive.ts
```diff
@@ -34,9 +34,9 @@ const MOVE_THRESHOLD = 15;
             'onTouchMove($event.touches[0].clientX, $event.touches[0].clientY)',
         '(touchstart.silent.passive)':
             'onTouchStart($event.touches[0].clientX, $event.touches[0].clientY)',
-        '(document:pointerdown.silent)': 'closeDropdown($event.target)',
-        '(document:contextmenu.capture.silent)': 'closeDropdown($event.target)',
-        '(document:keydown.esc)': 'closeDropdown($event.target)',
+        '(document:pointerdown.silent)': 'closeDropdown($event.currentTarget)',
```
waterplea, line 37: It was:
```ts
    @HostListener('document:pointerdown.silent', ['$event.target'])
    @HostListener('document:contextmenu.capture.silent', ['$event.target'])
    @HostListener('document:keydown.esc', ['$event.currentTarget'])
```
So `currentTarget` only on the last one.

projects/addon-doc/index.ts
```diff
@@ -29,7 +28,6 @@ export const TuiAddonDoc = [
     TuiDocPageTabConnector,
     TuiDocNavigation,
     TuiDocMain,
-    TuiDocText,
```
waterplea, line 32: Maybe we should deprecate and keep it? In case somebody was using it

waterplea (approved):

projects/core/directives/dropdown/dropdown-context.directive.ts
splincode: fixed

projects/addon-doc/index.ts
splincode: fixed

nsbarsukov (approved):

PR #8646, opened 2024-08-22
chore(demo): add example for `tuiIconsProvider`

splincode:
image

image


projects/demo/used-icons.ts
```diff
@@ -54,6 +54,8 @@ export const TUI_USED_ICONS = [
     '@tui.message-square',
     '@tui.info-filled',
     '@tui.info',
+    '@tui.badge-russian-ruble',
```
waterplea, line 57: Let's try to use icons we already use to simplify proprietary demo

projects/demo/src/modules/components/icon/examples/4/server.md
```diff
@@ -0,0 +1,17 @@
+```ts
+// ...
+
+const serverConfig = mergeApplicationConfig(config, {
+  providers: [
+    importProvidersFrom(ServerModule),
+    provideServerRendering(),
+    // ..
+    tuiIconsProvider({
```
waterplea, line 9: Let's fix the name of underlying token

waterplea (approved):

projects/demo/used-icons.ts
splincode: fixed

vladimirpotekhin (approved):

PR #8648, opened 2024-08-22
feat(layout): `Search` add new component

waterplea:
Closes #

### DIFF
--- a/projects/demo/src/modules/components/items-with-more/examples/1/index.ts
+++ b/projects/demo/src/modules/components/items-with-more/examples/1/index.ts
@@ -1,5 +1,5 @@
 import {NgForOf} from '@angular/common';
-import {Component} from '@angular/core';
+import {Component, inject} from '@angular/core';
 import {changeDetection} from '@demo/emulate/change-detection';
 import {encapsulation} from '@demo/emulate/encapsulation';
 import {TuiItem} from '@taiga-ui/cdk';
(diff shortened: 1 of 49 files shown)


splincode (approved):

splincode (approved):

projects/demo/src/modules/components/items-with-more/examples/1/index.ts
```diff
@@ -13,15 +13,7 @@ import {TuiChip, TuiItemsWithMore} from '@taiga-ui/kit';
...
-    protected readonly items = [
-        'John Cleese',
-        'Eric Idle',
-        'Graham Chapman',
-        'Michael Palin',
-        'Terry Gilliam',
-        'Terry Jones',
-    ];
-
+    protected readonly items = inject<readonly string[]>('Pythons' as any);
```
vladimirpotekhin, line 16: nit: what about token with a factory for `Pythons`? to avoid "as any"

projects/demo/src/modules/components/search/examples/import/template.md
```diff
@@ -0,0 +1,8 @@
+```html
+<header tuiNavigationHeader>Header content</header>
+<aside [tuiNavigationAside]="expanded">Sidebar content</aside>
+<main tuiNavigationMain>
+  <nav tuiNavigationNav>Breadcrumbs, tabs, buttons</nav>
+  Main content
+</main>
```
vladimirpotekhin, line 7: update md files

vladimirpotekhin (approved):

projects/demo/src/modules/components/items-with-more/examples/1/index.ts
waterplea: Then we would need to import it, I made it so it's specifically easier to open on StackBlitz. We can use token instead of string, but then we would need an import, if you can – you can add it later, I'm not really sure how to do it properly.

PR #8650, opened 2024-08-22
fix(cdk): `tuiMapper` should be stricter

splincode:


waterplea:
I believe it's on purpose to not break old projects with untyped mappers. Let's not change this in 3.x, since in 4.x it's already working as it should. I'm worried we might not take some case into consideration and inadvertently break somebody.

Issue #8644, opened 2024-08-22
🐞 - Reorder issue


Groxot:
### Playground Link

https://taiga-ui.dev/components/reorder

### Description

https://app.screencastify.com/v3/watch/yPpNO1D7HFDNXmM884XV

### Taiga UI version

4.3.0

### Which browsers have you used?

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

aleksey-mu:
FYI it's working fine in v3.92

PR #8663, opened 2024-08-22
fix(kit): handle pointer event

aleksey-mu:
Close https://github.com/taiga-family/taiga-ui/issues/8644

This issue is only presented in v4. When I compared it with v3.92 I noticed, that somehow in v4 `element` is coming as `PointerEvent` instead of `undefined` when we drag element out of container.
Because of that `filter` function returned `true` because of insufficient condition (`!!element`)
I suppose this issue might have occurred because of some changes in `ng-event-plugins`

| v3.92 (Angular 12.2.18) | v4.3 (Angular 16.2.14) |
| -------- | ------- |
| image | image |


splincode (approved):

projects/kit/components/tiles/tiles.component.ts
```diff
@@ -65,7 +65,12 @@ export class TuiTilesComponent {
     }
 
     private filter(element?: Element): element is Element {
-        return !!this.element && !!element && this.element !== element;
+        return (
+            !!this.element &&
+            !!element &&
+            this.element !== element &&
+            !(element instanceof PointerEvent)
```
splincode, line 72: `instanceof Event`

projects/kit/components/tiles/tiles.component.ts
```diff
@@ -65,7 +65,12 @@ export class TuiTilesComponent {
     }
 
     private filter(element?: Element): element is Element {
-        return !!this.element && !!element && this.element !== element;
```
waterplea, line 68: This is wrong, there should never be a `PointerEvent` here as an argument. The issue was cause by moving `@HostBinding` to `host` above, it used to not have any arguments and now it passes `$event` when it shouldn't. Please remove it from line 33

aleksey-mu: I see, it uses `onEnter()` from tile.
Fixed

waterplea (approved):

PR #8680, opened 2024-08-24
chore(demo): ignore white background

splincode:
# Before

https://github.com/user-attachments/assets/014cc62e-330b-4faf-904d-25087cc9a1f4

# After

https://github.com/user-attachments/assets/a4fb38fa-7324-41d4-bcd6-d5fca5e25cf0

### DIFF
--- a/projects/demo/src/styles/styles.less
+++ b/projects/demo/src/styles/styles.less
@@ -9,6 +9,10 @@ body {
     scroll-behavior: smooth;
     font: var(--tui-font-text-m);
 
+    &:has(tui-root[tuiTheme='dark']) {
+        background: #222;
+    }
+
     @media (prefers-reduced-motion) {
         scroll-behavior: auto;
     }


projects/core/styles/theme/palette.less
```diff
@@ -1,4 +1,5 @@
-[tuiTheme='dark'] {
+[tuiTheme='dark'],
+&:root:has(tui-root[tuiTheme='dark']) {
```
waterplea, line 2: Please only add this to the demo:
1) palette should not depend on where you put `tuiTheme` attribute
2) if you write it like that it will be broken for browsers not supporting `:has`, since they will skip entire rule if one of the selectors with `,` is not supported

splincode: fixed

waterplea (approved):

vladimirpotekhin (approved):

PR #8683, opened 2024-08-26
fix(kit): reorder tiles

splincode:


projects/kit/components/tiles/tiles.component.ts
```diff
@@ -30,7 +30,7 @@ import {BehaviorSubject, debounce, filter, map, Subject, timer} from 'rxjs';
     ],
     host: {
         '[class._dragged]': 'element',
-        '(pointerleave.silent)': 'rearrange($event)',
+        '(pointerleave.silent)': 'rearrange(element)',
```
waterplea, line 33: There should be nothing:
image

splincode: fixed

vladimirpotekhin (approved):

waterplea:
Merged external contributor PR since it was there earlier

PR #8756, opened 2024-08-29
feat(core): support dark palette for active document

splincode:
image

image


waterplea:
This would be breaking change, let's postpone this until v.5

PR #8757, opened 2024-08-29
feat(layout): `Navigation` add `Drawer`

waterplea:
Close #


splincode (approved):

nsbarsukov (approved):

hakimio:
@waterplea how does the new `drawer` component compares to the old `addon-mobile` `Sidebar`?

waterplea:
@hakimio new drawer is specifically for the burger type menu in `Navigation` set of components. We also plan to add a more general use right side drawer later. `Sidebar` from addon-mobile is more like a basic container.

Issue #476, opened 2021-06-09
[FEATURE] Cover components with harnesses


waterplea:
# 🚀 Feature request

To make it easy to test Taiga UI components and apps that use them we should provide [component harnesses](https://material.angular.io/cdk/test-harnesses/overview) for all our components.

waterplea:
This issue is rather big and it's fine to make many smaller PRs covering single component.

alvaromartmart:
I'm interested in gaining some experience with developing component harnesses so I'd be happy to contribute to this.
However, I'm not sure whether @zarghamkhandev is working on every one of them or tracking progress somehow. Please let me know if I can take one of the components 👀

waterplea:
> I'm interested in gaining some experience with developing component harnesses so I'd be happy to contribute to this.
> However, I'm not sure whether @zarghamkhandev is working on every one of them or tracking progress somehow. Please let me know if I can take one of the components 👀 

I believe he's not working on any more harnesses at the moment so just pick a component you'd like to try and let us know here in a comment so nobody else takes it.

zarghamkhandev:
@alvaromartmart feel free to take any component, I will take `input` components next. 🚀

alvaromartmart:
I'll start with the `action` component since it looks like a simple one, plus it doesn't have unit tests so I'll add them using the harness

alvaromartmart:
FYI I'll create a harness for the `filter` component

zarghamkhandev:
Instead of `input`, I will take `toggle` component next.

zarghamkhandev:
I'll write harness for `accordion` component next.

splincode:
@waterplea is it actual now? 
I suspect more efficient testing now is to use playwright or cypress

alvaromartmart:
> @waterplea is it actual now? I suspect more efficient testing now is to use playwright or cypress

my 2cents: they serve different purposes - component harnesses are useful for developers consuming Taiga UI in their unit/integration tests. Plus playwright doesn't (yet) support angular component testing.

splincode:
@alvaromartmart maybe, but lately I don't write unit tests anymore, as all issues are closed by playwright e2e and cypress component testing

alvaromartmart:
> @alvaromartmart maybe, but lately I don't write unit tests anymore, as all issues are closed by playwright e2e and cypress component testing

I understand that, by are you speaking as a taiga-ui maintainer or as an end user?
[...]

PR #8760, opened 2024-08-30
feat(testing): adds size harness for button

mohan-mu:
Partially implements https://github.com/taiga-family/taiga-ui/issues/476

### DIFF
--- a/projects/testing/core/button.harness.ts
+++ b/projects/testing/core/button.harness.ts
@@ -16,7 +16,7 @@ export class TuiButtonHarness extends TuiComponentHarness {
         return (await this.host()).hasClass(input);
     }
 
-    public async getSize(): Promise<string | null> {
-        return (await this.host()).getAttribute('data-size');
+    public async getSize(): Promise<string> {
+        return (await (await this.host()).getAttribute('data-size')) ?? '';
     }
 }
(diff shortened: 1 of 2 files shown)


splincode (approved):

projects/testing/core/button.harness.ts
```diff
@@ -15,4 +15,8 @@ export class TuiButtonHarness extends TuiComponentHarness {
     public async hasClass(input: string): Promise<boolean> {
         return (await this.host()).hasClass(input);
     }
+
+    public async getSize(): Promise<string | null> {
```
waterplea, line 19: Size is never `null`, I think we can add `!` there

waterplea (approved):

projects/testing/core/button.harness.ts
mohan-mu: Sorry I didn't understand `we can add ! there` !  

 I have to enfore `null` because of `getAttribute`  return type. 

`getAttribute(name: string): Promise<string | null>;`

waterplea: `return (await this.host()).getAttribute('data-size')!;`

waterplea: I meant adding `!` at the end. I know it might return `null`, but the button is designed in such a way that this is never possible, so we shouldn't make testing buttons harder by having them to work around `null` there, since it's never going to happen.

splincode: Or just

```
return (await this.host()).getAttribute('data-size') ?? '';
```

mohan-mu: Got it Done

PR #8786, opened 2024-09-02
feat(kit): `InputPassword` add new version, deprecate legacy

waterplea:
Close #


splincode (approved):

projects/legacy/components/input-password/input-password.module.ts
```diff
@@ -14,6 +14,9 @@ import {PolymorpheusOutlet, PolymorpheusTemplate} from '@taiga-ui/polymorpheus';
 import {TuiInputPasswordComponent} from './input-password.component';
 import {TuiInputPasswordDirective} from './input-password.directive';
 
+/**
+ * @deprecated use {@link TuiInputPassword} instead
```
nsbarsukov, line 18: nitpicking:
```suggestion
 * @deprecated use {@link TuiInputPassword} from {@link @taiga-ui/kit} instead
```

nsbarsukov (approved):

projects/legacy/components/input-password/input-password.module.ts
waterplea: WebStorm clicks through automatically

PR #8787, opened 2024-09-02
chore(demo): drop tinkoff word

splincode:


waterplea (approved):

vladimirpotekhin (approved):

projects/demo/src/modules/components/breadcrumbs/examples/2/index.ts
```diff
@@ -41,8 +41,8 @@ export default class Example {
             link: 'https://github.com/topics/angular',
         },
         {
-            caption: 'Tinkoff',
-            link: 'https://github.com/tinkoff',
+            caption: 'TramvaiJS',
```
vladimirpotekhin, line 44: nit: `Angular > tramvai ? > Taiga > Components` 
It seems that the Tramvai doesn't fit the logic
maybe it's better to find other word, or just drop this

PR #8801, opened 2024-09-03
feat(layout): `Cell` add border-radius

mdlufy:
Close #

### DIFF
--- a/projects/layout/components/cell/cell.styles.less
+++ b/projects/layout/components/cell/cell.styles.less
@@ -1,8 +1,8 @@
 @import '@taiga-ui/core/styles/taiga-ui-local';
 
 [tuiCell] {
-    --tui-height: calc(var(--tui-height-s) - 0.5rem);
     --t-padding: 0.125rem 1rem;
+    --t-radius: var(--tui-radius-s);
 
     .button-clear();
 
@@ -14,6 +14,7 @@
     color: var(--tui-text-primary);
     padding: var(--t-padding);
     min-block-size: var(--t-height);
+    border-radius: var(--t-radius);
 
     &:disabled {
         pointer-events: none;
@@ -139,6 +140,7 @@
     &[data-size='l'] {
         --t-height: calc(var(--tui-height-l) - 1rem);
         --t-padding: 0.5rem 1rem;
+        --t-radius: var(--tui-radius-l);
 
         & > *:not(:first-child),
         & > [tuiAccessories] > *:not(:first-child) {


splincode (approved):

projects/layout/components/cell/cell.styles.less
```diff
@@ -81,6 +81,10 @@
     }
 
     &[data-size='s'] {
+        &:hover {
```
vladimirpotekhin, line 84: I think we need to add borders to the cell, not just for hover state.

projects/layout/components/cell/cell.styles.less
```diff
@@ -81,6 +81,10 @@
...
+            border-radius: 0.5rem;
```
vladimirpotekhin, line 85: probably we can use approach as below with --t-height, --t-padding
like 
`--t-radius: ...`

also please check the proprietary radiuses, maybe they fit the layout

projects/layout/components/cell/cell.styles.less
```diff
@@ -137,6 +141,11 @@
     }
 
     &[data-size='l'] {
+        border-radius: var(--tui-radius-l);
+
+        --tui-radius-s: 0.5rem;
```
waterplea, line 146: We shouldn't override those variables.

mdlufy: yes, it was for debug ))

projects/layout/components/cell/cell.styles.less
mdlufy: fixed

projects/layout/components/cell/cell.styles.less
mdlufy: fixed

vladimirpotekhin (approved):

projects/layout/components/cell/cell.styles.less
```diff
@@ -3,6 +3,7 @@
 [tuiCell] {
     --tui-height: calc(var(--tui-height-s) - 0.5rem);
```
waterplea, line 4: Is this used anywhere?

waterplea (approved):

projects/layout/components/cell/cell.styles.less
mdlufy: No, it used `--t-height` only
It's misspel and maybe rename `--tui-height` to `--t-height` on host OR just remove `--tui-height` ?

image
image

PR #8823, opened 2024-09-04
chore(demo): `Appearance` add all options

waterplea:
Close #


projects/core/directives/appearance/appearance.options.ts
```diff
@@ -1,8 +1,31 @@
...
+    | 'whiteblock'
+    | 'outline'
+    | 'flat'
+    | 'destructive'
+    | 'neutral'
+    | 'error'
+    | 'success'
+    | 'warning'
+    | 'info'
+    | ({} & string);
```
splincode, line 25: `Record<never, never>`

splincode (approved):

projects/core/directives/appearance/appearance.options.ts
waterplea, line 25:
```suggestion
    | (Record<never, never> & string);
```

vladimirpotekhin (approved):

PR #8825, opened 2024-09-04
feat(kit): `Segmented` allow vertical orientation

waterplea:
Close #


projects/kit/components/segmented/segmented.style.less
```diff
@@ -12,13 +12,13 @@ tui-segmented {
         --t-padding: 0.625rem;
         --t-gap: 0.5rem;
         --t-margin: -0.125rem;
+        --t-height: var(--tui-height-s);
```
vladimirpotekhin, line 15: Are screenshots valid?

vladimirpotekhin (approved):

projects/kit/components/segmented/segmented.style.less
waterplea: No, my mistake. I'll fix it and address design changes in next commit

PR #8830, opened 2024-09-04
feat(addon-doc): adds keybindings to the search field

TheKenkei:


projects/addon-doc/components/navigation/navigation.component.ts
```diff
@@ -165,6 +164,16 @@ export class TuiDocNavigation {
         this.openActivePageGroup();
     }
 
+    protected onFocusSearch(event: KeyboardEvent): void {
+        if (
+            event.code === 'Slash' &&
```
waterplea, line 169: Would `(window:keydown.slash)` work?

projects/addon-doc/components/navigation/navigation.component.ts
```diff
@@ -165,6 +164,16 @@ export class TuiDocNavigation {
...
+            !(this.doc.activeElement instanceof HTMLInputElement)
```
waterplea, line 170: Let's check that nothing is focused instead, `this.doc.activeElement === this.doc.body`, just in case.

projects/addon-doc/components/navigation/navigation.component.ts
TheKenkei: No, I put together a [stackblitz ](https://stackblitz.com/edit/angular-6xnn6k?file=src%2Fapp%2Fapp.component.ts)to demonstrate

projects/addon-doc/components/navigation/navigation.component.ts
TheKenkei: Then we will lose the ability to quickly switch to the search if, for example, a button is in focus. For example, when clicking on Typescript here. Do you really want this?
image

waterplea: Ok, it's just a demo portal anyway, I doubt there would be any problems because of this

waterplea (approved):

splincode (approved):

PR #8910, opened 2024-09-09
chore: add maskito and new textfield example

splincode:


projects/demo/src/modules/components/textfield/examples/4/index.html
```diff
@@ -0,0 +1,11 @@
+<tui-textfield>
+    <strong>π</strong>
+    -value
+
+    <input
+        placeholder="3,14"
+        tuiTextfield
+        [maskito]="options"
+        [(ngModel)]="value"
+    />
```
vladimirpotekhin, line 10: example doesn't work properly
caret jumps
backspace doesn't work

https://github.com/user-attachments/assets/94e3dd1c-d49b-4f30-9f19-289fd26dfaa3

projects/demo/src/modules/components/textfield/examples/4/index.ts
```diff
@@ -0,0 +1,25 @@
...
+    imports: [FormsModule, MaskitoDirective, TuiIcon, TuiTextfield, TuiTooltip],
+    templateUrl: './index.html',
+    styleUrls: ['./index.less'],
+    encapsulation,
+    changeDetection,
+})
+export default class Example {
+    protected value = Math.PI.toFixed(48);
+    protected readonly options = maskitoNumberOptionsGenerator({
+        decimalSeparator: ', ',
```
vladimirpotekhin, line 21: decimal separator should not be more than 1 character, I think that's the problem with the example

projects/demo/src/modules/components/textfield/examples/4/index.ts
```diff
@@ -0,0 +1,25 @@
...
+import {changeDetection} from '@demo/emulate/change-detection';
+import {encapsulation} from '@demo/emulate/encapsulation';
+import {MaskitoDirective} from '@maskito/angular';
+import {maskitoNumberOptionsGenerator} from '@maskito/kit';
+import {TuiIcon, TuiTextfield} from '@taiga-ui/core';
+import {TuiTooltip} from '@taiga-ui/kit';
+
+@Component({
+    standalone: true,
+    imports: [FormsModule, MaskitoDirective, TuiIcon, TuiTextfield, TuiTooltip],
```
vladimirpotekhin, line 12: TuiTooltip is unused?

projects/demo/src/modules/components/textfield/examples/4/index.ts
splincode: fixed

projects/demo/src/modules/components/textfield/examples/4/index.ts
splincode: fixed

vladimirpotekhin (approved):

mdlufy (approved):

projects/demo/src/modules/components/textfield/index.html
```diff
@@ -15,6 +15,7 @@
             *ngFor="let example of examples; let index = index"
             [component]="index + 1 | tuiComponent"
             [content]="index + 1 | tuiExample"
+            [fullsize]="index === 3"
```
waterplea, line 18: Maybe make it short instead? Long inputs do not look that good, there's no need for such a big number

waterplea (approved):

projects/demo/src/modules/components/textfield/index.html
splincode: fixed

PR #8912, opened 2024-09-09
fix(kit): add overload for `TuiFilterByInputPipe`

### LINKED ISSUE #8844: 🐞 - `TuiFilterByInput`: variable of type `TuiStringMatcher<T>` is not assignable as `TuiFilterByInput` pipe argument
### Playground Link

_No response_

### Description

Trying to do the following:
```html
[items]="items | tuiFilterByInput: itemMatcher"
```
where item matcher is defined as the following:
```ts
protected readonly itemMatcher: TuiStringMatcher<T> =
        (item: T, search: string): boolean => (item[this.displayProperty] as string).toLowerCase().startsWith(search.toLowerCase());
```
results in `Angular` compiler type error when Angular `strictTemplates` option is enabled.

### Angular version

18.2

### Taiga UI version

4

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
Close #8844

### DIFF
--- a/projects/demo/src/modules/pipes/filter-by-input/examples/2/index.ts
+++ b/projects/demo/src/modules/pipes/filter-by-input/examples/2/index.ts
@@ -2,6 +2,7 @@ import {Component, inject} from '@angular/core';
 import {FormControl, FormGroup, ReactiveFormsModule} from '@angular/forms';
 import {changeDetection} from '@demo/emulate/change-detection';
 import {encapsulation} from '@demo/emulate/encapsulation';
+import type {TuiStringMatcher} from '@taiga-ui/cdk';
 import {
     TuiDataListWrapper,
     TuiFilterByInputPipe,
@@ -27,7 +28,7 @@ interface User {
     encapsulation,
     changeDetection,
 })
-export default class Example {
+export default class Example<T extends User> {
     protected readonly items = inject<readonly string[]>('Pythons' as any);
 
     protected readonly users = [
@@ -36,7 +37,7 @@ export default class Example {
         {id: 3, name: 'Graham Chapman'},
         {id: 4, name: 'Michael Palin'},
         {id: 5, name: 'Terry Gilliam'},
-    ];
+    ] as T[]; // only template compile check
 
     protected readonly form = new FormGroup({
         user: new FormControl<User | null>(null),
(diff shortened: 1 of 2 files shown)


projects/kit/pipes/filter-by-input/filter-by-input.pipe.ts
```diff
@@ -25,6 +25,7 @@ export class TuiFilterByInputPipe implements PipeTransform {
     private readonly textfield = inject(TuiTextfieldComponent, {optional: true});
     private readonly host = inject(TUI_DATA_LIST_HOST);
 
+    public transform<T>(items: readonly T[], matcher?: TuiStringMatcher<T>): readonly T[];
     public transform<T>(items: T, matcher?: TuiStringMatcher<TuiArrayElement<T>>): T;
```
hakimio, line 29: Do we really need the second one with `TuiArrayElement`? Does anything break if you only leave the new overload?

projects/demo/src/modules/pipes/filter-by-input/examples/2/index.ts
```diff
@@ -27,7 +28,7 @@ interface User {
     encapsulation,
     changeDetection,
 })
-export default class Example {
+export default class Example<T extends User> {
```
hakimio, line 31: In this case it could also be `T extends User = User`. Then it might not be necessary to add `as T[]`.

splincode: image

projects/kit/pipes/filter-by-input/filter-by-input.pipe.ts
splincode: fixed

projects/demo/src/modules/pipes/filter-by-input/examples/2/index.ts
hakimio: Ok, nvm. Bad assumption 🙂

hakimio (approved):

mdlufy (approved):

PR #8913, opened 2024-09-09
refactor: drop async pipe in `tui-dialogs`

splincode:


projects/core/components/dialog/dialogs.component.ts
```diff
@@ -33,6 +33,9 @@ import {TUI_DIALOGS} from './dialog.tokens';
 })
 export class TuiDialogs {
     protected readonly el = tuiInjectElement();
-    protected readonly dialogs$: Observable<ReadonlyArray<TuiPopover<any, any>>> =
-        inject(TUI_DIALOGS);
+    protected readonly dialogs = toSignal<ReadonlyArray<TuiPopover<any, any>>>(
```
vladimirpotekhin, line 36: it looks like the overlay is now blinking

https://github.com/user-attachments/assets/fb2a8eba-4769-4c7e-bfdf-34740f51fec1

splincode: fixed

vladimirpotekhin (approved):

projects/core/components/dialog/dialogs.template.html
```diff
@@ -1,4 +1,4 @@
-<ng-container *ngIf="dialogs$ | async as dialogs">
+<ng-container *ngIf="dialogs() as dialogs">
```
waterplea, line 1: Let's drop the container and just use `dialogs()` multiple times.

splincode: @waterplea https://github.com/taiga-family/taiga-ui/pull/8913#discussion_r1750107775

waterplea (approved):

PR #8914, opened 2024-09-09
chore(demo): add line clamp example with cdk virtual scroll

splincode:


mdlufy (approved):

projects/demo/src/modules/components/line-clamp/examples/6/index.ts
```diff
@@ -0,0 +1,220 @@
...
+
+@Component({
+    standalone: true,
+    imports: [
+        AsyncPipe,
+        CdkFixedSizeVirtualScroll,
+        CdkVirtualForOf,
+        CdkVirtualScrollViewport,
+        NgIf,
+        TuiChip,
```
vladimirpotekhin, line 21: unused ? and TuiNotification

vladimirpotekhin (approved):

projects/demo/src/modules/components/line-clamp/examples/6/index.ts
splincode: fixed

PR #8928, opened 2024-09-10
refactor: support `use-unknown-in-catch-callback-variable`

splincode:


projects/demo/src/main.ts
```diff
@@ -9,4 +9,4 @@ if (environment.production) {
     enableProdMode();
 }
 
-bootstrapApplication(App, config).catch((err) => console.error(err));
+bootstrapApplication(App, config).catch((err: unknown) => console.error(err));
```
waterplea, line 12: catch(console.error)?

waterplea (approved):

vladimirpotekhin (approved):

projects/demo/src/main.ts
splincode: image

by default `catch` throw `any`

image

PR #8937, opened 2024-09-10
refactor: support `single-character-alternation`

splincode:
image


splincode:
@waterplea just
image
image

vladimirpotekhin (approved):

Issue #3925, opened 2023-03-20
📚 - tuiTableFilters doesn't work with FormArray


Killjoyer:
### Which @taiga-ui/* package(s) are the source of the bug?

addon-table

### Please provide a link to a minimal reproduction of the bug

https://stackblitz.com/edit/angular-4tmzqc-gaann9?file=src/app/app.component.ts

### Is this issue blocking you?

Blocking

### Description

In the example "add" button modifies inner structure, but table won't update. It works without " | tuiTableFilters | async" (it errors, but still shows new rows).

### Angular version

12.2.17

### Taiga UI version

3.20.0

### Which browsers have you used?

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Which operating systems have you used?

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

waterplea:
Per personal discussion it was determined it is not a bug in Taiga UI but rather a configuration issue. But we need to add an example to the demo so people can use it more easily.

gennadiipetrov:
Hello. I will take it.

PR #8951, opened 2024-09-10
chore(demo): `table-filters` add example with form array

gennadiipetrov:
An example for FormArray was added to TableFilters(AddonTable)

Close # [3925](https://github.com/taiga-family/taiga-ui/issues/3925)


splincode:
@gennadiipetrov please run `npx eslint . --fix`
image

splincode (approved):

PR #8954, opened 2024-09-10
refactor: support `consistent-destructuring`

splincode:


mdlufy (approved):

projects/addon-commerce/components/input-card-group/input-card-group.component.ts
```diff
@@ -194,7 +194,7 @@ export class TuiInputCardGroup
     }
 
     public override writeValue(value: TuiCard | null): void {
-        const {bin} = this;
+        const bin = this.bin;
```
waterplea, line 197: Why not?

splincode: fixed

PR #8974, opened 2024-09-11
fix(cdk): remove global styles patching from auto-focus directive

vladimirpotekhin:
Close #


splincode (approved):

projects/demo/src/modules/components/dialog/examples/7/search-example/search-dialog-example.component.ts
```diff
@@ -11,6 +11,7 @@ import {injectContext} from '@taiga-ui/polymorpheus';
     templateUrl: './search-dialog-example.template.html',
     styleUrls: ['./search-dialog-example.component.less'],
     changeDetection,
+    providers: [tuiAutoFocusOptionsProvider({preventScroll: true})],
```
splincode, line 14: image

add info please

projects/cdk/directives/auto-focus/handlers/default.handler.ts
```diff
@@ -29,7 +29,9 @@ export class TuiDefaultAutofocusHandler extends AbstractTuiAutofocusHandler {
                     skipWhile(Boolean),
                     take(1),
                 ),
-            ).subscribe(() => this.element.focus({preventScroll: true}));
+            ).subscribe(() =>
+                this.element.focus({preventScroll: this.options.preventScroll}),
+            );
         } else {
             this.element.focus({preventScroll: true});
```
splincode, line 36: Should we add `{preventScroll: this.options.preventScroll}` for `this.element.focus({preventScroll: true});`?

vladimirpotekhin: no, that would be a breaking change for the buttons

projects/core/styles/mixins/mixins.less
```diff
@@ -129,3 +129,20 @@
     overflow: hidden;
     padding: 0;
 }
+
+// prevent scrolling to focused element on IOS
+.tui-prevent-ios-scroll() {
```
mdlufy, line 134: Why we need prevent scroll?

vladimirpotekhin: This is for fixed textfields like searchbar. IOS Safari tries to scroll to the textfield and the interface shakes

PR #8989, opened 2024-09-12
feat(kit): `Like` add new component

### LINKED ISSUE #8472: 🚀 - Create directive `tuiLike`
### Description

Based on a `Checkbox` with an appearance on the host, with a custom color for the icon if `:checked`. 
With icons from DI with an example in the demo of replacing with a star, it is possible that the inputs to the icons should be added to make it easier to redefine. 

image

mdlufy:
Close #8472

### DIFF
--- a/projects/demo/src/modules/components/like/examples/import/template.md
+++ b/projects/demo/src/modules/components/like/examples/import/template.md
@@ -1,0 +1,6 @@
+```html
+<input
+  tuiLike
+  type="checkbox"
+/>
+```
--- a/projects/kit/components/like/like.component.ts
+++ b/projects/kit/components/like/like.component.ts
@@ -1,0 +1,60 @@
+import {
+    ChangeDetectionStrategy,
+    Component,
+    inject,
+    Input,
+    ViewEncapsulation,
+} from '@angular/core';
+import type {TuiStringHandler} from '@taiga-ui/cdk/types';
+import {tuiIsString} from '@taiga-ui/cdk/utils/miscellaneous';
+import {
+    tuiAppearanceOptionsProvider,
+    TuiWithAppearance,
+} from '@taiga-ui/core/directives/appearance';
+import {tuiInjectIconResolver} from '@taiga-ui/core/tokens';
+import type {TuiSizeS} from '@taiga-ui/core/types';
+
+import {TUI_LIKE_OPTIONS, type TuiLikeOptions} from './like.options';
+
+@Component({
+    standalone: true,
+    selector: 'input[tuiLike][type=checkbox]',
+    template: '',
+    styles: ['@import "@taiga-ui/kit/styles/components/like.less";'],
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    providers: [tuiAppearanceOptionsProvider(TUI_LIKE_OPTIONS)],
+    hostDirectives: [TuiWithAppearance],
+    host: {
+        tuiLike: '',
+        '[attr.data-size]': 'size',
+        '[attr.data-mode]': '""',
+        '[style.--t-icon-color]': 'color',
+        '[style.--t-unchecked-icon]': 'getIcon("unchecked")',
+        '[style.--t-checked-icon]': 'getIcon("checked")',
+    },
+})
+export class TuiLike {
+    private readonly options = inject(TUI_LIKE_OPTIONS);
+    private readonly resolver = tuiInjectIconResolver();
+
+    @Input('tuiLike')
+    public color = '';
+
+    @Input()
+    public uncheckedIcon: TuiStringHandler<TuiSizeS> | string =
+        this.options.icons.unchecked;
+
+    @Input()
+    public checkedIcon: TuiStringHandler<TuiSizeS> | string = this.options.icons.checked;
+
+    @Input()
+    public size: TuiSizeS = this.options.size;
+
+    protected getIcon(state: keyof TuiLikeOptions['icons']): string {
+        const option = state === 'checked' ? this.checkedIcon : this.uncheckedIcon;
+        const icon = tuiIsString(option) ? option : option(this.size);
+
+        return icon && `url(${this.resolver(icon)})`;
+    }
+}
--- a/projects/kit/components/like/like.options.ts
+++ b/projects/kit/components/like/like.options.ts
@@ -1,0 +1,22 @@
+import type {TuiStringHandler} from '@taiga-ui/cdk/types';
+import {tuiCreateOptions} from '@taiga-ui/cdk/utils/di';
+import type {TuiAppearanceOptions} from '@taiga-ui/core/directives';
+import type {TuiSizeS} from '@taiga-ui/core/types';
+
+export interface TuiLikeOptions extends TuiAppearanceOptions {
+    readonly size: TuiSizeS;
+    readonly icons: Readonly<{
+        checked: TuiStringHandler<TuiSizeS> | string;
+        unchecked: TuiStringHandler<TuiSizeS> | string;
+    }>;
+}
+
+export const [TUI_LIKE_OPTIONS, tuiLikeOptionsProvider] =
+    tuiCreateOptions<TuiLikeOptions>({
+        size: 'm',
+        appearance: 'secondary',
+        icons: {
+            unchecked: '@tui.heart',
+            checked: '@tui.heart-filled',
+        },
+    });
--- a/projects/cdk/constants/used-icons.ts
+++ b/projects/cdk/constants/used-icons.ts
@@ -46,6 +46,8 @@ export const TUI_USED_ICONS = [
     '@tui.file',
     '@tui.trash',
     '@tui.phone',
+    '@tui.heart',
+    '@tui.heart-filled',
     '@tui.star',
     '@tui.calendar',
     '@tui.rotate-ccw-square',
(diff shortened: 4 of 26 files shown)


splincode (approved):

mdlufy:
Maked based on screenshots from #8472 

If use `TuiAppearance` secondary on host, it will be:

https://github.com/user-attachments/assets/a52c9904-a0f6-42ea-85fd-60f82482ed80

mdlufy:
Migrated to `data-mode="''"`

https://github.com/user-attachments/assets/db4cb226-207f-48c6-a702-b2e97afadeb1

projects/kit/components/like/like.options.ts
```diff
@@ -0,0 +1,28 @@
...
+    appearance: 'secondary',
+    icons: {
+        default: '@tui.heart',
+        checked: '@tui.heart-filled',
+    },
+};
+
+export const TUI_LIKE_OPTIONS = tuiCreateToken(TUI_LIKE_DEFAULT_OPTIONS);
+
+export function tuiLikeOptionsProvider(options: Partial<TuiLikeOptions>): Provider {
```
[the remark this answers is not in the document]
mdlufy: done

projects/kit/components/like/like.options.ts
```diff
@@ -0,0 +1,28 @@
...
+import type {TuiStringHandler} from '@taiga-ui/cdk/types';
+import {tuiCreateToken, tuiProvideOptions} from '@taiga-ui/cdk/utils/miscellaneous';
+import type {TuiAppearanceOptions} from '@taiga-ui/core/directives';
+import type {TuiSizeS} from '@taiga-ui/core/types';
+
+export interface TuiLikeOptions extends TuiAppearanceOptions {
+    readonly size: TuiSizeS;
+    readonly icons: Readonly<{
+        checked: TuiStringHandler<TuiSizeS> | string;
+        default: TuiStringHandler<TuiSizeS> | string;
```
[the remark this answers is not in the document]
mdlufy: good idea, done

projects/demo/src/modules/components/like/examples/import/template.md
```diff
@@ -0,0 +1,6 @@
+```html
+<input
+  tuiLike
+  type="checkbox"
```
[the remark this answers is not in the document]
mdlufy: Yes, we can. I think if we don't need to redefine `CheckboxControlValueAccessor`, it will be working nice

projects/kit/components/like/like.style.less
```diff
@@ -0,0 +1,64 @@
...
+        content: '';
+        display: block;
+        background: currentColor;
+        mask: var(--t-default-icon) no-repeat center/contain;
+        transition:
+            transform var(--tui-duration) ease-in-out,
+            mask 0s var(--tui-duration) ease-in-out;
+    }
+
+    &:checked::before {
```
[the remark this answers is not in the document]
mdlufy: good idea, done and added animation

projects/demo/src/modules/components/like/examples/import/template.md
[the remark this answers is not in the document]
mdlufy: I checked out, it looks like `'[attr.type]': '"checkbox"'` not works correctly with NgModel (

With manual `type=checkbox`:

```
<input
    tuiLike="var(--tui-status-negative)"
    type="checkbox"
    [checked]="value"
    [(ngModel)]="value"
/>
```

https://github.com/user-attachments/assets/bfddc97b-4c7b-4849-bd73-66189f8027e3



With `type=checkbox` on host:
```
host: {
        '[attr.type]': '"checkbox"',
        ...
    },
```

```
<input
    tuiLike="var(--tui-status-negative)"
    [checked]="value"
    [(ngModel)]="value"
/>
```

https://github.com/user-attachments/assets/05975c75-3ccd-4bae-8964-6791a8510aa1

projects/kit/components/like/like.style.less
```diff
@@ -0,0 +1,88 @@
...
+ * data-size — size (default: 'm') ('s' | 'm')
+ *
+ * @vars
+ * --t-icon-color - custom color for `:checked` state
+ * --t-unchecked-icon — default state icon
+ * --t-checked-icon — checkmark icon
+ *
+ * @example
+ * <input tuiLike />
+ * <input tuiLike="var(--tui-status-negative)" />
```
waterplea, line 34: Let's not use this example because this is done with Angular, not with pure CSS

projects/kit/components/like/like.component.ts
```diff
@@ -0,0 +1,51 @@
...
+        '[style.--t-unchecked-icon]': 'getIcon("unchecked")',
+        '[style.--t-checked-icon]': 'getIcon("checked")',
+    },
+})
+export class TuiLike {
+    private readonly options = inject(TUI_LIKE_OPTIONS);
+    private readonly resolver = tuiInjectIconResolver();
+
+    @Input('tuiLike')
+    public iconColor: string | '' = '';
```
waterplea, line 40: Can probably be just `color` for brevity

projects/kit/components/like/like.style.less
```diff
@@ -0,0 +1,88 @@
+@import '@taiga-ui/core/styles/taiga-ui-local';
+
+@keyframes tuiLikeScale {
```
waterplea, line 3: We should we move this file to /styles/components

projects/kit/components/like/like.style.less
```diff
@@ -0,0 +1,88 @@
...
+        &::before {
+            opacity: 0;
+        }
+
+        &::after {
+            opacity: 1;
+            animation: tuiLikeScale var(--tui-duration) forwards;
+        }
+
+        &:not([tuiLike=''])::after {
```
waterplea, line 79: I'm not sure this attribute is still there in prod build, that's why we have `:host { tuiButton: '' }` in all such cases. Maybe just `color: var(--t-icon-color, inherit)`?

projects/kit/components/like/like.style.less
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    &:checked {
+        &::before {
+            opacity: 0;
+        }
+
+        &::after {
+            opacity: 1;
+            animation: tuiLikeScale var(--tui-duration) forwards;
```
waterplea, line 76: I think we do not need the animation to bounce like that, we can do with with easing and transition:
https://easings.net

projects/kit/components/like/like.style.less
```diff
@@ -0,0 +1,88 @@
...
+ * Checkbox, Switch, Radio, Appearance
+ */
+[tuiLike] {
+    --t-size: var(--tui-height-m);
+    --t-border-width: 0.75rem;
+
+    inline-size: var(--t-size);
+    block-size: var(--t-size);
+    color: var(--tui-text-action);
+    background-color: var(--tui-background-neutral-2);
```
waterplea, line 46: Shouldn't it come from appearance?

projects/kit/components/like/like.style.less
```diff
@@ -0,0 +1,88 @@
...
+    margin: 0;
+    flex-shrink: 0;
+
+    &::before,
+    &::after {
+        .fullsize();
+        .transition(~'transform, mask, opacity, color');
+
+        content: '';
+        display: block;
```
waterplea, line 59: Unnecessary since we have `.fullsize();`

projects/kit/components/like/like.style.less
```diff
@@ -0,0 +1,88 @@
...
+    border: var(--t-border-width) transparent solid;
+    border-radius: 100%;
+    cursor: pointer;
+    margin: 0;
+    flex-shrink: 0;
+
+    &::before,
+    &::after {
+        .fullsize();
+        .transition(~'transform, mask, opacity, color');
```
waterplea, line 56: I don't think mask needs transition here

projects/kit/components/like/like.style.less
mdlufy: Okay, but why are we doing this?

projects/kit/components/like/like.style.less
mdlufy: `color: var(--t-icon-color, inherit)` unfortunately not working due to css priority.

image


 It seems like, we don't attach `color` to prevent such behaviour or use current solution with `&:not([tuiLike=''])`. Are there another options?

projects/kit/components/like/like.component.ts
mdlufy: done

projects/kit/components/like/like.style.less
mdlufy: removed

projects/kit/components/like/like.style.less
mdlufy: Yes, it's unnesessary, removed

projects/kit/components/like/like.style.less
mdlufy: yes, removed

projects/kit/components/like/like.style.less
mdlufy: removed

projects/kit/components/like/like.style.less
mdlufy: Without scaling, only transition for opacity and color ? It look nice with scaling

mdlufy: So far, I've removed scaling, will we do it?

vladimirpotekhin (approved):

projects/kit/components/like/like.style.less
waterplea: These are pure CSS components that rely on Angular only for quality of life such as attributes binding and icons resolving. But they can work without Angular and can be shared later with the react team.

waterplea (approved):

PR #9001, opened 2024-09-12
fix(kit): block user selection in skeleton

### LINKED ISSUE #9000: 🐞 - [Skeleton] Should block content selection in a Skeleton
### Playground Link

https://taiga-ui.dev/directives/skeleton

### Description

**Current behavior:**
- Try to select the content (text/images) across the skeleton section.
- Copy with right-click context menu or <kbd>Ctrl</kbd>/<kbd>Cmd</kbd> + <kbd>C</kbd>.
- We are able to select and copy content which is invisible and should be inert.

**Expected behavior:** User should not be able to interact with hidden/loading content.

**Suggested fixes:**
- A simple fix is to add [`user-select: none`](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select). This is unfortunately not supported on Safari yet, but should be fine as a progressive enhancement.
- An additional/complete fix would be to add [`pointer-events: none`](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events).
- Additionally, we should have the `aria-busy` attribute, but there are nuances here for which I have raised a separate issue: #8999

### Angular version

Any/Latest

### Taiga UI version

Latest

- [ ] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

ajitzero:
Close #9000

### DIFF
--- a/projects/kit/directives/skeleton/skeleton.style.less
+++ b/projects/kit/directives/skeleton/skeleton.style.less
@@ -17,6 +17,8 @@
         box-shadow: none !important;
         filter: contrast(0) brightness(0);
         animation: tuiSkeleton ease-in-out 1s infinite alternate;
+        user-select: none;
+        pointer-events: none;
 
         &[data-tui-skeleton] {
             background: transparent !important;


splincode (approved):

projects/kit/directives/skeleton/skeleton.style.less
```diff
@@ -17,16 +17,22 @@
         box-shadow: none !important;
         filter: contrast(0) brightness(0);
         animation: tuiSkeleton ease-in-out 1s infinite alternate;
+        user-select: none;
+        pointer-events: none;
 
         &[data-tui-skeleton] {
             background: transparent !important;
+            user-select: none;
```
waterplea, line 25: I don't think we need it here, it's the same element that already has those rules. Also, I believe, they are inherited by `::before` so it's not needed there too.

waterplea (approved):

projects/kit/directives/skeleton/skeleton.style.less
ajitzero: I was thinking the same about this tag. Let me verify the `::before thing`

ajitzero: Updated! Please review again @waterplea 
(Force-pushed again to correct scope for commitlint)

PR #9022, opened 2024-09-14
fix(cdk): unsubscribe from swipe service when directive is destroyed

splincode:


projects/cdk/directives/swipe/swipe.service.ts
```diff
@@ -52,6 +54,7 @@ export class TuiSwipeService extends Observable<TuiSwipeEvent> {
                         return null;
                     }),
                     filter(tuiIsPresent),
+                    takeUntilDestroyed(destroyRef),
```
vladimirpotekhin, line 57: Looks like we don't need to use destroyRef in observable constructor. It's enough to return subscription (as TearDown Logic)
same for pan service

splincode: fixed

vladimirpotekhin (approved):

mdlufy (approved):

PR #9034, opened 2024-09-15
fix(layout): `CardLarge` remove `Cell` negative margins and fix by spec

mdlufy:
Close #


splincode (approved):

mdlufy:
Сlosed in favor https://github.com/taiga-family/taiga-ui/pull/9040

PR #9048, opened 2024-09-16
chore(demo): add algolia docsearch

vladimirpotekhin:
Close #

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -28,7 +28,8 @@
         "svgs",
         "contentalign",
         "inputpassword",
-        "inputcopy"
+        "inputcopy",
+        "docsearch"
     ],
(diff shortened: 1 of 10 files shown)


vladimirpotekhin:
I suggest merging with the bundlemon warning now. In the future, if the search works fine, we can get rid of the external CSS and vendor code and create our own search dialog (with algolia logo and link).

.cspell.json
```diff
@@ -28,7 +28,8 @@
         "svgs",
         "contentalign",
         "inputpassword",
-        "inputcopy"
+        "inputcopy",
+        "docsearch"
```
splincode, line 32: revert

projects/demo/src/modules/app/search/index.html
```diff
@@ -0,0 +1 @@
+<div id="docsearch"></div>
```
nsbarsukov, line 1: Could we attach this `id` directly on `<tui-algolia-search />` and remove this template-file ?

nsbarsukov (approved):

Issue #9057, opened 2024-09-17
🚀 - Migrate `Demo` components to the new `API` page approach


mdlufy:
### Description

Migrate other `Demo` components to the new `API` page approach as in [tuiBlock](https://github.com/taiga-family/taiga-ui/pull/9015/files#diff-9e68d47bcdfc52469f2ddc01c346a398d8698891ea9ad3f7d49fe5b34a031502)

Should be visible main properties, appearances and icons
<img width="1119" alt="image" src="https://github.com/user-attachments/assets/b5fabe3d-d932-4885-a666-281f6b157a8e">

mdlufy:
Need to fix #9206

mdlufy:
We need this in `InputCard` and `InputPhoneInternational`

shiv9604:
Hi @waterplea & @splincode, This issue seems to have a higher priority than the other open issues. I would like your suggestion for picking other issues or contributing to this issue according to the next release planning.

waterplea:
@shiv9604 you can pick components one by one and migrate them to the new API components. Just don't take any from the legacy package. That would be very helpful, thank you!

shiv9604:
Sure @waterplea. Thanks for your suggestion!

PR #9067, opened 2024-09-17
chore(demo): `Dropdown` add reuse component & migrate example to new approach

mdlufy:
Partially solved #9057

### DIFF
--- a/projects/demo/src/components/appearance/index.html
+++ b/projects/demo/src/components/appearance/index.html
@@ -12,7 +12,7 @@
     </td>
 </tr>
 <tr
-    *ngIf="tuiDocAppearance && !hiddenOptions.includes('appearance')"
+    *ngIf="!hiddenOptions.includes('appearance')"
     name="[appearance]"
     tuiDocAPIItem
     type="string"
--- a/projects/demo/src/components/appearance/index.html
+++ b/projects/demo/src/components/appearance/index.html
@@ -12,6 +12,7 @@
     </td>
 </tr>
 <tr
+    *ngIf="tuiDocAppearance && !hiddenOptions.includes('appearance')"
     name="[appearance]"
     tuiDocAPIItem
     type="string"
@@ -22,6 +23,7 @@
     Appearance of the element
 </tr>
 <tr
+    *ngIf="!hiddenOptions.includes('state')"
     name="[tuiAppearanceState]"
     tuiDocAPIItem
     type="TuiInteractiveState | null"
@@ -31,6 +33,7 @@
     Manual interactive state override
 </tr>
 <tr
+    *ngIf="!hiddenOptions.includes('focus')"
     name="[tuiAppearanceFocus]"
     tuiDocAPIItem
     type="boolean | null"
@@ -40,6 +43,7 @@
     Manual focus state override
 </tr>
 <tr
+    *ngIf="!hiddenOptions.includes('mode')"
     name="[tuiAppearanceMode]"
     tuiDocAPIItem
     type="string | string[] | null"
--- a/projects/demo/src/components/dropdown/index.html
+++ b/projects/demo/src/components/dropdown/index.html
@@ -1,0 +1,75 @@
+<tr>
+    <td colspan="3">
+        <span tuiTitle>
+            <a
+                tuiLink
+                [routerLink]="routes.Dropdown"
+            >
+                <strong>Dropdown</strong>
+            </a>
+            <span tuiSubtitle>Applied as a host directive</span>
+        </span>
+    </td>
+</tr>
+<tr
+    name="[(tuiDropdownOpen)]"
+    tuiDocAPIItem
+    type="boolean"
+    [(value)]="open"
+>
+    Manual control/monitor for the dropdown being visible
+</tr>
+<tr
+    name="[tuiDropdownAlign]"
+    tuiDocAPIItem
+    type="tuiDropdownAlign"
+    [items]="aligns"
+    [(value)]="align"
+>
+    Align of dropdown (does not work together with
+    <code>limitWidth === 'fixed'</code>
+    )
+</tr>
+<tr
+    name="[tuiDropdownDirection]"
+    tuiDocAPIItem
+    type="TuiVerticalDirection | null"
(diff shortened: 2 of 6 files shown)


splincode (approved):

projects/demo/src/components/dropdown/index.html
```diff
@@ -0,0 +1,75 @@
...
+<tr
+    name="[tuiDropdownMinHeight]"
+    tuiDocAPIItem
+    type="number"
+    [(value)]="maxHeight"
+>
+    Maximum height of dropdown
+</tr>
+<tr
+    name="[tuiDropdownMinHeight]"
```
waterplea, line 69: Copypasted name :)

projects/demo/src/modules/directives/dropdown/index.html
```diff
@@ -78,24 +78,27 @@
...
+                type="PolymorpheusContent"
             >
                 Content
-            </ng-template>
-            <dropdown-documentation />
-        </tui-doc-documentation>
+            </tr>
+            <tbody
+                #dropdown
+                tuiDocDropdown
```
waterplea, line 99: Looks like we need to be able to show/hide certain options. For example, here we do not need `[(open)]` and in appearance we also need to able to hide appearance, because it is not changeable for checkbox/radio/switch, although other appearance options are applicable. In appearance let's hide it if we passed empty array.

projects/demo/src/components/dropdown/index.html
```diff
@@ -0,0 +1,75 @@
...
+                [routerLink]="routes.Dropdown"
+            >
+                <strong>Dropdown</strong>
+            </a>
+            <span tuiSubtitle>Applied as a host directive</span>
+        </span>
+    </td>
+</tr>
+<tr
+    name="[tuiDropdownOpen]"
```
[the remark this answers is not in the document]
mdlufy: fixed

projects/demo/src/components/dropdown/index.html
mdlufy: fixed

projects/demo/src/modules/directives/dropdown/index.html
mdlufy: Added `hiddenOptions`

mdlufy:
Problem with setting options by url


https://github.com/user-attachments/assets/e7eba53d-d9eb-4f7c-914e-662467f1a801

projects/demo/src/components/appearance/index.html
```diff
@@ -12,6 +12,7 @@
     </td>
 </tr>
 <tr
+    *ngIf="tuiDocAppearance && !hiddenOptions.includes('appearance')"
```
waterplea, line 15: No need to check `tuiDocAppearance`. When it's empty it means we need to show all appearances (stored in `appearances`)

projects/demo/src/components/dropdown/index.html
```diff
@@ -0,0 +1,82 @@
...
+            >
+                <strong>Dropdown</strong>
+            </a>
+            <span tuiSubtitle>Applied as a host directive</span>
+        </span>
+    </td>
+</tr>
+<tr
+    *ngIf="!hiddenOptions.includes('open')"
+    name="[(tuiDropdownOpen)]"
```
waterplea, line 16: This one is called `[(tuiDropdownOpen)]` on its own, when it's a host directive it is aliased as `[(open)]`. I think it will be more often added as a host directive so let's name it `[(open)]`

projects/demo/src/components/dropdown/index.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    @Input()
+    public hiddenOptions: Array<TuiLooseUnion<keyof TuiDropdownOptions>> = [];
+
+    public open = false;
+
+    public align: TuiDropdownAlign = 'left';
+    public direction: TuiVerticalDirection | null = null;
+    public limitWidth: TuiDropdownWidth = 'fixed';
+    public minHeight = 80;
```
waterplea, line 46: I think we can indeed inject default options and just override the limitWidth

projects/demo/src/modules/directives/dropdown/index.ts
```diff
@@ -16,6 +17,7 @@ import {DropdownDocumentation} from '../../components/abstract/dropdown-document
         TuiActiveZone,
         TuiButton,
         TuiDemo,
```
splincode, line 19: `TuiDocDropdown` should be inside `TuiDemo`

projects/demo/src/components/appearance/index.html
mdlufy: https://github.com/taiga-family/taiga-ui/pull/9067#discussion_r1766579682
`In appearance let's hide it if we passed empty array` 
But I don't understand you..

projects/demo/src/modules/directives/dropdown/index.ts
mdlufy: Other components do not need it, but only the dropdowns examples. If we put it inside `TuiDemo`, it seems that this will be an extra import for those who do not need it?

projects/demo/src/components/dropdown/index.html
mdlufy: renamed

projects/demo/src/components/dropdown/index.ts
mdlufy: done

projects/demo/src/components/appearance/index.html
waterplea: You added `hiddenOptions` array, it is more flexible than my original suggestion. Just remove `tuiDocAppearance && ` from `*ngIf` and it will be fine

mdlufy: nice, removed

waterplea (approved):

projects/demo/src/modules/directives/dropdown/index.ts
splincode: Oh, sorry

splincode (approved):

PR #9080, opened 2024-09-18
chore(demo): correct gap between push in basic example

### LINKED ISSUE #9060: 🐞 - broken gap
### Playground Link

/components/push

### Description

image

### Angular version

latest

### Taiga UI version

latest

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

shiv9604:
Close #9060

### DIFF
--- a/projects/demo/src/modules/components/push/examples/1/index.less
+++ b/projects/demo/src/modules/components/push/examples/1/index.less
@@ -10,6 +10,14 @@
     max-inline-size: 100%;
     border-radius: var(--tui-radius-l);
     margin-bottom: 1rem;
+
+    > * {
+        margin-bottom: 1rem;
+
+        &:last-child {
+            margin-bottom: 0;
+        }
+    }
 }
 
 .push {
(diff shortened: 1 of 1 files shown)


splincode (approved):

projects/demo/src/modules/components/push/examples/1/index.less
```diff
@@ -10,6 +10,14 @@
     max-inline-size: 100%;
     border-radius: var(--tui-radius-l);
     margin-bottom: 1rem;
+
```
waterplea, line 13: Please use the fix I suggested in the issue. The problem is not with gap per say, it's the background of the wrapper that's missing. I believe the gap is correct.

shiv9604: A wrapper contains 2 push elements as child elements. If we use the suggested fix, it merges both push elements. Can you confirm is wrapper necessary for each push element?

waterplea: Yes, the idea of this example is to show a list of push elements as in a dropdown when clicking on Bell icon with notifications, hence first one having (X) icon as if it's there to close the whole list.

shiv9604: Thanks for clarifying the requirements Alex, I have resolved the PR comment. Please checkout.

waterplea (approved):

Issue #9083, opened 2024-09-18
🤖 - Add SSR on stackblitz playground by default


splincode:
### Description

https://stackblitz.com/edit/taiga-ui-multi-select-arrow-animate-njfymb?file=src%2Fmain.server.ts

PR #9084, opened 2024-09-18
chore(demo): add external stackblitz playground

splincode:
Close https://github.com/taiga-family/taiga-ui/issues/9083


projects/demo/src/modules/app/stackblitz/starter/index.ts
```diff
@@ -1,56 +1,16 @@
...
-            {
-                newWindow: false,
-                openFile: appPrefix`app.template.html`,
-                hideExplorer: true,
-            },
-        );
-    }
+export default class Page {
+    protected readonly redirect = inject(WA_WINDOW).location.replace(
+        'https://stackblitz.com/~/github.com/taiga-family/taiga-playground?file=src/app/app.component.html',
```
waterplea, line 14: I think we only need it when somebody enters taiga-ui.dev/stackblitz manually, in header we can just have a regular target="_blank" link to this address, right?

waterplea (approved):

waterplea:
Does this PR address this? https://github.com/taiga-family/taiga-ui/issues/9083

projects/demo/src/modules/app/stackblitz/starter/index.ts
splincode: right

splincode: fixed

projects/demo/src/modules/app/app.template.html
```diff
@@ -47,10 +47,11 @@
             <a
                 appearance="icon"
                 iconStart="assets/icons/stackblitz.svg"
+                target="_blank"
```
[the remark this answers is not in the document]
splincode: fixed

projects/demo/src/modules/app/stackblitz/starter/index.ts
```diff
@@ -1,56 +1,17 @@
...
-                        "@import '@taiga-ui/core/styles/taiga-ui-local.less';",
-                },
-            },
-            {
-                newWindow: false,
-                openFile: appPrefix`app.template.html`,
-                hideExplorer: true,
-            },
-        );
+export default class Page {
```
nsbarsukov, line 13: Do we actually need this page for the new approach ?

Could we just update these lines
https://github.com/taiga-family/taiga-ui/blob/bfe18a1b6776df638037bf19cc8230879de84268/projects/demo/src/modules/app/pages.ts#L52-L58

to the following way ?

```ts
{
    section: 'Documentation',
    title: 'StackBlitz',
    keywords: 'reproduce, issue, bug, sandbox, playground, test',
    route: environment.playground,
    target: '_blank',
},
```

splincode: `Do we actually need this page for the new approach ?` @waterplea fyi

waterplea: I believe it's nice to have `https://taiga-ui.dev/stackblitz` as an alias to that playground. We definitely shared it here and there, let's keep it working.

projects/demo/src/modules/app/app.template.html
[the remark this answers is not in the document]
mdlufy: Not enough only `noreferrer`? Lighthouse not satisfy spec?

https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer
`It indicates that no referrer information is to be leaked when following the link and also implies the noopener keyword behavior under the same conditions.`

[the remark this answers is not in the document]
splincode: fixed

vladimirpotekhin (approved):

PR #9108, opened 2024-09-19
feat(addon-doc): add aria-label for dark mode button

splincode:
## Before

image

## After

image

### DIFF
--- a/projects/addon-doc/components/main/main.template.html
+++ b/projects/addon-doc/components/main/main.template.html
@@ -11,6 +11,7 @@
         <ng-content select="tuiDocHeader" />
         <button
             appearance="secondary"
+            aria-label="Switch between dark and light mode"
             size="s"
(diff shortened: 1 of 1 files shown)


projects/addon-doc/components/main/main.template.html
```diff
@@ -11,6 +11,7 @@
         <ng-content select="tuiDocHeader" />
         <button
             appearance="secondary"
+            aria-label="Switch between dark and light mode"
```
waterplea, line 14: You can just write this directly inside the button, that's exactly what we have font-size: 0 in icon buttons

nsbarsukov (approved):

projects/addon-doc/components/main/main.template.html
splincode: What do you mean about directly?

PR #9123, opened 2024-09-20
fix(legacy): `MultiSelect` empty line when non empty `placeholder` and `valueContent`

### LINKED ISSUE #9112: 🐞 - MultiSelect. When inserting "placeholder", a blank line appears
### Code Example URL

https://taiga-ui.dev/components/multi-select/API?valueContent$=1&stringify$=1&tuiTextfieldSize=s&placeholder=asdasd&valueContent=

### Description

1) Follow the link: https://taiga-ui.dev/components/multi-select/API?valueContent$=1&stringify$=1&tuiTextfieldSize=s&placeholder=asdasd&valueContent=
2) Enter the "placeholder". 
3) Select values.
4) Remove focus.
image

### Taiga UI version

4.6.0

### Browser Used

Chrome

### OS Used

macOS

mdlufy:
Close #9112

### DIFF
--- a/projects/legacy/components/input-tag/input-tag.component.ts
+++ b/projects/legacy/components/input-tag/input-tag.component.ts
@@ -1,4 +1,4 @@
-import type {AfterContentChecked, AfterContentInit, QueryList} from '@angular/core';
+import type {QueryList} from '@angular/core';
 import {
     ChangeDetectionStrategy,
     Component,
@@ -88,11 +88,7 @@ const TAG_VERTICAL_SPACE_REM = 0.125;
 })
 export class TuiInputTagComponent
     extends AbstractTuiMultipleControl<string>
-    implements
-        TuiFocusableElementAccessor,
-        TuiDataListHost<string>,
-        AfterContentInit,
-        AfterContentChecked
+    implements TuiFocusableElementAccessor, TuiDataListHost<string>
 {
     @ViewChild(TuiDropdownOpen)
     private readonly dropdown?: TuiDropdownOpen;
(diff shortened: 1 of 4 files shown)


splincode (approved):

projects/legacy/components/input-tag/input-tag.component.ts
```diff
@@ -113,6 +114,9 @@ export class TuiInputTagComponent
     @ContentChild(TuiDataListDirective, {read: TemplateRef})
     protected readonly datalist?: TemplateRef<TuiContext<TuiActiveZone>>;
 
+    @ContentChild(PolymorpheusOutlet, {read: ElementRef})
+    protected readonly valueContent?: ElementRef<HTMLSpanElement>;
```
waterplea, line 118: Type is wrong, just use `PolymorpheusOutlet` and type can be `unknown`. Outlet is a structural directive, it does not have an `ElementRef`, not sure how that works, probably it reaches up for closest `ElementRef`. If you would ask me this question on an interview I'd say it would be `null` :)

waterplea (changes requested):

projects/legacy/components/input-tag/input-tag.component.ts
mdlufy: Still, it gets to the ElementRef, but not the span element with the text, but to some kind of artifact from the directive, interestingly

is it better to just change to `unknown` type or, for clarity, hang some kind of tag on the content wrapper and not get tied to PolymerphousOutlet?

image

waterplea (approved):

PR #9168, opened 2024-09-24
fix(demo): rename alert `status` to `appearance`

ZhukAnna:
Close #

### DIFF
--- a/projects/demo/src/modules/components/alert/index.html
+++ b/projects/demo/src/modules/components/alert/index.html
@@ -58,7 +58,7 @@
                 Content
             </ng-template>
             <ng-template
-                documentationPropertyName="status"
+                documentationPropertyName="appearance"
                 documentationPropertyType="string"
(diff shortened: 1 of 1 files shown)


splincode (approved):

projects/demo/src/modules/components/alert/index.html
```diff
@@ -58,7 +58,7 @@
                 Content
             </ng-template>
             <ng-template
-                documentationPropertyName="status"
+                documentationPropertyName="appearance"
```
splincode, line 61: also add here

image

waterplea: Add what?

waterplea (approved):

projects/demo/src/modules/components/alert/index.html
splincode: Sorry, thought about `<button [appearance]>`

PR #9227, opened 2024-09-26
chore: revert stackblitz

splincode:


projects/demo/src/environments/environment.prod.ts
```diff
@@ -1,6 +1,4 @@
 export const environment = {
     production: true,
     ym: 87890624,
-    playground:
-        'https://stackblitz.com/~/github.com/taiga-family/taiga-playground?file=src/app/app.component.html',
```
waterplea, line 5: A good full-fledged repository is also great to have, I wish we could have this link somewhere in the docs too.

waterplea (approved):

vladimirpotekhin (approved):

PR #9230, opened 2024-09-26
chore(demo): support interpolate-size

splincode:


waterplea:
I think it is a way too fresh feature to even try, let's not introduce style branching like that in the demo.

PR #9237, opened 2024-09-27
chore(demo): `Textfield` add reuse component & migrate `TextfieldController` to new approach

mdlufy:
Partially solved #9057


mdlufy:
@waterplea we need to add API page for https://taiga-ui.dev/components/textfield, right?

splincode (approved):

projects/demo/src/components/textfield/index.html
```diff
@@ -0,0 +1,57 @@
...
+<tr
+    *ngIf="!hiddenOptions.includes('size')"
+    name="[tuiTextfieldSize]"
+    tuiDocAPIItem
+    type="TuiSizeS | TuiSizeL"
+    [items]="sizeVariants"
+    [(value)]="size"
+>
+    Size
+</tr>
```
mdlufy, line 57: @waterplea do the texfields need any more options, or is that all?

waterplea: These are old ones, we only need `tuiTextfieldSize` and `tuiTextfieldCleaner`.

mdlufy: Remove other old, except `tuiTextfieldSize`, `tuiTextfieldCleaner`

projects/demo/src/components/textfield/index.html
```diff
@@ -0,0 +1,32 @@
+<tr>
+    <td colspan="3">
+        <span tuiTitle>
+            <a
+                tuiLink
+                [routerLink]="routes.Textfield"
+            >
+                <strong>Textfield</strong>
```
waterplea, line 8: Let's name it `TextfieldOptions` and I'm not sure we need a link to textfield here, it's just 2 options, I don't know if people need to go investigate what bigger picture are they a part of :)

projects/demo/src/modules/components/textfield/index.html
```diff
@@ -20,5 +20,69 @@
         />
     </ng-template>
 
-    <tui-setup *pageTab="'Setup'" />
+    <ng-template pageTab>
+        <tui-doc-demo>
+            <tui-textfield
+                [iconEnd]="icons.iconEnd"
```
waterplea, line 26: We also need `filler` here. Let's not show other missing inputs for now (content and stringify)

waterplea (commented):
We need this in `InputCard` and `InputPhoneInternational`, but let's add the "Control" block first, the one with readOnly/invalid/disabled, otherwise it will look empty there :)

projects/demo/src/components/textfield/index.html
mdlufy: Yes, renamed and removed link

projects/demo/src/modules/components/textfield/index.html
mdlufy: Done

mdlufy:
> We need this in `InputCard` and `InputPhoneInternational`, but let's add the "Control" block first, the one with readOnly/invalid/disabled, otherwise it will look empty there :)

Okey, let's just create it here, and add to `InputCard` and `InputPhoneInternational` in another PR

projects/demo/src/components/control/index.html
```diff
@@ -0,0 +1,36 @@
...
+    <td colspan="3">
+        <span tuiTitle>
+            <strong>ControlOptions</strong>
+            <span tuiSubtitle>Applied as a host directive</span>
+        </span>
+    </td>
+</tr>
+
+<tr
+    *ngIf="!hiddenOptions.includes('readonly')"
```
waterplea, line 11: I think it's `readOnly` with capital `O`

projects/demo/src/components/control/index.html
```diff
@@ -0,0 +1,36 @@
...
+    type="boolean"
+    [(value)]="disabled"
+>
+    Disabled state
+</tr>
+<tr
+    *ngIf="!hiddenOptions.includes('invalid')"
+    name="[invalid]"
+    tuiDocAPIItem
+    type="boolean"
```
waterplea, line 32: It's `boolean | null` so we can manually set `true`/`false` invalid state or leave to the native behavior using `null`.

projects/demo/src/components/control/index.html
```diff
@@ -0,0 +1,36 @@
...
+    Readonly state
+</tr>
+<tr
+    *ngIf="!hiddenOptions.includes('disabled')"
+    name="[disabled]"
+    tuiDocAPIItem
+    type="boolean"
+    [(value)]="disabled"
+>
+    Disabled state
```
waterplea, line 26: We should probably add a note to use `control.disable()/.enable()` for reactive forms like we had in old docs.

projects/demo/src/components/control/index.html
```diff
@@ -0,0 +1,36 @@
+<tr>
+    <td colspan="3">
+        <span tuiTitle>
+            <strong>ControlOptions</strong>
```
waterplea, line 4: Just `Control`, I think

projects/demo/src/components/control/index.ts
```diff
@@ -0,0 +1,25 @@
...
+    selector: 'tbody[tuiDocControl]',
+    imports: [NgIf, TuiDocAPIItem, TuiTitle],
+    templateUrl: './index.html',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TuiDocControl {
+    protected readonly routes = DemoRoute;
+
+    @Input()
+    public hiddenOptions: Array<TuiLooseUnion<keyof AbstractControl>> = [];
```
waterplea, line 20: It's not properties of `AbstractControl`, it's just these 3 hardcoded values. For `Textfield` they come from here:
https://github.com/taiga-family/taiga-ui/blob/main/projects/core/components/textfield/textfield.directive.ts#L35-L39
And for controls like `InputCard` or `InputPhoneInternational` it comes from our abstract control class:
https://github.com/taiga-family/taiga-ui/blob/main/projects/cdk/classes/control.ts#L84-L89

projects/demo/src/components/textfield/index.ts
```diff
@@ -0,0 +1,28 @@
...
+    standalone: true,
+    selector: 'tbody[tuiDocTextfield]',
+    imports: [NgIf, TuiDocAPIItem, TuiTitle],
+    templateUrl: './index.html',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TuiDocTextfield {
+    private readonly options = inject(TUI_TEXTFIELD_OPTIONS);
+
+    protected readonly routes = DemoRoute;
```
waterplea, line 19: No longer needed

projects/demo/src/modules/components/textfield/index.html
```diff
@@ -20,5 +20,58 @@
...
+            </tui-textfield>
+        </tui-doc-demo>
+        <table tuiDocAPI>
+            <tr
+                name="[(ngModel)]"
+                tuiDocAPIItem
+                type="string"
+                [(value)]="value"
+            >
+                Value
```
waterplea, line 51: Let's add a note like `(or reactive control directives)`

projects/demo/src/components/control/index.html
mdlufy: Renamed

projects/demo/src/components/control/index.html
mdlufy: Type updated

projects/demo/src/components/control/index.html
mdlufy: Updated

projects/demo/src/components/control/index.ts
mdlufy: Yes, i seen. I think it would be cumbersome to derive these types from abstract class. 
Let's use union `'readOnly' | 'disabled' | 'invalid'` to simplify?

projects/demo/src/components/textfield/index.ts
mdlufy: Removed

projects/demo/src/components/control/index.html
mdlufy: > like we had in old docs

I can't find this notification in old docs. 
Will it be enough just to add text: "Use `control.disable()/.enable()` for reactive forms" ?

projects/demo/src/modules/components/textfield/index.html
mdlufy: Added

splincode (approved):

projects/demo/src/components/control/index.ts
```diff
@@ -0,0 +1,20 @@
...
+@Component({
+    standalone: true,
+    selector: 'tbody[tuiDocControl]',
+    imports: [NgIf, TuiDocAPIItem, TuiTitle],
+    templateUrl: './index.html',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TuiDocControl {
+    @Input()
+    public hiddenOptions: Array<'disabled' | 'invalid' | 'readOnly'> = [];
```
splincode, line 15: ReadonlyArray

projects/demo/src/components/control/index.ts
waterplea: Yes, that's what I think as well.

projects/demo/src/components/control/index.html
waterplea: image

You can write it as you see fit, we just want to make sure people understand this is a form thing, not our input, and for template driven form you're supposed to use `[disabled]` while for reactive forms you should call `control.disable()/.enable()` or create your own `[disabled]` directive that does so under the hood declaratively.

projects/demo/src/components/control/index.html
waterplea: I noticed change detection is broken there, when I set it to `true` it only updates when I click away to trigger change detection.

projects/demo/src/components/control/index.ts
mdlufy: Type updated

projects/demo/src/components/control/index.html
mdlufy: Okay, I get it :)

projects/demo/src/components/control/index.html
mdlufy: And [readOnly] too

mdlufy: Created #9389 9389

waterplea (approved):

splincode (approved):

PR #9243, opened 2024-09-27
fix(core): `Textfield` has change detection problems for `[filler]`

### LINKED ISSUE #9238: 🐞 - `Textfield` has change detection problems for filler
### Code Example URL

https://github.com/taiga-family/maskito/blob/time-meridiem-feature/projects/demo/src/pages/kit/time/examples/4-time-segments-min-max/component.ts

### Description

https://github.com/user-attachments/assets/b69561ad-0f0a-4390-8a18-92590a46aae4

### Reproduction
```html
<tui-textfield filler="HH:MM">
    <input
        tuiTextfield
        [(ngModel)]="value"
    />
</tui-textfield>
```

https://github.com/taiga-family/maskito/blob/755375e7f111e153c9e67f8ff12f35667efebe9c/projects/demo/src/pages/kit/time/examples/4-time-segments-min-max/component.ts#L14-L32

### Taiga UI version

4.8.1

### Browser Used

Chrome

### OS Used

macOS

nsbarsukov:
Fixes #9238

### DIFF
--- a/projects/core/components/textfield/textfield.component.ts
+++ b/projects/core/components/textfield/textfield.component.ts
@@ -8,6 +8,7 @@ import {
     forwardRef,
     inject,
     Input,
+    signal,
     ViewChild,
     ViewContainerRef,
     ViewEncapsulation,
@@ -65,6 +66,8 @@ import {TuiWithTextfieldDropdown} from './textfield-dropdown.directive';
     },
 })
 export class TuiTextfieldComponent<T> implements TuiDataListHost<T> {
+    // TODO: refactor to signal inputs after Angular update
+    private readonly filler = signal('');
     private readonly autoId = tuiInjectId();
     private readonly el = tuiInjectElement();
     private readonly open = tuiDropdownOpen();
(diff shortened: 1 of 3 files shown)


splincode (approved):

splincode:
Could you add cypress component test?

nsbarsukov:
> Could you add cypress component test?

I am not sure that this case could be covered by simple test with some assertion.
It should be at least screenshot test. Screenshot testing is not supported by our infrastructure.

projects/core/components/textfield/textfield.directive.ts
```diff
@@ -35,6 +37,11 @@ export class TuiTextfieldBase<T> implements OnChanges {
     @Input()
     public invalid: boolean | null = null;
 
+    public nativeValue = toSignal(
+        fromEvent(this.el, 'input', () => this.el.value),
```
waterplea, line 41: Didn't know you could add mapping function like that 😱

projects/core/components/textfield/textfield.component.ts
```diff
@@ -81,6 +84,21 @@ export class TuiTextfieldComponent<T> implements TuiDataListHost<T> {
...
+        const value = this.directive?.nativeValue() || '';
+        const filledValue = value + this.filler().slice(value.length);
+
+        return filledValue.length > value.length ? filledValue : '';
+    });
+
+    protected showFiller = computed(() =>
+        Boolean(
+            this.focused() &&
+                this.computedFiller() &&
```
waterplea, line 97: Weird tabbing 🤔

waterplea (approved):

projects/core/components/textfield/textfield.component.ts
nsbarsukov: Unfortunately, prettier does it (

https://github.com/user-attachments/assets/a06f4c4c-a307-405d-a285-c05886116f6d

The only way to bypass it is drop `Boolean` and add some `!!`

```ts
protected showFiller = computed(
    () =>
        this.focused() &&
        !!this.computedFiller() &&
        (!!this.directive?.nativeValue() || !this.input?.nativeElement.placeholder),
);
```

What do you think ?

waterplea: I guess it looks better

splincode: You can report this issue to [prettier](https://github.com/prettier/prettier)

nsbarsukov: > You can report this issue to [prettier](https://github.com/prettier/prettier)

@splincode done
https://github.com/prettier/prettier/issues/16709

PR #9261, opened 2024-09-29
chore(demo): drop deprecated directive from landing

splincode:


waterplea (approved):

projects/demo/src/modules/app/landing/index.html
```diff
@@ -1,7 +1,6 @@
 <img
     alt="taiga"
     src="assets/images/taiga.svg"
-    tuiLoading="lazy"
```
vladimirpotekhin, line 4: shouldn't we keep the loading="lazy" attribute?

splincode: fixed

vladimirpotekhin (approved):

PR #9262, opened 2024-09-29
feat(cdk): add `TuiRepeatTimes` pipe

splincode:
@waterplea I want to use `@for` in Angular 18 and I don't want use `*tuiRepeatTimes`

I can happy with it

image

+ strict types

image

### DIFF
--- a/projects/cdk/pipes/repeat-times/index.ts
+++ b/projects/cdk/pipes/repeat-times/index.ts
@@ -1,0 +1,12 @@
+import {Pipe, type PipeTransform} from '@angular/core';
+import {tuiClamp} from '@taiga-ui/cdk/utils';
+
+@Pipe({
+    standalone: true,
+    name: 'tuiRepeatTimes',
+})
+export class TuiRepeatTimesPipe implements PipeTransform {
+    public transform(n: number): number[] {
+        return Array.from({length: tuiClamp(n, 0, n)}, (_, i) => i);
+    }
+}
--- a/projects/cdk/pipes/index.ts
+++ b/projects/cdk/pipes/index.ts
@@ -2,5 +2,6 @@ export * from '@taiga-ui/cdk/pipes/filter';
 export * from '@taiga-ui/cdk/pipes/is-present';
 export * from '@taiga-ui/cdk/pipes/keys';
 export * from '@taiga-ui/cdk/pipes/mapper';
+export * from '@taiga-ui/cdk/pipes/repeat-times';
 export * from '@taiga-ui/cdk/pipes/replace';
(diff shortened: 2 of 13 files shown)


waterplea (approved):
Cool idea. Do we really need those excessive types, though?

splincode:
`need those excessive types,`

It helps with some clamp case

```
// @ts-expect-error
expect(pipe.transform(-1)).toEqual([]);
```

https://catchts.com/range-numbers

splincode:
In the future TypeScript add range type to standard library, I think 
https://github.com/microsoft/TypeScript/issues/54925

projects/cdk/pipes/repeat-times/index.ts
```diff
@@ -0,0 +1,14 @@
...
+import type {TuiIterationRange} from '@taiga-ui/cdk/types';
+
+type Range<From extends number, To extends number> = TuiIterationRange<From, To, 1>;
+
+@Pipe({
+    standalone: true,
+    name: 'tuiRepeatTimes',
+})
+export class TuiRepeatTimesPipe implements PipeTransform {
+    public transform<N extends number>(length: N): Range<1, N> {
```
nsbarsukov, line 11: Should `0` be a valid option as minimum value of range ?
```suggestion
    public transform<N extends number>(length: N): Range<0, N> {
```

It can useful for cases with dynamic show/hide logic

projects/cdk/types/range.ts
nsbarsukov, line 1: lgtm :D

image

nsbarsukov (approved):

projects/cdk/pipes/repeat-times/index.ts
splincode: I can't use `Range<0, N - 1>`

waterplea:
> `need those excessive types,`
> 
> It helps with some clamp case
> 
> ```
> // @ts-expect-error
> expect(pipe.transform(-1)).toEqual([]);
> ```
> 
> https://catchts.com/range-numbers

But it's a pipe, are there any benefits for it though? Seems like it's a lot of added complexity for little to no benefit.

splincode:
`But it's a pipe, are there any benefits for it though? Seems like it's a lot of added complexity for little to no benefit.`


image
image

splincode:
@waterplea 
Okey, I will drop types

projects/cdk/pipes/repeat-times/index.ts
splincode: fixed

projects/cdk/types/range.ts
splincode: dropped

waterplea (approved):

PR #9330, opened 2024-10-03
fix(cdk): `Control` react to `markAsTouched` on Angular 18+

waterplea:
Fixes #

### DIFF
--- a/projects/cdk/classes/control.ts
+++ b/projects/cdk/classes/control.ts
@@ -16,6 +16,7 @@ import {tuiProvide, tuiPure} from '@taiga-ui/cdk/utils';
 import {
     delay,
     distinctUntilChanged,
+    EMPTY,
     filter,
     map,
     merge,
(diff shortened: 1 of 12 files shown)


projects/cdk/classes/control.ts
```diff
@@ -71,7 +72,9 @@ export abstract class TuiControl<T> implements ControlValueAccessor {
                 map(() => this.control.control),
                 filter(Boolean),
                 distinctUntilChanged(),
-                switchMap((c) => merge(c.valueChanges, c.statusChanges)),
+                switchMap((c) =>
```
splincode, line 75: `as any` so sad (( why not `as unknown as what-do-you-need`

projects/demo/src/modules/components/breadcrumbs/examples/3/index.html
```diff
@@ -0,0 +1,36 @@
+<h3>Truncate</h3>
+<tui-breadcrumbs>
+    <ng-container *ngFor="let item of items">
+        <button
+            *tuiItem
```
splincode, line 5: please, add type=button

splincode (approved):

projects/cdk/classes/control.ts
waterplea: Because it makes no difference and is just longer

PR #9356, opened 2024-10-05
fix(addon-table): forgot add peer dependency

splincode:


waterplea:
I don't think this is how we should fix it, I think we should get rid of the dependency since it's a legacy package. I'll take a look on Monday, should be possible one way or another.

Issue #8244, opened 2024-07-29
🚀 - Allow tuiSortable to configured dynamically


evantrimboli:
### Description

Consider the following component structure. Note this isn't intended to be 100% correct, just for illustration.

As far as I can tell, because it's not possible to dynamically add a directive, to toggle sorting using a condition, I need to branch on whether it's sortable and then repeat the th for a second time and duplicate the code.

I'm not sure of the best way to solve it, but it would be nice to be able to provide a flag to the directive which would toggle it on or off.

```
interface Column {
  readonly id: string
  readonly title: string
  readonly sortable: boolean
  // Other configs
}

@Component(
  selector: 'my-table',
  template: `
  <table>
    <thead tuiHead>
      <tr tuiThGroup>
      @for (c of columns(); track c.id) {
        @if (c.sortable) {
         <th tuiTh *tuiHead="c.id" tuiSortable>
            {{ c.title }}
          </th>
        } @else {
          <th tuiTh *tuiHead="c.id">
            {{ c.title }}
          </th>
        }
      }
      </tr>
    </thead>
  </table>
`
);
export Class MyTable {
  readonly columns = input<readonly Column[]>([]);
}
```

waterplea:
We need to allow passing true / false to t`uiSortable` and then look for keys only among enabled ones in `TuiTableSortBy`

tsironis13:
Is something ready to consume regarding this? I need to enable sorting only for columns which are configured to be sortable.

tsironis13:
> > We need to allow passing true / false to `tuiSortable` and then look for keys only among enabled ones in `TuiTableSortBy`
> 
> Here's what needs to be done, you can make a PR.

@waterplea  I am on it, thanks

waterplea:
@evantrimboli could you clarify the issue please, the sorting is enabled dynamically by passing `[sorter]` value. `tuiSortable` is just a directive to connect sorter and column key when you operate with keys rather than sorters, for example, for server side sorting.

PR #9382, opened 2024-10-08
feat(addon-table): disable column sorting on demand

tsironis13:
Fixes https://github.com/taiga-family/taiga-ui/issues/8244

image
The idea is to enable sorting only for columns where the `[requiredSort]="true"` is provided

### DIFF
--- a/projects/addon-table/components/table/th/th.template.html
+++ b/projects/addon-table/components/table/th/th.template.html
@@ -1,5 +1,5 @@
 <button
-    *ngIf="sorter && table; else content"
+    *ngIf="requiredSort && table; else content"
     type="button"
     class="t-sort"
     [class.t-sort_sorted]="isCurrent"
--- a/projects/demo/src/modules/components/table/examples/4/index.html
+++ b/projects/demo/src/modules/components/table/examples/4/index.html
@@ -53,7 +53,6 @@
             <tr tuiThGroup>
                 <th
                     *tuiHead="'name'"
-                    tuiSortable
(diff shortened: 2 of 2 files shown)


splincode (approved):

waterplea:
This is not what the fix needs to be. The issue was about making `tuiSortable` dynamic. But looking at it again I don't think we need it at all. All it does it connect sorter with key, I think it is already working as expected. This current PR is definitely not right, `requiredSort` means when you click the column it switches from asc/desc and never to unsorted, unlike other columns with sorter which disable sorting on 3rd click.

tsironis13:
> This is not what the fix needs to be. The issue was about making `tuiSortable` dynamic. But looking at it again I don't think we need it at all. All it does it connect sorter with key, I think it is already working as expected. This current PR is definitely not right, `requiredSort` means when you click the column it switches from asc/desc and never to unsorted, unlike other columns with sorter which disable sorting on 3rd click.

Ok, can we disable sorting dynamically for some columns with the current codebase? 
image
my point is, if we have a way to provide this UI functionality to the end user and I am asking because I am new to the project and maybe I am missing something. Otherwise, it would be nice to create a new PR and provide this option also.

waterplea:
@tsironis13 sorry, I was sure I posted the explanation in the issue but it looks like I only typed it and forgot to press "Send" :) I ended up doing this PR myself because as I looked into the actual issue it turned out to be not as simple as I thought. It's going to be fixed in today's release.

PR #9407, opened 2024-10-09
fix(core): do not override passed empty id

splincode:
In my tests I wanna use empty id, but now it doesn't work as expected

```html
<input id="" />
```


waterplea:
That wouldn't work because if you omit `id` completely it's empty string, not `null` or `undefined`.

splincode:
`That wouldn't work because if you omit id completely it's empty string, not null or undefined.`

ha, that's what I needed

waterplea:
We use auto id because we need input to be linked with label even when no id is provided.

splincode:
`be linked with label even when no id is provided.` hm, okey

PR #9408, opened 2024-10-09
fix(core): support provide empty icon value

splincode:
I want pass empty icons for screenshot testing

```ts
tuiIconsProvider({
  '@tui.heart': ''
})
```

### DIFF
--- a/projects/core/tokens/icon-resolver.ts
+++ b/projects/core/tokens/icon-resolver.ts
@@ -23,12 +23,10 @@ export function tuiInjectIconResolver(): TuiStringHandler<string> {
     const icons = inject(TUI_ICON_REGISTRY);
     const resolver = inject(TUI_ICON_RESOLVER);
 
-    return (icon) => (!icon || icon.includes('/') ? icon : icons[icon] || resolver(icon));
+    return (icon) =>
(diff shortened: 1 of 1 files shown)


projects/core/tokens/icon-resolver.ts
```diff
@@ -23,12 +23,10 @@ export function tuiInjectIconResolver(): TuiStringHandler<string> {
     const icons = inject(TUI_ICON_REGISTRY);
     const resolver = inject(TUI_ICON_RESOLVER);
 
-    return (icon) => (!icon || icon.includes('/') ? icon : icons[icon] || resolver(icon));
+    return (icon) =>
+        !icon || icon.includes('/') ? icon : (icons[icon] ?? resolver(icon));
```
mdlufy, line 27: For an inexperienced user, this looks rather cumbersome, I would generally rewrite it via if..

mdlufy (approved):

projects/core/tokens/icon-resolver.ts
waterplea: I think it's ok. We typically write ternary every time we return value that can be expressed as `a ?  b : c`, even if there are some brackets involved.

waterplea (approved):

PR #9409, opened 2024-10-09
refactor: prefer nullish

splincode:


projects/legacy/components/tag/tag.component.ts
```diff
@@ -104,7 +104,7 @@ export class TuiTagComponent {
     }
 
     protected get displayText(): string {
-        return this.editedText === null ? this.value : this.editedText;
+        return this.editedText ?? this.value;
     }
```
mdlufy, line 108: This is not an equivalent replacement, if `this.editedText` is `undefined`. It's okay for us?

mdlufy (approved):

projects/legacy/components/tag/tag.component.ts
splincode: @waterplea why we use `=== null` instead of `== null`?

projects/kit/components/filter/test/filter.component.spec.ts
```diff
@@ -21,7 +21,7 @@ class ItemWithBadge {
     }
 
     public valueOf(): number | null {
-        return this.badgeValue !== undefined ? this.badgeValue : null;
+        return this.badgeValue ?? null;
```
mdlufy, line 24: This is not an equivalent replacement, if `this.badgeValue` is `null`. It's okay for us?

splincode: if `this.badgeValue` is `null` then 

`null !== undefined ? null : null;` -> `null`

vladimirpotekhin (approved):

PR #9435, opened 2024-10-10
chore: drop explicit dependency

splincode:


waterplea:
We need it there for WebStorm to pick WebTypes

splincode:
@waterplea why not just add

```
 "web-types": "node_modules/@taiga-ui/event-plugins/web-types.json"
```

waterplea:
> @waterplea why not just add
> 
> ```
>  "web-types": "node_modules/@taiga-ui/event-plugins/web-types.json"
> ```

I don't know. It works as is for now, not sure we need to touch it

PR #9438, opened 2024-10-10
chore(demo): `InputPhoneInternational` update `API` page

mdlufy:
Partially solved #9057

### DIFF
--- a/projects/addon-doc/components/api/api-item.style.less
+++ b/projects/addon-doc/components/api/api-item.style.less
@@ -54,6 +54,11 @@
 .t-input {
     min-inline-size: 10rem;
     margin-block-start: -0.625rem;
+
+    code {
+        overflow: hidden;
+        white-space: nowrap;
+    }
 }
 
 @media @tui-mobile {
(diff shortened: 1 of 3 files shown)


splincode (approved):

projects/addon-doc/components/api/api-item.style.less
```diff
@@ -54,6 +54,11 @@
 .t-input {
     min-inline-size: 10rem;
     margin-block-start: -0.625rem;
+
+    code {
+        overflow: hidden;
+        white-space: nowrap;
+    }
```
mdlufy, line 61: Maybe more preferably add directly to https://github.com/taiga-family/taiga-ui/blob/main/projects/addon-doc/components/main/main.style.less#L44-L75

waterplea: Let's leave it here for now, and if we meet it again – we'll move it.

projects/demo/src/modules/components/input-phone-international/index.ts
```diff
@@ -23,6 +27,10 @@ import {InheritedDocumentation} from '../abstract/inherited-documentation';
         InheritedDocumentation,
         ReactiveFormsModule,
         TuiDemo,
+        TuiDocControl,
+        TuiDocDropdown,
+        TuiDocIcons,
+        TuiDocTextfield,
```
waterplea, line 33: Please also remove `tuiProvide(ABSTRACT_PROPS_ACCESSOR, PageComponent),` and `extends AbstractExampleTuiControl` from this file

mdlufy: Removed

waterplea (approved):

PR #9443, opened 2024-10-10
refactor: simplify configuration code

splincode:


mdlufy (approved):

projects/cdk/schematics/utils/is-ci.ts
```diff
@@ -1,5 +1,8 @@
 /// <reference types="node" />
 
+/**
+ * @deprecated: drop in v5.0
+ */
 export function tuiIsCI(): boolean {
```
mdlufy, line 6: Would it be equivalent to `process.env.CI` ?

splincode: yes

PR #9446, opened 2024-10-11
feat(core): add dialog-factory

MillerSvt:
Fixes #

### DIFF
--- a/projects/core/components/dialog/dialog.factory.ts
+++ b/projects/core/components/dialog/dialog.factory.ts
@@ -5,40 +5,60 @@ import type {Observable} from 'rxjs';
 import type {TuiDialogContext, TuiDialogOptions} from './dialog.interfaces';
 import {TuiDialogService} from './dialog.service';
 
-type ReplaceNever<T> = [T] extends [never] ? void : T;
+type SingleUnionOrNever<T, U = T> = [T] extends [never]
+    ? never
+    : T extends U
+      ? [U] extends [T]
+          ? T
+          : never
+      : never;
 
 type ReplaceAny<T> = 0 extends T & 1 ? void : T;
 
-type ExtractDialogData<T> = ReplaceNever<
-    {
-        [K in keyof T]: ReplaceAny<T[K]> extends TuiDialogContext<any, infer D>
-            ? D
-            : never;
-    }[keyof T]
->;
+type ContextKeys<T> = {
+    [K in keyof T]: ReplaceAny<T[K]> extends TuiDialogContext<any, any> ? K : never;
+}[keyof T];
 
-type ExtractDialogResult<T> = ReplaceNever<
-    {
-        [K in keyof T]: ReplaceAny<T[K]> extends TuiDialogContext<infer R, any>
-            ? R
-            : never;
-    }[keyof T]
->;
+type AssertNotMultipleContexts<T, K extends keyof T> = [K] extends [never]
+    ? new () => T
+    : [SingleUnionOrNever<K>] extends [never]
+      ? 'Component has multiple context. Cannot determine the type...'
+      : new () => T;
+
+type ExtractDialogData<T, K extends keyof T = ContextKeys<T>> = [K] extends [never]
+    ? void
+    : [SingleUnionOrNever<K>] extends [never]
+      ? never
+      : T[K] extends TuiDialogContext<any, infer D>
+        ? D
+        : void;
+
+type ExtractDialogResult<T, K extends keyof T = ContextKeys<T>> = [K] extends [never]
+    ? void
+    : [SingleUnionOrNever<K>] extends [never]
+      ? never
+      : T[K] extends TuiDialogContext<infer R, any>
+        ? R
+        : void;
 
 export function tuiDialog<
     T,
-    D extends ExtractDialogData<T>,
-    R extends ExtractDialogResult<T>,
+    K extends ContextKeys<T>,
+    D extends ExtractDialogData<T, K>,
+    R extends ExtractDialogResult<T, K>,
 >(
-    component: new () => T,
+    component: AssertNotMultipleContexts<T, K>,
     options?: Partial<Omit<TuiDialogOptions<D>, 'data'>>,
 ): (data: D) => Observable<R> {
     const dialogService = inject(TuiDialogService);
     const injector = inject(INJECTOR);
 
     return (data) =>
-        dialogService.open(new PolymorpheusComponent(component, injector), {
-            ...options,
-            data,
-        });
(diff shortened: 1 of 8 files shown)


projects/core/components/dialog/dialog.factory.ts
```diff
@@ -0,0 +1,43 @@
...
+
+type ExtractDialogResult<T> = ReplaceNever<
+    {
+        [K in keyof T]: ReplaceAny<T[K]> extends TuiDialogContext<infer R, any>
+            ? R
+            : never;
+    }[keyof T]
+>;
+
+export function tuiDialog<
```
splincode, line 27: Cooler than my futile attempts

https://github.com/taiga-family/taiga-ui/pull/6175/files
https://github.com/taiga-family/taiga-ui/pull/7621

splincode (approved):

projects/core/components/dialog/test/dialog.factory.spec.ts
```diff
@@ -0,0 +1,155 @@
+import * as fs from 'node:fs';
```
splincode, line 1: @MillerSvt Maybe we can use something 

https://github.com/tsdjs/tsd or https://github.com/garronej/tsafe or `// @ts-expect-error`

what do you think?

MillerSvt: Hmm, I forgot about `@ts-expect-error`. I'll check it out next week.

MillerSvt: See what happened...

projects/demo/src/modules/components/dialog/examples/2/dialog-example/dialog-example.component.ts
```diff
@@ -36,12 +36,13 @@ import {injectContext} from '@taiga-ui/polymorpheus';
 })
 export class DialogExample {
     private readonly dialogs = inject(TuiDialogService);
-    private readonly context = injectContext<TuiDialogContext<number, number>>();
 
     protected value: number | null = null;
     protected name = '';
     protected items = [10, 50, 100];
 
+    public readonly context = injectContext<TuiDialogContext<number, number>>();
```
nsbarsukov, line 44: Why access modifiers was changed ?

projects/demo/src/modules/components/dialog/examples/9/pay-modal/pay-modal.component.ts
```diff
@@ -74,18 +74,22 @@ export class PayModal implements OnInit {
...
     protected payProcessing$ = new BehaviorSubject(false);
-    protected amount: number = this.context?.data?.amount ?? 0;
+    protected amount = 0;
     protected readonly PAYMENT_MODE = PaymentMode;
 
+    public readonly context =
+        injectContext<TuiDialogContext<void, DataForPayCardModal>>();
+
+    constructor() {
+        this.amount = this.context?.data?.amount ?? 0;
```
nsbarsukov, line 90: Could you explain the reasons for these line changes, please  ?

projects/core/components/dialog/dialog.factory.ts
```diff
@@ -0,0 +1,64 @@
...
+): (data: D) => Observable<R> {
+    const dialogService = inject(TuiDialogService);
+    const injector = inject(INJECTOR);
+
+    return (data) =>
+        dialogService.open(
+            new PolymorpheusComponent(component as new () => T, injector),
+            {
+                ...options,
+                data,
```
nsbarsukov, line 61: Why only `data`? Could we provide an opportunity to override any "default " option ?

projects/core/components/dialog/test/dialog.factory.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+    // component with context
+    class TestComponent {
+        public readonly someContextProp!: TuiDialogContext<string, number>;
+    }
+
+    const dialog = tuiDialogMock(TestComponent);
+
+    dialog(123).subscribe((_value: string) => {});
+    // @ts-expect-error TS2554: Expected 1 arguments, but got 0
+    dialog();
```
nsbarsukov, line 31: Probably, creating dialog without any `data` should be valid too 🤔 
Guys, what do you think about it ?

projects/demo/src/modules/components/dialog/examples/2/dialog-example/dialog-example.component.ts
MillerSvt: We can only use public fields as context. Otherwise, we won't be able to determine the associated types.

projects/demo/src/modules/components/dialog/examples/9/pay-modal/pay-modal.component.ts
MillerSvt: 1. The field is changed from protected to public, as explained in the comment above.
2. Since public fields must be placed after protected ones, I changed their order.
3. After the change in order, the amount field cannot read the fields declared below it. Therefore, the only way to do it is in the constructor.

projects/core/components/dialog/test/dialog.factory.spec.ts
MillerSvt: You have a context with string data. So, you cannot create a dialog without passing a string to the data.

For dialogs that do not require input data, you must specify the data type as void or omit the entire context, as shown in the example above.

projects/core/components/dialog/dialog.factory.ts
MillerSvt: I think this task might be more complicated than it seems. This is because TypeScript omits void arguments if there are no arguments behind them. Otherwise, if we have void data, we must specify undefined as the data. This is not the desired behavior.

projects/demo/src/modules/components/dialog/examples/9/pay-modal/pay-modal.component.ts
waterplea: Sounds like it's better to just use public for amount, it's a demo anyway. Looks/reads better than constructor.

projects/core/components/dialog/dialog.factory.ts
waterplea: I think it's good as is, no need to override options here since we already provide our own options when creating this. I doubt we would need the same dialog to open in one size and another conditionally. And even if we do, we can just store it twice with different options.

waterplea (approved):

vladimirpotekhin (approved):

nsbarsukov (approved):

PR #9450, opened 2024-10-11
chore: enable i18n lint

splincode:


projects/demo/src/modules/pipes/field-error/examples/5/index.ts
```diff
@@ -42,6 +42,7 @@ export default class Example {
 
     constructor() {
         this.form = this.fb.group({
+            // eslint-disable-next-line i18n/no-russian-character
             text: ['русский текст', Validators.required],
```
splincode, line 46: image

package.json
```diff
@@ -80,10 +80,6 @@
             "plugin:@taiga-ui/experience/entry-points"
         ],
         "rules": {
-            "i18n/no-greek-character": "off",
```
waterplea, line 83: Russian I understand, so there's no Russian "c" hidden in code by accident, but what's wrong with others?

splincode: `but what's wrong with others`

prevent contribution like this

image
image

splincode: fixed

waterplea (approved):

PR #9477, opened 2024-10-13
refactor(kit): drop `async` pipe and `let` directive

splincode:
Fix for flaky
image

### DIFF
--- a/projects/kit/components/tree/components/tree/tree.component.ts
+++ b/projects/kit/components/tree/components/tree/tree.component.ts
@@ -1,4 +1,4 @@
-import {AsyncPipe, NgForOf, NgIf} from '@angular/common';
+import {NgForOf, NgIf} from '@angular/common';
 import type {DoCheck, TrackByFunction} from '@angular/core';
 import {
     ChangeDetectionStrategy,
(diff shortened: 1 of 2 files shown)


projects/kit/components/tree/components/tree/tree.component.ts
```diff
@@ -48,10 +40,12 @@ export class TuiTreeComponent<T> implements DoCheck {
     @ViewChild(forwardRef(() => TuiTreeComponent))
     protected readonly child?: TuiTreeComponent<T>;
 
-    protected readonly children$ = this.check$.pipe(
-        startWith(null),
-        map(() => this.handler(this.value)),
-        distinctUntilChanged(),
+    protected readonly children = toSignal(
```
waterplea, line 43: `toSignal` immediately subscribes and at that time there's no `value`. I have a feeling this changes probably breaks something.

splincode: Don't worry, be happy :)

mdlufy:
> Fix for image

@splincode why tests not failed, if it fixes anything case?

splincode:
@mdlufy flaky

projects/kit/components/tree/components/tree/tree.component.ts
waterplea: Unless we can be sure, I don't think we should merge that.

PR #9491, opened 2024-10-14
chore(demo): fancy design for landing page

splincode:
image


waterplea:
I think that makes it harder to read (

splincode:
@waterplea what if 
image

or

image

waterplea:
I don't know, that seems a bit excessive.

PR #9510, opened 2024-10-15
feat(addon-doc): improve active zone for copy button

splincode:
### Before

image

### After

image

### DIFF
--- a/projects/addon-doc/components/code/index.html
+++ b/projects/addon-doc/components/code/index.html
@@ -12,8 +12,8 @@
     <div class="t-code-actions">
         <button type="button"
             tuiIconButton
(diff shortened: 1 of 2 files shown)


projects/addon-doc/components/code/index.less
```diff
@@ -50,6 +50,9 @@
     justify-content: center;
     align-items: center;
     flex-direction: row-reverse;
+    padding: 0.375rem;
```
waterplea, line 53: Can we just use whiteblock appearance and a bigger size?

splincode: fixed

waterplea (approved):

PR #9522, opened 2024-10-16
chore(styles): `tui-space` increase specifity

mdlufy:
Fixes #

### DIFF
--- a/projects/styles/markup/tui-space.less
+++ b/projects/styles/markup/tui-space.less
@@ -3,7 +3,7 @@
 .generate-space(@direction, @i: 0) when (@i =< @sizes) {
     .generate-space(@direction, @i + 1);
 
-    &_@{direction}-@{i} {
+    &_@{direction}-@{i}&_@{direction}-@{i} {
         & when (@direction = all) {
             margin: 0.25rem * @i;
         }


splincode (approved):

projects/styles/markup/tui-space.less
```diff
@@ -36,7 +36,7 @@
     }
 }
 
-.tui-space {
+.tui-space.tui-space {
```
waterplea, line 39: That wouldn't work as it would generate `.tui-space.tui-space_horizontal-x`. We need to update line 6 instead

mdlufy: Updated

projects/styles/markup/tui-space.less
```diff
@@ -3,7 +3,7 @@
 .generate-space(@direction, @i: 0) when (@i =< @sizes) {
     .generate-space(@direction, @i + 1);
 
-    &_@{direction}-@{i} {
+    &&_@{direction}-@{i} {
```
waterplea, line 6: That would do the same. We need something like `&_@{direction}-@{i}&_@{direction}-@{i}`

mdlufy: Oh, yes, updated
But now it looks some unsightly (

mdlufy: but I'm not sure if it will work. We need to check it out
image

mdlufy: @waterplea checked, yes, it's working:

image

waterplea (approved):

PR #9623, opened 2024-10-28
fix(kit): `InputDate` hold initial value with min/max defined

mdlufy:
Relates to #9582 

When textfield has initial control `value` (less than `min` or more than `max`), mask triggered and value transformed to min/max, what shouldn't be according to https://github.com/taiga-family/maskito/issues/1189#issuecomment-2394981669

### DIFF
--- a/projects/kit/components/input-date/input-date.component.ts
+++ b/projects/kit/components/input-date/input-date.component.ts
@@ -14,7 +14,7 @@ import {
     ViewChild,
 } from '@angular/core';
 import {NgControl} from '@angular/forms';
-import {MASKITO_DEFAULT_OPTIONS, MaskitoOptions} from '@maskito/core';
+import {MaskitoOptions} from '@maskito/core';
 import {maskitoDateOptionsGenerator} from '@maskito/kit';
 import {
     AbstractTuiNullableControl,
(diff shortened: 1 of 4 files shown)


projects/kit/components/input-date/input-date.component.ts
```diff
@@ -148,10 +149,28 @@ export class TuiInputDateComponent
     }
 
     get computedMin(): TuiDay {
+        /**
+         * TODO: we can delete this workaround in v4.0
+         * after solving this issue:
+         * https://github.com/taiga-family/maskito/issues/604
+         */
+        if (this.value && this.control?.pristine) {
+            return this.options.min;
```
mdlufy, line 158: Maybe it's better to direcrly return `TUI_FIRST_DAY`/`TUI_LAST_DAY`, because if provide option through DI, it will trigger value changes - and we don't want that

@n-barsukov what do you think about it?

nsbarsukov: Yep, I think `TUI_FIRST_DAY` / `TUI_LAST_DAY` is better 👍

mdlufy: Updated

waterplea (approved):

vladimirpotekhin (approved):

PR #9638, opened 2024-10-30
fix(kit): `Stepper` fix flashing styles

### LINKED ISSUE #9632: 🐞 - flashing default styles for stepper
### Reproduction url

https://stackblitz.com/edit/angular-mvs6s3

### Description

Default styles of tui-stepper are flashing activeItemIndex = -1
Can be reproduced - https://stackblitz.com/edit/angular-mvs6s3

### Taiga UI version

3.96.0

### Browser Used

Chrome

### OS Used

macOS

vladimirpotekhin:
Fixes #9632 

The first time, when the query list is empty yet, indexOf for all steps returns `-1`

Снимок экрана 2024-10-30 в 10 58 07

If we set `-1` to the `[activeItemIndex]` input of the stepper, it will cause the styles to flash since all steps are active for a millisecond.

https://github.com/user-attachments/assets/308f054e-abae-4457-8d02-18a0b1316cf6

### DIFF
--- a/projects/kit/components/stepper/stepper.component.ts
+++ b/projects/kit/components/stepper/stepper.component.ts
@@ -77,9 +77,11 @@ export class TuiStepperComponent {
     }
 
     public indexOf(step: HTMLElement): number {
-        return tuiGetOriginalArrayFromQueryList(this.steps).findIndex(
+        const index = tuiGetOriginalArrayFromQueryList(this.steps).findIndex(
             ({nativeElement}) => nativeElement === step,
         );
+
+        return index < 0 ? NaN : index;
     }
 
     public isActive(index: number): boolean {


projects/kit/components/stepper/stepper.component.ts
```diff
@@ -77,9 +77,11 @@ export class TuiStepperComponent {
     }
 
     public indexOf(step: HTMLElement): number {
-        return tuiGetOriginalArrayFromQueryList(this.steps).findIndex(
+        const index = tuiGetOriginalArrayFromQueryList(this.steps).findIndex(
             ({nativeElement}) => nativeElement === step,
         );
+
+        return index < 0 ? NaN : index;
```
mdlufy, line 84: Сan we add a test?

mdlufy (approved):

projects/kit/components/stepper/stepper.component.ts
vladimirpotekhin: no, unfortunately, at the time of the screenshot all the colors are normal

waterplea (approved):

PR #9650, opened 2024-10-31
fix(legacy): `InputDate` incorrect value after backspace

mdlufy:
When `value.length === DATE_FILLER_LENGTH (10)`, value may contain not only date string, but also letters string. 

```
this.value =
            value.length !== DATE_FILLER_LENGTH
                ? null
                : TuiDay.normalizeParse(value, this.dateFormat.mode);
```

This was not taken into account and as a result, an attempt was made to parse the date from the string and the value `01.01.0000` was obtained 

https://github.com/user-attachments/assets/92334503-3034-4294-b132-fc2bf195819c

### DIFF
--- a/projects/legacy/components/input-date/input-date.component.ts
+++ b/projects/legacy/components/input-date/input-date.component.ts
@@ -8,10 +8,10 @@ import {
 } from '@angular/core';
 import {takeUntilDestroyed} from '@angular/core/rxjs-interop';
 import type {MaskitoOptions} from '@maskito/core';
-import {MASKITO_DEFAULT_OPTIONS} from '@maskito/core';
+import {MASKITO_DEFAULT_OPTIONS, maskitoTransform} from '@maskito/core';
 import {maskitoDateOptionsGenerator} from '@maskito/kit';
 import {tuiAsControl} from '@taiga-ui/cdk/classes';
-import {TUI_FALSE_HANDLER} from '@taiga-ui/cdk/constants';
+import {TUI_FALSE_HANDLER, TUI_LETTER_REGEXP} from '@taiga-ui/cdk/constants';
 import type {TuiDateMode} from '@taiga-ui/cdk/date-time';
 import {
     DATE_FILLER_LENGTH,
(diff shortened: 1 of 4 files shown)


projects/legacy/components/input-date/input-date.component.ts
```diff
@@ -174,9 +174,13 @@ export class TuiInputDateComponent
         }
 
         this.value =
-            value.length !== DATE_FILLER_LENGTH
+            value.length !== DATE_FILLER_LENGTH || TUI_LETTER_REGEXP.test(value)
                 ? null
                 : TuiDay.normalizeParse(value, this.dateFormat.mode);
+
+        if (TUI_LETTER_REGEXP.test(this.nativeValue)) {
+            this.nativeValue = maskitoTransform(this.nativeValue, this.computedMask);
```
nsbarsukov, line 182: Why do we should use mask if textfield value contains any letters ? 🤔

mdlufy: When mask object changes from mask for any value to mask for date (when `this.activeItem` return `false`), it applies only when user changed next value, but not current. If we don't trigger it manually, it would be:
"Until today" -> (press firstly backspace) -> "Until toda" -> (press backspace secondly) -> ""

But we need: "Until today" -> (press firstly backspace) -> "", then we need to dispatch `maskitoTransform` to apply updated mask

nsbarsukov: Why not this?
```
this.nativeValue = ''
```

If we apply maskito date mask for any text value (not date string) – it will always return empty string

projects/legacy/components/input-date/input-date.component.ts
```diff
@@ -174,9 +174,13 @@ export class TuiInputDateComponent
         }
 
         this.value =
-            value.length !== DATE_FILLER_LENGTH
+            value.length !== DATE_FILLER_LENGTH || TUI_LETTER_REGEXP.test(value)
                 ? null
                 : TuiDay.normalizeParse(value, this.dateFormat.mode);
+
+        if (TUI_LETTER_REGEXP.test(this.nativeValue)) {
```
vladimirpotekhin, line 181: do we really need an extra regex? looks like we can check if a string value is in the input using existing `this.activeItem` getter

```
 if (this.activeItem) {
      this.nativeValue = '';
 }

  this.value =
            value.length !== DATE_FILLER_LENGTH || this.activeItem
                ? null
                : TuiDay.normalizeParse(value, this.dateFormat.mode);
```

mdlufy: Yes, it looks like it works in a similar way
Updated

vladimirpotekhin (approved):

waterplea (approved):

PR #9654, opened 2024-10-31
chore(docs): #9644 Added top position example for `ActionBar`

### LINKED ISSUE #9644: 🚀 -  Add example of top position for `ActionBar` in demo
### Description

Now `ActionBar` can be open only at bottom of the page. It would be great for some situations to add top position, too.
I suggest to add flag that takes position as a param.
Ready to do it by myself.

Image

40oleg:
Fixes #9644

This MR add example in the demo how to place `ActionBar` to the top on page
Discussed with @vladimirpotekhin and @waterplea
Example: https://taiga-ui.dev/components/actions-bar#top-position (URL will be available after merge)

image


40oleg:
One of test have to fail because I add a new test to check if position really on top

projects/demo-playwright/tests/kit/action-bar/action-bar.spec.ts
```diff
@@ -17,4 +17,15 @@ test.describe('ActionBar', () => {
 
         await expect(actionBarExample).toHaveScreenshot('01-actions-bar.png');
     });
+
+    test('should show on top', async ({page}) => {
+        await tuiGoto(page, DemoRoute.ActionBar);
+        const example = page.locator('#top-position');
+        const showActionBarButton = example.getByTestId('open-table-bar-on-top').first();
+
+        await showActionBarButton.click();
```
40oleg, line 26: The reason for the errors below is that I am adding a test to check if the position is on top

projects/demo/src/modules/components/action-bar/examples/3/index.less
```diff
@@ -0,0 +1,4 @@
+tui-action-bar {
+    top: 1rem;
+    bottom: unset !important;
```
vladimirpotekhin, line 3: it looks like `!important` is unnecessary

waterplea: It would be if we use class instead of a tag selector, since :host inside the component has higher specificity

projects/demo/src/modules/components/action-bar/examples/3/index.html
```diff
@@ -0,0 +1,35 @@
...
+        type="button"
+        (click)="open.set(false)"
+    >
+        Remove
+    </button>
+    <button
+        iconStart="@tui.x"
+        tuiIconButton
+        type="button"
+        [appearance]="isMobile() ? 'glass' : 'icon'"
```
waterplea, line 30: Maybe we should simplify the example since it should focus on positioning, rather than on content?

projects/demo/src/modules/components/action-bar/examples/3/index.less
40oleg: > It would be if we use class instead of a tag selector, since :host inside the component has higher specificity

It works fine without !important

40oleg:
> # Failed tests ❌
> ### Before (main) ← Diff → After (local)
> **tests-addon-doc-navigation-bbc63-getting-started-light-mode--chromium/01-tui-doc-navigation-light-mode.diff.png**
> 
> **tests-addon-doc-navigation-fa9c7--getting-started-dark-mode--chromium/02-tui-doc-navigation-dark-mode.diff.png**
> 
> **tests-demo-demo.pw-Demo-components-carousel-chromium/components-carousel/3.diff.png**
> 
> 
> (updated for commit [0ddc6d0](https://github.com/taiga-family/taiga-ui/commit/0ddc6d093d5b721a68d6b4d40c80d35836472ab1))

I suppose it is problem of all PRs?

waterplea (approved):

vladimirpotekhin (approved):

PR #9657, opened 2024-11-01
fix(addon-mobile): `TuiElasticStickyService` teardown observable zone fix

nezar-khamidekh:


vladimirpotekhin (approved):

splincode (approved):

projects/addon-mobile/directives/elastic-sticky/elastic-sticky.service.ts
```diff
@@ -35,6 +43,7 @@ export class TuiElasticStickyService extends Observable<number> {
                                     0,
                                 ),
                             ),
+                            tuiZoneOptimized(this.zone),
```
waterplea, line 46: Should we add `distinctUntilChanged`? Otherwise it works like just a zonefull stream

nezar-khamidekh: done

waterplea (approved):

PR #9681, opened 2024-11-04
fix: remove default styling from `a` tag in card component

### LINKED ISSUE #9664: 🐞 - `CardMedium` as anchor has some default browser styles
### Reproduction url

https://stackblitz.com/edit/card-medium-as-link-bug?file=src%2Fapp%2Fapp.template.html

### Description

```html
<a tuiAppearance="neutral" tuiCardMedium href="#">
  <tui-avatar src="@tui.star" appearance="primary" />

  <h2 tuiTitle>Dynamic QR‑code</h2>
</a>
```

returns the following card

Image

**TODO:**
Remove default browser color highlighting.

### Taiga UI version

4.13.0

### Browser Used

Chrome

### OS Used

macOS

michalgrzegorczyk-dev:
Fixes #9664

### DIFF
--- a/projects/layout/components/card/card.styles.less
+++ b/projects/layout/components/card/card.styles.less
@@ -26,9 +26,9 @@
         .scrollbar-hidden();
     }
 
-    :where(a&) {
-        color: inherit;
-    }
+    //:where(a&) {
+    //    color: inherit;
+    //}
 }
 
 [tuiCardMedium] {
--- a/projects/layout/components/card/card.styles.less
+++ b/projects/layout/components/card/card.styles.less
@@ -25,6 +25,10 @@
     > * {
         .scrollbar-hidden();
     }
+
+    :where(a&) {
+        color: inherit;
+    }
 }
 
 [tuiCardMedium] {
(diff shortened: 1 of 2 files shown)


projects/layout/components/card/card.styles.less
```diff
@@ -25,6 +25,10 @@
     > * {
         .scrollbar-hidden();
     }
+
+    :where(a&) {
+        color: inherit;
```
waterplea, line 30: We should just set `color: var(--tui-text-primary)` all the time, no matter what tag, since colors that come from appearances have higher specificity anyway, due to double attribute selector `[tuiAppearance][data-appearance='xxx']`

waterplea: Oops, I see, `neutral` appearance does not set color. It's a bug, all status appearances should have `color: var(--tui-text-primary)` here:
https://github.com/taiga-family/taiga-ui/blob/main/projects/core/styles/theme/appearance/status.less#L11

waterplea (approved):

splincode (approved):

PR #9702, opened 2024-11-06
chore: do not split previous diff png file

splincode:


scripts/visual-testing/combine-playwright-failed-screenshots.ts
```diff
@@ -19,8 +20,15 @@ const DIFF_IMAGE_POSTFIX = '-diff.png';
     }
 
     const imagesPaths: string[] = filesOrDirs
-        .filter((x) => x.isFile() && x.name.endsWith('.png'))
+        .filter(
+            (x) =>
+                x.isFile() &&
+                x.name.endsWith('.png') &&
+                !/-retry\d$/.exec(x.name) &&
```
nsbarsukov, line 27: This check duplicates twice now.

Let move it to the toppest level

```ts
const filesOrDirs = readdirSync(rootPath, {
        withFileTypes: true,
    }).filter(x => !/-retry\d$/.exec(x.name));
```

nsbarsukov (approved):

PR #9777, opened 2024-11-18
fix(kit): fix calendar range presets filtration

bondarvladislave:
**Summary**
I need to add a preset for the calendar: "Previous Period". This preset ensures that the selected period has the same length and directly precedes the current selection.

**For example:**

- If the selected range is 14–15, the preset will suggest 12–13.
- If the selected range is 10–15 (length 6), the preset will suggest 4–9.

**Implementation Details**
Ensures that the previous period matches the `minLength` and `maxLength` of the current selection.
Adjusts the range so the new period is directly adjacent to the current selection, maintaining its exact length.

**Issue**
https://stackblitz.com/edit/angular-ckrt12?file=src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.template.html
In the example at the link, the dates are different; it will always be Today - 2 days.
The length ends up being 3 days inclusive.
The preset will be:
`from: today - (3 + 2)` and `to: today - 3`

The library's filtering logic rejects the preset dates in some cases.

telegram-cloud-photo-size-2-5312075690281136275-x

However, I can select these dates manually, and in that case, the preset label is applied to the input.

telegram-cloud-photo-size-2-5312075690281136276-x

**For example:**
When the selected range is 14–15, the preset suggests 12–13.
`minLength = 2`, but `12 + 2 = 14`.
The library's mapper treats this as invalid because 14 does not satisfy `daySameOrBefore` with the "to" date, which is day 15.

telegram-cloud-photo-size-2-5312075690281136277-x

**Proposed Fix**
In the filtering logic, it seems necessary to include an adjustment like `append({ day: -1 })` to correctly validate the preset's range.

### DIFF
--- a/projects/kit/components/calendar-range/test/calendar-range.component.spec.ts
+++ b/projects/kit/components/calendar-range/test/calendar-range.component.spec.ts
@@ -10,6 +10,7 @@ import type {ComponentFixture} from '@angular/core/testing';
 import {TestBed} from '@angular/core/testing';
 import {FormControl, NgControl} from '@angular/forms';
 import {By} from '@angular/platform-browser';
+import type {TuiDayLike} from '@taiga-ui/cdk';
 import {
     TUI_LAST_DAY,
     tuiControlValue,
@@ -40,7 +41,9 @@ describe('rangeCalendarComponent', () => {
                 [items]="items"
                 [markerHandler]="markerHandler"
                 [max]="max"
+                [maxLength]="maxLength"
                 [min]="min"
+                [minLength]="minLength"
                 [value]="value"
                 (valueChange)="onRangeChange($event)"
             />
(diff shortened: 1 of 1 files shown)


projects/kit/components/calendar-range/calendar-range.component.ts
```diff
@@ -170,7 +170,10 @@ export class TuiCalendarRange implements OnInit, OnChanges {
         ...items.filter(
             (item) =>
                 (minLength === null ||
-                    item.range.from.append(minLength).daySameOrBefore(item.range.to)) &&
+                    item.range.from
+                        .append(minLength)
+                        .append({day: -1})
+                        .daySameOrBefore(item.range.to)) &&
```
splincode, line 176: @mdlufy Can we implement tests for this case?

mdlufy: Yes, it would be good to add a test,  I think we can do it

mdlufy: @bondarvladislave do you have a desire to add a test?)

splincode (approved):

mdlufy (approved):

PR #9818, opened 2024-11-22
fix(kit): `InputInline` has broken scroll in Safari & Firefox

### LINKED ISSUE #7552: 🐞 - `InputInline` has broken scroll in Safari / Firefox

nsbarsukov:
Fixes #7552

### DIFF
--- a/projects/kit/components/input-inline/input-inline.component.ts
+++ b/projects/kit/components/input-inline/input-inline.component.ts
@@ -8,7 +8,6 @@ import {
 import {NgControl} from '@angular/forms';
 import {TuiLet} from '@taiga-ui/cdk/directives/let';
 import {tuiControlValue} from '@taiga-ui/cdk/observables';
-import {tuiIsElement, tuiIsInput} from '@taiga-ui/cdk/utils/dom';
 import {defer} from 'rxjs';
 
 @Component({
@@ -19,21 +18,10 @@ import {defer} from 'rxjs';
     styleUrls: ['./input-inline.style.less'],
     encapsulation: ViewEncapsulation.None,
     changeDetection: ChangeDetectionStrategy.OnPush,
-    host: {
-        '(scroll.capture)': 'onScroll($event.target)',
-    },
 })
 export class TuiInputInline {
     @ContentChild(NgControl)
     private readonly control?: NgControl;
 
     protected readonly value$ = defer(() => tuiControlValue(this.control!));
-
-    protected indent = -1;
-
-    protected onScroll(target: EventTarget | null): void {
-        if (tuiIsElement(target) && tuiIsInput(target)) {
-            this.indent = -target.scrollLeft - 1; // -1 for Safari (see styles)
-        }
-    }
 }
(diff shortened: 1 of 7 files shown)


projects/kit/components/input-inline/input-inline.component.ts
```diff
@@ -19,21 +18,10 @@ import {defer} from 'rxjs';
...
     @ContentChild(NgControl)
     private readonly control?: NgControl;
 
     protected readonly value$ = defer(() => tuiControlValue(this.control!));
-
-    protected indent = -1;
-
-    protected onScroll(target: EventTarget | null): void {
-        if (tuiIsElement(target) && tuiIsInput(target)) {
-            this.indent = -target.scrollLeft - 1; // -1 for Safari (see styles)
```
nsbarsukov, line 36: We can replace this logic with 

```ts
protected indent$ = merge(
    fromEvent(this.element, 'scroll', {capture: true}), // Only Chrome emits this event for <input />
    // Chrome does not fire `selectionchange` on deletion (but do it on insertion) ¯\_(ツ)_/¯
    fromEvent(this.element, 'selectionchange', {capture: true}), // Firefox + Safari fallback
    // Chrome & Safari scrolls input to the beginning on blur
    // Firefox – keeps the same scrollLeft position on blur
    fromEvent(this.element, 'focusout').pipe(
        concatMap(
            (x) => of(x).pipe(delay(0)), // Without delay `scrollLeft` is not yet updated
        ),
    ),
).pipe(
    map(({target}) => target),
    filter((x): x is HTMLInputElement => tuiIsElement(x) && tuiIsInput(x)),
    map(
        ({scrollLeft}) => -scrollLeft - 1, // -1 for Safari (see styles)
    ),
);
```

However, it still does fully solves all issue for Safari / Firefox.
Try to insert many characters and the use many Backspaces then.
The content will slightly shift to the left: textContent of ghost-span updates faster than textIndent calculations (shrinking of ghost-span should be AFTER textIndent calculations).
https://github.com/taiga-family/taiga-ui/blob/432a4867f60c99fa266cd1bfaadf2e983d7c4984/projects/kit/components/input-inline/input-inline.template.html#L2-L6
Even if we delay update of textContent  (by adding delay(0) to `indent$` stream) - insertion of new character will be broken (we should increase size of ghost-span BEFORE textIndent calculations).

splincode (approved):

waterplea (approved):

splincode (approved):

PR #9820, opened 2024-11-22
fix(layout): `Navigation` fix aside width according to spec

vladimirpotekhin:
Fixes #

### DIFF
--- a/projects/experimental/components/navigation/aside.style.less
+++ b/projects/experimental/components/navigation/aside.style.less
@@ -17,7 +17,7 @@ aside[tuiNavigationAside] {
         width: 14.375rem;
 
         + main[tuiNavigationMain] {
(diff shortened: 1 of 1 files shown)


projects/experimental/components/navigation/aside.style.less
```diff
@@ -14,7 +14,7 @@ aside[tuiNavigationAside] {
     color: var(--tui-text-01-night);
 
     &._expanded {
-        width: 13.75rem;
+        width: 14.375rem;
 
         + main[tuiNavigationMain] {
             max-width: calc(100% - 13.75rem);
```
splincode, line 20: `13.75rem` does this value depend on the value above?

splincode (approved):

mdlufy (approved):

PR #9838, opened 2024-11-25
fix(legacy): `InputDateTime` validators not triggered when value change

### LINKED ISSUE #9839: 🐞 - InputDateTime validation
### Reproduction url

https://stackblitz.com/edit/angular-ncy3dq

### Description

Validators on tui-input-date-time only calls when entered date is valid.
In this case there is no opportunity to show validation message on invalid input.
For example, when default values of control is [null, null], user inputs 11 we should show validation message, but can't, because custom validator doesn't called.

### Taiga UI version

4.16.0

### Browser Used

Chrome

### OS Used

macOS

mdlufy:
Fixes #9839

### DIFF
--- a/projects/demo-playwright/tests/legacy/input-date-time/input-date-time.pw.spec.ts
+++ b/projects/demo-playwright/tests/legacy/input-date-time/input-date-time.pw.spec.ts
@@ -310,4 +310,37 @@ test.describe('InputDateTime', () => {
             });
         });
     });
+
+    test.describe('Examples', () => {
+        let example!: Locator;
+        let documentationPage!: TuiDocumentationPagePO;
+        let inputDateTime!: TuiInputDateTimePO;
+
+        test.beforeEach(async ({page}) => {
+            await tuiGoto(page, DemoRoute.InputDateTime);
+
+            documentationPage = new TuiDocumentationPagePO(page);
+            example = documentationPage.apiPageExample;
+
+            inputDateTime = new TuiInputDateTimePO(
+                example.locator('tui-input-date-time'),
+            );
+        });
+
+        test('With validator: enter incomplete date -> validator error', async () => {
+            example = documentationPage.getExample('#with-validator');
+            inputDateTime = new TuiInputDateTimePO(
+                example.locator('tui-input-date-time'),
+            );
+
+            await inputDateTime.textfield.clear();
+            await inputDateTime.textfield.fill('11');
+            await inputDateTime.textfield.blur();
+
+            await expect(example).toHaveScreenshot(
+                '04-input-data-time-with-validator.png',
+                {animations: 'allow'},
+            );
+        });
+    });
 });
--- a/projects/demo-playwright/tests/legacy/input-date-time/input-date-time.pw.spec.ts
+++ b/projects/demo-playwright/tests/legacy/input-date-time/input-date-time.pw.spec.ts
@@ -337,6 +337,7 @@ test.describe('InputDateTime', () => {
             await inputDateTime.textfield.fill('11');
             await inputDateTime.textfield.blur();
 
+            // allow animations to capture tui-error validation message on screenshot
             await expect(example).toHaveScreenshot(
                 '04-input-data-time-with-validator.png',
                 {animations: 'allow'},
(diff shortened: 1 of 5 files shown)


projects/demo-playwright/tests/legacy/input-date-time/input-date-time.pw.spec.ts
```diff
@@ -310,4 +310,37 @@ test.describe('InputDateTime', () => {
...
+                example.locator('tui-input-date-time'),
+            );
+
+            await inputDateTime.textfield.clear();
+            await inputDateTime.textfield.fill('11');
+            await inputDateTime.textfield.blur();
+
+            await expect(example).toHaveScreenshot(
+                '04-input-data-time-with-validator.png',
+                {animations: 'allow'},
```
mdlufy, line 342: To capture validation message `Value is invalid` on scheenshot

image

nsbarsukov, line 342: Should we add some wait before this screenshot for test stability ?

```suggestion
            await page.waitForTimeout(100);
            
            await expect(example).toHaveScreenshot(
                '04-input-data-time-with-validator.png',
                {animations: 'allow'},
```

splincode: Add comment to line

splincode (approved):

projects/demo/src/modules/components/input-date-time/examples/7/index.ts
```diff
@@ -0,0 +1,38 @@
...
+    encapsulation,
+    changeDetection,
+})
+export default class Example {
+    protected readonly testForm = new FormGroup({
+        testValue: new FormControl(
+            [new TuiDay(2017, 2, 15), null],
+            completeDateTimeValidator,
+        ),
+    });
```
nsbarsukov, line 37: Let's simplify these lines (to focus user on the main points):

```suggestion
    protected readonly control = new FormControl(
            [new TuiDay(2017, 2, 15), null],
            completeDateTimeValidator,
    );
```

\+ use only `[formControl]="control"` inside template

projects/demo/src/modules/components/input-date-time/index.html
```diff
@@ -115,6 +115,20 @@ <h3>DI-tokens for input-configurations:</h3>
...
+            id="with-validator"
+            heading="With validator"
+            [component]="7 | tuiComponent"
+            [content]="7 | tuiExample: 'html,ts'"
+            [description]="withValidatorDescription"
+        >
+            <ng-template #withValidatorDescription>
+                Create custom validator function to verify input value. Custom
+                <code>completeDateTimeValidator</code>
+                using to check if data is complete
```
nsbarsukov, line 129: I wonder if this description is really required 🤔

projects/demo-playwright/tests/legacy/input-date-time/input-date-time.pw.spec.ts
mdlufy: I do not know, is it really necessary? It is still working

splincode: why we use `{animations: 'allow'}` here

projects/demo/src/modules/components/input-date-time/index.html
mdlufy: Do you think to remove description at all?

projects/demo/src/modules/components/input-date-time/examples/7/index.ts
mdlufy: Done

projects/demo-playwright/tests/legacy/input-date-time/input-date-time.pw.spec.ts
mdlufy: Added

projects/demo/src/modules/components/input-date-time/index.html
mdlufy: I looked at it again and I think it's obvious here and the description is really redundant
Removed 👍

nsbarsukov (approved):

PR #9846, opened 2024-11-26
fix(kit): `InputDateTime` validators not triggered when value change

### LINKED ISSUE #9839: 🐞 - InputDateTime validation

mdlufy:
Fixes #9839

### DIFF
--- a/projects/demo-playwright/tests/kit/input-date-time/input-date-time.spec.ts
+++ b/projects/demo-playwright/tests/kit/input-date-time/input-date-time.spec.ts
@@ -241,4 +241,37 @@ describe('InputDateTime', () => {
             });
         });
     });
+
+    test.describe('Examples', () => {
+        let example!: Locator;
+        let documentationPage!: TuiDocumentationPagePO;
+        let inputDateTime!: TuiInputDateTimePO;
+
+        test.beforeEach(async ({page}) => {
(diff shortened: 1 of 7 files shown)


projects/demo-playwright/tests/kit/input-date-time/input-date-time.spec.ts
```diff
@@ -241,4 +241,37 @@ describe('InputDateTime', () => {
...
+                example.locator('tui-input-date-time'),
+            );
+
+            await inputDateTime.textfield.clear();
+            await inputDateTime.textfield.fill('11');
+            await inputDateTime.textfield.blur();
+
+            await expect(example).toHaveScreenshot(
+                '04-input-data-time-with-validator.png',
+                {animations: 'allow'},
```
mdlufy, line 273: To capture validation message `Value is invalid` on scheenshot

image

splincode (approved):

PR #9847, opened 2024-11-26
fix(kit): `CalendarRange` should not require extra click to start new range after same day range

### LINKED ISSUE #7550: 🐞 - Additional click required in InputDateRange for same day period
### Playground Link

_No response_

### Description

If a period of one day is selected in InputDateRange (for example `[TuiDay.currentLocal(), TuiDay.currentLocal()]` or manually select same day for start and end) when you open the date selection in the calendar, the first click changes the end date, and only after that you can change the start date along with the end date.
When a period of several days is initially selected, the first click changes the start date, and only then the end date.

This can be reproduced in API section in docs for InputDateRange.

### Angular version

17.3

### Taiga UI version

3.80

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [X] Android

nsbarsukov:
Fixes #7550

### Problem 
Low-level component `TuiCalendarSheet` decides if user completes selection of range period or not:
https://github.com/taiga-family/taiga-ui/blob/031f55054aec8f856f61f37461d7690ef1aaca99/projects/core/components/calendar/calendar-sheet.component.ts#L38-L39
https://github.com/taiga-family/taiga-ui/blob/031f55054aec8f856f61f37461d7690ef1aaca99/projects/core/components/calendar/calendar-sheet.component.ts#L124-L125

`TuiCalendarSheet` is dumb and does not have any internal state – it only depends on input property `this.value` 
https://github.com/taiga-family/taiga-ui/blob/031f55054aec8f856f61f37461d7690ef1aaca99/projects/core/components/calendar/calendar-sheet.component.ts#L57-L58
and emits output event `dayClick` on any day click
https://github.com/taiga-family/taiga-ui/blob/031f55054aec8f856f61f37461d7690ef1aaca99/projects/core/components/calendar/calendar-sheet.component.ts#L69-L70

However, at the present time (if `TuiCalendarSheet` is used inside `CalendarRange`) `this.value` is always `TuiDayRange | null` (higher-level component `CalendarRange` always sends only the such kinds of value). So, **it is impossible to distinguish if `this.value` with same day period** (e.g. `TuiDayRange(today, today)`) **is finished selection of range period** (it happens if user clicks twice on the same day) **or it is just an intermediate value for incomplete selection**.

### Could we just listen to `click` event inside `TuiCalendarSheet` to solve issue?
No. `CalendarRange` includes two `TuiCalendarSheet`-s. User can start selection on the 1st sheet and finish it on the second one. I am not sure that different instances of `TuiCalendarSheet` should "communicate" between each others.

### New proposed approach
Don't use `TuiDayRange(sameDay, sameDay)` for incomplete date range.
Incomplete date range – `TuiDay`.
Complete date range with the same day (user clicks twice on the same day) – `TuiDayRange`.

So, `CalendarRange` / `MobileCalendar[single=true]` can pass `TuiDay` inside `TuiCalendarSheet` to indicate incomplete range selection.

### DIFF
--- a/projects/kit/components/calendar-range/calendar-range.component.ts
+++ b/projects/kit/components/calendar-range/calendar-range.component.ts
@@ -10,10 +10,11 @@ import {
 } from '@angular/core';
 import {takeUntilDestroyed} from '@angular/core/rxjs-interop';
 import {TUI_FALSE_HANDLER} from '@taiga-ui/cdk/constants';
-import type {TuiDay, TuiDayLike} from '@taiga-ui/cdk/date-time';
+import type {TuiDayLike} from '@taiga-ui/cdk/date-time';
 import {
     TUI_FIRST_DAY,
     TUI_LAST_DAY,
+    TuiDay,
     TuiDayRange,
     TuiMonth,
 } from '@taiga-ui/cdk/date-time';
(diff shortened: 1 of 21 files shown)


projects/kit/components/calendar-range/calendar-range.component.ts
```diff
@@ -125,12 +137,6 @@ export class TuiCalendarRange implements OnInit, OnChanges {
         this.selectedPeriod = period;
     }
 
-    public ngOnChanges(): void {
-        if (!this.value) {
-            this.initDefaultViewedMonth();
-        }
-    }
-
```
mdlufy, line 133: When inputs (for example `defaultViewedMonth`) updated, we need to run `this.initDefaultViewedMonth()`. Not only `value` input

waterplea (approved):

splincode (approved):

PR #9865, opened 2024-11-29
fix(cdk): optional chaining for tuiControlValue subscription

### LINKED ISSUE #9758: 🐞 - v3 experimental TuiSegmented throws error on initialization
### Reproduction url

https://stackblitz.com/edit/stackblitz-starters-pf2rkh?file=src%2Fmain.ts

### Description

After updating from 3.94 to 3.98, `tui-segmented` component started throwing the following error on initialization

`ERROR TypeError: Cannot read properties of undefined (reading 'valueChanges')`

Stack trace leads to the following code

```js
// node_modules/@taiga-ui/cdk/fesm2015/taiga-ui-cdk-observables.js
function tuiControlValue(control) {
  return new Observable((subscriber) => {
    if (!control.valueChanges) { // <--
      throw new TuiValueChangesException();
    }
    return control.valueChanges.pipe(startWith(control.value)).subscribe(subscriber);
  });
}
```

May be caused by #9074?

### Taiga UI version

3.98.0

### Browser Used

Chrome

### OS Used

Windows

dequubi:
Fixes #9758

Solution from #8605

### DIFF
--- a/projects/cdk/exceptions/index.ts
+++ b/projects/cdk/exceptions/index.ts
@@ -8,5 +8,6 @@ export * from './owner-document.exception';
 export * from './pure.exception';
 export * from './table-sort-key.exception';
 export * from './ts-parser.exception';
+export * from './value-changes.exception';
 export * from './value-present.exception';
 export * from './xml-parsing.exception';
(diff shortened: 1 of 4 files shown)


dequubi:
Can not seem to reproduce test failure locally. The only diff I get is this.

image

waterplea (approved):
Should we keep the exception just in case not to introduce breaking change?

vladimirpotekhin (approved):

PR #9883, opened 2024-12-02
refactor: support nested diff directories

splincode:


projects/demo/src/modules/components/input-inline/examples/1/index.html
```diff
@@ -1,27 +1 @@
-<form [formGroup]="testForm">
```
splincode, line 1: check failed snapshots

vladimirpotekhin (approved):

nsbarsukov (approved):

PR #9886, opened 2024-12-02
feat(legacy): limit `tui-tag` width to `80%` to avoid wrapping in selects

### LINKED ISSUE #4255: 🐞 - Empty line in multi-select in case of long tags
### Playground Link

https://stackblitz.com/edit/angular-ycryww?file=src%2Fapp%2Fapp.component.ts

### Description

When tag values in multi-select are very long, there is an empty line (even if select not focused).

### Angular version

12.0.0

### Taiga UI version

2.83.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
Fixes #4255

### Before

image

### After

image

### DIFF
--- a/projects/demo/src/modules/components/multi-select/examples/1/index.ts
+++ b/projects/demo/src/modules/components/multi-select/examples/1/index.ts
@@ -9,7 +9,7 @@ import {TuiDataListWrapper} from '@taiga-ui/kit';
 import {TuiMultiSelectModule, TuiTextfieldControllerModule} from '@taiga-ui/legacy';
 
 const ITEMS: readonly string[] = [
-    'Luke Skywalker',
+    'Luke Skywalker and a long time ago in a galaxy far, far away..',
     'Leia Organa Solo',
     'Darth Vader',
     'Han Solo',
--- a/projects/legacy/components/input-tag/input-tag.style.less
+++ b/projects/legacy/components/input-tag/input-tag.style.less
@@ -82,6 +82,10 @@
     :host[data-size='l']:not(._label-outside) & {
         padding: 0;
     }
+
+    tui-tag:last-of-type {
+        max-inline-size: 80%;
+    }
 }
 
 .t-content {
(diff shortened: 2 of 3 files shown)


projects/legacy/components/input-tag/input-tag.style.less
```diff
@@ -82,6 +82,10 @@
     :host[data-size='l']:not(._label-outside) & {
         padding: 0;
     }
+
+    & > tui-tag:nth-child(1):has(+ .t-input-wrapper) {
```
waterplea, line 86: Good idea, but it breaks if there are tags before:
image
How about we just make :last-child tag max width 80%?

splincode: @waterplea fixed

projects/legacy/components/input-tag/input-tag.style.less
```diff
@@ -82,6 +82,10 @@
     :host[data-size='l']:not(._label-outside) & {
         padding: 0;
     }
+
+    & > tui-tag:has(+ .t-input-wrapper) {
```
waterplea, line 86: No need for `:has` stuff, just the `:last-child` — we don't want a single tag to only take up 80% of the input if it is alone on the row, we only want the last one which is right next to input

splincode: I don't understand how it should work
image

image

waterplea: It's not `:last-child`, apparently:
image
`:last-of-type` should work.

splincode: fixed

waterplea (approved):

vladimirpotekhin (approved):

projects/demo/src/modules/components/multi-select/examples/1/index.ts
```diff
@@ -9,7 +9,7 @@ import {TuiDataListWrapper} from '@taiga-ui/kit';
 import {TuiMultiSelectModule, TuiTextfieldControllerModule} from '@taiga-ui/legacy';
 
 const ITEMS: readonly string[] = [
-    'Luke Skywalker',
+    'Luke Skywalker and a long time ago in a galaxy far, far away..',
```
vladimirpotekhin, line 12: nit: looks like one example of a long tag is enough for the test

splincode: fixed

PR #9888, opened 2024-12-03
fix(core): `DropdownOpen` has CD problems after switching of `tuiDropdownEnabled` (`false`=>`true`)

nsbarsukov:
### Reproduction
```html
<tui-input [formControl]="control">
  Enter 3 characters
  
  @if (control.value.length > 2) {
    <tui-data-list *tuiDataList> Blah blah </tui-data-list>
  }
</tui-input>
```

Enter `123`

**Expected behavior**: dropdown appears
**Actual behavior**: no dropdown

Explore more:
https://stackblitz.com/edit/input-with-datalist-v4-bug?file=src%2Fapp%2Fapp.template.html

### DIFF
--- a/projects/demo-cypress/src/tests/input-with-datalist.cy.ts
+++ b/projects/demo-cypress/src/tests/input-with-datalist.cy.ts
@@ -1,0 +1,58 @@
+import {AsyncPipe, NgIf} from '@angular/common';
+import {ChangeDetectionStrategy, Component} from '@angular/core';
+import {FormControl, ReactiveFormsModule} from '@angular/forms';
+import {TuiDataList, TuiRoot} from '@taiga-ui/core';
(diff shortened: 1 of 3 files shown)


projects/demo-cypress/src/tests/input-with-datalist.cy.ts
```diff
@@ -0,0 +1,58 @@
...
+        </tui-root>
+    `,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TestInputWithDatalist {
+    protected readonly control = new FormControl('');
+}
+
+describe('Input + Datalist', () => {
+    describe('Shows dropdown only textfield contains >2 characters', () => {
```
nsbarsukov, line 28:
```suggestion
    describe('Shows dropdown only if textfield contains >2 characters', () => {
```

PR #9889, opened 2024-12-03
fix(core): prevent scrolling by arrow keys the whole page in safari

### LINKED ISSUE #9837: 🐞 - Different keyboard behavior when focusing in dropdown
### Reproduction url

https://taiga-ui.dev/components/data-list#submenu

### Description

https://github.com/user-attachments/assets/812f708c-51d0-46d7-b2c3-47155a5ca3fc

### Taiga UI version

latest

### Browser Used

Safari

### OS Used

macOS

splincode:
Fixes #9837


waterplea:
This is because dumb Safari does not focus buttons on click unless you toggle checkbox in OS settings to do so. We cannot assume we need to take over ALL arrow up/down from entire page just because data list is visible.

PR #9904, opened 2024-12-04
fix(addon-table): sort icons are reversed

### LINKED ISSUE #9884: 🐞 - `Table` sort icons in table header are reversed
### Reproduction url

https://taiga-ui.dev/components/table

### Description

If you check `th` directive source code, you'll see that the sort icons are reversed:
https://github.com/taiga-family/taiga-ui/blob/76d8d01006481564500ec1e1f328fe748b6d9879/projects/addon-table/components/table/th/th.component.ts#L73

`direction === 1` means `ascending` order and `-1` means `descending` order, but the code is the opposite way:
```ts
this.table?.direction === 1
    ? this.options.sortIcons.desc
    : this.options.sortIcons.asc;
```

### Taiga UI version

4.17.0

### Browser Used

Chrome

### OS Used

Windows

hakimio:
Fixes #9884

### DIFF
--- a/projects/addon-table/components/table/table.options.ts
+++ b/projects/addon-table/components/table/table.options.ts
@@ -2,8 +2,14 @@ import type {Provider} from '@angular/core';
 import {tuiCreateToken, tuiProvideOptions} from '@taiga-ui/cdk/utils/miscellaneous';
 import type {TuiSizeL, TuiSizeS} from '@taiga-ui/core/types';
 
+export const TuiSortDirection = {
+    Asc: 1,
+    Desc: -1,
+} as const;
+export type TuiSortDirection = (typeof TuiSortDirection)[keyof typeof TuiSortDirection];
+
 export interface TuiTableOptions {
-    readonly direction: -1 | 1;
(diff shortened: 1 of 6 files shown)


hakimio:
@waterplea @vladimirpotekhin can you review the PR?

projects/addon-table/components/table/table.options.ts
```diff
@@ -2,8 +2,14 @@ import type {Provider} from '@angular/core';
 import {tuiCreateToken, tuiProvideOptions} from '@taiga-ui/cdk/utils/miscellaneous';
 import type {TuiSizeL, TuiSizeS} from '@taiga-ui/core/types';
 
+export const TuiDirection = {
```
nsbarsukov, line 5: We already have many `Tui*Direction` types/constants: `TuiSwipeDirection`, `TuiGradientDirection`, `TuiHintDirection` etc.
New constant with just single word `direction` can be rather confusing.

Let's rename this const `TuiDirection` => `TuiSortDirection`.

waterplea (approved):

PR #9908, opened 2024-12-04
feat(legacy): add `itemsHidden` prop to `tui-input-time`

### LINKED ISSUE #3519: 🚀 - InputTime hide dropdown list items by settings flag

kit

kit

### Description
When I using InputTime component with  [items] and [strict] properties, I have a dropdown list of items, and rounding to nearest value from items list.
I want to suggest making it possible to disable the display of the dropdown menu by input.

splincode:
Fixes #3519

### DIFF
--- a/projects/demo/src/modules/components/input-time/index.html
+++ b/projects/demo/src/modules/components/input-time/index.html
@@ -103,6 +103,7 @@
                     [focusable]="focusable"
                     [formControl]="control"
                     [items]="items"
+                    [itemsHidden]="itemsHidden"
                     [itemSize]="itemSize"
                     [mode]="mode"
                     [pseudoFocus]="pseudoFocused"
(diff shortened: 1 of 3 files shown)


projects/demo/src/modules/components/input-time/index.html
```diff
@@ -158,6 +159,14 @@ <h3>DI-tokens for input-configurations:</h3>
             >
                 Items to choose
             </ng-template>
+            <ng-template
+                documentationPropertyMode="input"
+                documentationPropertyName="itemsHidden"
+                documentationPropertyType="boolean"
+                [(documentationPropertyValue)]="itemsHidden"
+            >
+                Disable the display of the dropdown list
```
nsbarsukov, line 168: Should we just use `inject(TuiDropdownOpen, {optional: true}).tuiDropdownEnabled` instead of creating new prop ? 🤔

waterplea: I'm not sure `tuiDropdownEnabled` exists for legacy controls, this prop will only be needed in legacy, we would definitely implement this differently once we rewrite `InputTime` to new `Textfield`

waterplea (approved):

nsbarsukov (approved):

PR #9909, opened 2024-12-04
feat(kit): add `itemsHidden` prop to `tui-input-time`

### LINKED ISSUE #3519: 🚀 - InputTime hide dropdown list items by settings flag

splincode:
Fixes #3519

### DIFF
--- a/projects/demo/src/modules/components/input-time/input-time.template.html
+++ b/projects/demo/src/modules/components/input-time/input-time.template.html
@@ -100,6 +100,7 @@
                     [focusable]="focusable"
                     [formControl]="control"
                     [items]="items"
+                    [itemsHidden]="itemsHidden"
                     [itemSize]="itemSize"
                     [mode]="mode"
                     [pseudoFocus]="pseudoFocused"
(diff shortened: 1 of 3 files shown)


projects/demo/src/modules/components/input-time/input-time.template.html
```diff
@@ -155,6 +156,14 @@ <h3>DI-tokens for input-configurations:</h3>
             >
                 Items to choose
             </ng-template>
+            <ng-template
+                documentationPropertyMode="input"
+                documentationPropertyName="itemsHidden"
+                documentationPropertyType="boolean"
+                [(documentationPropertyValue)]="itemsHidden"
+            >
+                Disable the display of the dropdown list
```
nsbarsukov, line 165: Should we just use `inject(TuiDropdownOpen, {optional: true}).tuiDropdownEnabled` instead of creating new prop ? 🤔

splincode: I think most people are on version 3, so it would be great to have this option already.

nsbarsukov (approved):

splincode:
@waterplea could you review please?

waterplea (approved):

PR #9932, opened 2024-12-06
fix(legacy): `InputDate{Range|Time}` + `input[tuiTextfieldLegacy]` + `tuiTextfieldFiller` has CD problems

### LINKED ISSUE #9642: 🐞 - `PrimitiveTextfield` with projected `<input tuiTextfieldLegacy />` + `tuiTextfieldFiller` has change detection problems
### Reproduction url

https://taiga-ui.dev/components/input-date#date-localization

### Description

1. Open any example with projection of `<input tuiTextfieldLegacy />` inside `PrimitiveTextfield`:
    - https://taiga-ui.dev/components/input-date#date-localization
    - https://taiga-ui.dev/components/input-date-range#base
2. Erase value

https://github.com/user-attachments/assets/8e01dd99-fec3-4f6b-a203-40354dc7ea0e

   

### Taiga UI version

4.13.0

### Browser Used

Chrome

### OS Used

macOS

nsbarsukov:
Fixes #9642

image

### DIFF
--- a/projects/legacy/components/input-date-range/input-date-range.component.ts
+++ b/projects/legacy/components/input-date-range/input-date-range.component.ts
@@ -3,6 +3,7 @@ import {
     Component,
     inject,
     Input,
+    signal,
     ViewChild,
 } from '@angular/core';
 import {takeUntilDestroyed} from '@angular/core/rxjs-interop';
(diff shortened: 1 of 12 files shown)


projects/legacy/components/input-date-range/input-date-range.component.ts
```diff
@@ -86,6 +87,7 @@ export class TuiInputDateRangeComponent
     private readonly mobileCalendar = inject(TUI_MOBILE_CALENDAR, {optional: true});
     private readonly options = inject(TUI_INPUT_DATE_OPTIONS);
     private readonly textfieldSize = inject(TUI_TEXTFIELD_SIZE);
+    private readonly nativeValue = signal(this.nativeFocusableElement?.value || '');
```
waterplea, line 90: It will always be `''`, because it's a view child, it doesn't exist in constructor. Are we sure it works right when there is a value in the input initially?

waterplea (approved):

splincode (approved):

projects/legacy/components/input-date-range/input-date-range.component.ts
nsbarsukov: fixed

PR #9955, opened 2024-12-10
feat(core): enable TuiHint hover when TuiHintManual is null

### LINKED ISSUE #9935: 🚀 - Ability to set `tuiHintManual` as null
### Description

Ability to set `tuiHintManual` as null will provide users with a dynamic usage of the `tuiHintManual` in tuiHint. 

So, whenever someone has a case where they want to sometimes use `tuiHintManual` and sometimes use the tuiHint auto show/hide they can achieve that buy sending [tuiHintManual]='alwaysShow ? true : null'. 

Right now it only accepts boolean values.

Thank you.

MuhammedSmani:
Fixes #9935


splincode (approved):

projects/core/directives/hint/hint-manual.directive.ts
```diff
@@ -12,19 +13,22 @@ import {TuiHintHover} from './hint-hover.directive';
 })
 export class TuiHintManual extends TuiDriver implements OnChanges {
     private readonly hover = inject(TuiHintHover);
-    private readonly stream$ = new BehaviorSubject(false);
+    private readonly stream$ = new BehaviorSubject<boolean | null>(false);
```
waterplea, line 16: You shouldn't change the type, just use `!!` to cast `null` to `false` when you call `next`.

projects/core/directives/hint/hint-manual.directive.ts
```diff
@@ -15,16 +16,19 @@ export class TuiHintManual extends TuiDriver implements OnChanges {
...
 
     @Input()
-    public tuiHintManual = false;
+    public tuiHintManual: boolean | null = false;
 
     public readonly type = 'hint';
 
     constructor() {
-        super((subscriber) => this.stream$.subscribe(subscriber));
+        super((subscriber: Subscriber<boolean | null>) =>
```
waterplea, line 24: This change is wrong, it should remain as it was

waterplea:
@MuhammedSmani would you be able to address my last comment and fix the unit test?

waterplea (approved):

PR #9960, opened 2024-12-10
chore(demo): always check options by host in virtual scroll

### LINKED ISSUE #5725: 🐞 - paste selection does not work as expected in virtual scroll ComboBox
### Playground Link

_No response_

### Description

Options that were not previously displayed in virtual scroll ComboBox can not be correctly selected when option title (full-matched with some option name) is pasted from clipboard.

Steps to reproduce:
1. Open https://taiga-ui.dev/components/combo-box#virtual
2. Copy `Zimbabwe` to your clipboard
3. Paste clipboard value to ComboBox
4. Scroll to the end of list
5. `Zimbabwe` is not selected

Expected behavior:
- `Zimbabwe` is selected (checkmark should be shown)

How to reproduce expected behavior:
1. Copy `Albania` to your clipboard
2. Paste clipboard value to ComboBox
3. `Albania` is selected, checkmark is shown

P.S. Partial-matching clipboard value works as expected. E.g. `Zimbabw` or `Albani` trigger search/filtering as expected. When you paste (or manually search) such value, matching option would be selected correctly.

### Angular version

14.0.0

### Taiga UI version

3.46.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

splincode:
Fixes #5725

### DIFF
--- a/projects/demo/src/modules/components/combo-box/examples/5/index.html
+++ b/projects/demo/src/modules/components/combo-box/examples/5/index.html
@@ -5,7 +5,6 @@
         <!--8 = data list padding-->
         <cdk-virtual-scroll-viewport
             *tuiLet="countries | tuiFilterByInput as items"
-            appendOnly
             tuiScrollable
             class="scroll"
             [itemSize]="48"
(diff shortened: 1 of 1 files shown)


waterplea:
This doesn't really fix the issue. It makes virtual scroll recreate every item every time, which is ok here, but I think we should make it work for regular reusing mode too. That would require us to rethink the approach completely. I plan to do it when I refactor combo box to the new textfield so let's not merge this, otherwise I might forget :)

Issue #5105, opened 2023-08-14
🚀 - TuiTable - allow to resize columns in case of >100% width


meis0n:
### Description

Currently the component doesn't allow to resize columns in case of total sum of initial columns' widths > 100%.

- Component allows to resize columns up to 100% total width in case of empty cell content 
- Component adds horizontal scroll in case of existing wide cell content AND suppresses resizing

Example of suppressed resizing - https://stackblitz.com/edit/taiga-hwoqvn?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.module.ts

**Feature request**:

- Allow to exceed 100% width threshhold while adding horizontal scroll BUT while keeping the option to resize the columns
- Implement new input (e.g. [(width)]) to control the column's width programmatically 
- Implement new table event (e.g. (onResize)) to get the column's width after a user successfully resizes a column with [resizable]="true" attribute

PR #9961, opened 2024-12-10
feat(addon-table): allow resize more than 100% width of table

splincode:
Fixes https://github.com/taiga-family/taiga-ui/issues/5105

https://github.com/user-attachments/assets/61595f87-8b1f-4a32-a608-ff54cddbc091

### DIFF
--- a/projects/demo/src/modules/components/table/examples/8/index.html
+++ b/projects/demo/src/modules/components/table/examples/8/index.html
@@ -1,0 +1,31 @@
+<table tuiTable>
+    <thead>
+        <tr tuiThGroup>
+            <th
+                tuiTh
+                [resizable]="true"
+            >
+                Name
+            </th>
+            <th
+                tuiTh
+                [resizable]="true"
+            >
+                Age
+            </th>
+            <th tuiTh>Description</th>
+        </tr>
+    </thead>
+    <tbody tuiTbody>
+        <tr *ngFor="let item of data">
+            <td tuiTd>{{ item.name }}</td>
+            <td tuiTd>{{ item.balance }}</td>
+            <td
+                tuiTd
+                [style.white-space]="'nowrap'"
+            >
+                {{ item.description }}
+            </td>
+        </tr>
+    </tbody>
+</table>
--- a/projects/demo/src/modules/components/table/index.ts
+++ b/projects/demo/src/modules/components/table/index.ts
@@ -19,5 +19,6 @@ export default class Page {
         'Virtual scroll',
         'Dynamic columns',
         'Footer',
+        'Resize a large table',
     ];
 }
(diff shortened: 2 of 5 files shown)


projects/demo/src/modules/components/table/index.ts
```diff
@@ -19,5 +19,6 @@ export default class Page {
         'Virtual scroll',
         'Dynamic columns',
         'Footer',
+        'Allow resize more than 100%',
```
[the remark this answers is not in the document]
splincode: fixed

splincode:
@waterplea updated

waterplea (approved):

projects/demo/src/modules/components/table/examples/8/index.html
```diff
@@ -0,0 +1,60 @@
...
+                *tuiCell="'b'"
+                tuiTd
+            >
+                b
+            </td>
+            <td
+                *tuiCell="'c'"
+                tuiTd
+            >
+                content content
```
vladimirpotekhin, line 50: nit: It would be nice to make a meaningful example. I mean something like "name", "age", "Lorem ipsum dolor sit" instead of  "a", "b", "c", "contentcontentcontent"

projects/demo/src/modules/components/table/examples/1/index.ts
```diff
@@ -18,10 +18,12 @@ export default class Example {
         {
             name: 'Alex Inkin',
             balance: 1323525,
+            country: 'UAE',
         },
         {
             name: 'Roman Sedov',
             balance: 423242,
+            country: 'Russia',
```
vladimirpotekhin, line 26: something went wrong 

https://github.com/user-attachments/assets/546c6c8e-9d96-47da-9706-e3c6d309a3a9

splincode: https://github.com/taiga-family/taiga-ui/issues/7379

Yes, I removed this problem from the description, it remains unsolved

projects/demo/src/modules/components/table/examples/8/index.html
splincode: `contentcontentcontent` is required for overflow line

projects/demo/src/modules/components/table/examples/1/index.ts
waterplea: Yes, let's do this in another PR. I'm thinking when we first start to drag a column, we should fix current width of all columns. Then we can adjust the column next to the one we drag to the same amount but inversed (when we extend column 2 by 10px we should shrink column 3 by same 10px, when we shrink column 2 — we should extend column 3). That should work, but then we should probably just make a directive for entire table, not for specific columns for this to work.

waterplea (approved):

vladimirpotekhin (approved):

PR #9962, opened 2024-12-10
fix(core): emit tuiDropdownOpenChange on distinct values

### LINKED ISSUE #9609: 🐞 - Multi-Select - tuiDropdownOpenChange emitting two events
### Reproduction url

https://stackblitz.com/edit/angular-u8tdmj?file=src%2Fapp%2Fapp.component.html

### Description

When opening and closing the multi-select dropdown two events are being emitted from the tuiDropdownOpenChange output. Check stackblitz for example.

### Taiga UI version

3.96

### Browser Used

Chrome

### OS Used

macOS

shiv9604:
Fixes #9609

Description : 
1. We used `Subject` because we are concerned about the latest value compared to maintaining the last state.
2. Created the `emitOpenChange` method to centralize & handle responsibility for emitting values.
3. Replaced previous value emission logic with the currently integrated flow.

### DIFF
--- a/projects/core/directives/dropdown/dropdown-open-legacy.directive.ts
+++ b/projects/core/directives/dropdown/dropdown-open-legacy.directive.ts
@@ -1,5 +1,5 @@
 import {Directive, Input, Output} from '@angular/core';
-import {BehaviorSubject} from 'rxjs';
+import {distinctUntilChanged, Subject} from 'rxjs';
 
 /**
  * @deprecated TODO: remove in v.5 when legacy controls are dropped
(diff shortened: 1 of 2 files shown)


splincode (approved):

projects/core/directives/dropdown/dropdown-open-legacy.directive.ts
```diff
@@ -10,11 +10,19 @@ import {BehaviorSubject} from 'rxjs';
         '[tuiDropdownOpen]:not([tuiDropdown]),[tuiDropdownOpenChange]:not([tuiDropdown])',
 })
 export class TuiDropdownOpenLegacy {
+    private readonly openStateSub = new Subject<boolean>();
+
     @Output()
-    public readonly tuiDropdownOpenChange = new BehaviorSubject(false);
+    public readonly tuiDropdownOpenChange = this.openStateSub
+        .asObservable()
```
waterplea, line 17: Unnecessary, pipe would already turn it into Observable

waterplea (approved):

projects/core/directives/dropdown/dropdown-open-legacy.directive.ts
shiv9604: Thanks for pointing that out.
Resolved!

PR #9987, opened 2024-12-13
Create & Integrate `tuiContainer` layout component 

### LINKED ISSUE #9721: 🐞 - Demo should not use any of "tui-" classes other than spacings
### Reproduction url

https://taiga-ui.dev/form

### Description

Some of the core layouting building blocks used in [Forms](https://taiga-ui.dev/form) is missing documentation. These elements (like row, skeleton, table) seem to be useful for a lot of other contexts, but some more documentation how they are intended to use them would be helpful.

### Taiga UI version

4.14.0

### Browser Used

Firefox

### OS Used

Windows

shiv9604:
Fixes #9721 

Changes Description : 
1. Created the `tuiContainer` layout to replace depreciating `tui-container` [styles](https://github.com/taiga-family/taiga-ui/blob/v3.x/projects/styles/taiga-ui-global.less).
2. Migrated depreciating `tui-container` styles with layout component.

### DIFF
--- a/projects/layout/components/container/container.directive.ts
+++ b/projects/layout/components/container/container.directive.ts
@@ -7,7 +7,7 @@ import {
 } from '@angular/core';
 import {tuiWithStyles} from '@taiga-ui/cdk/utils/miscellaneous';
 
-type ContainerType = 'adaptive' | 'fixed' | 'fullWidth' | 'menu';
+type ContainerType = '' | 'adaptive' | 'fixed' | 'fullwidth' | 'menu';
 
 @Component({
     standalone: true,
--- a/projects/layout/components/container/container.directive.ts
+++ b/projects/layout/components/container/container.directive.ts
@@ -7,8 +7,6 @@ import {
 } from '@angular/core';
 import {tuiWithStyles} from '@taiga-ui/cdk/utils/miscellaneous';
 
-type ContainerType = '' | 'adaptive' | 'fixed' | 'fullwidth' | 'menu';
-
 @Component({
     standalone: true,
     template: '',
@@ -32,6 +30,6 @@ class TuiContainerStyles {}
 export class TuiContainer {
     protected readonly nothing = tuiWithStyles(TuiContainerStyles);
 
-    @Input('tuiContainer')
-    public type!: ContainerType;
+    @Input({alias: 'tuiContainer', required: true})
+    public type!: '' | 'adaptive' | 'fixed' | 'fullwidth' | 'menu';
 }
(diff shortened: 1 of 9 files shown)


projects/layout/components/container/container.directive.ts
```diff
@@ -0,0 +1,37 @@
...
+    host: {
+        tuiContainer: '',
+        '[attr.data-type]': 'type',
+    },
+})
+export class TuiContainer {
+    protected readonly nothing = tuiWithStyles(TuiContainerStyles);
+
+    @Input('tuiContainer')
+    public type!: ContainerType;
```
splincode, line 36: `@Input({ alias: tuiContainer, required: true })`

shiv9604: I agree with alias & required parameter, Inline typing for multiple string values would be okay?

shiv9604: Resolved.

waterplea:
Hi @shiv9604. Sorry if I wasn't clear on that — please do not introduce new components as replacements. We would have our designers spec out all necessary layout components before we implement them, we do not have a goal to 1 to 1 replace all global styles with layout components. I'm not sure we will have container component at all in the future. For now if you want to help with this issue, you can get rid of `tui-table` (using `tuiTable` component instead) and try to drop all `tui-form` classes from this example:
https://taiga-ui.dev/form
and migrate it to this component as second example without using any `tui-form` classes:
https://taiga-ui.dev/components/form
Any styles you would need to recreate to achieve that — keep them in the example styles.

P.S. also form component should be moved to layout section of the demo (change path in `DemoRoute` constant and section changed in `const pages: DocRoutePages`).

shiv9604:
I can understand @waterplea, Can you just confirm the following things for the same ticket.

1. tui-form - I understand requirements about this-form but I am able to see usage of those in multiselect example 9 & pipes field error example, should I take care of them as well?

2. tui-table - I wil go ahead for creation & integration of tuiTable as it’s not existing there.

3. tui-List - Actually I have implemented tui-list layout component & I was integrating in demo wherever it’s used, Is there any possibility I can takeover on it and complete it thoroughly as just integration of it is remaining from my side for few demo examples.

waterplea:
1. Can probably be just replaced with styles directly in those examples or even completely dropped, I don't think they are really necessary there 
2. Thank you
3. I'm worried it will be different in the specs than it is now, not sure they are ready yet.

shiv9604:
1. Clear!
2. Most Welcome!
3. I will keep at local as of now & I will focus on `tui-form` & `tui-table`, eventually whenever you get clarity for the `tui-list` layout component specs, I would appreciate it if that ticket can be assigned to me so I can continue or extend my existing implementation for it if possible.

PR #10026, opened 2024-12-18
fix(legacy): wrong prefix placement in table context

### LINKED ISSUE #10022: 🐞 - Incorrect prefix placement when using tui-input inside tuiTable
### Reproduction url

https://stackblitz.com/edit/angular-vtimgmun

### Description

Appears only when label is provided for tui-input.

Image

### Taiga UI version

4.18.0

### Browser Used

Chrome

### OS Used

Windows

splincode:
Fixes #10022

image

### DIFF
--- a/projects/legacy/components/primitive-textfield/primitive-textfield.template.html
+++ b/projects/legacy/components/primitive-textfield/primitive-textfield.template.html
@@ -50,7 +50,7 @@
         </div>
         <div class="t-wrapper">
             <label
-                *ngIf="hasPlaceholder"
+                *ngIf="hasPlaceholder && !isContextTable"
                 automation-id="tui-primitive-textfield__placeholder"
                 class="t-placeholder"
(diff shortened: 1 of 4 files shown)


projects/legacy/components/primitive-textfield/primitive-textfield.template.html
```diff
@@ -50,7 +50,7 @@
         </div>
         <div class="t-wrapper">
             <label
-                *ngIf="hasPlaceholder"
+                *ngIf="hasPlaceholder && !isContextTable"
```
waterplea, line 53: 1) === 'table' can be included in `hasPlaceholder` getter
2) Don't we need placeholder in table when the value is empty? Check out this example:
https://taiga-ui.dev/components/input#table

splincode: fixed

PR #10056, opened 2024-12-25
feat(core): add `injector` to tuiDialog options

MillerSvt:


splincode (approved):

projects/core/components/dialog/dialog.factory.ts
```diff
@@ -41,17 +42,23 @@ type ExtractDialogResult<T, K extends keyof T = ContextKeys<T>> = [K] extends [n
...
     R extends ExtractDialogResult<T, K>,
 >(
     component: AssertNotMultipleContexts<T, K>,
-    options?: Partial<Omit<TuiDialogOptions<D>, 'data'>>,
+    {injector, ...options}: Partial<Options<D>> = {},
 ): (data: D) => Observable<R> {
-    const dialogService = inject(TuiDialogService);
-    const injector = inject(INJECTOR);
+    if (!injector) {
+        assertInInjectionContext(tuiDialog);
```
waterplea, line 57: Do we need it? Isn't it already inside `inject` method?

splincode:
```ts
ngDevMode && assertInInjectionContext(tuiDialog);
```

MillerSvt: No, no one does that

splincode: I figure out, but does we really need `readable error` in production instead of development mode?

MillerSvt: This is not a warning, this is an assertion, so the prod should have this error.
But readability on prod will be worse, due to minification of the code.
But I still don't see the point in complicating the condition here.

waterplea (approved):

PR #10077, opened 2024-12-27
chore(schematics): valid migration for badged content

splincode:
Part of #10017

### DIFF
--- a/projects/cdk/schematics/ng-update/v4/steps/templates/migrate-badged-content.ts
+++ b/projects/cdk/schematics/ng-update/v4/steps/templates/migrate-badged-content.ts
@@ -120,7 +120,7 @@ function migrateColor({
 
     const value = attr.value;
     const insertTo = sourceCodeLocation.startTag?.endOffset ?? 0;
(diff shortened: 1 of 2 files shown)


projects/cdk/schematics/ng-update/v4/steps/templates/migrate-badged-content.ts
```diff
@@ -120,13 +120,14 @@ function migrateColor({
 
     const value = attr.value;
     const insertTo = sourceCodeLocation.startTag?.endOffset ?? 0;
+    const hasBinding = attr ? isBinding(attr) : false;
```
nsbarsukov, line 123:
```suggestion
    const hasBinding = Boolean(attr) && isBinding(attr);
```

nsbarsukov (approved):

projects/cdk/schematics/ng-update/v4/steps/templates/migrate-badged-content.ts
splincode: fixed

vladimirpotekhin (approved):

PR #10083, opened 2024-12-27
#9057 - Migrate `tui-alert` API docs to new API 

shiv9604:
Fixes #9057 

Changes Description : 
- Migrate `tui-alert` API docs to new to new API


splincode (approved):

projects/demo/src/modules/components/alert/index.ts
```diff
@@ -13,7 +14,7 @@ import {AlertExampleWithData} from './examples/4';
 
 @Component({
     standalone: true,
-    imports: [TuiButton, TuiDemo],
+    imports: [TuiButton, TuiDemo, TuiDocAPI, TuiDocAPIItem],
```
waterplea, line 17: Unnecessary imports

waterplea (approved):

projects/demo/src/modules/components/alert/index.ts
shiv9604: Resolved!

Issue #9752, opened 2024-11-13
🐞 - `Textfield` `open` input doesn't work as expected


hakimio:
### Reproduction url

https://stackblitz.com/edit/angular-3dfsuh?file=src%2Fapp%2Fapp.template.html

### Description

Textfield dropdown opens even when textfield's `open` input is set to `false`.
[Might also be an issue with `TuiDropdownOpen` directive.]
Check the stackblitz link for reproduction.

Issue is also present in the Taiga demo app:
https://github.com/taiga-family/taiga-ui/blob/d2c4a998e6ab640e386d2150994e2ceaee4a98c0/projects/addon-doc/components/navigation/navigation.template.html#L6

Maybe as a workaround `tuiDropdownEnabled` could be exposed from the `Textfield` component?

### Taiga UI version

4.15.0

### Browser Used

Chrome

### OS Used

Windows

waterplea:
Yes, that would be not a workaround but a proper fix — open is a two way binding and can be internally updated, `tuiDropdownEnabled` is there to control built-in opening mechanic like clicks on buttons or arrows on inputs. We need to expose it with `TuiWithDropdownOpen` and add to the docs. We do not need to add aliasing to the name in this case because it is not used in the directive selector.

PR #10101, opened 2024-12-29
feat(core): added `tuiDropdownEnabled` to exposed inputs of `TuiWithDropdownOpen`

lightest:
fixes: #9752


waterplea (approved):

splincode (approved):

hakimio:
@lightest Navigation search needs to be fixed since now the `canOpen` restriction doesn't work as it should:
https://github.com/taiga-family/taiga-ui/blob/d2c4a998e6ab640e386d2150994e2ceaee4a98c0/projects/addon-doc/components/navigation/navigation.template.html#L6
and it might be a good idea to add tests for the search in navigation component.

waterplea:
I'll take a look after new year, this is not really the same issue as I believe the intent there was not to just disable opening the dropdown but also closing it when the condition is not met.

PR #10110, opened 2025-01-02
feat(core): disable dark theme for print mode

MillerSvt:


projects/core/styles/theme/palette.less
```diff
@@ -138,3 +131,28 @@
...
+        color-scheme: light;
+    }
+
+    &:root,
+    [tuiTheme='light'] {
+        .light();
+    }
+}
+
+@media screen {
```
waterplea, line 152: print?

projects/core/styles/theme/palette.less
```diff
@@ -138,3 +131,28 @@
...
+    [tuiTheme] {
```
waterplea, line 153: :root as well?

waterplea (approved):

splincode (approved):

PR #10116, opened 2025-01-04
#9057 - Migrate `tui-badged-content` doc API to new API

shiv9604:
Fixes #9057 

Changes Description : 
- Migrate badged content doc api to new api.

### DIFF
--- a/projects/demo/src/modules/components/badged-content/index.html
+++ b/projects/demo/src/modules/components/badged-content/index.html
@@ -35,7 +35,7 @@
                 />
             </tui-badged-content>
         </tui-doc-demo>
-        <span tuiTitle="l">CSS customization</span>
+        <h1 [style.font]="'--tui-font-heading-4;'">CSS customization</h1>
         <table tuiDocAPI>
             <tr
                 name="style.--t-radius"
@@ -43,7 +43,9 @@
                 type="string"
                 [items]="radiusVariants"
                 [(value)]="radius"
-            ></tr>
+            >
+                Border radius
+            </tr>
         </table>
     </ng-template>
 
--- a/projects/demo/src/modules/components/badged-content/index.html
+++ b/projects/demo/src/modules/components/badged-content/index.html
@@ -35,7 +35,7 @@
                 />
             </tui-badged-content>
         </tui-doc-demo>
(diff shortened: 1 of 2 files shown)


projects/demo/src/modules/components/badged-content/index.html
```diff
@@ -35,17 +35,18 @@
...
             </tui-badged-content>
         </tui-doc-demo>
-        <tui-doc-documentation heading="CSS customization">
-            <ng-template
-                documentationPropertyMode="input"
-                documentationPropertyName="style.--t-radius"
-                documentationPropertyType="string"
-                [documentationPropertyValues]="radiusVariants"
-                [(documentationPropertyValue)]="radius"
+        <h1 [style.font]="'--tui-font-heading-4;'">CSS customization</h1>
```
waterplea, line 38: Semantically `h1` can be only one per page, we shouldn't use it, let's change it to `h3`. Also let's use `tuiTitle="m"` instead of customizing font style, or `tuiHeader` – see what looks better/closer to what we had in old component.

projects/demo/src/modules/components/badged-content/index.ts
```diff
@@ -13,6 +14,8 @@ import {TuiAvatar, TuiBadgedContent, TuiBadgeNotification} from '@taiga-ui/kit';
         TuiBadgedContent,
         TuiBadgeNotification,
         TuiDemo,
+        TuiDocAPI,
```
waterplea, line 17: Unnecessary imports

waterplea (approved):

splincode (approved):

projects/demo/src/modules/components/badged-content/index.html
shiv9604: Yeah, I Agree about the header. I thought on the component level so implemented h1.

I have tried with `tuiTitle="m"` but it wasn't matching the previous styling, but it seems to be near to it with `tuiHeader="xxl"` which I implemented now.

Resolved!

projects/demo/src/modules/components/badged-content/index.ts
shiv9604: Resolved!

PR #10117, opened 2025-01-04
#9057 - Migrate `tui-bar-chart` docs API to new API

shiv9604:
Fixes #9057 

Changes Description : 
- Migrate bar chart docs api to new api.

### DIFF
--- a/projects/demo/src/modules/components/bar/index.ts
+++ b/projects/demo/src/modules/components/bar/index.ts
@@ -2,11 +2,12 @@ import {Component} from '@angular/core';
 import {changeDetection} from '@demo/emulate/change-detection';
 import {TuiDemo} from '@demo/utils';
 import {TuiBar} from '@taiga-ui/addon-charts';
+import {TuiDocAPI, TuiDocAPIItem} from '@taiga-ui/addon-doc';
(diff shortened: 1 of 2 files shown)


projects/demo/src/modules/components/bar/index.ts
```diff
@@ -2,11 +2,12 @@ import {Component} from '@angular/core';
 import {changeDetection} from '@demo/emulate/change-detection';
 import {TuiDemo} from '@demo/utils';
 import {TuiBar} from '@taiga-ui/addon-charts';
+import {TuiDocAPI, TuiDocAPIItem} from '@taiga-ui/addon-doc';
 import type {TuiSizeL, TuiSizeS} from '@taiga-ui/core';
 
 @Component({
     standalone: true,
-    imports: [TuiBar, TuiDemo],
+    imports: [TuiBar, TuiDemo, TuiDocAPI, TuiDocAPIItem],
```
waterplea, line 10: You do not need to import them, they are already included in `TuiDemo`. Same comment applies to all other PRs

waterplea (approved):

projects/demo/src/modules/components/bar/index.ts
shiv9604: Thanks! It was insightful. 

Resolved.

PR #10129, opened 2025-01-09
fix(addon-commerce): redundant sign symbol when value is zero

splincode:


splincode:
it turns out to be a contribution to the community

Issue #10127, opened 2025-01-09
🐞 - tuiAmount should not show a negative sign when the value is zero


Zheleznyak-F:
### Reproduction url

https://stackblitz.com/edit/65qvljbj?file=src%2Fapp%2Fapp.template.html

### Description

Wrong view of small negative number with decimal part e.g. "-0.83" displays like "-0". Negative sign must not be in rounded zero number

Also actual for 3.x version with tui-money component

### Taiga UI version

3.x+

### Browser Used

Chrome

### OS Used

macOS

Zheleznyak-F:
![Image](https://github.com/user-attachments/assets/80ff5c68-6580-4afc-b6f8-b795122edb43)

rashiddok:
I'll take it

PR #10147, opened 2025-01-13
fix(addon-commerce): tuiAmount hiding negative sign when value is zero

rashiddok:
Fixes [#10127](https://github.com/taiga-family/taiga-ui/issues/10127)


splincode (approved):

splincode:
please add an example
image

vladimirpotekhin (approved):

PR #10212, opened 2025-01-22
feat(layout): `Header` add body-l and body-m sizes

vladimirpotekhin:
Fixes #

### DIFF
--- a/projects/demo/src/modules/components/header/examples/1/index.html
+++ b/projects/demo/src/modules/components/header/examples/1/index.html
@@ -1,4 +1,4 @@
-<div tuiHeader="xxl">
+<div tuiHeader="h1">
     <h1 tuiTitle>
         Title
         <div tuiSubtitle>Subtitle</div>
@@ -16,7 +16,7 @@
     </div>
 </div>
 
-<div tuiHeader="xl">
+<div tuiHeader="h2">
     <h2 tuiTitle>
         Title
         <div tuiSubtitle>Subtitle</div>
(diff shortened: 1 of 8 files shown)


splincode (approved):

projects/demo/src/modules/components/header/examples/1/index.html
```diff
@@ -1,4 +1,4 @@
-<div tuiHeader="xxl">
+<div tuiHeader="h1">
```
waterplea, line 1: Maybe ngFor?

projects/demo/src/modules/components/header/examples/2/index.less
```diff
@@ -9,3 +9,7 @@
     margin-top: 0.5rem;
     max-inline-size: 30rem;
 }
+
+[tuiLink] {
+    font: var(--tui-font-text-s);
```
waterplea, line 14: Should we fix the default font inside header built-in styles?

projects/layout/components/header/header.directive.ts
```diff
@@ -46,6 +59,19 @@ export class TuiHeader {
     protected readonly options = inject(TUI_HEADER_OPTIONS);
     protected readonly nothing = tuiWithStyles(TuiHeaderStyles);
 
+    // @deprecated: remove alpha sizes in 5.0
```
waterplea, line 62: This wouldn't mark the whole input as deprecated for everybody?

waterplea (approved):

projects/demo/src/modules/components/header/examples/1/index.html
vladimirpotekhin: I'll do it in a separate PR

PR #10226, opened 2025-01-23
chore(demo): `box-sizing`

splincode:


waterplea:
I'm not sure we want that, either way we will miss a bug for different behavior, better to not miss the default, I think, even if this reset is common

Issue #9836, opened 2024-11-25
🐞 - Dropdown in tabs closes immediately in chrome, but not in safari


splincode:
### Reproduction url

https://taiga-ui.dev/navigation/tabs

## Safari

https://github.com/user-attachments/assets/f80c0033-6a05-425b-9683-1d0112ff968d

## Chrome

https://github.com/user-attachments/assets/4eac4488-1330-44ed-b992-bbf9588a970a

### Taiga UI version

latest

### Browser Used

Safari

### OS Used

macOS

splincode:
@waterplea It looks like `focusin` and `focusout` events are not [triggered](https://github.com/taiga-family/taiga-ui/blob/main/projects/cdk/tokens/active-element.ts#L66) in Safari. Which in turn does not send a signal [here](https://github.com/taiga-family/taiga-ui/blob/main/projects/core/directives/dropdown/dropdown-open.directive.ts#L80).

waterplea:
I'm not sure that's the reason. I know that only Chrome triggers `focusout` when a focused element is removed from the DOM, but we specifically made `ActiveZone` so that it does not count such events as leaving zone 🤔

PR #10232, opened 2025-01-24
[WIP] fix(cdk): focusout event not firing in safari

splincode:
Fixes https://github.com/taiga-family/taiga-ui/issues/9836


waterplea:
Let's keep this open for now until I have time to take a look myself. That's a delicate piece of code and while this change fixes the issue I want to first investigate it myself before adding this workaround.

PR #10242, opened 2025-01-27
feat(demo): `Button` add example with two labels

mdlufy:
Fixes #


projects/demo/src/modules/components/button/examples/7/index.html
```diff
@@ -0,0 +1,16 @@
+<button
+    tuiButton
```
waterplea, line 2: Let's add type="button" just in case

waterplea (approved):

splincode (approved):

Issue #3915, opened 2023-03-19
🐞 - tuiSortByChange and directionChange methods should not been called together


xvs32x:
### Which @taiga-ui/* package(s) are the source of the bug?

addon-table

### Please provide a link to a minimal reproduction of the bug

https://stackblitz.com/edit/angular-ddzbqa

### Is this issue blocking you?

Non-Blocking

### Description

Hi there, at first, thank you for amazing UI Library. It's the greatest!

In my application router query params and pagination are synced, so query string is a source of truth for the pagination. 
It looks like when a user change something on the page (sort, direction, etc...), at first, I use router to change the query string, and then, state from query string changes the pagination on a page. 

When I try to change the sort column, both of methods (tuiSortByChange and directionChange) try to change query params, but Angular can't handle it. It just skips the second changing, because the previous one hasn't finished yet. Thats why we cant change the sort colum.

 I'd like to change the way how it works. For example, use one event instead of two. For instance, it works the same in Angular Material (https://material.angular.io/components/sort/api#MatSort) We have only one event (sortChange) there for both - sort and direction. And it works well with query params.  

Thank you!

### Angular version

14.0.0

### Taiga UI version

3.20.0

### Which browsers have you used?

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

### Which operating systems have you used?

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

loxy:
Any news on this?

waterplea:
> Any news on this?

Not yet. Looks like we need to add a singular `{ sorter, direction }` and `{ sortBy, orderBy }` input-output.

loxy:
Maybe we can work on this. But decision if we use your table implementation is not taken yet.

I saw that the direction comes directly from the table directive whereas the sorting is applied by `tuiSortBy`. Does that need to change then?

waterplea:
No, table has `sorter` and `direction` and there are also "semantic" directives `tuiSortBy` to use key, rather than sorter and `tuiDirectionOrder` to do desc/asc, rather than -1/1

PR #10268, opened 2025-01-28
feat(addon-table): add new output event with sort and direction changes

splincode:
Fixes https://github.com/taiga-family/taiga-ui/issues/3915

### DIFF
--- a/projects/demo/src/modules/components/table/examples/4/index.html
+++ b/projects/demo/src/modules/components/table/examples/4/index.html
@@ -58,8 +58,7 @@
         [columns]="columns"
         [direction]="(direction$ | async) || 1"
         [tuiSortBy]="sorter$ | async"
-        (directionChange)="direction$.next($event)"
-        (tuiSortByChange)="sorter$.next($event!)"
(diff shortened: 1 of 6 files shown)


projects/demo/src/modules/components/table/examples/4/index.html
```diff
@@ -55,8 +55,7 @@
         [columns]="columns"
         [direction]="(direction$ | async) || 1"
         [tuiSortBy]="sorter$ | async"
-        (directionChange)="direction$.next($event)"
-        (tuiSortByChange)="sorter$.next($event!)"
+        (tuiSortAndOrderChange)="change($event)"
```
waterplea, line 58: Let's rename it to `tuiSortChange` for brevity?

waterplea (commented):
We also need similar combined event for `Table` itself with `sorter` and `direction`

projects/demo/src/modules/components/table/examples/4/index.html
splincode: fixed

waterplea (approved):

PR #10270, opened 2025-01-28
feat(legacy): input tag converts the last element to tag after paste event

### LINKED ISSUE #10264: 🚀 - InputTag should convert the last element to tag after paste event
### Description

Strings with separators like "one,two,three" and "one,two,three," should be treated in the same manner. InputTag should convert them into 3 tags: one, two, and three.

At this moment, a string without a separator after the last element doesn't get converted into a tag until the user unfocuses InputTag

Current behavior:
Image

Desired behavior:
Image

rashiddok:
Fixes #10264


rashiddok:
Should I make PR for v3?

splincode (approved):

waterplea (approved):

PR #10288, opened 2025-01-30
feat(kit): `InputSlider` refactor to new `Textfield`

nsbarsukov:
```html
<tui-textfield>
  <input
    tuiInputSlider
    prefix="$"
    postfix="%"
    [min]="0"
    [max]="100"
    [tuiNumberFormat]="{precision: 3}"
    [(ngModel)]="value"
  />

  <input
    tuiSlider
    type="range"
    [keySteps]="keySteps"
    [segments]="segments"
    [step]="step"
  />
</tui-textfield>
```


projects/addon-doc/components/api/api-item.component.ts
```diff
@@ -71,7 +71,7 @@ export class TuiDocAPIItem<T> implements OnInit {
     @Output()
     public readonly valueChange = new EventEmitter<T>();
 
-    public ngOnInit(): void {
+    public ngAfterViewInit(): void {
```
nsbarsukov, line 74: TODO(n.barsukov): move to separate PR

projects/kit/components/input-number/input-number.component.ts
```diff
@@ -149,6 +146,10 @@ export class TuiInputNumber extends TuiControl<number | null> {
         ),
     );
 
+    public readonly min = signal(this.options.min);
+    public readonly max = signal(this.options.max);
+    public readonly step = signal(this.options.step);
```
nsbarsukov, line 151: Make `step` protected again

projects/kit/components/input-slider/input-slider.style.less
```diff
@@ -0,0 +1,34 @@
...
+    display: none !important;
+}
+
+tui-textfield._hide-buttons {
+    --t-input-number-offset-end: 0px;
+}
+
+tui-textfield._hide-buttons [tuiInputNumber]._with-buttons {
+    border-radius: inherit;
+}
```
nsbarsukov, line 12: Oops, It is codebase from my previous iteration of new `InputSlider`.

Not relevent for the latests version of `InputSlider` 😄

nsbarsukov: https://github.com/taiga-family/taiga-ui/pull/10288/commits/2bfb1a514c515a312dd56b489bf74e8921714b76

projects/core/styles/components/textfield.less
```diff
@@ -182,12 +182,12 @@ tui-textfield {
...
     }
 
-    &._with-template input,
+    &._with-template input:first-of-type,
     &._with-template select {
         color: transparent !important;
     }
 
-    input:defined,
+    input:defined:first-of-type,
```
waterplea, line 190: You can drop `:defined` it's just for the specificity

projects/kit/components/slider/slider.component.ts
```diff
@@ -64,10 +64,26 @@ export class TuiSliderComponent {
...
 
+    public set max(x: number) {
+        this.el.max = String(x);
+    }
+
+    public get step(): number {
+        return Number(this.el.step) || 1;
+    }
+
+    public set step(x: number) {
```
waterplea, line 83: Maybe it makes sense to add those as inputs + host binding instead of getter/setter couples relying on native props?

projects/kit/components/input-slider/input-slider.component.ts
```diff
@@ -0,0 +1,145 @@
...
+        slider && this.keyStepsTransformer() // For non-linear slider step means percentage
+            ? ((this.max() - this.min()) / 100) * slider.step()
+            : (slider?.step() ?? 1),
+    );
+
+    protected disabled = computed(
+        () => (this.inputNumber()?.disabled() || this.inputNumber()?.readOnly()) ?? false,
+    );
+
+    protected safeInputNumberValue = toSignal(
```
waterplea, line 71: That's a long ass name, can we just call it `value` as it seems like this is the actual value of the control that we operate with in this context?

projects/kit/components/input-slider/input-slider.component.ts
```diff
@@ -0,0 +1,145 @@
...
+            this.inputNumber()?.setValue(
+                this.keyStepsTransformer()?.toControlValue(value) ?? value,
+            );
+        }
+    }
+
+    protected onStep(coefficient: number): void {
+        const slider = this.slider();
+
+        if (this.textfield.focused() && slider) {
```
waterplea, line 127: When can it not be focused? Why do we use slider value? Can we just use value?

projects/kit/components/input-slider/input-slider.component.ts
```diff
@@ -0,0 +1,145 @@
...
+    private keyStepsTransformer: WritableSignal<TuiValueTransformer<
+        number,
+        number
+    > | null> = signal(null);
+
+    private readonly inputNumber = signal<TuiInputNumber | null>(null);
+    private readonly slider = signal<TuiSliderComponent | null>(null);
+    protected readonly textfield = inject(TuiTextfieldComponent);
+
+    protected min = computed(() => this.inputNumber()?.min() ?? 0);
```
waterplea, line 59: I wish we could get rid of this trick somehow, since because we use computed here we have a mess with types inside slider

projects/kit/components/input-slider/input-slider.component.ts
```diff
@@ -0,0 +1,145 @@
...
+        '(change)': 'onSliderInput($event)',
+        '(click)': 'textfield.input.nativeElement.focus()',
+        '(focusout)': 'onBlur()',
+        '(keydown.arrowUp)': 'onStep(1)',
+        '(keydown.arrowDown)': 'onStep(-1)',
+    },
+})
+export class TuiInputSliderDirective {
+    private readonly injector = inject(INJECTOR);
+    private keyStepsTransformer: WritableSignal<TuiValueTransformer<
```
waterplea, line 50: We can use type inside `signal<...>(null)` to shorten this

projects/kit/components/input-slider/input-slider.component.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    protected min = computed(() => this.inputNumber()?.min() ?? 0);
+    protected max = computed(() => this.inputNumber()?.max() ?? 100);
+    protected step = computed((slider = this.slider()) =>
+        slider && this.keyStepsTransformer() // For non-linear slider step means percentage
+            ? ((this.max() - this.min()) / 100) * slider.step()
+            : (slider?.step() ?? 1),
+    );
+
+    protected disabled = computed(
```
waterplea, line 67: Can we make it `interactive` and reversed, since it's about both disabled and readOnly?

projects/kit/components/input-slider/input-slider.component.ts
```diff
@@ -0,0 +1,145 @@
...
+            filter(tuiIsPresent),
+            switchMap(({value}) =>
+                toObservable(value, {injector: this.injector}).pipe(startWith(value())),
+            ),
+            pairwise(),
+            map(([prev, cur]) => cur ?? prev!),
+        ),
+    );
+
+    protected readonly textfieldToSliderSync = effect(() => {
```
waterplea, line 82: nit: Let's add `(slider = this.slider())` so that early return is the first line in the actual function? That seems more readable this way, what do you think?

projects/kit/components/input-slider/input-slider.component.ts
```diff
@@ -0,0 +1,145 @@
...
+    protected set keyStepsSetter({transformer}: TuiSliderKeyStepsBase) {
+        this.keyStepsTransformer = transformer;
+    }
+
+    protected onSliderInput({target}: Event): void {
+        if (tuiIsElement(target) && tuiIsInput(target) && target.type === 'range') {
+            const value = Number(target.value);
+
+            this.inputNumber()?.setValue(
+                this.keyStepsTransformer()?.toControlValue(value) ?? value,
```
waterplea, line 119: nit: How about we write such things like:
```ts
const transformed = this.keyStepsTransformer()?.toControlValue(value) ?? value;

this.inputNumber()?.setValue(transformed);
```
That's 1 line shorter and more readable.

projects/kit/components/input-slider/input-slider.component.ts
```diff
@@ -0,0 +1,145 @@
...
+        slider.el.setAttribute('tabindex', '-1');
+    }
+
+    @ContentChild(TuiSliderKeyStepsBase)
+    protected set keyStepsSetter({transformer}: TuiSliderKeyStepsBase) {
+        this.keyStepsTransformer = transformer;
+    }
+
+    protected onSliderInput({target}: Event): void {
+        if (tuiIsElement(target) && tuiIsInput(target) && target.type === 'range') {
```
waterplea, line 115: Early return since it's more than one line under if?

projects/kit/components/input-slider/input-slider.component.ts
```diff
@@ -0,0 +1,145 @@
...
+            const value = Number(target.value);
```
waterplea, line 116: `HTMLInputElement` has `.valueAsNumber`, would it work here instead of parsing?

projects/kit/components/input-slider/input-slider.component.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    protected disabled = computed(
+        () => (this.inputNumber()?.disabled() || this.inputNumber()?.readOnly()) ?? false,
+    );
+
+    protected safeInputNumberValue = toSignal(
+        toObservable(this.inputNumber).pipe(
+            filter(tuiIsPresent),
+            switchMap(({value}) =>
+                toObservable(value, {injector: this.injector}).pipe(startWith(value())),
```
waterplea, line 75: Do we really need such a crazy chain? ) Looks like this could be simplified somehow.

projects/kit/components/slider/helpers/slider-key-steps.directive.ts
```diff
@@ -1,24 +1,59 @@
...
     standalone: true,
     selector: 'input[tuiSlider][keySteps]',
+})
+export class TuiSliderKeyStepsBase {
+    private readonly slider = inject<TuiSliderComponent>(
+        forwardRef(() => TuiSliderComponent),
+    );
+
+    // eslint-disable-next-line no-restricted-syntax
+    @HostBinding('attr.aria-valuemin')
```
waterplea, line 28: `host: {}`?

projects/kit/components/slider/helpers/slider-key-steps.directive.ts
```diff
@@ -1,24 +1,59 @@
...
+    // eslint-disable-next-line no-restricted-syntax
+    @HostBinding('attr.aria-valuemax')
+    protected max: number | null = this.slider.max();
+
+    public transformer = signal<TuiValueTransformer<number, number> | null>(null);
+
+    @Input()
+    public set keySteps(steps: TuiKeySteps | null) {
+        this.transformer.set(steps && tuiCreateKeyStepsTransformer(steps, this.slider));
+        this.min = steps?.[0][1] ?? null;
```
waterplea, line 40: Unnecessary fallback if we allow it to be `undefined`? We only use it for aria, right? So we can probably do that

projects/kit/components/slider/helpers/slider-key-steps.directive.ts
```diff
@@ -1,24 +1,59 @@
...
+        this.max = steps?.[steps.length - 1]?.[1] ?? null;
+    }
+}
+
+@Directive({
+    standalone: true,
+    selector:
+        'input[tuiSlider][keySteps][ngModel],input[tuiSlider][keySteps][formControl],input[tuiSlider][keySteps][formControl]',
```
waterplea, line 48: formControlName in the last one

projects/kit/components/slider/helpers/slider-key-steps.directive.ts
```diff
@@ -1,24 +1,59 @@
...
     providers: [tuiFallbackValueProvider(0)],
     host: {
-        '[attr.aria-valuenow]': 'value()',
-        '[attr.aria-valuemin]': 'min',
-        '[attr.aria-valuemax]': 'max',
+        '[value]': 'this.value()',
+        '[attr.aria-valuenow]': 'transformer.toControlValue(this.value())',
```
waterplea, line 52: This guy is missing in non-control version of the directive. I wonder if it's a problem worth spending time on 🤔

projects/kit/components/slider/helpers/slider-key-steps.directive.ts
```diff
@@ -1,24 +1,59 @@
...
         '[disabled]': 'disabled()',
         '(blur)': 'onTouched()',
-        '(input)': 'updateControlValue()',
-        '(change)': 'updateControlValue()',
+        '(input)': 'onChange($event.target.value)',
+        '(change)': 'onChange($event.target.value)',
     },
 })
 export class TuiSliderKeySteps extends TuiControl<number> {
```
waterplea, line 59: We should rename this to `TuiSliderKeyStepsControlDirective` and add an array `TuiSliderKeySteps = [base, control]`, otherwise people will not be attached without an explicit import

projects/kit/components/slider/helpers/slider-key-steps.directive.ts
```diff
@@ -1,24 +1,59 @@
...
 import {TuiSliderComponent} from '../slider.component';
 import type {TuiKeySteps} from './key-steps';
-import {tuiKeyStepValueToPercentage, tuiPercentageToKeyStepValue} from './key-steps';
+import {tuiCreateKeyStepsTransformer} from './key-steps';
 
 @Directive({
     standalone: true,
     selector: 'input[tuiSlider][keySteps]',
+})
+export class TuiSliderKeyStepsBase {
```
waterplea, line 22: Let's rename it to `TuiSliderKeyStepsDirective` so people will know not to import it separately

projects/kit/components/slider/helpers/slider-key-steps.directive.ts
```diff
@@ -27,45 +62,7 @@ export class TuiSliderKeySteps extends TuiControl<number> {
...
-
-    private transformToNativeValue(controlValue: number): number {
-        const {min, max} = this.slider;
-        const newValuePercentage = tuiKeyStepValueToPercentage(
-            controlValue,
-            this.keySteps,
-        );
-
-        return (newValuePercentage * (max - min)) / 100 + min;
+    public set keySteps(steps: TuiKeySteps) {
```
waterplea, line 65: We can maybe inject the other directive and use its transformer instead of this setter?

projects/kit/components/slider/slider.component.ts
```diff
@@ -57,15 +74,9 @@ export class TuiSliderComponent {
...
-    }
-
-    public get min(): number {
-        return Number(this.el.min);
-    }
+        const range = this.max() - this.min();
 
-    public get max(): number {
-        return Number(this.el.max || 100);
+        return (range && (this.value - this.min()) / range) || 0;
```
waterplea, line 79: Why change that part? Seemed easier to read as a one-liner to me :)

projects/kit/components/slider/slider.component.ts
```diff
@@ -84,6 +95,31 @@ export class TuiSliderComponent {
         this.el.value = `${newValue}`;
     }
 
+    @Input({
+        alias: 'min',
+        transform: (x: number | string) => coerceNumberProperty(x), // TODO: remove `transform` in v5
```
waterplea, line 100: Can it be just `transform: coerceNumberProperty`?

projects/kit/components/slider/slider.component.ts
```diff
@@ -84,6 +95,31 @@ export class TuiSliderComponent {
...
+    protected set minSetter(x: number) {
+        isWritableSignal(this.min) && this.min.set(x);
+    }
+
+    @Input({
+        alias: 'max',
+        transform: (x: number | string) => coerceNumberProperty(x), // TODO: remove `transform` in v5
+    })
+    protected set maxSetter(x: number) {
+        isWritableSignal(this.max) && this.max.set(x);
```
waterplea, line 111: Can it be `this.max.set?.(x)`?

projects/kit/components/slider/helpers/slider-key-steps.directive.ts
nsbarsukov: Both directives are already part of 

```ts
export const TuiInputSlider = [...] as const;
```

Why do we need to create one more entity ?

projects/kit/components/slider/slider.component.ts
nsbarsukov: Native slider behaves in the such way

https://github.com/taiga-family/taiga-ui/blob/55b4fb84b65ce8194ffcd00fa91e66285aff62b3/projects/kit/components/slider/test/slider.spec.ts#L176-L189

- https://github.com/taiga-family/taiga-ui/pull/1703

projects/kit/components/slider/slider.component.ts
nsbarsukov: It will throw error:

```
Error: node_modules/@angular/cdk/coercion/index.d.ts:30:1 - error NG1010:
Input transform function cannot have multiple signatures
  Value is a reference to 'coerceNumberProperty'.

30 export declare function coerceNumberProperty(value: any): number;
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  node_modules/@angular/cdk/coercion/index.d.ts:30:25
    30 export declare function coerceNumberProperty(value: any): number;
                               ~~~~~~~~~~~~~~~~~~~~
    Reference is declared here.
```

projects/kit/components/slider/slider.component.ts
nsbarsukov: It will throw ts-error:

```
Error: projects/kit/components/slider/slider.component.ts:111:18 - error TS2339:
Property 'set' does not exist on type 'WritableSignal<number> | Signal<number>'.
  Property 'set' does not exist on type 'Signal<number>'.

111         this.max.set?.(x);
                     ~~~

```

projects/kit/components/slider/helpers/slider-key-steps.directive.ts
nsbarsukov: It will not be inherited :(

projects/kit/components/slider/helpers/slider-key-steps.directive.ts
waterplea: Oh, I forgot we already have this, sorry:
```ts
export const TuiSlider = [
    TuiSliderComponent,
    TuiSliderThumbLabel,
    TuiSliderKeySteps,
    TuiSliderReadonly,
] as const;
```

projects/kit/components/slider/slider.component.ts
waterplea: No, I mean why moving `range` in a separate const? Why not:
```ts
return (this.value - this.min()) / (this.max() - this.min()) || 0;
```

projects/kit/components/slider/slider.component.ts
waterplea: Ok. Would be nice to get rid of duality of Writable and non-Writable then.

projects/kit/components/input-slider/input-slider.component.ts
nsbarsukov: It is an option only for `computed(...)`. Effect has its own first argument (`onCleanup` callback) 😢

projects/kit/components/slider/slider.component.ts
nsbarsukov: It is not interchangeable code.
If  `this.max() - this.min()` return `0` => all division returns `Infinity`.

projects/kit/components/slider/helpers/slider-key-steps.directive.ts
nsbarsukov: Update this logic 

https://github.com/taiga-family/taiga-ui/pull/10288/commits/2fddfb73ad465a51517c6ca03168bfe493f9370a

Now, both `Slider` + `InputSlider` have `attr.aria-valuenow`

projects/kit/components/input-slider/input-slider.component.ts
nsbarsukov: Is it better ?

```ts
function computedWithPrev<T>(
    fn: (prev?: T) => T,
    options?: CreateComputedOptions<T>,
): Signal<T> {
    let prev: T;

    return computed<T>(() => {
        prev = fn(prev);

        return prev;
    }, options);
}
```

```ts
protected value = computedWithPrev<number>(
    (prev) => this.inputNumber()?.value() ?? prev!,
);
```

projects/kit/components/slider/helpers/slider-key-steps.directive.ts
nsbarsukov: It should be overwritten on every input change.
I could use `effect` instead.

Is this solution is better
```ts
private readonly keyStepsBase = inject(TuiSliderKeyStepsBase);
protected readonly $ = effect(
    () => (this.transformer = this.keyStepsBase.transformer()),
);
```

than this one ?
```ts
private readonly slider = inject<TuiSliderComponent>(
    forwardRef(() => TuiSliderComponent),
);

@Input()
public set keySteps(steps: TuiKeySteps) {
    this.transformer = tuiCreateKeyStepsTransformer(steps, this.slider);
}
```

projects/kit/components/slider/slider.component.ts
nsbarsukov, line 1: Something goes wrong here!

image

nsbarsukov: https://stackblitz.com/edit/slider-async-max-bug?file=src%2Fmain.ts

projects/demo/src/modules/components/input-slider-legacy/index.html
```diff
@@ -0,0 +1,238 @@
+<tui-doc-page
+    header="InputSliderLegacy"
+    package="LEGACY"
+    type="components"
```
vladimirpotekhin, line 4: please add "DEPRECATED" badge and link to the new variant

nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/10288/commits/05c18430f6f4aa2c0d07784541f74f0b9202b87e

projects/kit/components/input-slider/input-slider.component.ts
nsbarsukov: New version of the component does not have this problem ✅

projects/kit/components/input-slider/input-slider.component.ts
nsbarsukov: UPDATE: this problem is not relevant for new version of the component.
The previous logic is replaced by the following lines:

```ts
private readonly controlTransformer = inject(TuiValueTransformer, {self: true});
private readonly value = computed(() =>
    this.controlTransformer.toControlValue(this.inputNumber.value()),
);
```

projects/kit/components/input-slider/input-slider.directive.ts
```diff
@@ -0,0 +1,137 @@
...
+        (x) => x instanceof TuiSliderComponent,
+    );
+
+    private readonly controlTransformer = inject(TuiValueTransformer, {self: true});
+    private readonly value = computed(() =>
+        this.controlTransformer.toControlValue(this.inputNumber.value()),
+    );
+
+    private readonly keyStepsTransformer = computed<TuiValueTransformer<
+        number,
```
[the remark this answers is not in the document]
nsbarsukov: https://github.com/taiga-family/taiga-ui/pull/10288/commits/01c90d6551975f02cbf3da1055446e246a4b9cfe

projects/kit/components/input-slider/input-slider.directive.ts
```diff
@@ -0,0 +1,137 @@
...
+        number
+    > | null>(() => this.slider()?.keySteps?.transformer() ?? null);
```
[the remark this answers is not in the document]
nsbarsukov: https://github.com/taiga-family/taiga-ui/pull/10288/commits/01c90d6551975f02cbf3da1055446e246a4b9cfe

projects/kit/components/input-slider/input-slider.directive.ts
```diff
@@ -0,0 +1,137 @@
...
+
+        const subscription = fromEvent(slider.el, 'input', (x) => x.target)
+            .pipe(filter(tuiIsElement), filter(tuiIsInput))
+            .subscribe((x) => this.onSliderInput(x.valueAsNumber));
+
+        onCleanup(() => subscription.unsubscribe());
+    });
+
+    constructor() {
+        this.textfield.options.cleaner.set(false);
```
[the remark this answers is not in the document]
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/10288/commits/4be35425ff5cfa90c15398bceaadf5ca1dfcc26a

projects/kit/components/slider/slider.component.ts
```diff
@@ -56,6 +58,11 @@ export class TuiSliderComponent {
         }
     }
 
+    @tuiPure
+    public get keySteps(): TuiSliderKeyStepsBase | null {
+        return this.injector.get(TuiSliderKeyStepsBase, null);
```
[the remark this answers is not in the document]
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/10288/commits/fef7cf394186faaa7954f3de4708c58b5d00d878 and https://github.com/taiga-family/taiga-ui/pull/10288/commits/521c6c6d975a79a2a8efb30f5bed186c2f2f4af4

waterplea (approved):

PR #10307, opened 2025-02-04
feat(layout): `Cell` add disabled state

vladimirpotekhin:
Fixes #

### DIFF
--- a/projects/layout/components/cell/cell.styles.less
+++ b/projects/layout/components/cell/cell.styles.less
@@ -16,7 +16,18 @@
     min-block-size: var(--t-height);
     border-radius: var(--t-radius);
 
-    &:disabled {
+    &:disabled,
+    &[data-state='disabled'] {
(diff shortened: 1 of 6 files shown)


splincode (approved):

projects/layout/components/cell/cell.styles.less
```diff
@@ -213,7 +228,9 @@
     a[tuiCell]:not([tuiSurface]):hover,
     button[tuiCell]:not([tuiSurface]):hover,
     label[tuiCell]:not([tuiSurface]):hover {
-        background: var(--tui-background-neutral-1);
-        cursor: pointer;
+        &:not(:disabled, [data-state='disabled']):not(:has(input:disabled)) {
```
waterplea, line 231: You cannot use `:has` here or it will break this whole rule for old browsers

waterplea (approved):

PR #10311, opened 2025-02-04
feat(icons-lucide): add new package

splincode:


waterplea:
Let's not confuse users with 2 same packages until v5 is out

PR #10324, opened 2025-02-06
fix(kit): input should be below text

codrone:
Fixes # 
https://github.com/taiga-family/taiga-ui/blob/main/projects/kit/styles/components/block.less#L90


waterplea (approved):

splincode (approved):

waterplea:
Thank you! Please move `z-index` above `min-block-size` to satisfy linter

PR #10341, opened 2025-02-10
fix(core): use `TUI_DARK_MODE` instead of element in dropdown

olegnowikow:
Coorrects change theme in dropdown.
Before:

https://github.com/user-attachments/assets/570d95bc-fa34-41cf-8f2a-9b6d6ed8eb90

After:

https://github.com/user-attachments/assets/89c06fd2-47a7-4e7f-b812-7a5a9e617058


projects/core/directives/dropdown/dropdown.component.ts
```diff
@@ -60,9 +68,26 @@ export class TuiDropdownComponent {
...
+    protected themeSignal = signal(
+        this.directive.el.closest('[tuiTheme]')?.getAttribute('tuiTheme'),
+    );
+
+    // TODO(v5): delete it
+    protected readonly updateTheme = effect(() => {
+        this.themeSignal.set(
+            this.directive.el.closest('[tuiTheme]')?.getAttribute('tuiTheme'),
+        );
+    }, TUI_ALLOW_SIGNAL_WRITES);
```
waterplea, line 90: Most of this code doesn't do anything :) Here's the same thing shorter:

```ts
    protected readonly darkMode = inject(TUI_DARK_MODE);

    // TODO(v5): use `TUI_DARK_MODE` instead of element attribute to get current theme
    protected theme = computed(() => {
        // @ts-ignore
        const _ = this.darkMode();

        return this.directive.el.closest('[tuiTheme]')?.getAttribute('tuiTheme');
    });
```

olegnowikow:
```ts
const _ = this.darkMode();
 ```
 mb just use a call without assignment, to eslint correct?

waterplea: Sure, if lint is ok with it

waterplea (approved):

vladimirpotekhin (approved):

PR #10345, opened 2025-02-11
[WIP] feat(kit): `InputMonth` refactor to new `Textfield`

nsbarsukov:
```ts
@Component({
  standalone: true,
  imports: [
    TuiTextfield,
    TuiInputMonth,
  ],
})
export class Page {}
```

```html
<tui-textfield>
  <input
    tuiInputMonth
    [max]="max"
    [min]="min"
    [disabledItemHandler]="disabledItemHandler"
    [(ngModel)]="value"
  />
</tui-textfield>
```


nsbarsukov:
@splincode fixed

splincode (approved):

projects/demo/src/modules/components/input-month/examples/2/index.ts
```diff
@@ -1,19 +1,23 @@
...
+            // Callback has the first argument – size of the textfield box ('s' | 'm' | 'l')
+            icon: () => '',
+        }),
+    ],
 })
 export default class Example {
-    protected readonly testForm = new FormGroup({
-        testValue: new FormControl<TuiMonth | null>(null),
-    });
+    protected value: number | null = null;
```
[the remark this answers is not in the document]
nsbarsukov: Good catch!

I've added example + fix validation problem via creating new `TuiWithChildNativeValidator` host-directive.

projects/kit/components/input-month/input-month.directive.ts
```diff
@@ -0,0 +1,136 @@
...
+@Directive({
+    standalone: true,
+    selector: 'input[tuiInputMonth]',
+    providers: [tuiAsControl(TuiInputMonthDirective), tuiFallbackValueProvider(null)],
+    hostDirectives: [TuiWithTextfield],
+    host: {
+        '(click)': 'toggleDropdown()',
+        '(blur)': 'onTouched()',
+        '[disabled]': 'disabled()',
+        // Make <input /> readonly (without readonly appearance)
```
waterplea, line 40: Not sure we need this comment here, how about we add [inputmode='none'] { caret-color: transparent } into textfield styles? Do we ever need it with visible caret?

projects/kit/components/input-month/input-month.directive.ts
```diff
@@ -0,0 +1,136 @@
...
+import type {TuiCalendarMonth} from '@taiga-ui/kit/components/calendar-month';
+import {TUI_MONTH_FORMATTER} from '@taiga-ui/kit/tokens';
+
+import {TUI_INPUT_MONTH_OPTIONS} from './input-month.options';
+import {TuiNativeMonthPicker} from './native-month-picker/native-month-picker.component';
+
+@Directive({
+    standalone: true,
+    selector: 'input[tuiInputMonth]',
+    providers: [tuiAsControl(TuiInputMonthDirective), tuiFallbackValueProvider(null)],
```
waterplea, line 34: Do we need `null` as a fallback? Isn't it a default?

projects/kit/components/input-month/input-month.directive.ts
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    @Input('max')
+    public set maxSetter(x: TuiMonth | null) {
+        this.max.set(x);
+    }
+
+    public override setDisabledState(): void {
+        super.setDisabledState();
+        this.open.set(false);
```
waterplea, line 114: We will need this everywhere, I wonder if we can somehow reuse it in an abstract class

projects/kit/components/input-month/native-month-picker/native-month-picker.style.less
```diff
@@ -0,0 +1,17 @@
...
+    opacity: 0;
+    border: 0;
+
+    &::-webkit-calendar-picker-indicator {
+        .fullsize(); // Extends clickable area for icon to open native calendar
+    }
+}
+
+tui-textfield:has([tuiInputMonth]) .t-clear {
+    z-index: 1;
```
waterplea, line 16: Does it make any sense for clear not to be above everything else out of the box? Maybe we can just apply this rule all the time

projects/kit/components/input-month/native-month-picker/native-month-picker.style.less
```diff
@@ -0,0 +1,17 @@
+@import '@taiga-ui/core/styles/taiga-ui-local';
+
+// TODO: add tui-textfield:has([tuiInputMonth])
```
waterplea, line 3: tui-textfield input[tuiInputMonth] + .t-content input[type='month']
Maybe like that?

projects/kit/components/input-month/input-month.directive.ts
```diff
@@ -0,0 +1,136 @@
...
+        if (this.interactive() && !this.nativePicker?.enabled) {
+            this.open.update((x) => !x);
+        }
+    }
+
+    protected clear(): void {
+        this.onChange(null);
+
+        if (this.nativePicker?.enabled) {
+            this.nativePicker.showPicker();
```
waterplea, line 127: What if I just want to remove the value – on android it would be a full on modal window. I think it's ok not to toggle it automatically on cleaner tap

projects/kit/components/input-month/input-month.directive.ts
```diff
@@ -0,0 +1,136 @@
...
+        this.max.set(x);
+    }
+
+    public override setDisabledState(): void {
+        super.setDisabledState();
+        this.open.set(false);
+    }
+
+    protected toggleDropdown(): void {
+        if (this.interactive() && !this.nativePicker?.enabled) {
```
waterplea, line 118: If we manage to do this somehow differently we can get rid of that circular dependency and injector

projects/demo/src/modules/components/input-month/examples/4/index.ts
```diff
@@ -0,0 +1,27 @@
...
+})
+export default class Example {
+    @ViewChild(TuiDropdownOpen)
+    protected readonly dropdown?: TuiDropdownOpen;
+
+    protected value: TuiMonth | null = null;
+
+    protected chooseTheOnlyCorrectOption(): void {
+        this.value = new TuiMonth(1998, 2);
+        this.dropdown?.toggle(false);
```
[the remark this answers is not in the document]
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/10345/commits/78cbd5203964835f1463fc4ff044c0ecfee8e14e

I've used signal instead `[(open)]` because of change detection (otherwise, I'll need to use `this.cdr.markForCheck`).

projects/kit/components/input-month/input-month.directive.ts
nsbarsukov: https://github.com/taiga-family/taiga-ui/pull/10345/commits/c2580fce6cc9ac7226bc714eeca3eaae1dc34286

projects/kit/components/input-month/input-month.directive.ts
nsbarsukov: fixed in https://github.com/taiga-family/taiga-ui/pull/10345/commits/3e93bc54d56b6d8e5a72124aabd2f15c9e111133

projects/kit/components/input-month/native-month-picker/native-month-picker.style.less
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/10345/commits/57b3855e2c9dba9441bc6bc2f5db2365dc45139e

projects/core/components/textfield/textfield.component.ts
```diff
@@ -137,6 +140,11 @@ export class TuiTextfieldComponent<T> implements TuiDataListHost<T> {
         this.open.set(false);
     }
 
+    @ContentChild(TUI_AUXILIARY, {descendants: true})
```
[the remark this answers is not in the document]
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/10345/commits/5020fcb9c1b82e6fbce005764a6d93f7de926506

projects/kit/components/input-month/input-month.directive.ts
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/10345/commits/756d45357deb599faf2b516f040edbe4af8ca4d5

projects/core/components/textfield/textfield-auxiliary.ts
```diff
@@ -0,0 +1,21 @@
+import type {Provider, Signal, Type} from '@angular/core';
+import {computed, inject} from '@angular/core';
+import {tuiCreateToken, tuiProvide} from '@taiga-ui/cdk/utils/miscellaneous';
+
+import {TuiTextfieldComponent} from './textfield.component';
+
+export const TUI_AUXILIARY = tuiCreateToken(null);
+
+export function tuiAsAuxiliary(x: Type<unknown>): Provider {
+    return tuiProvide(TUI_AUXILIARY, x, true);
```
waterplea, line 10: I don't think it makes sense to provide it as multi token, we only need an array because there can be many of such content children, not because there can be multiple of auxiliaries on one injector.

projects/core/components/textfield/textfield.component.ts
```diff
@@ -137,6 +141,11 @@ export class TuiTextfieldComponent<T> implements TuiDataListHost<T> {
         this.open.set(false);
     }
 
+    @ContentChildren(TUI_AUXILIARY, {descendants: true})
+    protected set auxiliariesSetter(x: ReadonlyArray<Type<unknown>>) {
```
waterplea, line 145: Is it readonly array here and not query list? Are we sure it is triggered everytime it changes and we shouldn't instead listen to changes on the query list?

projects/kit/components/input-month/input-month.directive.ts
```diff
@@ -0,0 +1,120 @@
...
+@Directive({
+    standalone: true,
+    selector: 'input[tuiInputMonth]',
+    providers: [tuiAsControl(TuiInputMonthDirective)],
+    hostDirectives: [TuiWithTextfield],
+    host: {
+        '(click)': 'toggleDropdown()',
+        '(blur)': 'onTouched()',
+        '[disabled]': 'disabled()',
+        inputmode: 'none',
```
waterplea, line 31: Should we sort those like we do in the template? attribute -> binding -> event?

projects/kit/components/input-month/input-month.directive.ts
```diff
@@ -0,0 +1,120 @@
...
+            this.onChange(month);
+            this.open.set(false);
+        });
+
+        onCleanup(() => subscription?.unsubscribe());
+    });
+
+    public readonly min = signal<TuiMonth | null>(null);
+    public readonly max = signal<TuiMonth | null>(null);
+    public readonly calendar = tuiInjectAuxiliary<TuiCalendarMonth>();
```
waterplea, line 78: Technically we want not the first item but `instanceof TuiCalendarMonth` one

projects/kit/components/input-month/input-month.directive.ts
```diff
@@ -0,0 +1,120 @@
...
+    public readonly nativePickerEnabled =
+        tuiInjectElement<HTMLInputElement>().type === 'month' && inject(TUI_IS_MOBILE);
```
waterplea, line 80: Do we no longer reset it back to text? Is that ok?

waterplea (approved):

vladimirpotekhin (approved):

Issue #6576, opened 2024-01-26
🐞 - Reorder - items moves incorrectly when continuously dragging or when the container is overflowed


baltzarmattsson:
### Playground Link

https://stackblitz.com/edit/angular-tiprds?file=src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.component.html

### Description

Hi, thanks for a great library. The items moves weirdly when either moving the item continuously or when the container overflows.

**Moving continously**:

[c](https://github.com/taiga-family/taiga-ui/assets/20445589/05c0d2b3-d8e0-4a82-a7c5-cbce72dceff2)

**Overflow** (I mousewheel down when reaching item 14)

[overflow](https://github.com/taiga-family/taiga-ui/assets/20445589/ff372e5b-ee03-4318-9e96-f500abb93390)

### Angular version

16.0.0

### Taiga UI version

3.40.0

### Which browsers have you used?

- [X] Chrome
- [X] Firefox
- [ ] Safari
- [X] Edge

### Which operating systems have you used?

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

waterplea:
It is working as expected in terms of it switches two items positions, not moves your current item in the new place while shifting all items in between. It does look a bit weird when dragging fast like that, I wonder what can we do about it that does not require rewriting the overall principle of tiles, a component which is underneath 🤔

40oleg:
May I take this problem?

waterplea:
> May I take this problem?

What's your plan for it?

kulagind:
Hi there! I'll take it. I suppose it consists of 2 parts - add possibility of shifting behaviour (leaving current "swap" behaviour by default) and fixing scroll while drag element in long list

waterplea:
@kulagind let's only fix the first part for now, I'm not sure we should add scroll into the drag. The scrolling containers can be very diverse, it would be easier to externally add some invisible DIV when an item is dragged and do some action like scrolling when you hover over it than try and guess what to do inside the dragging directives.

PR #10360, opened 2025-02-13
feat(kit): `Tiles` add reorder strategies

kulagind:
Fixes https://github.com/taiga-family/taiga-ui/issues/6576.


projects/demo/src/modules/components/reorder/index.html
```diff
@@ -14,5 +14,33 @@
...
+
+            <li class="tui-list__item">
+                <p>Add to the template:</p>
+
+                <tui-doc-code [code]="templateCode" />
+            </li>
+
+            <li class="tui-list__item">
+                <p>
+                    By default reordering has swap behaviour. You can overwrite it by providing either
```
waterplea, line 33: I don't think we ever need to override it in `Reorder`. It makes sense for it to always be in `shift` mode

projects/kit/components/tiles/tiles-reorder.providers.ts
```diff
@@ -0,0 +1,29 @@
+import type {FactoryProvider} from '@angular/core';
```
waterplea, line 1: That's quite a lot of tokens, how about we create only one: `TUI_TILE_REORDER` and 2 imported methods `tuiTileSwap` and `tuiTileShift`. Swap will be default value for that token and in reorder we will provide shift one. Otherwise it looks a bit over-engineered.

waterplea (approved):

vladimirpotekhin (approved):

PR #10362, opened 2025-02-13
chore: add basic a11y test

vladimirpotekhin:
Fixes #


splincode (approved):

projects/addon-doc/components/code/index.html
```diff
@@ -15,6 +15,7 @@
             appearance="outline-grayscale"
             size="s"
             class="t-copy-button"
+            title="Copy"
```
nsbarsukov, line 18: Should we use the following token ?

https://github.com/taiga-family/taiga-ui/blob/24572b1cdc7e0bbd48b3bcbf7faf3960e80c4372/projects/kit/tokens/i18n.ts#L41

projects/demo/src/modules/components/input-inline/examples/4/index.html
```diff
@@ -1,13 +1,13 @@
 <ul class="task">
-    <ol>
```
nsbarsukov, line 2: Oops 😄

nsbarsukov (approved):

projects/addon-doc/components/code/index.html
waterplea: Also we can put it in content of tuiIconButton instead of title that shows on hover.

BTW, what's up with this part, a few lines above, why isn't it formatted properly?
```html
        <button type="button"
            tuiIconButton
```

projects/demo/src/modules/components/input/examples/6/index.html
```diff
@@ -44,6 +44,7 @@
                 <td tuiTd>
                     <tui-textfield tuiChevron>
                         <select
+                            aria-label="color"
```
waterplea, line 47: We should probably internalize it based on `placeholder` and aria-placeholder?

projects/layout/components/search/search-filters.template.html
```diff
@@ -8,9 +8,11 @@
     #rem
     class="t-rem"
 ></div>
+<!-- TODO: title i18n -->
 <button
     *ngIf="overflown() < templates.length"
     appearance="secondary"
+    title="Filters"
```
waterplea, line 15: Should be as plain text here:
https://github.com/taiga-family/taiga-ui/blob/main/projects/layout/components/search/search-filter.component.ts#L21

projects/demo/src/modules/components/input/examples/6/index.html
vladimirpotekhin: I'm not sure, the placeholder and the label may be different

projects/addon-doc/components/code/index.ts
```diff
@@ -41,6 +42,7 @@ export class TuiDocCode {
         inject(TUI_DOC_EXAMPLE_MARKDOWN_CODE_PROCESSOR);
 
     protected readonly copy$ = new Subject<void>();
+    protected readonly copyText$ = inject(TUI_COPY_TEXTS).pipe(map(([copy]) => copy));
```
waterplea, line 45: Let's start using toSignal instead of async pipe for such cases

waterplea (approved):
Would be nice to eventually enable 'label' rule and 'nested-interactive', if it's possible to add exceptions to it inline where we have to use it.

PR #10388, opened 2025-02-17
fix(layout): `Card` add default background color

waterplea:
Fixes #

### DIFF
--- a/projects/layout/components/card/card.styles.less
+++ b/projects/layout/components/card/card.styles.less
@@ -11,6 +11,7 @@
     text-decoration: none;
     overscroll-behavior: contain;
     overflow: hidden;
+    background: var(--tui-background-elevation-2);
 
     [tuiTitle],
     [tuiSubtitle] {


vladimirpotekhin (approved):

splincode (approved):

ankemp:
Honestly this should've been marked as a breaking change, or at least had a migration attached to it. Completely changed the look of my application on updated from 4.24 to 4.25.

waterplea:
> Honestly this should've been marked as a breaking change, or at least had a migration attached to it. Completely changed the look of my application on updated from 4.24 to 4.25.

Were you not applying any background color to the card at all? This was the lowest specificity style, didn't expect anybody to use cards just for the padding/flex gap 🤔

ankemp:
Most of the time, no I was not applying a background, and when I was it was inherited from appearance.

Here are some examples:
localhost_4200_tasks_new(Pixel 7) (1)
localhost_4200_tasks_new(Pixel 7)

waterplea:
> Most of the time, no I was not applying a background, and when I was it was inherited from appearance.

Appearance has higher specificity as would any other rule. Card without appearances/custom styles does not do anything except adding padding and gaps. I'm sorry it affected you, we did not expect anybody to use it that way. Was it easy for you to fix? I would rather keep them that way so they can be used as is on base-alt background:

image
image

PR #10397, opened 2025-02-18
chore(demo): add example

### LINKED ISSUE #10394: 📚 - add example with small icons
### What is the affected URL?

https://taiga-ui.dev/components/icon

### Description

we need to add example with 16x16 icon

### Browser Used

Chrome

### OS Used

macOS

splincode:
Fixes #10394

image

### DIFF
--- a/projects/demo/src/modules/components/icon/examples/9/index.html
+++ b/projects/demo/src/modules/components/icon/examples/9/index.html
@@ -32,5 +32,4 @@ container is 24x24 and icon is 16x16
 <tui-icon
     icon="@tui.settings"
     [style.border-width.rem]="0.25"
-    [style.font-size.rem]="1.5"
 />
--- a/projects/demo/src/modules/components/icon/examples/9/index.html
+++ b/projects/demo/src/modules/components/icon/examples/9/index.html
@@ -1,0 +1,36 @@
+<tui-icon
+    icon="@tui.settings"
+    [style.font-size.rem]="4"
+/>
+
+<tui-icon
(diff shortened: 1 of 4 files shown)


projects/demo/src/modules/components/icon/examples/9/index.less
```diff
@@ -0,0 +1,42 @@
+:host {
+    display: flex;
+    flex-flow: row wrap;
+    align-items: center;
+    justify-content: center;
+    gap: 0.625rem;
+
+    --x1-size: 1rem;
```
waterplea, line 8: I think variables overcomplicate example, let's skip them. Maybe we can even inline styles like [style.font-size.rem]="1" so that it's all in one line and in one place in HTML all available at a glance?

vladimirpotekhin (approved):

projects/demo/src/modules/components/icon/examples/9/index.less
splincode: fixed

projects/demo/src/modules/components/icon/examples/9/index.html
```diff
@@ -0,0 +1,36 @@
...
+<hr />
+
+Reverse scaling,
+<br />
+container is 24x24 and icon is 16x16
+
+<tui-icon
+    icon="@tui.settings"
+    [style.border-width.rem]="0.25"
+    [style.font-size.rem]="1.5"
```
waterplea, line 35: Unnecessary line

PR #10427, opened 2025-02-21
chore(demo): improve stability of anchor scrolling

nsbarsukov:


splincode (approved):

projects/demo/src/modules/app/abstract.app.ts
```diff
@@ -42,12 +45,6 @@ export abstract class AbstractDemo implements OnInit {
     protected abstract readonly storage: Storage;
     protected abstract readonly router: Router;
 
-    private readonly element = tuiInjectElement();
-
-    protected readonly pageLoaded = inject(TUI_DOC_PAGE_LOADED)
-        .pipe(takeUntilDestroyed())
-        .subscribe(() => this.element.classList.add('_loaded'));
```
nsbarsukov, line 49: It does not make sense because `TUI_DOC_PAGE_LOADED` emmited `false` on initialization and immediately added `_loaded` class (even when application is not ready).

vladimirpotekhin (approved):

PR #10430, opened 2025-02-21
fix(core): fix dropdown when host is svg element

vladimirpotekhin:
Fixes #

### DIFF
--- a/projects/core/directives/dropdown/dropdown.component.ts
+++ b/projects/core/directives/dropdown/dropdown.component.ts
@@ -72,7 +72,9 @@ export class TuiDropdownComponent {
     protected readonly sub = inject(TuiPositionService)
         .pipe(
             takeWhile(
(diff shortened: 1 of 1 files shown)


projects/core/directives/dropdown/dropdown.component.ts
```diff
@@ -72,7 +72,9 @@ export class TuiDropdownComponent {
     protected readonly sub = inject(TuiPositionService)
         .pipe(
             takeWhile(
-                () => this.directive.el.isConnected && !!this.directive.el.offsetParent,
+                () =>
```
splincode, line 75: could you add e2e test please?

splincode (approved):

projects/core/directives/dropdown/dropdown.component.ts
vladimirpotekhin: don't want to clutter up the demo, non target case

PR #10431, opened 2025-02-21
fix(core): `DataList` fix empty content

### LINKED ISSUE #10425: `data-list` empty-content with <a> tag
### Reproduction url

- 

### Description

if in tui-data-list input emptyContent to ng-template containing <a> tag - emptyContent starts to be deleted spontaneously.

Image

we should change that to `a[tuiOption],button[tuiOption],label[tuiOption]`

### Taiga UI version

last

### Browser Used

Chrome

### OS Used

macOS

vladimirpotekhin:
Fixes #10425

### DIFF
--- a/projects/core/components/data-list/data-list.component.ts
+++ b/projects/core/components/data-list/data-list.component.ts
@@ -115,6 +115,8 @@ export class TuiDataListComponent<T>
     }
 
     private get elements(): readonly HTMLElement[] {
-        return Array.from(this.el.querySelectorAll('a,button,input'));
+        return Array.from(
+            this.el.querySelectorAll('a[tuiOption],button[tuiOption],input[tuiOption]'),
+        );
     }
 }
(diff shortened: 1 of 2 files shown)


projects/core/components/data-list/data-list.component.ts
```diff
@@ -115,6 +115,8 @@ export class TuiDataListComponent<T>
     }
 
     private get elements(): readonly HTMLElement[] {
-        return Array.from(this.el.querySelectorAll('a,button,input'));
+        return Array.from(
+            this.el.querySelectorAll('a[tuiOption],button[tuiOption],input[tuiOption]'),
```
nsbarsukov, line 119: I have doubts about `input[tuiOption]` 🤔 
Should we leave it as it was ?

https://github.com/taiga-family/taiga-ui/blob/b5cafe5cea5f761a1183c3764b334723b95609c8/projects/core/components/data-list/option.component.ts#L25

nsbarsukov (approved):

PR #10442, opened 2025-02-24
feat(addon-mobile): `SheetDialog` add `fullscreen`

mdlufy:
`SheetDialog` add `fullscreen` option, add example and update API page

image
image

### DIFF
--- a/projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
+++ b/projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
@@ -5,7 +5,6 @@ import {
     Component,
     ElementRef,
     inject,
-    ViewChild,
     ViewChildren,
 } from '@angular/core';
 import {EMPTY_QUERY} from '@taiga-ui/cdk/constants';
@@ -37,8 +36,6 @@ function isCloseable(this: TuiSheetDialogComponent<unknown>): boolean {
     host: {
         '[@tuiSlideInTop]': 'slideInTop',
         '[style.--tui-offset.px]': 'context.offset',
-        '[style.--t-bar.px]': 'bar?.nativeElement.clientHeight',
-        '[style.--t-heading.px]': 'heading?.nativeElement.clientHeight',
         '[class._closeable]': 'context.closeable === true',
         '[class._fullscreen]': 'context.fullscreen === true',
         '(document:touchstart.passive.zoneless)': 'onPointerChange(1)',
@@ -54,12 +51,6 @@ export class TuiSheetDialogComponent<I> implements AfterViewInit {
 
     private readonly el = tuiInjectElement();
     private pointers = 0;
-
-    @ViewChild('bar')
-    protected readonly bar?: ElementRef<HTMLElement>;
-
-    @ViewChild('heading')
-    protected readonly heading?: ElementRef<HTMLElement>;
 
     protected readonly slideInTop = {
         value: '',
--- a/projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
+++ b/projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
@@ -40,7 +40,7 @@ function isCloseable(this: TuiSheetDialogComponent<unknown>): boolean {
         '[style.--t-bar.px]': 'bar?.nativeElement.clientHeight',
         '[style.--t-heading.px]': 'heading?.nativeElement.clientHeight',
         '[class._closeable]': 'context.closeable === true',
-        '[class._fullsize]': 'context.fullsize === true',
+        '[class._fullscreen]': 'context.fullscreen === true',
         '(document:touchstart.passive.zoneless)': 'onPointerChange(1)',
         '(document:touchend.zoneless)': 'onPointerChange(-1)',
         '(document:touchcancel.zoneless)': 'onPointerChange(-1)',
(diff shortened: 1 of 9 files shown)


splincode (approved):

projects/addon-mobile/components/sheet-dialog/sheet-dialog.options.ts
```diff
@@ -12,6 +12,7 @@ export interface TuiSheetDialogOptions<I = undefined> {
     readonly offset: number;
     readonly stops: readonly string[];
     readonly bar: boolean;
+    readonly fullsize: boolean;
```
vladimirpotekhin, line 15: let's name this option `fullscreen`, like the size in the dialogs and like in figma

projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
```diff
@@ -51,6 +55,12 @@ export class TuiSheetDialogComponent<I> implements AfterViewInit {
     private readonly el = tuiInjectElement();
     private pointers = 0;
 
+    @ViewChild('bar')
+    protected readonly bar?: ElementRef<HTMLElement>;
+
+    @ViewChild('heading')
+    protected readonly heading?: ElementRef<HTMLElement>;
```
vladimirpotekhin, line 62: can we try to do it via css only, without bar and heading query?

mdlufy: I'm tried to use only css, but i think it's not possible: if the bar can still be considered a fixed size with paddings (it's string), but there are heading that can be `PolymorpheusContent` with own dynamic size, than we need to calc

Do you have any ideas how to calc without querying?

https://github.com/taiga-family/taiga-ui/blob/main/projects/addon-mobile/components/sheet-dialog/sheet-dialog.template.html#L19-L21
image

vladimirpotekhin: We need the sheet and content to fill all the available space, right? At first glance, something like this is enough:
```
   &._fullscreen {
        .t-sheet {
            display: flex;
            flex-direction: column;
            min-height: 100%;
        }

        .t-content {
            flex-grow: 1;
        }
    }
    ```
+ remove t-content `::after` in fullscreen mode to remove stops

Comment (mdlufy) [reply projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts:62]: Yes, if we want only sheet fills fullscreen it wil be almost enough :)

If we want content, which is placed by user, to be filled all available space and growed to down, we need to add `display: flex, flex-direction: column` to `.t-content`. In addition, it's will be usefull for positioning sticky button in the bottom.

It's also worth considering that sheet's without `closeable: true` (when false) sheet will be without [styles](https://github.com/taiga-family/taiga-ui/blob/main/projects/addon-mobile/components/sheet-dialog/sheet-dialog.style.less#L24-L29) that makes host `block`, so it will be some incorrect sheet size (see the video). Therefore, we need to have `display: block` on the host when use fullscreen too.

And I think, that would be nice to have a support stops even in fullscreen? It seems to work by default with fullscreen too.

By the way, I didn't notice any difference in availability `.t-content::after`, when tui-sheet-dialog is `display: block`. With it's `flex`, it works incorrect (demonstrated in video). We can remove it, but it doesn't seem to make a difference.

In result, it looks like code below will be enough for cases with both `closeable: true` and `closeable: false`, and to support the positioning of the sticky button at the bottom of the screen. What do you think about this?

```
&._fullscreen {
        display: block;

        .t-sheet {
            display: flex;
            flex-direction: column;
            min-block-size: 100%;
        }

        .t-content {
            display: flex;
            flex-direction: column;
            flex-grow: 1;
        }
    }
```

I do not know why I did not do this initially, but decided to use `--tui-viewport-height` instead. Maybe we're missing some case that I took into account last time, or maybe I just didn't think to make it easier :)

Comment (mdlufy) [reply projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts]: https://github.com/user-attachments/assets/9261eff4-0852-4bc9-a04b-4be815c45a6a

File: projects/addon-mobile/components/sheet-dialog/sheet-dialog.style.less [hunk 3]
```diff
@@ -107,6 +123,7 @@
     isolation: isolate;
     padding-block-end: calc(1.5rem + env(safe-area-inset-bottom));
     border-radius: inherit;
+    box-sizing: border-box;
```
vladimirpotekhin, line 126: why do we need this?

mdlufy: Looks like it's no longer necessary

mdlufy: Removed

projects/addon-mobile/components/sheet-dialog/sheet-dialog.style.less
```diff
@@ -28,6 +28,22 @@
...
+    &._fullscreen {
+        display: block;
+
+        .t-sheet {
+            display: flex;
+            flex-direction: column;
+            min-block-size: 100%;
+        }
+
+        .t-content {
```
waterplea, line 41: I don't like that enabling fullscreen changes display of the content. It doesn't seem like it would break anything, but that shouldn't depend on the mode. Let's add TODO here to make those styles default for `.t-content` regardless of fullscreen in v5

waterplea (approved):

projects/addon-mobile/components/sheet-dialog/sheet-dialog.style.less
mdlufy: Comment added

vladimirpotekhin (approved):

PR #10543, opened 2025-03-07
chore(demo): `Icon` add example inline svg

mdlufy:
Add new example
image

```
<tui-icon [icon]="icon" />
```

```
protected readonly icon = `data:image/svg+xml;charset=UTF-8,${encodeURIComponent(
        '<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="#000" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="m5 12 7-7 7 7"/><path d="M12 19V5"/></svg>',
)}`;
```

### DIFF
--- a/projects/demo/src/modules/components/icon/examples/10/index.ts
+++ b/projects/demo/src/modules/components/icon/examples/10/index.ts
@@ -6,12 +6,12 @@ import {TuiIcon} from '@taiga-ui/core';
 @Component({
     standalone: true,
     imports: [TuiIcon],
-    templateUrl: './index.html',
+    template: '<tui-icon [icon]="icon" />',
     encapsulation,
     changeDetection,
 })
 export default class Example {
     protected readonly icon = `data:image/svg+xml;charset=UTF-8,${encodeURIComponent(
-        '<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="#000" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="m5 12 7-7 7 7"/><path d="M12 19V5"/></svg>',
+        '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none" stroke="#000" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="m5 12 7-7 7 7"/><path d="M12 19V5"/></svg>',
     )}`;
 }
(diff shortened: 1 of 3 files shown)


projects/demo/src/modules/components/icon/examples/10/index.ts
```diff
@@ -0,0 +1,17 @@
+import {Component} from '@angular/core';
+import {changeDetection} from '@demo/emulate/change-detection';
+import {encapsulation} from '@demo/emulate/encapsulation';
+import {TuiIcon} from '@taiga-ui/core';
+
+@Component({
+    standalone: true,
+    imports: [TuiIcon],
+    templateUrl: './index.html',
```
nsbarsukov, line 9: I think we can just inline single-line template 

```suggestion
    template: '<tui-icon [icon]="icon" />',
```

What do you think about it ?

projects/demo/src/modules/components/icon/index.html
```diff
@@ -34,6 +34,13 @@
             [component]="9 | tuiComponent"
             [content]="9 | tuiExample: 'html,ts'"
         />
+
+        <tui-doc-example
+            id="dynamic-construct-svg"
+            heading="Dynamic construct svg"
```
nsbarsukov, line 40: Let's call it `Inline Svg`

The such naming was used for the previous generation of `Icon`-component:
https://taiga-ui.dev/v3/icons/bundled#all
I think it was rather self-descriptive.

splincode (approved):

projects/demo/src/modules/components/icon/examples/10/index.ts
```diff
@@ -0,0 +1,17 @@
...
+    encapsulation,
+    changeDetection,
+})
+export default class Example {
+    protected readonly icon = `data:image/svg+xml;charset=UTF-8,${encodeURIComponent(
+        '<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="#000" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="m5 12 7-7 7 7"/><path d="M12 19V5"/></svg>',
```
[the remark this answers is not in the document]
mdlufy: Yes, you're so attentive :)

projects/demo/src/modules/components/icon/examples/10/index.ts
mdlufy: «Brevity is the sister of talent»

nsbarsukov (approved):

PR #10553, opened 2025-03-10
chore(demo): add example with `InputMonth` + `TUI_MONTH_FORMATTER` customization

nsbarsukov:


projects/demo/src/modules/components/input-month/examples/5/index.ts
```diff
@@ -1,18 +1,35 @@
...
+                    month: 'short',
+                });
+
+                return of(
+                    (x: TuiMonth | null) =>
+                        (x && formatter.format(x.toLocalNativeDate())) ?? '',
+                );
+            },
+        },
+    ],
```
splincode, line 31: what if

```ts
    providers: [
        tuiMonthFormatterProvider(({locale}) =>
            of(
                (x: TuiMonth | null) =>
                    (x &&
                        Intl.DateTimeFormat(locale, {
                            year: '2-digit',
                            month: 'short',
                        }).format(x.toLocalNativeDate())) ??
                    '',
            ),
        ),
    ],
```

splincode (approved):

projects/demo/src/modules/components/input-month/examples/5/index.ts
nsbarsukov: 1. No need to recreate `Intl.DateTimeFormat` on every callback execution
2. `inject(LOCALE_ID)` is not the only way to get locale (application can use its own custom token or etc)

I think no need to create additional entity `tuiMonthFormatterProvider` (at least, for now)

waterplea (approved):

PR #10577, opened 2025-03-12
fix(core): resolve hint's rattling when scrolling

### LINKED ISSUE #10208: 🐞 - Hints on mobile might overflow
### Reproduction url

https://taiga-ui.dev/components/pulse#popover

### Description

Mobile hints should move so that they never fall out of the screen:

Image

Also we should position them absolutely, like dropdowns, so they do not shake when the whole page is scrolled.

### Taiga UI version

4

### Browser Used

Chrome

### OS Used

macOS

holy1crab:
Fixes #10208

--------------------
Before:

https://github.com/user-attachments/assets/4e582f28-cbda-41b8-8448-e747328de11e

After:

https://github.com/user-attachments/assets/d56dbe4b-f213-4481-a4d3-ac6372e93da5

### DIFF
--- a/projects/core/directives/hint/hint.component.ts
+++ b/projects/core/directives/hint/hint.component.ts
@@ -104,20 +104,27 @@ export class TuiHintComponent<C = any> {
     }
 
     @tuiPure
-    private apply(top: string, left: string, beakTop: number, beakLeft: number): void {
-        this.el.style.top = top;
-        this.el.style.left = left;
-        this.el.style.setProperty('--t-top', `${beakTop}%`);
-        this.el.style.setProperty('--t-left', `${beakLeft}%`);
+    private applyElementStyles(top: number, left: number, isFixed: boolean): void {
+        this.el.style.top = tuiPx(top);
+        this.el.style.left = tuiPx(left);
+        this.el.style.position = isFixed ? 'fixed' : 'absolute';
+    }
+
+    @tuiPure
+    private applyBeakStyles(top: number, left: number): void {
+        this.el.style.setProperty('--t-top', `${top}%`);
+        this.el.style.setProperty('--t-left', `${left}%`);
         this.el.style.setProperty(
             '--t-rotate',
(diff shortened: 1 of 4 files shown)


projects/cdk/utils/dom/get-element-offset.ts
```diff
@@ -25,3 +25,14 @@ export function tuiGetElementOffset(
 
     return {offsetTop, offsetLeft};
 }
+
+export function tuiGetOffsetParentOffset(el?: HTMLElement): {
```
holy1crab, line 29: Maybe it's not looks like `utils` method but atm the same code is used in `dropdown`, `hint` and `line-clamp-position`

projects/core/directives/hint/hint.component.ts
```diff
@@ -112,42 +110,78 @@ export class TuiHintComponent<C = any> {
...
+                hintRect.top + BEAK_EDGE_OFFSET < accessorRect.bottom)
         );
     }
 
+    private positionBeak(accessorRect: DOMRect, hintRect: DOMRect): void {
+        const beakRotate =
+            accessorRect.left >= hintRect.right || hintRect.left >= accessorRect.right;
+        const hintCenterX = hintRect.width / 2;
+        const hintCenterY = hintRect.height / 2;
+        const [beakTop, beakLeft] = this.vvs.correct([
```
holy1crab, line 145: I'm not sure about `this.vvs.correct` in this place. It was previously there. I believe the beak's position inside already fixed parent should not be fixed again.

projects/core/directives/hint/hint.directive.ts
```diff
@@ -66,6 +68,11 @@ export class TuiHintDirective<C>
         }
     }
 
+    @tuiPure // todo: how to invalidate in case of changed styles
```
holy1crab, line 71: I took it from `TuiDropdownDirective` as an example. On the one hand this code is executed in every frame when hint is visible and `pure` looks suitable.

waterplea:
Closing this in favor of https://github.com/taiga-family/taiga-ui/issues/14266 which I will take care of

PR #10585, opened 2025-03-12
chore(kit): refactor dropdown logic for non-editable `InputMonth` textfield

nsbarsukov:


projects/cdk/utils/dom/is-element-editable.ts
```diff
@@ -1,5 +1,8 @@
 import {tuiIsTextfield} from './element-checks';
 
 export function tuiIsElementEditable(element: HTMLElement): boolean {
-    return (tuiIsTextfield(element) && !element.readOnly) || !!element.isContentEditable;
+    return (
+        (tuiIsTextfield(element) && !element.readOnly && element.inputMode !== 'none') ||
```
nsbarsukov, line 5: Relates
https://github.com/taiga-family/taiga-ui/blob/03068fb0c0de2ffd4f1785a0c4cbf4abe763a186/projects/core/styles/components/textfield.less#L34-L35

vladimirpotekhin (approved):

PR #10592, opened 2025-03-14
fix(core): synchronize driver's state

### LINKED ISSUE #9941: 🐞 - Unable to open tuiDropdownContext menu after manually closing it
### Reproduction url

https://taiga-ui.dev/directives/dropdown-context#basic

### Description

Steps to reproduce:

1. RMB on the gear icon
2. Close dropdown via X button
3. RMB on the gear icon again

Expected behavior:

- Context menu opens again

Actual:

- Context menu doesn't open

The only way to open it again is to click anywhere else beforehand.

Found in v3, same behavior in v4.

### Taiga UI version

4.17.0

### Browser Used

Chrome

### OS Used

Windows

splincode:
Fixes #9941

### DIFF
--- a/projects/core/directives/dropdown/dropdown.directive.ts
+++ b/projects/core/directives/dropdown/dropdown.directive.ts
@@ -1,3 +1,4 @@
+import {coerceArray} from '@angular/cdk/coercion';
 import type {AfterViewChecked, ComponentRef, OnChanges, OnDestroy} from '@angular/core';
 import {
     ChangeDetectorRef,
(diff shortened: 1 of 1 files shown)


waterplea (approved):

projects/core/directives/dropdown/dropdown.directive.ts
```diff
@@ -112,5 +120,9 @@ export class TuiDropdownDirective
             this.ref.set(null);
             this.service.remove(ref);
         }
+
+        this.drivers
+            .filter((driver) => driver?.type === this.type)
```
waterplea, line 125: No need to check the type as they are all dropdown drivers. However sometimes directives provide themselves as drivers like `TuiDropdownSelection` or `TuiDropdownHover`. Let's put `TODO` on line 59 to think of a better solution later, I suppose it's good enough for now.

waterplea (approved):

projects/core/directives/dropdown/dropdown.directive.ts
splincode: fixed

mdlufy (approved):

PR #10636, opened 2025-03-21
feat(addon-charts): add `tapColumn` output event in `tui-bar-chart`

### LINKED ISSUE #10617: 🚀 - `BarChart`. Add a click event to the bar
### Description

Hi, we need a click event on a separate bar of the chart to perform an action on it, for example, forming a link and redirect on it. We were offered this hack https://stackblitz.com/edit/angular-x9aktkjd, but I want it to be possible out of the box

splincode:
Fixes #10617

### DIFF
--- a/projects/addon-charts/components/bar-chart/bar-chart.template.html
+++ b/projects/addon-charts/components/bar-chart/bar-chart.template.html
@@ -7,6 +7,8 @@
     [tuiHintContext]="{$implicit: index}"
     [tuiHintDescribe]="getHintId(index)"
     [tuiHintHost]="hintHost"
+    (click)="tapColumn.emit(index)"
+    (keydown.enter)="tapColumn.emit(index)"
     (mouseenter.once)="(0)"
 >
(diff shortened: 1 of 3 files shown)


projects/addon-charts/components/bar-chart/bar-chart.template.html
```diff
@@ -24,6 +25,7 @@
             [size]="size"
             [tabIndex]="hintContent ? 0 : -1"
             [value]="set"
+            (keydown.enter)="tapColumn.emit(index)"
```
waterplea, line 28: nit: I think it bubbles, so it can be moved to the same place as click

waterplea (approved):

projects/addon-charts/components/bar-chart/bar-chart.template.html
splincode: fixed

mdlufy (approved):

PR #10641, opened 2025-03-21
feat(kit): `ItemsWithMore` add multiline mode

vladimirpotekhin:
Fixes #

### DIFF
--- a/projects/demo/src/modules/components/items-with-more/examples/4/index.less
+++ b/projects/demo/src/modules/components/items-with-more/examples/4/index.less
@@ -1,0 +1,33 @@
+@import '@taiga-ui/core/styles/taiga-ui-local';
+
+:host {
+    display: block;
+    max-inline-size: 30rem;
+}
+
+.container {
+    max-block-size: 4.5rem;
+    overflow: hidden;
+
+    &_expanded {
+        max-block-size: 100%;
+    }
+}
+
+.item {
+    .transition(~'scale, opacity');
+
+    opacity: 0;
+    scale: 0.9;
+    margin: 0 0.25rem 0.25rem 0;
+
+    &_visible {
+        opacity: 1;
+        scale: 1;
+    }
+}
+
+.more {
+    inline-size: 5rem;
+    margin: 0 0.25rem 0.25rem 0;
+}
--- a/projects/demo/src/modules/components/items-with-more/examples/4/index.less
+++ b/projects/demo/src/modules/components/items-with-more/examples/4/index.less
@@ -15,19 +15,18 @@
 }
 
 .item {
-    .transition(~'scale, opacity');
+    .transition(~'transform, opacity');
 
     opacity: 0;
-    scale: 0.9;
+    transform: scale(0.9);
     margin: 0 0.25rem 0.25rem 0;
 
     &_visible {
         opacity: 1;
-        scale: 1;
+        transform: scale(1);
     }
 }
 
 .more {
-    inline-size: 5rem;
     margin: 0 0.25rem 0.25rem 0;
 }
--- a/projects/demo/src/modules/components/items-with-more/index.html
+++ b/projects/demo/src/modules/components/items-with-more/index.html
@@ -22,20 +22,21 @@
         <tui-doc-demo>
             <tui-items-with-more
                 [itemsLimit]="itemsLimit"
+                [linesLimit]="linesLimit"
                 [required]="required"
                 [side]="side"
             >
                 <ng-container *ngFor="let item of items">
                     <tui-chip
                         *tuiItem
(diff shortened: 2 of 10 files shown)


projects/demo/src/modules/components/items-with-more/examples/4/index.less
```diff
@@ -0,0 +1,33 @@
...
+        max-block-size: 100%;
+    }
+}
+
+.item {
+    .transition(~'scale, opacity');
+
+    opacity: 0;
+    scale: 0.9;
+    margin: 0 0.25rem 0.25rem 0;
```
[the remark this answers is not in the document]
vladimirpotekhin: No, unfortunately. The items-with-more work due to the width calculation. It doesn't work with gaps (

projects/kit/components/items-with-more/items-with-more.service.ts
```diff
@@ -19,7 +19,11 @@ export class TuiItemsWithMoreService extends Observable<number> {
         inject(ResizeObserverService, {self: true}),
     ).pipe(
         debounceTime(0, tuiZonefreeScheduler()),
-        map(() => this.getOverflowIndex()),
+        map(() =>
+            this.directive.linesLimit > 1
+                ? this.getMultilineOverflowIndex()
+                : this.getOverflowIndex(this.el.children),
```
waterplea, line 25: Let's use Array.from so that we do not have this duality in types

projects/kit/components/items-with-more/items-with-more.service.ts
```diff
@@ -89,4 +93,19 @@ export class TuiItemsWithMoreService extends Observable<number> {
 
         return -1;
     }
+
+    private getMultilineOverflowIndex(): number {
```
waterplea, line 97: Let's rename to `getOverflowIndexMultiline` so that you can tell at a glance they are two versions of the same thing

projects/kit/components/items-with-more/items-with-more.service.ts
```diff
@@ -89,4 +93,19 @@ export class TuiItemsWithMoreService extends Observable<number> {
...
+        const {children} = this.el;
+        const {linesLimit} = this.directive;
+        const items = Array.from(children) as HTMLElement[];
+        const rows = new Set(items.map((item) => item.offsetTop));
+        const lastRowOffset = Array.from(rows)[linesLimit - 1];
+
+        const firstItemLastRow = items.findIndex(
```
waterplea, line 104: nit: if we get those names shorter we can get rid of extra lines and spaces :)

projects/kit/components/items-with-more/items-with-more.directive.ts
```diff
@@ -1,24 +1,45 @@
...
+
+    public get side(): 'end' | 'start' {
+        return this.linesLimit > 1 ? 'end' : this.moreSide;
+    }
+
     public ngOnChanges(): void {
         this.change$.next();
     }
+
+    protected maxWidth(): number {
```
waterplea, line 42: Would this cause some unexpected spaces if there's a huge item that is way down the list and is hidden?

waterplea (approved):

projects/kit/components/items-with-more/items-with-more.directive.ts
```diff
@@ -1,24 +1,45 @@
...
     @Input()
-    public side: 'end' | 'start' = 'end';
+    public linesLimit = 1;
 
     // TODO: refactor to signal inputs after Angular update
     public readonly change$ = new Subject<void>();
 
+    @Input('side')
+    public set setSide(side: 'end' | 'start') {
+        this.moreSide = side;
```
waterplea, line 31: This is messy, I think we only need it in one place in the service, let's just check if we have multiline there?

projects/demo/src/modules/components/items-with-more/index.html
```diff
@@ -22,22 +22,23 @@
         <tui-doc-demo>
             <tui-items-with-more
                 [itemsLimit]="itemsLimit"
```
mdlufy, line 24: In the future, we may consider renaming to `max`/`limit` or  `maxLines`/`linesMax`. Just thinking out loud

projects/kit/components/items-with-more/items-with-more.service.ts
```diff
@@ -89,4 +93,17 @@ export class TuiItemsWithMoreService extends Observable<number> {
 
         return -1;
     }
+
+    private getOverflowIndexMultiline(): number {
```
mdlufy, line 97: It seems to me that it is possible to improve readability by adding spaces between lines)

mdlufy (approved):

PR #10651, opened 2025-03-24
chore(demo): add button group example with elastic container

splincode:


projects/demo/src/modules/components/button-group/examples/4/index.less
```diff
@@ -0,0 +1,32 @@
+@import '@taiga-ui/core/styles/taiga-ui-local';
+
+:host {
+    display: flex;
+    flex-direction: column;
+    gap: 1rem;
+    inline-size: 21.5625rem;
```
waterplea, line 7: Let's round those two

projects/demo/src/modules/components/button-group/examples/4/index.less
```diff
@@ -0,0 +1,32 @@
...
+    block-size: 21.875rem;
+    interpolate-size: allow-keywords;
```
waterplea, line 9: Unnecessary

waterplea (approved):

projects/demo/src/modules/components/button-group/examples/4/index.less
splincode: fixed

projects/demo/src/modules/components/button-group/examples/4/index.less
splincode: fixed

projects/kit/directives/button-group/button-group.style.less
```diff
@@ -1,13 +1,27 @@
 @import '@taiga-ui/core/styles/taiga-ui-local';
 
 [tuiButtonGroup] {
-    display: flex;
-    border-radius: var(--tui-radius-xl);
-    justify-content: center;
-    overflow: hidden;
+    &,
+    tui-elastic-container {
+        transition: all calc(var(--tui-duration-moderate, 300ms) / 3)
```
mdlufy, line 6: Maybe use `.transition(~'background, height, border-radius')` and redefine other props, such `transition-timing-function` and `transition-duration` ?

mdlufy (approved):

projects/kit/directives/button-group/button-group.style.less
waterplea: The mixin makes no sense when you redefine everything :)

mdlufy: I thought some properties from the mixin would be useful to us
Okey, no need mixin, if we redefine all

PR #10659, opened 2025-03-25
feat(kit): add `shift` output event to `carousel` component

splincode:
https://github.com/user-attachments/assets/97a34a5e-6830-4767-9ef0-89bdb447ac65

https://github.com/user-attachments/assets/0652668c-7451-4cf2-9711-48c699332acc

### DIFF
--- a/projects/kit/components/carousel/carousel.component.ts
+++ b/projects/kit/components/carousel/carousel.component.ts
@@ -105,7 +105,7 @@ export class TuiCarouselComponent {
     }
 
     protected get transform(): string {
-        return `translateX(${100 * this.getX()}%)`;
+        return `translateX(${100 * this.x}%)`;
     }
 
     @tuiPure
@@ -117,10 +117,6 @@ export class TuiCarouselComponent {
             minWidth: percent,
             maxWidth: percent,
         };
-    }
-
-    protected getX(): number {
-        return this.transitioned ? this.computedTranslate : this.translate;
     }
 
     protected onTransitioned(transitioned: boolean): void {
@@ -181,7 +177,11 @@ export class TuiCarouselComponent {
     }
 
     protected onShift(): void {
-        this.shift.emit(Math.abs(100 * this.getX()) - 100 * this.index);
+        this.shift.emit(Math.abs((this.x % 1) + 0.5) * 2);
+    }
+
+    private get x(): number {
+        return this.transitioned ? this.computedTranslate : this.translate;
     }
 
     private get computedTranslate(): number {
--- a/projects/kit/components/carousel/carousel.component.ts
+++ b/projects/kit/components/carousel/carousel.component.ts
@@ -83,6 +83,9 @@ export class TuiCarouselComponent {
     @Output()
     public readonly indexChange = new EventEmitter<number>();
 
+    @Output()
+    public readonly shift = new EventEmitter<number>();
+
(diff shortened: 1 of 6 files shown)


projects/kit/directives/button-group/button-group.style.less
```diff
@@ -18,6 +18,8 @@
 
     tui-elastic-container {
         inline-size: 100%;
+        transition: opacity calc(var(--tui-duration-moderate, 300ms) / 3)
```
waterplea, line 21: We should remove transition from opacity so that it does not lag when you drag with your finger, besides it breaks height transition.

projects/kit/components/carousel/carousel.component.ts
```diff
@@ -83,6 +85,9 @@ export class TuiCarouselComponent {
     @Output()
     public readonly indexChange = new EventEmitter<number>();
 
+    @Output()
+    public readonly shifted = new EventEmitter<number>();
```
waterplea, line 89: nit: let's call it `shift` as in `click` is the name of the event, not `clicked`

projects/kit/components/carousel/carousel.component.ts
```diff
@@ -54,9 +54,11 @@ import {TuiCarouselScroll} from './carousel-scroll.directive';
         '[class._transitioned]': 'transitioned',
         '[class._draggable]': 'draggable',
         '(touchstart)': 'onTransitioned(false)',
+        '(document:touchmove)': 'transitioned ? (0) : shift()',
         '(touchend)': 'onTransitioned(true)',
+        '(document:mousemove)': 'transitioned ? (0) : shift()',
         '(mousedown)': 'onTransitioned(false)',
-        '(document:mouseup.zoneless)': 'onTransitioned(true)',
+        '(document:mouseup.zoneless)': 'onTransitioned(true);',
```
waterplea, line 61: I think we do not really need those event changes, we can just emit values inside `onPan` and `onTransitioned`

projects/kit/directives/button-group/button-group.style.less
splincode: fixed

projects/kit/components/carousel/carousel.component.ts
splincode: fixed

projects/kit/components/carousel/carousel.component.ts
splincode: fixed

waterplea (approved):
Looks like we do not yet have the various duration variables in open source theme. Let's change it to our regular `.transition(~'...')` mixin for now and later we will rework our animations altogether so that in v5 we will have different durations and easing as per the specs everywhere.

projects/kit/components/carousel/carousel.component.ts
```diff
@@ -171,6 +176,14 @@ export class TuiCarouselComponent {
         this.updateIndex(this.index === this.items.length - 1 ? 0 : this.index + 1);
     }
 
+    protected onShift(): void {
+        this.shift.emit(Math.abs((this.x % 1) + 0.5) * 2);
+    }
+
+    private get x(): number {
```
mdlufy, line 183: Mathematical equation.. find `x` :)

mdlufy (approved):

PR #10664, opened 2025-03-26
chore(core): refactor some logic inside new version of `TuiOption`

nsbarsukov:


projects/core/components/data-list/data-list.style.less
```diff
@@ -74,6 +74,10 @@ tui-data-list {
         margin: var(--tui-data-list-margin) 0;
     }
 
+    [tuiOption]:not([new]) {
+        justify-content: space-between;
+    }
```
waterplea, line 79: Also `gap`? It should probably depend on the size too.

waterplea (approved):

mdlufy (approved):

projects/core/components/data-list/data-list.style.less
nsbarsukov: It will be added later

projects/core/components/data-list/data-list.style.less
```diff
@@ -74,6 +74,10 @@ tui-data-list {
         margin: var(--tui-data-list-margin) 0;
     }
 
+    [tuiOption]:not([new]) {
+        justify-content: space-between;
```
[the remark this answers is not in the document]
nsbarsukov: https://github.com/taiga-family/taiga-ui/pull/10664/commits/12726b395bbecf042a47a4aa8d54174f99b0154b

PR #10671, opened 2025-03-26
feat(layout): `ChipGroup` add new component

vladimirpotekhin:
Fixes #


splincode (approved):

projects/layout/components/chip-group/chip-group.styles.less
```diff
@@ -0,0 +1,40 @@
+@import '@taiga-ui/core/styles/taiga-ui-local';
+
+tui-chip-group,
+[tuiChipGroup] {
```
waterplea, line 4: Let's only have the attribute selector. I would like to remove this duality everywhere in next major version

projects/layout/components/chip-group/chip-group.styles.less
```diff
@@ -0,0 +1,40 @@
...
+    display: flex;
+    flex-wrap: wrap;
+
+    tui-chip,
```
waterplea, line 8: Can we change it to direct children? Might be helpful for thumbnail card, badges, not just chips. Maybe even rename it to `ItemGroup` as in `ItemsWithMore`?

projects/layout/components/chip-group/chip-group.directive.ts
```diff
@@ -0,0 +1,61 @@
...
+    protected readonly initialized = signal(false);
+
+    @Input()
+    public horizontal = false;
+
+    @Input()
+    public autoscroll = false;
+
+    public ngAfterViewInit(): void {
+        this.initialized.set(true);
```
waterplea, line 45: Maybe el.addClass so that we do not need to trigger change detection one more time?

projects/demo/src/modules/app/pages.ts
```diff
@@ -1002,10 +1002,17 @@ export const pages: DocRoutePages = [
...
+        keywords: 'chip, tag, тэг, badge',
         route: DemoRoute.Chip,
         meta: {},
     },
+    {
+        section: 'Components',
+        title: 'ItemGroup',
+        keywords: 'item, chip, group, list, tag, тэг, badge',
+        route: DemoRoute.ItemGroup,
+        meta: {},
```
waterplea, line 1014: name: 'tui-chip-group'?

projects/layout/components/index.ts
```diff
@@ -3,6 +3,7 @@ export * from '@taiga-ui/layout/components/block-details';
 export * from '@taiga-ui/layout/components/block-status';
 export * from '@taiga-ui/layout/components/card';
 export * from '@taiga-ui/layout/components/cell';
+export * from 'projects/layout/components/item-group';
```
waterplea, line 6: Oops

waterplea (approved):

PR #10689, opened 2025-03-28
fix(kit): ignore transition directly on tile

### LINKED ISSUE #10635: 🐞 - `Tiles` nested tiles do not work
### Reproduction url

https://stackblitz.com/edit/angular-catyxxap

### Description

For some reason there are issues with assigning internal '_dragged' class or something like that so the transform happens with transition when it shouldn't so it lags when you drag nested items

### Taiga UI version

4

### Browser Used

Chrome

### OS Used

macOS

splincode:
Fixes #10635

### DIFF
--- a/projects/kit/components/tiles/tile.service.ts
+++ b/projects/kit/components/tiles/tile.service.ts
@@ -98,8 +98,6 @@ export class TuiTileService implements OnDestroy {
 
         style.removeProperty('position');
         style.removeProperty('transition');
-
-        // caretaker note: I don't know why, but it's working now
         style.removeProperty('top');
         style.removeProperty('left');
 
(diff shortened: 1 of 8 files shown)


waterplea (changes requested):
This way when I let go of the tile I am dragging it snaps to its place without transition

waterplea (approved):
2 comments on the last example:
1. When we drag parent let's not add shadow to the children
2. When we drag children let's put z-index: 1 on the parent so that children are above next parent when you drag around

projects/kit/components/tiles/tile.service.ts
```diff
@@ -96,6 +97,12 @@ export class TuiTileService implements OnDestroy {
         const host = this.el.getBoundingClientRect();
 
         style.removeProperty('position');
+        style.removeProperty('transition');
+
+        // caretaker note: I don't know why, but it's working now
```
vladimirpotekhin, line 102: unhelpful comment

PR #10708, opened 2025-03-31
feat(kit): `TuiElastic` add directive 

mdlufy:
feat(kit): `TuiElastic` add directive


projects/kit/components/elastic-container/elastic.directive.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    constructor() {
+        effect((onCleanup) => {
+            this.mutations();
+
+            const previous = this.el.clientHeight;
+
+            this.renderer.setStyle(this.el, 'height', 'auto');
+
+            this.outerId = requestAnimationFrame(() => {
```
waterplea, line 47: I'm not sure you need rAF, you need to just trigger CSS reflow, I think getting clientHeight should be enough to apply auto height or maybe getBoundingClientRect()

PR #10709, opened 2025-03-31
feat(core): `Dialog` add event handler to change close button theme

mdlufy:
Fixes #


waterplea (approved):
Are we sure we need it? I know it's v3 but that's still quite dirty

splincode (approved):

mdlufy:
> Are we sure we need it? I know it's v3 but that's still quite dirty

Yep.. But are there another way to control close button appearance? We need change it, when works with dark image in dialog, otherwise, the button will not be visible

PR #10715, opened 2025-04-01
fix(core): do not use hardcoded `tui-hint` selector

splincode:


projects/core/directives/hint/hint.component.ts
```diff
@@ -104,7 +104,8 @@ export class TuiHintComponent<C = any> {
 
     protected onClick(target: HTMLElement): void {
         if (
-            (!target.closest('tui-hint') && !this.hint.el.contains(target)) ||
+            (!target.closest(this.el.tagName.toLowerCase()) &&
```
waterplea, line 107: It's case insensitive, no need for lowercasing

projects/kit/components/line-clamp/line-clamp-box.component.ts
```diff
@@ -5,7 +5,6 @@ import {PolymorpheusOutlet} from '@taiga-ui/polymorpheus';
 
 @Component({
     standalone: true,
-    selector: 'tui-hint',
```
waterplea, line 8: You still need some selector, otherwise it will be `ng-component` which we don't want

projects/core/directives/hint/hint.component.ts
splincode: fixed

vladimirpotekhin (approved):

waterplea (approved):

PR #10727, opened 2025-04-02
feat(addon-mobile) `SheetDialog` pass context and correct fullsize

mdlufy:
Fixes

### DIFF
--- a/projects/addon-mobile/components/sheet-dialog/sheet-dialog.style.less
+++ b/projects/addon-mobile/components/sheet-dialog/sheet-dialog.style.less
@@ -26,12 +26,13 @@
         .t-sheet {
             display: flex;
             flex-direction: column;
-            min-block-size: calc(100% - 1rem);
+            block-size: calc(100% - 1rem);
         }
 
         // TODO: make default in v5
         .t-content {
             display: flex;
+            block-size: 100%;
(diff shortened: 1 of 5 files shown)


waterplea (approved):

projects/addon-mobile/components/sheet-dialog/sheet-dialog.style.less
```diff
@@ -26,12 +26,13 @@
         .t-sheet {
             display: flex;
             flex-direction: column;
-            min-block-size: calc(100% - 1rem);
```
waterplea, line 29: Why this change? This would cause issues with long content
image

mdlufy: Hmm, I want children to work `flex: 1 0 50%` for onboarding, so all parents must have concrete height. I rely for `--tui-visual-viewport` for this purpose in v4 and i think to use flex instead of porting VisualViewport service to v3

mdlufy: It's seems like we need just add overflow-y: scroll, I'll take a closer look

splincode (approved):

PR #10746, opened 2025-04-04
feat(kit): `Slider` supports not equidistant `segments` order

### LINKED ISSUE #5299: 🚀 - `Slider` allows to set special ticks/segments order (not only equidistant)
### Description

`[segments]` accepts `number | number[]`

```html
<input
    tuiSlider
    type="range"
    [segments]="[0.2, 0.6, 0.8]"
/>
```

Every number in array - is ratio of slider's track length where to put tick.

LarisaGl:
Fixes #5299

### DIFF
--- a/projects/kit/components/slider/slider.component.ts
+++ b/projects/kit/components/slider/slider.component.ts
@@ -1,4 +1,11 @@
-import {ChangeDetectionStrategy, Component, inject, Input} from '@angular/core';
+import {
+    ChangeDetectionStrategy,
+    Component,
+    computed,
+    inject,
+    Input,
+    signal,
+} from '@angular/core';
 import {NgControl, NgModel} from '@angular/forms';
 import {tuiWatch} from '@taiga-ui/cdk/observables';
 import {tuiInjectElement} from '@taiga-ui/cdk/utils/dom';
@@ -25,7 +32,7 @@ import {TUI_SLIDER_OPTIONS} from './slider.options';
          */
         '(input)': '0',
         '[style.--tui-slider-track-color]': 'options.trackColor',
-        '[style.--tui-ticks-gradient]': 'ticksGradient',
+        '[style.--tui-ticks-gradient]': 'ticksGradient()',
         '[style.--tui-slider-fill-ratio]': 'valueRatio',
         '[attr.data-size]': 'size',
     },
@@ -34,12 +41,13 @@ export class TuiSliderComponent {
     private readonly control = inject(NgControl, {self: true, optional: true});
 
     protected readonly options = inject(TUI_SLIDER_OPTIONS);
+    protected readonly segments = signal<number[]>([1]);
+    protected readonly ticksGradient = computed((segments = this.segments()) =>
+        this.getTicksGradient(segments),
+    );
 
     @Input()
     public size: TuiSizeS = this.options.size;
-
-    @Input()
-    public segments: number[] | number = 1;
 
     public readonly el = tuiInjectElement<HTMLInputElement>();
     public readonly keySteps = inject(TuiSliderKeyStepsBase, {
@@ -58,6 +66,29 @@ export class TuiSliderComponent {
              */
             this.control.valueChanges?.pipe(tuiWatch(), take(1)).subscribe();
         }
+    }
+
+    // TODO(v5): use signal inputs
+    @Input({
+        alias: 'segments',
+        transform: (segments: number[] | number) => {
+            if (Array.isArray(segments)) {
+                const arr = segments.map((segment) => segment * 100);
+
+                if (arr[0] !== 0) {
+                    arr.unshift(0);
+                }
+
+                return arr;
+            }
+
+            return new Array(segments)
+                .fill(null)
+                .map((_, index) => (index * 100) / segments);
+        },
+    })
+    public set segmentsSetter(segments: number[]) {
+        this.segments.set(segments);
     }
 
     public get valueRatio(): number {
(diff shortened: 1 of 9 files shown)


waterplea:
Positioning seems a bit off on the screenshots

LarisaGl:
> Positioning seems a bit off on the screenshots

It's fixed. Issue was because of linear-gradients array from first decision needs an array for background-position-x
`
background-position-x: 0, @first-tick-offset, ..., @first-tick-offset, 0;
`
 and background-size too
`
background-size: @ticks-background-size, @ticks-background-size, ..., @ticks-background-size, auto;
`
But length of linear-gradients array is dynamic (css var is needed) and functions from less (range, each, loop, length...) don't work with dynamic values (css vars) :( There is an issue on Github. So had to change it to one linear-gradient. Not best decision, as I think. Make review, please, waiting for comments

projects/kit/components/slider/slider.component.ts
```diff
@@ -104,7 +104,41 @@ export class TuiSliderComponent {
         this.el.value = `${newValue}`;
     }
 
-    protected get segmentWidth(): number {
-        return 100 / Math.max(1, this.segments);
+    protected get ticksGradient(): string {
+        const segments = Array.isArray(this.segments)
+            ? this.segments.map((segment) => (segment / this.max) * 100)
```
nsbarsukov, line 109: After solving my previous comment this getter will become dependent only on `this.segments` (`this.max` will not be required).

Probably we can use `computed` to memoise all logic of this getter.
```ts
protected readonly segments = signal<number[]>([1]);
protected readonly ticksGradient = computed((segments = this.segments()) => {
    // ...
});

// [...]

// TODO(v5): use signal inputs
@Input({
    alias: 'segments',
    transform: (x: number[] | number) =>
        Array.isArray(x) ? x : new Array(x).fill(1).map((x) => 1 / x),
})
public set segmentsSetter(x: number[]) {
    this.segments.set(x);
}
```

projects/kit/components/slider/slider.component.ts
```diff
@@ -104,7 +104,41 @@ export class TuiSliderComponent {
...
+            : new Array(Math.max(1, this.segments))
+                  .fill(1)
+                  .map(
+                      (_, index) =>
+                          this.getSegmentWidth(this.segments as number) * (index + 1),
+                  );
+
+        if (segments.length === 1) {
```
nsbarsukov, line 117:
```suggestion
        if (segments.length <= 1) {
```

projects/kit/components/slider/slider.component.ts
```diff
@@ -104,7 +104,41 @@ export class TuiSliderComponent {
...
+            }
+
+            gradient = `
+                ${gradient}
+                var(--tui-text-tertiary) ${segment}% calc(${segment}% + var(--tui-ticks-thickness)),
+                transparent ${segment}% ${segments[index + 1]}%${index + 1 === segments.length - 1 ? ')' : ','}
+            `;
+        });
+
+        return gradient;
```
nsbarsukov, line 138:
```ts
return segments.slice(1).reduce(
    (acc, segment, index) =>
        `${acc}
        var(--tui-text-tertiary) ${segment}% calc(${segment}% + var(--tui-ticks-thickness)),
        transparent ${segment}% ${segments[index + 1]}%${index + 1 === segments.length - 1 ? ')' : ','}
    `,
    `linear-gradient(to right, transparent ${segments[0]}%`,
);
```

projects/kit/components/slider/slider.style.less
```diff
@@ -130,6 +125,8 @@
 }
 
 :host {
+    --tui-ticks-thickness: @ticks-thickness;
```
nsbarsukov, line 128: Let's use `--t-tick-thickness` (we use `--tui-***`-prefix for public variables and `--t-**` for private API).
I think `--t-tick-thickness` should be private one (for the 1st iteration, at least)

nsbarsukov: \+ use `--t-tick-thickness` everywhere inside this `less`-file (instead of static `less`-variable `@ticks-thickness`)

projects/kit/components/slider/slider.component.ts
LarisaGl: Thanks for comments! Interesting and a bit challenging task :) I have fixed all comments with little changes, make review, please

projects/kit/components/slider/slider.component.ts
```diff
@@ -60,6 +68,29 @@ export class TuiSliderComponent {
...
+                if (arr[0] !== 0) {
+                    arr.unshift(0);
+                }
+
+                return arr;
+            }
+
+            return new Array(segments)
+                .fill(null)
+                .map((_, index) => (index * 100) / segments);
```
nsbarsukov, line 87: I am not sure that `@Input`'s transform should be so huge 🤔 

Let's just transform single `number` to `number[]` (with ratios, not percentages).
```ts
@Input({
    alias: 'segments',
    transform: (x: number[] | number) =>
        Array.isArray(x) ? x : new Array(x).fill(null).map((_, i) => i / x),
})
public set segmentsSetter(segments: number[]) {
```

projects/kit/components/slider/slider.component.ts
```diff
@@ -60,6 +68,29 @@ export class TuiSliderComponent {
...
 
+    // TODO(v5): use signal inputs
+    @Input({
+        alias: 'segments',
+        transform: (segments: number[] | number) => {
+            if (Array.isArray(segments)) {
+                const arr = segments.map((segment) => segment * 100);
+
+                if (arr[0] !== 0) {
+                    arr.unshift(0);
```
nsbarsukov, line 79: Do we actually need to mutate array ?
I think we could drop this logic (and process initial zero inside `getTicksGradient`).

Also, let's update documentation
```diff
- length in range [0; 1]
+ length in range (0; 1)
```

Passing `0` & `1` (as a position for the tick) seems senseless anyway

projects/kit/components/slider/slider.component.ts
```diff
@@ -104,7 +135,18 @@ export class TuiSliderComponent {
...
+            return 'linear-gradient(to right, transparent 0 100%)';
+        }
+
+        return segments.slice(1).reduce(
+            (acc, segment, index, array) =>
+                `${acc}
+                var(--tui-text-tertiary) ${segment}% calc(${segment}% + var(--t-tick-thickness)),
+                transparent ${segment}% ${array[index + 1] ?? 100}%${array[index + 1] ? ',' : ')'}
+                `,
+            `linear-gradient(to right, transparent ${segments[0]}% ${segments[1]}%,`,
```
nsbarsukov, line 149: Should it be `0` and `segments[0]` (if we change range to `(0;1)` according to the previous comment) ?

```suggestion
            `linear-gradient(to right, transparent 0 ${segments[0]}%,`,
```

projects/kit/components/slider/slider.style.less
```diff
@@ -29,18 +28,13 @@
...
-    @first-tick-offset: ((@thumb-width - @ticks-thickness) / 2);
+    @first-tick-offset: calc((@thumb-width - var(--t-tick-thickness)) / 2);
     @ticks-background-size: calc(100% - @thumb-width);
 
     @hide-first-tick: linear-gradient(
         to right,
-        @thumb-color 0px (@first-tick-offset + @ticks-thickness),
-        transparent @ticks-thickness
-    );
-    @ticks-gradient: repeating-linear-gradient(
```
nsbarsukov, line 40: I wonder if `@hide-first-tick` is required if `repeating-linear-gradient` was replaced by `linear-gradient` ? 🤔

nsbarsukov (approved):

projects/kit/components/slider/slider.component.ts
```diff
@@ -34,13 +41,14 @@ export class TuiSliderComponent {
     private readonly control = inject(NgControl, {self: true, optional: true});
 
     protected readonly options = inject(TUI_SLIDER_OPTIONS);
+    protected readonly segments = signal<number[]>([1]);
+    protected readonly ticksGradient = computed((segments = this.segments()) =>
+        this.getTicksGradient(segments),
```
waterplea, line 46: can just be `this.getTicksGradient(this.segments())`. We move signals up that way if they are used multiple times or if we want to make sure they are definitely read so that computed depends on them.

waterplea (approved):

PR #10760, opened 2025-04-07
fix(core): `InputMonth` should hide Android text select handle

nsbarsukov:
https://taiga-ui.dev/components/input-month

android-text-select-handle


projects/core/components/textfield/select-like.directive.ts
```diff
@@ -8,6 +8,7 @@ import {Directive} from '@angular/core';
         // Click on cleaner icon does not trigger `beforeinput` event --> handle all kind of deletion in input event
         '(beforeinput)': '$event.inputType.includes("delete") || $event.preventDefault()',
         '(input.capture)': '$event.inputType?.includes("delete") && clear($event.target)',
+        '(pointerdown.prevent)': '0', // Hide Android text select handle (bubble marker below transparent caret)
```
waterplea, line 11: Wouldn't that also prevent focus?

waterplea (approved):

splincode (approved):

PR #10761, opened 2025-04-07
fix(core): `Textfield` has invalid width for `[tuiChevron]`'s icon

nsbarsukov:
### Previous behavior
before

### New behavior
after

### DIFF
--- a/projects/kit/directives/chevron/chevron.style.less
+++ b/projects/kit/directives/chevron/chevron.style.less
@@ -8,7 +8,20 @@ tui-icon[tuiChevron]::after {
 }
 
 [tuiChevron][tuiIcons]::after {
-    font-size: 1rem;
+    block-size: 1rem;
+    inline-size: 1rem;
+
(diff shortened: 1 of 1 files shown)


waterplea:
Icon is now too close to the right edge, needs some `margin-inline-end` added. The distance to all 3 edges is supposed to be the same:
image

splincode (approved):

nsbarsukov:
@waterplea updated (`margin-inline-end` was added according to the spec)

waterplea:
> @waterplea updated (`margin-inline-end` was added according to the spec)

Some other things got broken now, looks like we only need it in `tui-textfield` + fix input-phone

nsbarsukov:
> > @waterplea updated (`margin-inline-end` was added according to the spec)
> 
> Some other things got broken now, looks like we only need it in `tui-textfield` + fix input-phone

fixed

waterplea (approved):

PR #10767, opened 2025-04-08
refactor(core): add `autocomplete: 'off'` for `TuiSelectLike`

nsbarsukov:


projects/core/components/textfield/select-like.directive.ts
```diff
@@ -5,10 +5,16 @@ import {TUI_IS_ANDROID} from '@taiga-ui/cdk/tokens';
     standalone: true,
     host: {
         inputmode: 'none',
+        autocomplete: 'off',
```
nsbarsukov, line 8: Is it better to use `autocomplete: new-password` ? 🤔 

https://taiga-ui.dev/components/input#autofill

waterplea (approved):

vladimirpotekhin (approved):

PR #10774, opened 2025-04-09
fix(legacy): reuse `identityMatcher` inside `tuiMultiSelectGroup`

### LINKED ISSUE #10721: 🐞 - identityMatcher doesnot work in tuiMultiSelectGroup
### Reproduction url

https://stackblitz.com/edit/angular-rthgbrve?file=src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.template.html

### Description

if we have complex object in controk,  tui-data-list-wrapper with tuiMultiSelectGroup in template and provide identityMatcher in providers, the checkbox does not active in multi select

### Taiga UI version

4.31

### Browser Used

Chrome

### OS Used

macOS

splincode:
Fixes #10721

### DIFF
--- a/projects/legacy/components/multi-select/multi-select-group/multi-select-group.directive.ts
+++ b/projects/legacy/components/multi-select/multi-select-group/multi-select-group.directive.ts
@@ -7,6 +7,7 @@ import {
     TUI_DATA_LIST_HOST,
     tuiAsOptionContent,
 } from '@taiga-ui/core/components/data-list';
+import {TUI_ITEMS_HANDLERS} from '@taiga-ui/kit/tokens';
 import {AbstractTuiControl} from '@taiga-ui/legacy/classes';
 import {TuiMultiSelectOptionComponent} from '@taiga-ui/legacy/components/multi-select-option';
 import {PolymorpheusComponent} from '@taiga-ui/polymorpheus';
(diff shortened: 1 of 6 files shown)


projects/legacy/components/multi-select/multi-select-group/multi-select-group.directive.ts
```diff
@@ -28,16 +29,24 @@ export const TUI_MULTI_SELECT_OPTION = new PolymorpheusComponent(
...
+                const {identityMatcher, stringify} =
+                    inject(TUI_ITEMS_HANDLERS, {optional: true}) ?? {};
 
                 return (
                     multiSelect || {
+                        stringify,
+                        identityMatcher,
                         handleOption: (option) => {
+                            const state = tuiArrayToggle(
+                                host ? host.value : ngControl.control?.value || [],
```
waterplea, line 41: Looks like we should make a const for this line:
```ts
const value = host?.value ?? ngControl.control?.value ?? [];
```

waterplea (approved):

projects/legacy/components/multi-select/multi-select-group/multi-select-group.directive.ts
splincode: fixed

vladimirpotekhin (approved):

PR #10776, opened 2025-04-09
fix(addon-mobile): `SheetDialog` add `env(safe-area-inset-top)`

### LINKED ISSUE #10753: 🚀 - add safe-area-inset-top in tuiSheetDialog offset
### Description

When using tuiSheetDialog the safe top margin is not respected on fullscreen webview

Please, add default safe-area-inset-top in offset or add the ability to set offset by string

mdlufy:
Part of #10963 

Fixes #10753

### DIFF
--- a/projects/demo/src/modules/components/sheet-dialog/examples/4/index.less
+++ b/projects/demo/src/modules/components/sheet-dialog/examples/4/index.less
@@ -26,7 +26,7 @@
     gap: 1rem;
     text-align: center;
     margin: 0 -1rem -1rem;
-    padding: 1rem;
+    padding: 1rem 1rem calc(1rem + env(safe-area-inset-bottom));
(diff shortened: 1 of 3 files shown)


projects/addon-mobile/components/sheet-dialog/sheet-dialog.style.less
```diff
@@ -116,7 +116,7 @@
 .t-content {
     position: relative;
     isolation: isolate;
-    padding-block-end: calc(1.5rem + env(safe-area-inset-bottom));
```
waterplea, line 119: You need to rebase, pre-commit hook combined two paddings into one.

Also we only need this in fullscreen mode, in regular mode we need to add top padding to the offset of the sheet, not to the padding of its content.

waterplea (approved):

PR #10777, opened 2025-04-09
fix(kit): navigate to active element after first render for `tuiStepper`

### LINKED ISSUE #10737: 🐞 - `Stepper` error setting initial index
### Reproduction url

https://stackblitz.com/edit/angular-gih7qray

### Description

Initial setting of the `activeItemIndex` doesn't work, starts working if set with a delay

### Taiga UI version

4

### Browser Used

Chrome

### OS Used

macOS

splincode:
Fixes #10737

### DIFF
--- a/projects/kit/components/stepper/stepper.component.ts
+++ b/projects/kit/components/stepper/stepper.component.ts
@@ -62,7 +62,7 @@ export class TuiStepperComponent implements OnChanges {
         .pipe(takeUntilDestroyed())
         .subscribe(() => this.scrollIntoView(this.activeItemIndex));
 
-    @Input('activeItemIndex')
+    @Input()
     public activeItemIndex = 0;
 
     @Input()
--- a/projects/kit/components/stepper/stepper.component.ts
+++ b/projects/kit/components/stepper/stepper.component.ts
@@ -1,5 +1,5 @@
 import {AsyncPipe, NgIf} from '@angular/common';
-import type {QueryList} from '@angular/core';
+import type {OnChanges, QueryList} from '@angular/core';
 import {
     ChangeDetectionStrategy,
     ChangeDetectorRef,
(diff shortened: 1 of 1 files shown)


projects/kit/components/stepper/stepper.component.ts
```diff
@@ -72,12 +73,20 @@ export class TuiStepperComponent {
 
     @Input('activeItemIndex')
     public set activeIndex(index: number) {
+        this.itemIndex = index;
+
         if (this.steps.get(index)?.nativeElement) {
             this.activeItemIndex = index;
```
waterplea, line 79: Why do we only update internal index if there is such a step? That would cause out of sync situations when external index is A and internally we have B. Shouldn't we update it all the time and scroll to the element only when it's there?

Also what if we change from setter to `ngOnChanges`? Would content children be processed by then?

splincode: fixed

projects/kit/components/stepper/stepper.component.ts
```diff
@@ -58,24 +58,21 @@ export class TuiStepperComponent {
     private readonly speed = inject(TUI_ANIMATIONS_SPEED);
     private readonly destroyRef = inject(DestroyRef);
 
-    protected activeItemIndex = 0;
-
     protected readonly $ = inject(ResizeObserverService, {self: true})
         .pipe(takeUntilDestroyed())
         .subscribe(() => this.scrollIntoView(this.activeItemIndex));
 
+    @Input('activeItemIndex')
```
waterplea, line 65: Unnecessary alias

waterplea (approved):

projects/kit/components/stepper/stepper.component.ts
splincode: fixed

vladimirpotekhin (approved):

PR #10781, opened 2025-04-09
fix(kit): take correct dimensions inside `tui-line-clamp`

### LINKED ISSUE #10714: 🐞 - `LineClamp` incorrect calculation of `overflownChange`
### Reproduction url

https://stackblitz.com/edit/angular-zlkukbve

### Description

Component reports `true` when there's enough space and entire content is shown

### Taiga UI version

4

### Browser Used

Chrome

### OS Used

macOS

splincode:
Fixes #10714

### DIFF
--- a/projects/kit/components/line-clamp/line-clamp.component.ts
+++ b/projects/kit/components/line-clamp/line-clamp.component.ts
@@ -38,9 +38,6 @@ import {TUI_LINE_CLAMP_OPTIONS} from './line-clamp.options';
 import {TuiLineClampBox} from './line-clamp-box.component';
 import {TuiLineClampPositionDirective} from './line-clamp-position.directive';
 
-// 4px buffer for IE/Edge incorrectly rounding scrollHeight
-const BUFFER = 4;
-
 @Component({
     standalone: true,
     selector: 'tui-line-clamp',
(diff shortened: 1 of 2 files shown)


waterplea (approved):

vladimirpotekhin (approved):

projects/kit/components/line-clamp/line-clamp.style.less
```diff
@@ -18,4 +18,9 @@
     -webkit-box-orient: vertical;
     overflow: hidden;
     overflow-wrap: anywhere;
+
+    &,
+    &::ng-deep > * {
```
waterplea, line 23: This should be unnecessary if we rollback last condition here:
https://github.com/taiga-family/taiga-ui/blob/main/projects/kit/components/line-clamp/line-clamp.component.ts#L137
Please see why it was added and if we can get rid of it.

splincode: fixed

mdlufy (approved):

projects/kit/components/line-clamp/line-clamp.component.ts
```diff
@@ -151,6 +156,15 @@ export class TuiLineClamp implements DoCheck, AfterViewInit {
             this.height.set(this.outlet.nativeElement.scrollHeight + BUFFER);
         }
 
-        this.maxHeight.set(this.lineHeight * this.linesLimit$.value);
+        this.maxHeight.set(
+            (parseInt(
+                (this.outlet &&
+                    this.w.getComputedStyle(this.outlet.nativeElement).lineHeight) ??
```
mdlufy, line 162: `getComputedStyle` trigger forced sync reflow all, maybe we can rely on element.style if need only inline style?

splincode: fixed

waterplea (approved):

PR #10782, opened 2025-04-09
feat(addon-mobile): `DropdownMobile` supports new `Select`

nsbarsukov:


projects/demo/src/modules/components/select/index.html
```diff
@@ -17,6 +17,13 @@
         </a>
         element.
 
+        <tui-doc-example
+            id="dropdown-mobile"
+            heading="With DropdownMobile"
+            [component]="99 | tuiComponent"
+            [content]="99 | tuiExample"
```
nsbarsukov, line 24: TODO: move it before **"Native picker with disabled option"**-example

projects/addon-mobile/directives/dropdown-mobile/dropdown-mobile.directive.ts
```diff
@@ -16,6 +18,13 @@ import {TuiDropdownMobileComponent} from './dropdown-mobile.component';
                     ? TuiDropdownMobileComponent
                     : inject(TUI_DROPDOWN_COMPONENT, {skipSelf: true}),
         },
+        {
+            provide: TUI_OPTION_CONTENT,
+            useFactory: () =>
+                inject(TUI_IS_MOBILE)
+                    ? TuiOptionMobile
```
nsbarsukov, line 25: Inject also `TuiTextfieldComponent` for backward compatibility for legacy controls

projects/demo/src/modules/components/select/examples/6/index.html
nsbarsukov, line 1: The new example is inserted as the 6th position.
All existing examples are just moved (`previousIndex + 1`); their contents are not changed.

projects/addon-mobile/directives/dropdown-mobile/option-mobile/option-mobile.component.ts
```diff
@@ -0,0 +1,56 @@
...
+    private readonly option = inject<TuiOptionWithValue<T>>(TuiOptionWithValue, {
+        optional: true,
+    });
+
+    private readonly control = tuiInjectAuxiliary<TuiControl<T>>(
+        (x) => x instanceof TuiControl,
+    );
+
+    protected readonly select = tuiInjectAuxiliary<TuiControl<T>>(
+        (x) => x instanceof TuiSelectDirective,
```
waterplea, line 39: I'm not sure we want to rely on that, it would also be the same in combo box and in ButtonSelect eventually

projects/addon-mobile/directives/dropdown-mobile/option-mobile/option-mobile.component.ts
```diff
@@ -0,0 +1,56 @@
...
+    );
+
+    protected readonly selected = computed(
```
waterplea, line 42: We also need to think about future multi select case

nsbarsukov: It will be one additional `isMultiSelect ? arr.some(...) : identityMatcher`, yep

waterplea (approved):

splincode (approved):

PR #10791, opened 2025-04-10
feat(addon-doc): customize fullscreen buttons

splincode:
Part of UIKIT-7806


vladimirpotekhin (approved):

waterplea (approved):
Is it a breaking change though? Should we make them optional before v5?

PR #10795, opened 2025-04-11
chore(demo): add example for dropdownOpen

splincode:


waterplea:
First example of dropdown open already implements let-close and had the issue reproducible

PR #10800, opened 2025-04-11
feat(addon-doc): hide fullscreen button when device doesn't support fullscreen API

splincode:
image


mdlufy (approved):

mdlufy:
Maybe add fullscreen for API page too? Just an idea)

waterplea (approved):

PR #10803, opened 2025-04-11
fix(legacy): return possibility to use any data type in `TUI_DATE_TIME_VALUE_TRANSFORMER`

### LINKED ISSUE #10770: 🐞 - `tui-input-date-time` does not display initial data if `TUI_DATE_TIME_VALUE_TRANSFORMER` is working with a certain type
### Reproduction url

https://stackblitz.com/edit/angular-4rjxcfec

### Description

I use `TUI_DATE_TIME_VALUE_TRANSFORMER` to convert internal Taiga UI date values ​​to native `Date` and back.

After upgrading from Taiga UI 2 to Taiga UI 4, the `tui-input-date-time` component, which uses the transformer, stopped displaying the initial value. However, after dynamic update (via `setValue`), the value is displayed.

This is most likely due to these lines of code:

https://github.com/taiga-family/taiga-ui/blob/6ddb93cc4ea7d4ad050bc31c6eed9a3f5c78f564/projects/legacy/components/input-date-time/input-date-time.component.ts#L172-L176

Accordingly, the data types for which this bug is repeated must be non-enumerable at all, or they must not have a first element. With a regular string (for example, if you convert to ISO, and not to `Date`) everything works fine. But with `Date` it breaks, since it doesn’t have the `0` property.

### Taiga UI version

4.30.0

### Browser Used

Chrome

### OS Used

macOS

Markiewic:
Fixes #10770

### DIFF
--- a/projects/legacy/components/input-date-time/input-date-time.component.ts
+++ b/projects/legacy/components/input-date-time/input-date-time.component.ts
@@ -169,7 +169,11 @@ export class TuiInputDateTimeComponent
             return;
         }
 
-        if (value?.[0]) {
+        const transformed = this.valueTransformer
+            ? this.valueTransformer.fromControlValue(value)
+            : value;
+
+        if (transformed?.[0]) {
             super.writeValue(value);
         } else {
             super.writeValue(null);
--- a/projects/legacy/components/input-date-time/test/input-date-time.component.spec.ts
+++ b/projects/legacy/components/input-date-time/test/input-date-time.component.spec.ts
@@ -37,7 +37,7 @@ describe('InputDateTime', () => {
         public dateTimeComponent!: TuiInputDateTimeComponent;
 
         public readonly control = new FormControl<
-            string | [TuiDay | null, TuiDay | null] | null
+            Date | string | [TuiDay | null, TuiDay | null] | null
         >([new TuiDay(2021, 6, 12), null]);
 
         public min: TuiDay | [TuiDay, TuiTime] = TUI_FIRST_DAY;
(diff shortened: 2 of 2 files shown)


projects/legacy/components/input-date-time/input-date-time.component.ts
```diff
@@ -169,7 +169,11 @@ export class TuiInputDateTimeComponent
             return;
         }
 
-        if (value?.[0]) {
+        const transformed = this.valueTransformer
+            ? this.valueTransformer.fromControlValue(value)
+            : value;
+
+        if (transformed?.[0]) {
```
Markiewic, line 176: @waterplea, as you said in issue:
> We should not use `value` in there as it is the control value before it was ran through transformer. It should be this as the first line of the `writeValue` method there:
> ```ts
> const transformed = this.valueTransformer ? this.valueTransformer.fromControlValue(value) : value
> ```
> And later in the method `value` should be replaced with `transformed`.

I understand that `value` needs to be discarded in this method after it passes through the transformer, but I can't do that because later value will be needed again to pass here:
```ts
if (transformed?.[0]) {
    super.writeValue(value);
} else {
    super.writeValue(null);
}
```
I can't pass `transformed` there because inside `super.writeValue` it passes through the transformer again.

waterplea: Yes, you are correct. I forgot about that second run of transformer. Your code is fine, except on line 168 there's `value === this.value` — it also has to be `transformed` there.

waterplea: @Markiewic please move that `const transformed` to the beginning of the method and update `value === this.value` to `transformed === this.value` and we're good to merge 👍

waterplea (approved):

splincode (approved):

splincode:
@Markiewic Hello, could you add backport to v3.x branch?

Markiewic:
@splincode, ok, I'll make it

Markiewic:
@splincode, there is no need to do this in v3, as everything works correctly there.

https://stackblitz.com/edit/angular-iszq8ybj

Issue #10809, opened 2025-04-13
📚 - tui-tiles documentation returns wrong info in setup


Farbaks:
### What is the affected URL?

https://taiga-ui.dev/components/tiles/Setup

### Description

When trying to use the tui-tiles component, I realised that there was no API to to show to possible inputs and outputs, and when I tried following the setup tabe, the wrong input was passed into the component there. It takes items instead of order.

### Browser Used

Chrome

### OS Used

macOS

Farbaks:
PS: I can work on this and add in the right documentation for the API tab

PR #10810, opened 2025-04-13
chore: add API tab for tiles and correct setup tab

Farbaks:
Fixes https://github.com/taiga-family/taiga-ui/issues/10809


projects/demo/src/modules/components/tiles/index.html
```diff
@@ -34,5 +34,58 @@
         />
     </ng-template>
 
+    <ng-template pageTab>
+        <tui-doc-documentation heading="<tui-tiles>">
```
waterplea, line 38: Please use new API components like the ones here:
https://github.com/taiga-family/taiga-ui/blob/main/projects/demo/src/modules/components/button/index.html#L33

waterplea (approved):

splincode (approved):

PR #10817, opened 2025-04-14
feat(addon-doc): add fullscreen mode on API page

splincode:
image


waterplea:
I don't think it makes much sense on API tab, do we need it?

splincode:
@waterplea 
image

waterplea:
I think let's drop it

PR #10825, opened 2025-04-14
feat(core): update `tuiLink` appearance by spec

splincode:
image

### DIFF
--- a/projects/core/styles/components/link.less
+++ b/projects/core/styles/components/link.less
@@ -22,8 +22,8 @@
     border: none;
     cursor: pointer;
     font: inherit;
-    color: inherit;
-    text-decoration: none dashed currentColor;
+    color: var(--tui-text-primary);
+    text-decoration: underline solid currentColor;
     text-underline-offset: 0.2em;
     text-decoration-thickness: 0.7px;
     text-decoration-color: color-mix(in lch, currentColor, transparent);
@@ -59,13 +59,17 @@
         background: color-mix(in lch, currentColor 12%, transparent);
     });
 
-    @media @tui-mouse {
-        &[data-appearance='']:hover {
-            opacity: 0.7;
-        }
-    }
+    .appearance-hover({
+        text-decoration-color: currentColor;
+        text-decoration-line: underline;
+    });
 
-    &[data-appearance='']:active {
-        opacity: 0.7;
+    .appearance-active({
+        text-decoration-color: currentColor;
+        text-decoration-line: underline;
+    });
+
+    &[data-appearance='action'] {
+        text-decoration-line: none;
     }
 }
--- a/projects/demo/src/modules/components/link/examples/3/index.html
+++ b/projects/demo/src/modules/components/link/examples/3/index.html
@@ -1,42 +1,130 @@
-<a
-    href="#"
-    iconStart="@tui.heart"
-    tuiLink
-    [style.--tui-icon-size.rem]="1.5"
->
-    Large icon
-</a>
-<p tuiTheme="dark">
+<tui-notification class="tui-space_bottom-3">
+    Use textContent binding to avoid extra whitespace introduced by browser when closing tag wraps to a new line —
+    otherwise distance to the icons will be bigger than it should be
+</tui-notification>
+
+<section>
     <button
-        appearance="action-grayscale"
-        iconEnd="@tui.circle-help"
+        iconEnd="@tui.chevron-down"
+        iconStart="@tui.arrow-right"
         tuiLink
         type="button"
-        [pseudo]="true"
-        (click)="onClick($event)"
-    >
-        Pseudo-link
-    </button>
-</p>
-<button
-    appearance="action-grayscale"
-    iconEnd="@tui.circle-help"
-    tuiLink
-    type="button"
-    [pseudo]="true"
-    (click)="onClick($event)"
->
-    Pseudo-link
-</button>
-<p>
-    <a
-        href="#"
+        [textContent]="'Link with icons'"
+    ></button>
+    <button
+        iconEnd="@tui.chevron-down"
+        iconStart="@tui.arrow-right"
         tuiLink
-        [style.color]="'var(--tui-status-warning)'"
-    >
-        Custom color
-    </a>
-</p>
-(If you need a pseudo-link, use
-<code>button</code>
-element for correct semantics)
+        type="button"
+        [style.text-decoration-style]="'dashed'"
+        [textContent]="'Link with icons'"
+    ></button>
+</section>
+
+<section>
+    <button
+        appearance=""
+        iconEnd="@tui.chevron-down"
+        iconStart="@tui.arrow-right"
+        tuiLink
+        type="button"
+        [textContent]="'Link with icons'"
+    ></button>
+    <button
+        appearance=""
+        iconEnd="@tui.chevron-down"
+        iconStart="@tui.arrow-right"
+        tuiLink
+        type="button"
+        [style.text-decoration-style]="'dashed'"
+        [textContent]="'Link with icons'"
+    ></button>
+</section>
+
+<section>
+    <button
+        appearance="action-grayscale"
+        iconEnd="@tui.chevron-down"
+        iconStart="@tui.arrow-right"
+        tuiLink
+        type="button"
+        [textContent]="'Link with icons'"
+    ></button>
+    <button
+        appearance="action-grayscale"
+        iconEnd="@tui.chevron-down"
+        iconStart="@tui.arrow-right"
+        tuiLink
+        type="button"
+        [style.text-decoration-style]="'dashed'"
+        [textContent]="'Link with icons'"
+    ></button>
+</section>
+
+<section
+    tuiTheme="light"
+    [style.background]="'#fff'"
+>
+    <button
+        appearance=""
+        iconEnd="@tui.chevron-down"
+        iconStart="@tui.arrow-right"
+        tuiLink
+        type="button"
+        [textContent]="'Link with icons'"
+    ></button>
+    <button
+        appearance=""
+        iconEnd="@tui.chevron-down"
+        iconStart="@tui.arrow-right"
+        tuiLink
+        type="button"
+        [style.text-decoration-style]="'dashed'"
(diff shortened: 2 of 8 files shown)


waterplea:
This is not what we need to update the spec, all of these are already possible:
https://stackblitz.com/edit/angular-pqczn7zp
Keep in mind appearance="" just inherits text color so it works in all cases. What we need to update are underline styles and hover state. Currently appearance="" changes opacity on hover so that you can tell a link from regular text when they are the same color. Redesigned spec introduced underline instead. Also we have an input called [pseudo]="boolean" — I would love to deprecate it, all it does is adding `text-decoration-line: underline;` style. So what we need is:

1) Introduce underline that changes color on hover and that is NOT there in non-hovered "action" appearance
2) Remove all the examples of [pseudo] and instead use `text-decoration-style: dashed`

splincode:
@waterplea hello, updated 🙏

projects/core/styles/components/link.less
```diff
@@ -59,13 +58,24 @@
...
+
+    [tuiTheme='dark'] &[data-appearance=''] {
+        text-decoration-color: var(--tui-service-underline-grayscale, rgba(255, 255, 255, 0.32));
+        color: var(--tui-text-primary);
+    }
+
+    &[data-appearance=''],
+    [tuiTheme='light'] &[data-appearance=''] {
+        text-decoration: underline;
+        text-decoration-color: var(--tui-service-underline-grayscale, rgba(0, 0, 0, 0.32));
```
waterplea, line 73: Do not overcomplicate this for this minor detail, use `color-mix` with `currentColor` and `transparent` like before, just update line 26 to have `solid` be default.

Then in `data-appearance="action"` make `text-decoration-line: none` so it's hidden in that appearance, and keep `.appearance-hover(...)` rule that you've added, that's all.

projects/core/styles/components/link.less
```diff
@@ -25,7 +25,6 @@
     color: inherit;
     text-decoration: none dashed currentColor;
     text-underline-offset: 0.2em;
-    text-decoration-thickness: 0.7px;
```
waterplea, line 28: Please put it back, it helps dashed line have smaller dash length which better resemble design.

projects/core/styles/components/link.less
```diff
@@ -59,13 +58,24 @@
...
+    &[data-appearance=''],
+    [tuiTheme='light'] &[data-appearance=''] {
+        text-decoration: underline;
+        text-decoration-color: var(--tui-service-underline-grayscale, rgba(0, 0, 0, 0.32));
     }
 
-    &[data-appearance='']:active {
-        opacity: 0.7;
+    &[data-appearance='action'] {
+        .appearance-hover({
```
waterplea, line 77: Let's not do this, might be just a mistake in the specs, I notified designers.

projects/demo/src/modules/components/link/examples/3/index.html
```diff
@@ -12,7 +12,8 @@
         iconEnd="@tui.circle-help"
         tuiLink
         type="button"
-        [pseudo]="true"
+        [style.text-decoration-line]="'underline'"
```
waterplea, line 15: This line should be unnecessary as line would be hidden only in default `action` appearance.

projects/demo/src/modules/components/link/examples/3/index.html
```diff
@@ -23,7 +24,8 @@
     iconEnd="@tui.circle-help"
     tuiLink
     type="button"
-    [pseudo]="true"
+    [style.text-decoration-line]="'underline'"
```
waterplea, line 27: Same here

projects/demo/src/modules/components/link/examples/5/index.html
```diff
@@ -0,0 +1,118 @@
+<h3>Default</h3>
```
waterplea, line 1: Let's replace 3rd example with this instead. And remove primary on light, it's irrelevant to us as it is the same as primary (it's in the specs for other platforms because it has to be a separate appearance for them due to inability to add tuiTheme="light" on a single element). Also let's add one more entry here at the end with custom color showing that you can write `[style.color]="'var(--tui-status-negative)'"` and override the color.

projects/demo/src/modules/components/link/examples/5/index.ts
```diff
@@ -0,0 +1,18 @@
+import {Component} from '@angular/core';
+import {RouterLink} from '@angular/router';
+import {changeDetection} from '@demo/emulate/change-detection';
+import {encapsulation} from '@demo/emulate/encapsulation';
+import {DemoRoute} from '@demo/routes';
+import {TuiLink} from '@taiga-ui/core';
+
+@Component({
+    standalone: true,
+    imports: [RouterLink, TuiLink],
```
waterplea, line 10: RouterLink and routes aren't used?

projects/demo/src/modules/components/link/index.html
```diff
@@ -10,6 +10,7 @@
             *ngFor="let example of examples; let index = index"
             [component]="index + 1 | tuiComponent"
             [content]="index + 1 | tuiExample: 'html,ts'"
+            [fullsize]="index === 4"
```
waterplea, line 13: Let's not do that, instead add `width: max-content;` to the `div` inside example

projects/demo/src/modules/components/link/examples/5/index.html
```diff
@@ -0,0 +1,118 @@
...
+<h3>Text primary</h3>
+
+<div>
+    <button
+        appearance=""
+        iconEnd="@tui.chevron-down"
+        iconStart="@tui.arrow-right"
+        tuiLink
+        type="button"
+        [textContent]="'Link with icons'"
```
waterplea, line 31: Let's add a `tui-notification` explaining this:
```
Use textContent binding to avoid extra whitespace introduced by browser when closing tag wraps to a new line — otherwise distance to the icons will be bigger than it should be
```

projects/demo/src/modules/components/link/examples/3/index.html
splincode: image

 Without `[style.text-decoration-line]="'underline'"` this is an angular 16 bug

projects/demo/src/modules/components/link/examples/3/index.html
splincode: Without [style.text-decoration-line]="'underline'" this is an angular 16 bug

projects/core/styles/components/link.less
splincode: fixed

projects/core/styles/components/link.less
splincode: fixed

projects/core/styles/components/link.less
splincode: fixed

projects/demo/src/modules/components/link/examples/3/index.html
splincode: fixed

projects/demo/src/modules/components/link/examples/3/index.html
splincode: fixed

projects/demo/src/modules/components/link/examples/5/index.html
splincode: fixed

projects/demo/src/modules/components/link/examples/5/index.ts
splincode: fixed

projects/demo/src/modules/components/link/examples/5/index.html
splincode: fixed

projects/demo/src/modules/components/link/examples/5/index.ts
```diff
@@ -0,0 +1,17 @@
...
+@Component({
+    standalone: true,
+    imports: [TuiLink, TuiNotification],
+    templateUrl: './index.html',
+    styleUrls: ['./index.less'],
+    encapsulation,
+    changeDetection,
+})
+export default class Example {
+    protected readonly routes = DemoRoute;
```
waterplea, line 16: Not used

projects/core/styles/components/link.less
```diff
@@ -59,13 +59,15 @@
         background: color-mix(in lch, currentColor 12%, transparent);
     });
 
-    @media @tui-mouse {
-        &[data-appearance='']:hover {
-            opacity: 0.7;
-        }
+    .appearance-hover({
+        text-decoration-thickness: 1px;
```
waterplea, line 63: `text-decoration-color: currentColor` instead. And also let's add it for `.appearance-active(...)` so that it works on mobile too where there's no hover state. Also please add `text-decoration-line: underline` so that regular appearance action links have underline on hover.

projects/core/styles/components/link.less
```diff
@@ -23,7 +23,7 @@
     cursor: pointer;
     font: inherit;
     color: inherit;
```
waterplea, line 25: Let's set primary color here, instead of `[tuiTheme='dark'] &[data-appearance=''] {`

projects/demo/src/modules/components/link/index.ts
```diff
@@ -9,5 +9,11 @@ import {TuiDemo} from '@demo/utils';
     changeDetection,
 })
 export default class Page {
-    protected readonly examples = ['Basic', 'Icons', 'Variants', 'Long text'];
+    protected readonly examples = [
+        'Basic',
+        'Icons',
+        'Variants',
```
waterplea, line 15: Let's drop `Variants` example now, the new one covers it better

splincode: fixed

projects/core/styles/components/link.less
splincode: fixed

projects/demo/src/modules/components/link/examples/5/index.ts
splincode: fixed

waterplea (approved):

nsbarsukov (approved):

PR #10830, opened 2025-04-15
fix(addon-commerce): `InputCardGroup` fix autofill issues

### LINKED ISSUE #10603: 🐞 - [InputCardGroup] Wrong Autofill Behaviour
### Reproduction url

https://taiga-ui.dev/components/input-card-group

### Description

1. Go to taiga ui InputCardGroup expamples
2. Add save card with all data (number/expdate/cvc)
3. Try use autofill on any example, except "With saved cards"
4. On every example except "With saved cards", all fields were pasted. In "With saved cards" ExpDate wasnt pasted

reproduction artifact: 

https://github.com/user-attachments/assets/e66558f2-e7cb-48e3-aee0-04c613f6cac9

### Taiga UI version

4.28.0

### Browser Used

Safari

### OS Used

iOS

waterplea:
Fixes #10603


splincode (approved):

vladimirpotekhin (approved):

projects/addon-commerce/components/input-card-group/input-card-group.component.ts
```diff
@@ -198,6 +201,7 @@ export class TuiInputCardGroup
     @Input()
     public id = tuiInjectId();
 
+    /** @deprecated apparently "off" doesn't disable autocomplete */
     @Input()
     public autocomplete = this.options.autocomplete;
```
nsbarsukov, line 206: User can still set `autocomplete="new-password"`

https://taiga-ui.dev/components/input#autofill

nsbarsukov: (same comment for other similar deprecations)

nsbarsukov (approved):

projects/addon-commerce/components/input-card-group/input-card-group.component.ts
waterplea: It was boolean before, so no. In separate inputs we can override it if we want because the input is externally accessible. In here, let's just keep it always on.

PR #10868, opened 2025-04-18
revert: `tuiLink` appearance

splincode:


waterplea (approved):

mdlufy (approved):

projects/core/styles/components/link.less
```diff
@@ -22,8 +22,8 @@
     border: none;
     cursor: pointer;
     font: inherit;
-    color: var(--tui-text-primary);
-    text-decoration: underline solid currentColor;
+    color: inherit;
```
waterplea, line 25: I think this one we should keep, let's make it `var(--tui-text-primary);`

waterplea (approved):

projects/core/styles/components/link.less
splincode: fixed

Issue #10508, opened 2025-03-04
🐞 - Skeleton animation does not depend on animation speed


waterplea:
### Reproduction url

https://taiga-ui.dev/directives/skeleton

### Description

Right now it is 1s regardless of the settings, this is bad because when animations are disabled through OS system settings or by providing speed as 0 – it does not affect it. Plus we need to set `opacity: 0.06` initially so that disabling animations still shows it with correct transparence. That last change needs to also be implemented in v3 in experimental skeleton.

### Taiga UI version

4, 3

### Browser Used

Chrome

### OS Used

macOS

PR #10875, opened 2025-04-21
fix(kit): skeleton animation does not depend on animation speed

Anectem:
Fixes https://github.com/taiga-family/taiga-ui/issues/10508


projects/kit/directives/skeleton/skeleton.style.less
```diff
@@ -1,22 +1,11 @@
...
-    }
-}
-
 [tuiSkeleton] {
     &._skeleton {
         color: transparent;
         background: var(--tui-background-base) !important;
         box-shadow: none !important;
         filter: contrast(0) brightness(0);
-        animation: tuiSkeleton ease-in-out 1s infinite alternate;
```
waterplea, line 19: I don't think we need to do all that, maybe just add calc(var(--tui-duration) * 4) here?

* 4 seems like a good middle ground for now, as we use double the animation speed in proprietary projects.

Anectem: Sorry, i have missed that, your right

waterplea (approved):

projects/kit/directives/skeleton/skeleton.style.less
Anectem: it seems opacity has been added previously, or do you mean switch opacity places in keyframes?
https://github.com/taiga-family/taiga-ui/blob/104753cba57dcd58fe661ed60e81ac2ee71ed080/projects/kit/directives/skeleton/skeleton.style.less#L22

splincode (approved):

PR #10879, opened 2025-04-21
fix(core): `Textfield` icons should focus input & open dropdown on click

### LINKED ISSUE #10840: 🐞 - `Chevron` does not open dropdown on click (directly on icon)
### Reproduction url

https://taiga-ui.dev/components/textfield#dropdown

### Description

- https://taiga-ui.dev/components/textfield#dropdown
- https://taiga-ui.dev/components/select

https://github.com/user-attachments/assets/b9cab2cc-d1cf-4296-b6c3-c4cb6c413e52

### Taiga UI version

4.33.0

### Browser Used

Chrome

### OS Used

macOS

nsbarsukov:
Fixes #10840


projects/core/components/textfield/textfield.component.ts
```diff
@@ -180,4 +181,15 @@ export class TuiTextfieldComponent<T> implements TuiDataListHost<T>, AfterConten
     protected onResize({contentRect}: ResizeObserverEntry): void {
         this.el.style.setProperty('--t-side', tuiPx(contentRect.width));
     }
+
+    protected onClick(event: Event): void {
+        if (
+            this.el === event.target && // Click on ::before,::after pseudo-elements ([iconStart] / [iconEnd])
+            !this.input?.nativeElement.matches(':read-only')
```
nsbarsukov, line 188: Should we augment `TuiTextfieldAccessor` with `readonly`-property ? 🤔 

https://github.com/taiga-family/taiga-ui/blob/11c644f98f410a9457729e4619c2bf9e9768fa9d/projects/core/components/textfield/textfield-accessor.ts#L4-L6

splincode (approved):

projects/core/components/textfield/textfield.component.ts
waterplea: I don't think we need to expand accessor. A concerns to think about:

We are setting open to true regardless of whether dropdown is there or not, might get us into a sync issue if dropdown content appears later but the open status is `true` and we cannot tell it to show the dropdown because it already thinks it's showing it.

waterplea (approved):

PR #10891, opened 2025-04-22
fix(core): `Root` fix dropdowns when scrollable

### LINKED ISSUE #10870: 🐞 - `TuiRoot` is incompatible with `overflow: auto`
### Reproduction url

https://stackblitz.com/edit/tui-root-overflow-auto?file=src%2Fglobal_styles.less

### Description

1. Open StackBlitz
2. Click on textfield to open dropdown
3. Scroll page

https://github.com/user-attachments/assets/5b32614f-e1cd-4584-96f7-f47e4bb16382

### Taiga UI version

4.33.0

### Browser Used

Chrome

### OS Used

macOS

waterplea:
Fixes #10870

### DIFF
--- a/projects/core/directives/dropdown/dropdown.directive.ts
+++ b/projects/core/directives/dropdown/dropdown.directive.ts
@@ -13,7 +13,6 @@ import {takeUntilDestroyed} from '@angular/core/rxjs-interop';
 import {tuiZonefreeScheduler} from '@taiga-ui/cdk/observables';
 import type {TuiContext} from '@taiga-ui/cdk/types';
 import {tuiInjectElement} from '@taiga-ui/cdk/utils/dom';
-import {tuiPure} from '@taiga-ui/cdk/utils/miscellaneous';
 import type {TuiRectAccessor, TuiVehicle} from '@taiga-ui/core/classes';
 import {tuiAsRectAccessor, tuiAsVehicle} from '@taiga-ui/core/classes';
 import type {TuiPortalItem} from '@taiga-ui/core/types';
@@ -92,7 +91,6 @@ export class TuiDropdownDirective
                 : content;
     }
 
-    @tuiPure
     public get position(): 'absolute' | 'fixed' {
         return tuiCheckFixedPosition(this.el) ? 'fixed' : 'absolute';
     }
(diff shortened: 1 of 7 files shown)


splincode (approved):

nsbarsukov (approved):

projects/core/directives/dropdown/dropdown.directive.ts
```diff
@@ -92,7 +91,6 @@ export class TuiDropdownDirective
                 : content;
     }
 
-    @tuiPure
```
hakimio, line 95: @waterplea the position getter is called constantly when the dropdown is open. Isn't removal of memoization a bit of a performance issue? `tuiCheckFixedPosition()` should be a pretty expensive call.

Here is how it looks like when the dropdown is just open, without doing anything (`OnPush` CD and `zoneless`):
dropdown position calls

waterplea: tuiCheckFixedPosition is called only once as it is remembered by dropdown component upon opening:
https://github.com/taiga-family/taiga-ui/blob/main/projects/core/directives/dropdown/dropdown.component.ts#L72

hakimio: It's called here too: https://github.com/taiga-family/taiga-ui/blob/14a5625d77ab64f4b68937eb38fa5aedb05c5d43/projects/core/directives/dropdown/dropdown.component.ts#L63

hakimio: That `map()` function seems to be called constantly. All the calls seen in the gif come from the `map()`.
image

waterplea: > It's called here too:
> 
> https://github.com/taiga-family/taiga-ui/blob/14a5625d77ab64f4b68937eb38fa5aedb05c5d43/projects/core/directives/dropdown/dropdown.component.ts#L63

That's a bug, looks like we forgot to change to `this.position`. The fact that map function is called each frame is expected, no way to generally sync position otherwise until CSS anchors are a thing. Position can change due to gazillion factors from transitions/animations to scroll, DOM changes and resizable textareas.

PR #10893, opened 2025-04-22
feat(cdk): new `Obfuscate` pipe

mansurishmuhametov:
Fixes #

### DIFF
--- a/projects/cdk/pipes/obfuscate/obfuscate.pipe.ts
+++ b/projects/cdk/pipes/obfuscate/obfuscate.pipe.ts
@@ -1,0 +1,46 @@
+import type {PipeTransform} from '@angular/core';
+import {inject, Pipe} from '@angular/core';
+
+import {TUI_OBFUSCATE_OPTIONS} from './obfuscate.options';
+
+@Pipe({
+    standalone: true,
+    name: 'tuiObfuscate',
+})
+export class TuiObfuscatePipe implements PipeTransform {
+    private readonly option = inject(TUI_OBFUSCATE_OPTIONS);
+
+    /**
+     * Transforms the input value by obfuscating it according to the specified recipe or symbol.
+     *
+     * @param value The value to be obfuscated should be a string.
+     * @param recipe The name of the recipe to use for obfuscation or a single character symbol for custom obfuscation.
+     * @returns The obfuscated string.
+     * @throws Will throw an error if the specified recipe is not found.
+     */
+    public transform(value: string, recipe = ''): string {
+        if (!value) {
+            return value;
+        }
+
+        if (!recipe) {
+            return this.option.default(value);
+        }
+
+        const obfuscate = this.option.recipes[recipe];
+
+        if (obfuscate) {
+            return obfuscate(value);
+        }
+
+        if (recipe.length === 1) {
+            return this.option.default(value, recipe[0]);
+        }
+
+        const availableRecipes = Object.keys(this.option.recipes).sort();
+
+        throw new Error(
+            `Obfuscate recipe "${recipe}" not found. Available recipes: [${availableRecipes}]`,
+        );
+    }
+}
--- a/projects/cdk/utils/miscellaneous/obfuscate.ts
+++ b/projects/cdk/utils/miscellaneous/obfuscate.ts
@@ -1,0 +1,66 @@
+/**
+ * Obfuscates a string by replacing certain characters with a symbol.
+ *
+ * @param value the input string to obfuscate
+ * @param symbol the symbol for obfuscation
+ * @return the obfuscated string
+ *
+ * The function determines which characters to obfuscate using a regular expression and the string's length:
+ * - 8 or more: show first 2 and last 2 characters
+ * - 4 to 7: show first and last character
+ * - less than 4: obfuscate all characters
+ * - obfuscates only alphanumeric characters
+ */
+export function tuiObfuscate(value: string, symbol: string): string {
+    if (!value) {
+        return value;
+    }
+
+    const match = /[\p{L}\p{N}]/gu;
+
+    let visible = 0;
+    let obfuscateIndexes: number[] = getObfuscateIndexes(value, match);
+
+    if (obfuscateIndexes.length >= 8) {
+        visible = 2;
+    } else if (obfuscateIndexes.length >= 4) {
+        visible = 1;
+    }
+
+    obfuscateIndexes = obfuscateIndexes.slice(visible, obfuscateIndexes.length);
+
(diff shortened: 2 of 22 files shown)


projects/core/tokens/obfuscate-format.ts
```diff
@@ -0,0 +1,79 @@
+import type {InjectionToken, Provider} from '@angular/core';
+import {Optional, SkipSelf} from '@angular/core';
+import {tuiCreateToken} from '@taiga-ui/cdk/utils/miscellaneous';
+
+/**
+ * Defining the format for obfuscation
+ */
+export interface TuiObfuscateFormat {
```
waterplea, line 8: We do not want to introduce all this complexity here. What we want is:
1) A dictionary token of recipes
2) A pipe that accepts the key to that dictionary
3) A reasonable default behavior

A recipe will be just a callback `(value: string): string` and people will be free to set it up any way they like. For default behavior you can have a fallback recipe (also a token) that gets the key as input, kinda like that inside the pipe:
```js
return RECIPEES[key]?.(value) ?? FALLBACK(key, value);
```
And the default fallback would take `key[0]` as a char and replace all non-space, non-brackets, non-dash, non-plus symbols in the value with that char. It should leave out last 4 chars and some chars in the beginning. Maybe first block of actual chars like country code in the phone, up to 4 chars. So when you pass these values and an asterisk it has this result:
```
+1 (234) 567-8900  -> +1 (***) ***-8900
4200 0000 0000 1234 -> 4200 **** **** 1234
GB24BARC20201630093459 -> GB24**************3459
```
You would need to play a bit with the default fallback, that might just be a rough idea. The main point is we do not want it to be so thoroughly configurable. It should just be:
1) Specific recipes
2) Reasonable fallback
And people will provide complete implementations if default does not suit their needs. We do not want a swiss army knife here.

waterplea:
I think it should be in cdk, it's pretty low level

projects/core/tokens/obfuscate-format.ts
mansurishmuhametov: Agree that the solution needs to be more configurable. I tried to create a simpler solution that can be easily customized based on user needs. The core behavior allows users to use the pipe immediately without needing to configure token. However, if a user wants, they can add recipes by defining their own callbacks. Users can also override the default behavior.

projects/demo/src/modules/pipes/obfuscate/examples/1/index.ts
```diff
@@ -0,0 +1,20 @@
...
+@Component({
+    standalone: true,
+    imports: [TuiObfuscatePipe],
+    templateUrl: './index.html',
+    encapsulation,
+    changeDetection,
+    providers: [
+        tuiObfuscateOptionProvider({
+            recipes: {
+                city: (value: string): string => 'x'.repeat(value.length),
```
waterplea, line 15: nit: can be `({ length }) => ...`. Also do we need both :strings? I think it's already typed since we pass it to the helper function?

projects/demo/src/modules/pipes/obfuscate/examples/import/import.md
```diff
@@ -0,0 +1,24 @@
...
+  imports: [
+    // ...
+    TuiObfuscatePipe,
+  ],
+  //  ...
+  providers: [
+    tuiObfuscateOptionProvider({
+      default: (value: string, symbol?: string): string => (symbol ?? '*').repeat(value.length),
+      recipes: {
+        phone: (value: string): string => 'xxx',
```
waterplea, line 17: Are those `: string` necessary?

projects/demo/src/modules/pipes/obfuscate/examples/import/template.md
```diff
@@ -0,0 +1,4 @@
+```html
+Obfuscated value: {{'abc' | tuiObfuscate}} Obfuscated value: {{'abc' | tuiObfuscate: '#'}} Obfuscated value: {{'abc' |
```
waterplea, line 2: Let's break the lines for readability

projects/cdk/pipes/obfuscate/obfuscate.pipe.ts
```diff
@@ -0,0 +1,50 @@
...
+import {inject, Pipe} from '@angular/core';
+import type {TuiObfuscateOption} from '@taiga-ui/cdk/tokens';
+import {TUI_OBFUSCATE_OPTION} from '@taiga-ui/cdk/tokens';
+
+@Pipe({
+    standalone: true,
+    name: 'tuiObfuscate',
+})
+export class TuiObfuscatePipe implements PipeTransform {
+    private readonly option?: TuiObfuscateOption = inject(TUI_OBFUSCATE_OPTION);
```
waterplea, line 11: Why is it optional?

projects/cdk/tokens/obfuscate-option.ts
```diff
@@ -0,0 +1,45 @@
...
+    readonly default: (value: string, symbol?: string) => string;
+
+    /**
+     * Collection of named recipes for different obfuscation scenarios.
+     */
+    readonly recipes: Record<string, (value: string) => string>;
+}
+
+export const TUI_OBFUSCATE_DEFAULT_OPTION: TuiObfuscateOption = {
+    default: (value, symbol) => tuiObfuscate(value, symbol ?? '*'),
```
waterplea, line 23: symbol = '*'

projects/cdk/pipes/obfuscate/obfuscate.pipe.ts
```diff
@@ -0,0 +1,50 @@
...
+            return value;
+        }
+
+        if (!recipe) {
+            const obfuscate = this.option?.default;
+
+            return obfuscate ? obfuscate(value) : value;
+        }
+
+        const obfuscate = this.option?.recipes?.[recipe];
```
waterplea, line 32: recipes is also not optional

waterplea (approved):

projects/cdk/utils/miscellaneous/obfuscate.ts
```diff
@@ -0,0 +1,68 @@
...
+ */
+export function tuiObfuscate(value: string, symbol: string): string {
+    if (!value) {
+        return value;
+    }
+
+    const match = /[\p{L}\p{N}]/gu;
+
+    let showFirst = 0;
+    let showLast = 0;
```
vladimirpotekhin, line 21: seems like `showFirst` and `showLast` are always equal. Maybe combine them into one variable? `visible` for example

vladimirpotekhin (approved):

nsbarsukov (approved):

PR #10911, opened 2025-04-24
fix(kit): `InputMonth` with enabled native picker should never show desktop dropdown

nsbarsukov:


projects/kit/components/input-month/native-month-picker/native-month-picker.style.less
```diff
@@ -11,3 +11,7 @@ tui-textfield input[tuiInputMonth] + .t-content input[type='month'] {
         .fullsize(); // Extends clickable area for icon to open native calendar
     }
 }
+
+tui-textfield:has(input[type='month'])::after {
+    pointer-events: none;
```
nsbarsukov, line 16: Is it a good candidate for graceful degradation ?

waterplea (approved):

splincode (approved):

PR #10914, opened 2025-04-25
feat(kit): add loop mode to carousel

### LINKED ISSUE #3452: 🚀 - Scrolling mode for carousel

kit

### Description

Problem
Right now carousel rewinds to the first element when it reaches the end of a sequence.
I would like to have behaviour without rewind, just with continuous scrolling instead. 
Like here for example - https://coreui.io/angular/docs/components/carousel.

Proposal
I guess that the carousel could have a "scrollingMode" input parameter with "rewind" and "flow" values.
"rewind" for current behaviour and "flow" for suggested.

gan-h:
Hi everyone! This is my first pull request - I hope I covered all the bases. These code changes aim to add a "loop" mode to the carousel. 

Closes #3452

### DIFF
--- a/projects/demo-playwright/tests/kit/carousel/carousel.pw.spec.ts
+++ b/projects/demo-playwright/tests/kit/carousel/carousel.pw.spec.ts
@@ -49,4 +49,30 @@ test.describe('Carousel', () => {
             .soft(document.apiPageExample)
             .toHaveScreenshot('carousel-draggable.png');
     });
+
+    test('loop mode works in “Custom” example', async ({page}) => {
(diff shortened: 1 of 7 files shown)


vladimirpotekhin:
Thanks a lot for the contribution!
Returning to the first slide can already be achieved in the current version by controlling the carousel externally and setting the index to 0. What we need in the issue is an infinite carousel, as shown in the example. https://coreui.io/angular/docs/components/carousel/

We’re planning to implement this as part of a new component, and the team decided not to extend the existing one, so we’re going to close this PR. Thanks again for your idea and involvement!

PR #10924, opened 2025-04-28
chore: update readme

mdlufy:
Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,33 +204,7 @@ design your app with Taiga UI components.
                         src="https://raw.githubusercontent.com/MarsiBarsi/readme-icons/main/send.svg"
                 /></a>
             </div>
-        </td>
-        <td align="center">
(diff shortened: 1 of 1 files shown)


vladimirpotekhin (approved):

README.md
```diff
@@ -204,26 +204,33 @@ design your app with Taiga UI components.
                         src="https://raw.githubusercontent.com/MarsiBarsi/readme-icons/main/send.svg"
                 /></a>
             </div>
-        </td>
+        </td> 
         <td align="center">
-            <a href="https://github.com/mdlufy"
```
splincode, line 209: semantically, has anything changed?

splincode: I don't think the order matters, I make less impact and work part time in a team, so the order can be left as is :)

waterplea (approved):

PR #10966, opened 2025-05-13
feat(cdk): `Animated` add new directive, drop all Angular animations

waterplea:
Fixes #


projects/core/directives/dropdown/dropdown.component.ts
```diff
@@ -1,19 +1,18 @@
 import {ChangeDetectionStrategy, Component, computed, inject} from '@angular/core';
 import {takeUntilDestroyed} from '@angular/core/rxjs-interop';
+import {TuiAnimated} from '@taiga-ui/cdk';
```
nsbarsukov, line 3: Eslint did not autofix it 🤔 

```suggestion
import {TuiAnimated} from '@taiga-ui/cdk/directives/animated';
```

projects/demo/src/modules/app/app.config.ts
```diff
@@ -252,5 +253,9 @@ export const config: ApplicationConfig = {
             eventCoalescing: false,
             runCoalescing: false,
         }),
+        {
+            provide: REMOVE_STYLES_ON_COMPONENT_DESTROY,
+            useValue: false,
```
nsbarsukov, line 258: Do we need it here if `TuiRoot` already provides it ?

projects/layout/components/dynamic-header/dynamic-header.component.ts
```diff
@@ -1,23 +1,18 @@
-import type {AnimationOptions} from '@angular/animations';
 import {NgForOf, NgIf} from '@angular/common';
-import {ChangeDetectionStrategy, Component, computed, inject} from '@angular/core';
-import {tuiFadeIn} from '@taiga-ui/core/animations';
-import {TUI_ANIMATIONS_SPEED} from '@taiga-ui/core/tokens';
-import {tuiGetDuration, tuiToAnimationOptions} from '@taiga-ui/core/utils/miscellaneous';
+import {ChangeDetectionStrategy, Component, inject} from '@angular/core';
+import {TuiAnimated} from '@taiga-ui/cdk';
```
nsbarsukov, line 3:
```suggestion
import {TuiAnimated} from '@taiga-ui/cdk/directives/animated';
```

nsbarsukov: (+ fix all other similar cases)

nsbarsukov (approved):

projects/demo/src/modules/app/app.config.ts
waterplea: I was using it for tests by providing `true` here because that's the default in newer versions for Angular. Let's keep it, I just forgot to switch it back to `true`

vladimirpotekhin (approved):

hakimio:
@waterplea would be nice to have some documentation page describing how to use the new directive or even a migration script to migrate the old animations.

PR #10995, opened 2025-05-20
fix: tuiFormatSize hard codes ru-RU locale

### LINKED ISSUE #10957: 🐞 - tuiFormatSize hard codes ru-RU locale
### Reproduction url

n/a

### Description

I know that the file size options can be provided. 

However that means you need to duplicate the logic to have the number displayed with `.` as a decimal separator vs `,`. It seems like these should be abstracted so that the locale isn't hard coded in there.

### Taiga UI version

4.36.0

### Browser Used

Chrome

### OS Used

Windows

Nikikwiki:
Fixes #10957 

1. Added default locale by LOCALE_ID injection
2. Added input param to formatSize function


nsbarsukov:
Closing this draft because it is already fixed by the following PR:
* https://github.com/taiga-family/taiga-ui/pull/11004

Anyway, thanks for your desire to help ❤️

Issue #10963, opened 2025-05-13
🚀 - Support safe-area zones


mdlufy:
### Description

Support [safe-area](https://developer.mozilla.org/en-US/docs/Web/CSS/env) in components:
- sheet-dialog
- tui-dialog
- dropdown-mobile

mdlufy:
[Drawer](https://taiga-ui.dev/components/drawer) need to add `top: 0` in overlay by default in v5

PR #11001, opened 2025-05-22
feat(core): `Alert` add safe-area support

mdlufy:
Part of #10963

### DIFF
--- a/projects/core/components/alert/alert.providers.ts
+++ b/projects/core/components/alert/alert.providers.ts
@@ -2,7 +2,5 @@ import {inject} from '@angular/core';
 import {TUI_IS_MOBILE, tuiCreateTokenFromFactory} from '@taiga-ui/cdk';
 
 export const TUI_ALERT_POSITION = tuiCreateTokenFromFactory<string>(() =>
-    inject(TUI_IS_MOBILE)
-        ? 'calc(1rem + env(safe-area-inset-top)) 1rem 0 auto'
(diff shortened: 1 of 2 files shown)


projects/core/components/alert/alert.providers.ts
```diff
@@ -2,5 +2,7 @@ import {inject} from '@angular/core';
 import {TUI_IS_MOBILE, tuiCreateTokenFromFactory} from '@taiga-ui/cdk';
 
 export const TUI_ALERT_POSITION = tuiCreateTokenFromFactory<string>(() =>
-    inject(TUI_IS_MOBILE) ? '1rem 1rem 0 auto' : '2rem 3rem 0 auto',
+    inject(TUI_IS_MOBILE)
```
waterplea, line 5: If possible, it's better to take care of this in `tui-alerts`, so it's always there even if people override this

waterplea (approved):

projects/core/components/alert/alert.providers.ts
mdlufy: What's is tui-alerts in v3? I can't found it

mdlufy: Not working neither in tui-alert-host in v3 , nor in tui-alerts in v4

image



> If possible, it's better to take care of this in `tui-alerts`, so it's always there even if people override this

PR #11006, opened 2025-05-22
feat(core): `Alert` add safe-area-top support

mdlufy:
Part of #10963

### DIFF
--- a/projects/core/components/alert/alerts.style.less
+++ b/projects/core/components/alert/alerts.style.less
@@ -6,6 +6,7 @@ tui-alerts > .t-wrapper {
     display: flex;
     flex-direction: column;
     pointer-events: none;
+    padding-top: env(safe-area-inset-top);
 
     & > * {
         pointer-events: auto;


projects/core/components/alert/alerts.style.less
```diff
@@ -6,6 +6,8 @@ tui-alerts > .t-wrapper {
     display: flex;
     flex-direction: column;
     pointer-events: none;
+    block-size: calc(100% - env(safe-area-inset-top) - env(safe-area-inset-bottom));
```
waterplea, line 9: Mabye just 100% + box-sizing: border-box?

waterplea (approved):

projects/core/components/alert/alerts.style.less
mdlufy: Updated

projects/core/components/alert/alert.style.less
```diff
@@ -18,14 +18,17 @@
     &:not(:last-child) {
         margin-bottom: 0 !important;
     }
+
+    &:not(:last-child) > .t-wrapper {
+        margin-bottom: 0.75rem;
```
waterplea, line 23: Now if we provide `auto 3rem 2rem auto` as `TUI_ALERT_POSITION` so that alerts are aligned to the bottom and try closing the last one – the animation is a bit broken.

PR #11016, opened 2025-05-27
feat(layout): `Form` add default options

### LINKED ISSUE #11010: 🚀 - `tuiForm` has no default value
### Description

When using `tuiForm` we always need to pass the size. This directive primarily represents certain styles, so it would be more convenient if we didn't have to pass the size every time. For this purpose, the directive already has a default value (large). You can add the `tuiFormOptionsProvider` provider for convenient global override.

Stivin:
Added default values ​​and a provider to override directive `tuiForm` parameters at a global or more local level. This way we can define the required size once and not specify it every time in the templates.

Fixes #11010

### DIFF
--- a/projects/layout/components/form/form.directive.ts
+++ b/projects/layout/components/form/form.directive.ts
@@ -5,17 +5,23 @@ import {
     Directive,
     inject,
     Input,
+    signal,
     ViewEncapsulation,
 } from '@angular/core';
 import type {TuiStringHandler} from '@taiga-ui/cdk/types';
 import {tuiWithStyles} from '@taiga-ui/cdk/utils/miscellaneous';
-import {tuiButtonOptionsProvider} from '@taiga-ui/core/components/button';
-import {tuiNotificationOptionsProvider} from '@taiga-ui/core/components/notification';
-import {TuiTextfieldOptionsDirective} from '@taiga-ui/core/components/textfield';
-import type {TuiSizeL, TuiSizeS} from '@taiga-ui/core/types';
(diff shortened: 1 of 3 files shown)


projects/layout/components/form/form.directive.ts
```diff
@@ -61,10 +70,16 @@ class TuiFormStyles {}
     },
 })
 export class TuiForm {
+    protected readonly options = inject(TUI_FORM_OPTIONS);
+
     protected readonly nothing = tuiWithStyles(TuiFormStyles);
 
+    public size: TuiFormOptions['size'] = this.options.size;
+
     @Input('tuiForm')
```
waterplea, line 79: No need for an alias

projects/layout/components/form/form.directive.ts
```diff
@@ -39,20 +45,23 @@ class TuiFormStyles {}
     standalone: true,
     selector: '[tuiForm]',
     providers: [
-        tuiButtonOptionsProvider(TuiForm),
-        tuiNotificationOptionsProvider(TuiForm),
+        projectSize(TUI_BUTTON_OPTIONS, (size) => size),
+        projectSize(TUI_NOTIFICATION_OPTIONS, (size) => size),
         projectSize(TUI_HEADER_OPTIONS, (size) => HEADER_SIZE[size]!),
```
waterplea, line 50: What if we drop type of `HEADER_SIZE` and add `as const`? We can then change type in projectSize to `TuiHandler<TuiFormOptions['size'], string>` and get rid of `!`?

waterplea (approved):

PR #10980, opened 2025-05-15
[WIP] feat(kit): `ComboBox` refactor to new `Textfield`

nsbarsukov:
```html
<tui-textfield tuiChevron>
  <input
    tuiComboBox
    [(ngModel)]="value"
  />

  <tui-data-list-wrapper
    *tuiTextfieldDropdown
    [items]="items"
  />
</tui-textfield>
```

PR #11039, opened 2025-05-29
fix(kit): `ComboBox` should retrigger value matching on new items in datalist

### LINKED ISSUE #11077: 🚀 - ComboBox and string value from DataList
### Description

Hello!

I've encountered an issue when creating a `ComboBox` with a custom value in a `DataList`.

When using an array of values, everything works correctly as long as the values are not strings. However, if a string is used as a value, the entire flow breaks.

After reviewing the `ComboBox` source code, I noticed a strong dependency on `tuiIsString`. Is this behavior intentional? It currently requires significant changes to the logic even when only the type of value changes.

Here’s the relevant snippet from the ComboBox [directive](https://github.com/taiga-family/taiga-ui/blob/main/projects/kit/components/combo-box/combo-box.directive.ts):

```
protected readonly valueEffect = effect(() => {
    const value = this.value() ?? '';
    const stringified = tuiIsString(value)
        ? value
        : this.itemsHandlers.stringify()(value);
    const match = this.match(stringified);

    this.textfield.value.update((x) => stringified || x);

    if (match) {
        setTimeout((end = this.el.value.length) =>
            this.el.setSelectionRange(end, end),
        );
    }
}, TUI_ALLOW_SIGNAL_WRITES);
```

I’ve also created a minimal reproduction example on StackBlitz:
🔗 https://stackblitz.com/edit/angular-ryibuhbo-w7iirwwa?file=src%2Fapp%2Fapp.component.ts

Would appreciate any insights or clarification on whether this is expected behavior, or if support for string values could be improved.

Thanks in advance!

nsbarsukov:
Relates:
* https://github.com/taiga-family/taiga-ui/pull/10980#discussion_r2109058436

Closes #11077

### DIFF
--- a/projects/kit/components/combo-box/combo-box.directive.ts
+++ b/projects/kit/components/combo-box/combo-box.directive.ts
@@ -1,9 +1,17 @@
-import {Directive, effect, inject, Input, untracked} from '@angular/core';
+import {
+    computed,
+    Directive,
+    effect,
+    inject,
+    Input,
+    isSignal,
+    signal,
+    untracked,
+} from '@angular/core';
 import {tuiAsControl, TuiControl} from '@taiga-ui/cdk/classes';
 import {TUI_ALLOW_SIGNAL_WRITES, TUI_STRICT_MATCHER} from '@taiga-ui/cdk/constants';
 import type {TuiStringMatcher} from '@taiga-ui/cdk/types';
 import {tuiInjectElement} from '@taiga-ui/cdk/utils/dom';
-import {tuiIsString} from '@taiga-ui/cdk/utils/miscellaneous';
 import type {TuiDataListAccessor} from '@taiga-ui/core/components/data-list';
 import {tuiAsOptionContent} from '@taiga-ui/core/components/data-list';
 import type {TuiTextfieldAccessor} from '@taiga-ui/core/components/textfield';
@@ -37,7 +45,7 @@ import {TuiSelectOption} from '@taiga-ui/kit/components/select';
     host: {
         '[disabled]': 'disabled()',
         '(click)': 'toggleDropdown()',
-        '(input)': 'onInput($event.target.value)',
+        '(input)': 'onInput()',
         '(keydown.enter)': 'keydownEnter($event)',
     },
 })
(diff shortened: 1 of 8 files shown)


projects/kit/components/combo-box/combo-box.directive.ts
```diff
@@ -48,91 +57,113 @@ export class TuiComboBox<T>
...
-            : this.itemsHandlers.stringify()(value);
-        const match = this.match(stringified);
+    protected readonly matchingEffect = effect(() => {
+        const options = this.options();
+        const textfieldValue = this.textfield.value();
 
-        this.textfield.value.update((x) => stringified || x);
+        if (!options.length) {
+            if (!this.strict()) {
+                this.onChange(textfieldValue || null);
```
nsbarsukov: Is it better to create the separate effect to remove the such nesting ?

```ts
protected readonly nonStrictValueEffect = effect(() => {
    if (!this.options().length && !this.strict()) {
        this.onChange(this.textfield.value());
    }
}, TUI_ALLOW_SIGNAL_WRITES);

protected readonly matchingEffect = effect(() => {
    const options = this.options();

    if (!options.length) {
        return;
    }

    // [...]
}, TUI_ALLOW_SIGNAL_WRITES);
```

projects/kit/components/combo-box/combo-box.directive.ts
```diff
@@ -37,7 +46,7 @@ import {TuiSelectOption} from '@taiga-ui/kit/components/select';
     host: {
         '[disabled]': 'disabled()',
         '(click)': 'toggleDropdown()',
-        '(input)': 'onInput($event.target.value)',
+        '(input.debounce~0ms)': 'toggleDropdown(true)',
```
waterplea: Can we avoid using debounce plugin? I'm still kind of skeptical about it.

projects/kit/components/combo-box/combo-box.directive.ts
```diff
@@ -48,91 +57,113 @@ export class TuiComboBox<T>
...
+            this.setValue(this.options()[0]!);
             this.toggleDropdown(false);
         }
     }
 
-    private get options(): readonly T[] {
-        return this.datalist()?.getOptions() || [];
-    }
+    private stringify(value: T | string | null): string {
+        if (tuiIsString(value)) {
```
waterplea: Are we sure we want to skip it for strings? I believe we had it at like that some point and then people came asking if they can use stringify on strings IDs to show better presentational value.

projects/kit/components/combo-box/combo-box.directive.ts
```diff
@@ -48,91 +57,113 @@ export class TuiComboBox<T>
...
+
+    // TODO(v5): use signal input
+    @Input('matcher')
+    public set matcherSetter(x: TuiStringMatcher<T> | null) {
+        this.matcher.set(x ?? TUI_STRICT_MATCHER);
+    }
 
-    @Input()
-    public matcher: TuiStringMatcher<T> | null = TUI_STRICT_MATCHER;
+    public setValue(value: T | string | null): void {
```
waterplea: Do we really need to extend it to `| string | null` everywhere? Looks a bit messy at a glance

projects/kit/components/combo-box/combo-box.directive.ts
```diff
@@ -48,91 +57,113 @@ export class TuiComboBox<T>
...
+    // TODO(v5): use signal input
+    @Input('strict')
+    public set strictSetter(x: boolean) {
+        this.strict.set(x);
+    }
+
+    // TODO(v5): use signal input
+    @Input('matcher')
+    public set matcherSetter(x: TuiStringMatcher<T> | null) {
+        this.matcher.set(x ?? TUI_STRICT_MATCHER);
```
waterplea: Previously `null` was used to disable matching

projects/kit/components/combo-box/combo-box.directive.ts
waterplea: Yes, I think that's more readable. Except I'm not sure about `nonStrictValueEffect` name

projects/kit/components/combo-box/combo-box.directive.ts
nsbarsukov: It is required to use here
https://github.com/taiga-family/taiga-ui/blob/14a5625d77ab64f4b68937eb38fa5aedb05c5d43/projects/kit/components/combo-box/combo-box.directive.ts#L45
because otherwise we cannot 
```ts
if (!this.strict()) {
    this.onChange(this.textfield.value())
}
```

___

As for other cases (`writeValue` & `setValue`), we can remove `string` from its type without breaking build, yep.
But I think that should not do it for `writeValue` (because actually control can be really patched with string in non-strict mode). Probably, we should remove `string` for `setValue` - this method is used only by datalist (click on option) and by cleaner.

projects/kit/components/combo-box/combo-box.directive.ts
nsbarsukov: No problem. I've replaced it with the following lines:

```ts
host: {
    '(input)': 'onInput()',
}

// [...]

protected onInput(): void {
    setTimeout(() => this.toggleDropdown(true));
}
```

projects/kit/components/combo-box/combo-box.directive.ts
nsbarsukov: Fixed

projects/kit/components/combo-box/combo-box.directive.ts
nsbarsukov: Fixed

projects/kit/components/combo-box/combo-box.directive.ts
nsbarsukov: Fixed

PR #11046, opened 2025-05-30
feat(core): add variables to :host

MillerSvt:


projects/core/styles/theme/palette.less
```diff
@@ -144,13 +144,15 @@
     }
 
     &:root,
+    :host,
```
waterplea, line 147: Would having & for :host be useful? Technically would make it easier to conditionally apply it based on some selector on host element.

waterplea (approved):

projects/core/styles/theme/palette.less
MillerSvt: I don't know what `&` mean in style root. I'm not sure that it will work in Shadow DOM. I have no possibility to check this right now. If you sure, let's do it.

waterplea: We use it so we can toggle themes on demo portal on HTML by importing: 
```css
[data-theme="xxx"] {
  @import ...
}
```
When there's nothing above, & gracefully turns to nothing in that case. If you don't need anything like that, let's keep it as is, it can always be added later.

vladimirpotekhin (approved):

Issue #11028, opened 2025-05-28
📚 - `Animated` directive


waterplea:
### Description

We are moving away from Angular Animations. As a replacement a new directive `TuiAnimated` was added. It adds `tui-enter` class when the element appears and `tui-leave` class when the element disappears from DOM and sets duration and direction for animations. In order to have your element animated you can define CSS animations like that, for example:
```css
.tui-enter,
.tui-leave {
  animation-name: tuiFade;
}
```
We need to add a documentation page on the demo and in v5 when we will drop our animations we will need to provide migration guide in comments.

### Browser Used

Chrome

### OS Used

macOS

loxy:
Can you explain why you move away?

waterplea:
> Can you explain why you move away?

Angular itself is moving away:
https://angular.dev/guide/animations
There's no need for this package anymore, the only problem was :leave state and we worked around it with a directive until Angular comes up with an official solution.

PR #11068, opened 2025-06-05
chore(doc): add documentation for new Animated directive

Farbaks:
Fixes https://github.com/taiga-family/taiga-ui/issues/11028

### DIFF
--- a/projects/demo/src/modules/directives/animated/examples/1/index.html
+++ b/projects/demo/src/modules/directives/animated/examples/1/index.html
@@ -1,0 +1,67 @@
+<button
+    tuiButton
+    type="button"
+    (click)="step.set(step() || 1)"
+>
+    Show popover
+</button>
+<div
+    *tuiPopup="!!step()"
+    tuiAnimated
+    tuiAppearance="floating"
+    tuiCardLarge="compact"
+    class="popover"
+>
+    <ng-container *ngIf="step() === 1">
+        <h1 tuiTitle>
+            Leave your feedback
+            <span tuiSubtitle>It will only take 3 minutes</span>
+        </h1>
+        <tui-rating [(ngModel)]="rating" />
+    </ng-container>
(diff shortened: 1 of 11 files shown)


projects/demo/src/modules/directives/animated/examples/import/template.md
```diff
@@ -0,0 +1,10 @@
+```html
+<span class="t-wrapper">
+  <span
```
waterplea: I think we should wrap it in `@if` since that's how it will be used. Also let's remove wrapper and classes, this should be a minimal example

projects/demo/src/modules/directives/animated/examples/import/import.md
```diff
@@ -0,0 +1,15 @@
+```ts
+import {TuiAnimated} from '@taiga-ui/cdk/directives/animated';
```
waterplea: Just /cdk, it's tree shakable from the top level

projects/demo/src/modules/directives/animated/examples/1/index.html
```diff
@@ -0,0 +1,67 @@
+<button
```
waterplea, line 1: This example is rather extensive, I think we should show something simpler, otherwise new users might get overwhelmed and miss the point that this directive is just there to add animations on entering and leaving and doesn't even need any extra parameters.

projects/demo/src/modules/directives/animated/examples/1/index.less
```diff
@@ -0,0 +1,25 @@
+.popover {
+    position: fixed;
+    right: 1.5rem;
+    bottom: 1.5rem;
+    inline-size: 20rem;
+
+    --tui-from: translateX(100%);
```
waterplea: It's good to showcase this. Other than that, let's try to simplify example.

waterplea (approved):

splincode (approved):

PR #11069, opened 2025-06-05
[WIP] feat(kit): refactor `InputTime` to new `Textfield`

nsbarsukov:


projects/demo/src/modules/components/input-time/examples/5/index.ts
```diff
@@ -0,0 +1,28 @@
...
+        tuiInputTimeOptionsProvider({
+            icon: () => '',
+        }),
+    ],
+})
+export default class Example {
+    protected value: TuiTime | null = new TuiTime(9, 0);
+
+    protected isNight({hours}: TuiTime): boolean {
+        return hours >= 0 && hours < 8;
```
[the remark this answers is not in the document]
nsbarsukov: Good catch 😄

projects/demo/src/modules/components/input-time/examples/7/index.ts
```diff
@@ -0,0 +1,61 @@
...
+         *     [stringify]="..."
+         * />
+         * ```
+         */
+        tuiItemsHandlersProvider({
+            stringify: signal((x: TuiTime) => x.toString('HH:MM')),
+            identityMatcher: signal(
+                (a: TuiTime | null, b: TuiTime | null) => a?.valueOf() === b?.valueOf(),
+            ),
+            // disabledItemHandler: signal((x: TuiTime) => x.hours > 18),
```
splincode, line 43: is this still needed?

nsbarsukov: Yep, it is required to show developer one more possible way to enable `disabledItemHandler` by DI

projects/kit/components/input-time/input-time.options.ts
```diff
@@ -0,0 +1,23 @@
...
+export interface TuiInputTimeOptions
+    extends Required<Pick<MaskitoTimeParams, 'mode' | 'timeSegmentMaxValues'>> {
+    readonly icon: TuiHandler<TuiSizeL | TuiSizeS, string>;
+    readonly valueTransformer: TuiValueTransformer<TuiTime | null, any> | null;
+}
+
+export const TUI_INPUT_TIME_DEFAULT_OPTIONS: TuiInputTimeOptions = {
+    icon: () => '@tui.clock',
+    mode: 'HH:MM',
+    timeSegmentMaxValues: {},
```
waterplea, line 17: Can we shorten that somehow? Looks scary 😨

projects/kit/components/input-time/input-time.directive.ts
```diff
@@ -0,0 +1,165 @@
...
+        tuiAsTextfieldAccessor(TuiInputTimeDirective),
+        tuiAsAuxiliary(TuiInputTimeDirective),
+        tuiValueTransformerFrom(TUI_INPUT_TIME_OPTIONS),
+        tuiAsOptionContent(TuiSelectOption),
+    ],
+    hostDirectives: [TuiWithTextfield, MaskitoDirective],
+    host: {
+        '[disabled]': 'disabled()',
+        '[attr.inputMode]': 'inputMode()',
+        '(blur)': 'onTouched()',
```
waterplea: Do we still need it? I think I've added active zone in textfield to take care of that

projects/kit/components/input-time/input-time.directive.ts
```diff
@@ -0,0 +1,165 @@
...
+
+        this.control?.control?.updateValueAndValidity({emitEvent: false});
+        this.onChange(newValue);
+
+        if (newValue && newValue !== time) {
+            this.textfield.value.set(newValue?.toString(this.timeMode()));
+        }
+    }, TUI_ALLOW_SIGNAL_WRITES);
+
+    public readonly nativePickerEnabled =
```
waterplea: I think we call it just `native` in other similar cases

projects/kit/components/input-time/input-time.directive.ts
```diff
@@ -0,0 +1,165 @@
...
+    );
+
+    protected readonly filler = tuiDirectiveBinding(
+        TuiTextfieldComponent,
+        'fillerSetter',
+        computed((fillers = this.fillers()) => fillers?.[this.timeMode()] ?? ''),
+        {},
+    );
+
+    protected readonly inputMode = computed(() =>
```
waterplea: I think we can just have it numeric initially all the time and leave it to maskito plugin to update

projects/kit/components/input-time/input-time.directive.ts
```diff
@@ -0,0 +1,165 @@
...
+            this.computeMask({
+                ...this.options,
+                mode: this.timeMode(),
+                step: this.interactive() && !this.dropdown._content() ? 1 : 0,
+            }),
+        ),
+    );
+
+    protected readonly valueEffect = effect(() => {
+        const value = this.textfield.value();
```
waterplea: Do we need untracked here?

waterplea (approved):

projects/kit/components/input-time/input-time.directive.ts
```diff
@@ -0,0 +1,165 @@
...
+            value.length === this.timeMode().length ? TuiTime.fromString(value) : null;
+        const newValue =
+            this.acceptableValues().length && time
+                ? this.findNearestTime(time, this.acceptableValues())
+                : time;
+
+        this.control?.control?.updateValueAndValidity({emitEvent: false});
+        this.onChange(newValue);
+
+        if (newValue && newValue !== time) {
```
waterplea: Wouldn't this cause an infinite loop due to time being recreated from the string value an newValue being picked from the list, so === will never be true?

projects/kit/components/input-time/input-time.options.ts
nsbarsukov: It is naming of already existing parameter for `Time`-mask
https://maskito.dev/kit/time#time-segment-min-max

I thought it is nice to keep the same naming 🤔

projects/kit/components/input-time/input-time.directive.ts
nsbarsukov: ✅

projects/kit/components/input-time/input-time.directive.ts
nsbarsukov: ✅

projects/kit/components/input-time/input-time.directive.ts
nsbarsukov: ✅

projects/kit/components/input-time/input-time.directive.ts
nsbarsukov: Even if the such case happens, `this.textfield.value.set` will be patched with the same string and `effect` will not be triggered again (because effect depends on changes of `this.textfield.value()`; `this.textfield.value()` is a `string`)

projects/kit/components/input-time/input-time.directive.ts
nsbarsukov: This effect is replaced by `(input)` handler

https://github.com/taiga-family/taiga-ui/pull/11069/commits/05ad92a5774102c4341797cc88f1f2fec4bf5bf1

projects/kit/components/input-time/input-time.component.ts
```diff
@@ -0,0 +1,65 @@
...
+            default:
+                return 60;
+        }
+    });
+
+    protected setValue(value: string): void {
+        const mode = this.host.timeMode();
+        const time = TuiTime.fromString(value);
+
+        this.textfield.value.set(time.toString(mode));
```
nsbarsukov, line 64: `onChange` !

PR #11074, opened 2025-06-05
fix: replace InjectFlags with shallow copy to support angular v20

### LINKED ISSUE #11073: 🐞 - Angular v20 incompatibility with v3.x
### Reproduction url

-

### Description

Hey. On angular v20, project crashes with following error:

```
✘ [ERROR] No matching export in "node_modules/@angular/core/fesm2022/core.mjs" for import "InjectFlags"

    node_modules/@taiga-ui/cdk/fesm2015/taiga-ui-cdk-tokens.js:2:17:
      2 │ import { inject, InjectFlags, RendererFactory2, InjectionToken, PLATFORM_ID, ElementRef } from '@angular/core';
        ╵                  ~~~~~~~~~~~

✘ [ERROR] No matching export in "node_modules/@angular/core/fesm2022/core.mjs" for import "InjectFlags"

    node_modules/@taiga-ui/kit/fesm2015/taiga-ui-kit-components-input-date.js:3:55:
      3 │ import { Directive, Inject, ChangeDetectorRef, inject, InjectFlags, Component, ChangeDetectionStrategy, Optional, Self, ViewChild, Input, HostBindin...
        ╵                                                        ~~~~~~~~~~~

✘ [ERROR] No matching export in "node_modules/@angular/core/fesm2022/core.mjs" for import "InjectFlags"

    node_modules/@taiga-ui/kit/fesm2015/taiga-ui-kit-components-input-date-multi.js:2:36:
      2 │ import { ChangeDetectorRef, inject, InjectFlags, Component, ChangeDetectionStrategy, Optional, Self, Inject, ViewChild, Input, HostListener, HostBin...
        ╵                                     ~~~~~~~~~~~

✘ [ERROR] No matching export in "node_modules/@angular/core/fesm2022/core.mjs" for import "InjectFlags"

    node_modules/@taiga-ui/kit/fesm2015/taiga-ui-kit-components-input-date-range.js:3:36:
      3 │ import { ChangeDetectorRef, inject, InjectFlags, Component, ChangeDetectionStrategy, Optional, Self, Inject, ViewChild, forwardRef, Input, HostListe...
        ╵                                     ~~~~~~~~~~~
```

### Taiga UI version

v3.115.0

### Browser Used

Chrome

### OS Used

macOS

zip-fa:
Fixes #11073


projects/cdk/enums/inject-flags.ts
```diff
@@ -0,0 +1,18 @@
...
+     * Specifies that an injector should retrieve a dependency from any injector until reaching the
+     * host element of the current component. (Only used with Element Injector)
+     */
+    Host = 1,
+    /** Don't ascend to ancestors of the node requesting injection. */
+    Self = 2,
+    /** Skip the node that is requesting injection. */
+    SkipSelf = 4,
+    /** Inject `defaultValue` instead if token not found. */
+    Optional = 8,
```
nsbarsukov, line 21: Let's use the same syntax as [original one](https://github.com/angular/angular/blob/12.3.x/packages/compiler/src/core.ts#L261-L280):

```ts
export const enum InjectFlags {
  Default = 0,

  /**
   * Specifies that an injector should retrieve a dependency from any injector until reaching the
   * host element of the current component. (Only used with Element Injector)
   */
  Host = 1 << 0,
  /** Don't descend into ancestors of the node requesting injection. */
  Self = 1 << 1,
  /** Skip the node that is requesting injection. */
  SkipSelf = 1 << 2,
  /** Inject `defaultValue` instead if token not found. */
  Optional = 1 << 3,
}
```

nsbarsukov: Also, let's add prefix – `TuiInjectFlags`.

Otherwise, IDEs of TaigaUI's users will suggest to import it for legacy Angular versions (instead of official built-in one). We don't want to let that happen.

projects/cdk/enums/inject-flags.ts
```diff
@@ -0,0 +1,18 @@
+// In Angular v20 InjectFlags is no longer public api
+// This enum is a complete copy to keep things working
+// eslint-disable-next-line no-restricted-syntax,@typescript-eslint/naming-convention
+export enum InjectFlags {
```
nsbarsukov: Add `@deprecated` doc comment (it will make any usage of this INTERNAL type as strikethrough text for modern IDEs).

```suggestion
/**
 * In Angular v20 InjectFlags is no longer public api
 * This enum is a complete copy to keep things working
 * @deprecated Use `InjectFlags` from `@angular/core` instead!
 * @see https://github.com/angular/angular/blob/12.3.x/packages/compiler/src/core.ts#L261-L280
 */
// eslint-disable-next-line no-restricted-syntax
export enum TuiInjectFlags {
```

projects/cdk/enums/inject-flags.ts
zip-fa: 1) Renaming to TuiInjectFlags forces TS compiler to fail
2) Changing to const enum forces TS compiler to fail
3) Changing style to angular v20+ forces TS compiler to fail

```
TS2345: Argument of type InjectFlags is not assignable to parameter of type InjectFlags | undefined
```

In other words, the only working code is:

```ts
/**
 * In Angular v20 InjectFlags is no longer public api
 * This enum is a complete copy to keep things working
 * @deprecated Use `InjectFlags` from `@angular/core` instead!
 * @see https://github.com/angular/angular/blob/12.3.x/packages/compiler/src/core.ts#L261-L280
 */
// eslint-disable-next-line no-restricted-syntax,@typescript-eslint/naming-convention
export enum InjectFlags {
    /** Check self and check parent injector if needed */
    Default = 0,
    /**
     * Specifies that an injector should retrieve a dependency from any injector until reaching the
     * host element of the current component. (Only used with Element Injector)
     */
    Host = 1,
    /** Don't ascend to ancestors of the node requesting injection. */
    Self = 2,
    /** Skip the node that is requesting injection. */
    SkipSelf = 4,
    /** Inject `defaultValue` instead if token not found. */
    Optional = 8,
}
```

zip-fa:
Updated jsdoc, force-pushed the branch.

It is not possible to add Tui* prefix, make this enum `const` or change style to comply with angular v20 - we can't trick TS compiler, it throws an error:

```
TS2345: Argument of type InjectFlags is not assignable to parameter of type InjectFlags | undefined
```

zip-fa:
@nsbarsukov sorry for bothering, any chances this get merged? thank you

waterplea (approved):

PR #11091, opened 2025-06-10
feat(i18n): add arabic

serafy0:
Adds Arabic language support


waterplea (approved):
This looks good to me, however main issue with Arabic is that it's right to left. While we tried to make our DOM so that it supports this with `text-align` and `direction` CSS rules, but none of us is used to Arabic UIs so if you would be willing to try and make a StackBlitz using Arabic with some form and a few other UI components – that would be great and help us make sure Taiga UI works well in that circumstance.

PR #11104, opened 2025-06-11
chore(demo): add script for generate `llms-full.txt` & add `llms.txt`

### LINKED ISSUE #11053: 📚 - Suggestion: Add a docs/llms.txt file to support LLMs
### What is the affected URL?

_No response_

### Description

Hello!

I’ve been experimenting with Taiga UI and large language models (LLMs), and I’ve curated a list of references to the official documentation that I believe could be especially helpful for developers working with Taiga UI.

I’d like to propose adding a docs/llms.txt file — inspired by Angular’s [llms.txt initiative](https://github.com/angular/angular/blob/a70f03a9b40a4f1f0e554541f236270025b41501/adev/src/llms.txt#L4). This file would act as a directory of curated, up-to-date links to help both developers and LLMs surface accurate, high-quality information more easily.

This could be particularly valuable for newcomers and anyone relying on LLMs to understand or integrate Taiga UI into their projects.

Would the team be open to including such a file in the repository?

I’d be happy to open a PR.

Thanks so much for your time and for all the great work on Taiga UI!

### Browser Used

Chrome

### OS Used

macOS

mdlufy:
Fixes #11053


projects/demo/src/llms-full.txt
```diff
@@ -0,0 +1,19156 @@
...
+    styleUrls: ['./index.less'],
+    changeDetection,
+})
+export default class Page {
+    protected readonly examples = ['Sizes', 'Stacked'];
+
+    protected readonly valueVariants = [
+        [42],
+        [40, 30, 20, 10],
+        [13769, 10172, 3018, 2592],
```
nsbarsukov: Is it actually required for AI ?

projects/demo/src/llms-full.txt
nsbarsukov, line 1: Should we add this file to `.gitignore` (the same approach as for auto-generated `routes.txt`) ?
https://github.com/taiga-family/taiga-ui/blob/cdfddf40c4994c97a342125c666725a5d866a854/.gitignore#L40

I hope, it is not required to keep this file inside repository

projects/demo/src/llms-full.txt
```diff
@@ -0,0 +1,19156 @@
...
+    protected readonly handler = (item: Country): boolean =>
+        item.id.charCodeAt(1) % 3 === 0;
+}
+```
+
+---
+
+# components/ComboBoxLegacy
+
+- **Package**: `LEGACY`
```
nsbarsukov: Should we omit legacy/deprecated components from this file ?

projects/demo/src/llms-full.txt
```diff
@@ -0,0 +1,19156 @@
...
+
+    protected readonly exampleHtml = import('./examples/import/template.md?raw');
+
+    protected readonly example2 = {
+        HTML: import('./examples/2/index.html?raw'),
+        TypeScript: import('./examples/2/index.ts?raw'),
+        LESS: import('./examples/2/index.less?raw'),
+        'user.ts': import('./examples/2/user.ts?raw'),
+        'request.service.ts': import('./examples/2/request.service.ts?raw'),
+        'database-mock-data.ts': import('./examples/2/database-mock-data.ts?raw'),
```
nsbarsukov: Should be deleted ?

projects/demo/src/llms.txt
```diff
@@ -0,0 +1,262 @@
...
+- [Breakpoints](https://taiga-ui.dev/breakpoints)
+- [Colors](https://taiga-ui.dev/colors)
+- [Styles / Shadows](https://taiga-ui.dev/shadows)
+- [Styles / Lists](https://taiga-ui.dev/lists)
+- [Styles / Spaces](https://taiga-ui.dev/spaces)
+
+### Components
+
+- [Accordion](https://taiga-ui.dev/components/accordion)
+- [Accordion (deprecated)](https://taiga-ui.dev/components/accordion-deprecated)
```
nsbarsukov: Omit deprecated/legacy

projects/demo/src/llms-full.txt
mdlufy: What are you talking about?

projects/demo/src/llms-full.txt
mdlufy: Yes, it may not be necessary for AI, but now script parse full example code from index.ts file and it's not simple task to parse/remove concrete sections of code within a single file. I would suggest not to complicate script now, but to think about it in the future.

Now I think it would be nice to have llms-full.txt the file is in one form or another, so that it already works and can be tested, and then we can write out in more detail what we want to see in it and what not.

projects/demo/src/llms-full.txt
mdlufy: If you about these lines, it may not be necessary for AI, but these are examples of values that are still referenced in the template and that's why it's necessary for the full picture. Now script parse full example code from index.ts file and it's not simple task to parse/remove concrete sections of code within a single file. I would suggest not to complicate script now, but to think about it in the future.

Now I think it would be nice to have llms-full.txt the file is in one form or another, so that it already works and can be tested, and then we can write out in more detail what we want to see in it and what not.

projects/demo/src/llms-full.txt
mdlufy: As far as I know, not all legacy components are ready to be replaced yet. I suggest we keep it simple for now and have a complete file with all the components so that AI also has an understanding of them. I think it's better to have an overall pircutre of all the components and not overly complicate the script to start at the current stage

projects/demo/src/llms.txt
mdlufy: I'm thinking about how it's better.. The arguments to show it are the contents of all the components in llms-full.txt and all versions of legacy/non-legacy. The arguments against it are that AI can use them to generate code, but it will also warn that it is a legacy component. What if it doesn't have an analog yet? Legacy is better than no component at all.

I think we should leave all the components for now, and in the next iteration specify what we want to see and what we don't

projects/demo/src/llms-full.txt
mdlufy: Hmm.. Angular store both llms.txt and llms-full.txt and repo. But i can't found these files in bun.sh repo..

mdlufy: What are the arguments for not storing it? Just so that the repository doesn't get clogged up?

mdlufy: UPD: removed

vladimirpotekhin (approved):

PR #11131, opened 2025-06-20
fix(kit): `InputDate` fix textfield value after resetting

### LINKED ISSUE #11100: 🐞 - `tuiInputDate` problems resetting form values
### Reproduction url

https://stackblitz.com/edit/angular-nhb47vcp

### Description

We need to reset the form by calling the method. In this case, the value itself is reset, but the text (date) remains in the input field. But I also tried with a button that has a reset type. In this case, the text will disappear, but the label will hang in the air.

### Taiga UI version

4.40.0

### Browser Used

Chrome

### OS Used

macOS

vladimirpotekhin:
Fixes #11100

### DIFF
--- a/projects/demo-cypress/src/tests/input-date.cy.ts
+++ b/projects/demo-cypress/src/tests/input-date.cy.ts
@@ -1,0 +1,57 @@
+import {ChangeDetectionStrategy, Component} from '@angular/core';
+import {FormControl, FormsModule, ReactiveFormsModule} from '@angular/forms';
+import {TuiDay} from '@taiga-ui/cdk';
+import {TUI_ANIMATIONS_SPEED, TuiRoot, TuiTextfield} from '@taiga-ui/core';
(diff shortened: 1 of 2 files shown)


projects/demo-cypress/src/tests/input-date.cy.ts
```diff
@@ -0,0 +1,57 @@
...
+import {TUI_ANIMATIONS_SPEED, TuiRoot, TuiTextfield} from '@taiga-ui/core';
+import {TuiInputDate} from '@taiga-ui/kit';
+
+@Component({
+    standalone: true,
+    imports: [FormsModule, ReactiveFormsModule, TuiInputDate, TuiRoot, TuiTextfield],
+    template: `
+        <tui-root>
+            <tui-textfield>
+                <label tuiLabel>Choose a date</label>
```
nsbarsukov, line 13: Not required for test

nsbarsukov (approved):

PR #11143, opened 2025-06-23
fix(core): `Textfield` label transform only for `L` size

mdlufy:
Fixes #


projects/core/styles/components/textfield.less
```diff
@@ -153,6 +153,27 @@ tui-textfield {
         }
     }
 
+    &[data-size='l'] {
+        input:not([type='range']),
```
waterplea: Maybe we can use a local CSS variable like --t-label? So that we do not need to copy all those selectors

vladimirpotekhin (approved):

waterplea (approved):

PR #11145, opened 2025-06-23
fix(kit): fix TuiStatus for TuiBadge custom style

BrothersWar:
Fixes #


projects/kit/styles/components/badge.less
```diff
@@ -77,7 +77,6 @@ tui-badge,
     &[tuiStatus][data-appearance='neutral']::before {
```
waterplea: Looks like this whole block of rules is not necessary anymore.

splincode (approved):

mdlufy (approved):

PR #11170, opened 2025-06-25
fix(core): loader RTL layout support

### LINKED ISSUE #11169: 🐞 - TuiLoader RTL layout
### Reproduction url

https://stackblitz.com/edit/angular-zo2pvrvf?file=src%2Findex.html

### Description

TuiLoader is not displayed correctly when switched to RTL layout

Image

### Taiga UI version

4.41.0

### Browser Used

Chrome

### OS Used

Windows

dmarov:
Fixes #11169

### DIFF
--- a/projects/core/components/loader/loader.style.less
+++ b/projects/core/components/loader/loader.style.less
@@ -80,6 +80,11 @@
     animation: tuiFadeIn var(--tui-duration);
     font-size: 1rem; // to use em units as rem ones interchangeably (Safari)
 
+    @supports (inset-inline-start: -100%) {
+        left: unset;
+        inset-inline-start: -100%;
+    }
+
     &&_horizontal {
         flex-direction: row;
     }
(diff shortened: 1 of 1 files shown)


projects/core/components/loader/loader.style.less
```diff
@@ -65,16 +64,13 @@
 }
 
 .t-loader {
-    position: relative;
-    left: -100%;
+    position: absolute;
     display: flex;
+    inset: 0;
```
nsbarsukov: Bad safari support

* https://caniuse.com/mdn-css_properties_inset
* https://taiga-ui.dev/browser-support

projects/core/components/loader/loader.style.less
```diff
@@ -65,16 +64,16 @@
 }
 
 .t-loader {
-    position: relative;
-    left: -100%;
+    position: absolute;
```
waterplea: It is relative because it needs to stretch the container too when it's present. It's enough to just change `left` to `inset-inline-start`, as it would be in most of those RTL cases.

dmarov: Changed to `inset-inline-start`, but it has no support for safari 13.

dmarov: added `@supports` check, rtl will not work for safari 13

waterplea (approved):

waterplea:
Looks like that change to the margin is also needed

dmarov:
> Looks like that change to the margin is also needed

yes indeed, updated margins

vladimirpotekhin (approved):

PR #11173, opened 2025-06-26
feat(kit): `InputChip` add RTL support

waterplea:
Fixes #


projects/kit/components/input-chip/input-chip.directive.ts
```diff
@@ -116,12 +116,13 @@ export class TuiInputChipDirective<T>
     }
 
     protected scrollTo(): void {
+        const left = this.textfield.el.matches(':dir(rtl)')
+            ? -Number.MAX_SAFE_INTEGER
+            : Number.MAX_SAFE_INTEGER;
```
nsbarsukov: 😄

projects/core/components/textfield/textfield-multi/textfield-multi.style.less
```diff
@@ -55,6 +67,10 @@ tui-textfield[multi][multi] {
             clip-path: inset(0 0 0 calc(var(--t-start) / 2 - var(--t-padding) - 0.25rem));
             flex-wrap: nowrap;
 
+            &:dir(rtl) {
```
nsbarsukov, line 70: https://caniuse.com/css-dir-pseudo

projects/kit/components/input-chip/input-chip.directive.ts
```diff
@@ -116,13 +116,18 @@ export class TuiInputChipDirective<T>
...
 
     protected scrollTo(): void {
-        const left = this.textfield.el.matches(':dir(rtl)')
-            ? -Number.MAX_SAFE_INTEGER
-            : Number.MAX_SAFE_INTEGER;
+        let sign = 1;
+
+        try {
+            sign = this.textfield.el.matches(':dir(rtl)') ? -1 : 1;
+        } catch {}
```
splincode, line 123: Do we need to keep this in mind for next time?

```ts
function tuiGetSafetyDirSign(el: Element): -1 : 1 {
   try {
     return this.textfield.el.matches(':dir(rtl)') ? -1 : 1;
    } catch {
     return 1;
    }
}
```

and then

```ts
this.textfield.el.scrollTo({
    left: tuiGetSafetyDirSign(this.textfield.el) * Number.MAX_SAFE_INTEGER,
    top: Number.MAX_SAFE_INTEGER,
});
```

waterplea: I think it's fine for now, until we need to reuse it somewhere

PR #11190, opened 2025-06-26
fix(kit): `FilterByInput` should unfilter items only on unambiguous option match

### LINKED ISSUE #2346: 🐞 - Combo resets search prematurely

kit

https://taiga-ui.dev/components/combo-box#virtual

Non-Blocking

### Description

- Focus the countries box
- Type "Canad", wait.
- The list is filtered to only show Canada, which is correct.
- Now append "a" so the string reads "Canada".
- The list filtering has now been cleared and all options are visible.

Assume in my list I have the following values:
- Canada
- Canada 01
- Canada 02

By virtue of matching with the prefix, I never get the option to choose any other items that might be a partial match.

### Angular version

14.0.0

### Taiga UI version

2.57.0

- [X] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

nsbarsukov:
Fixes #2346

https://stackblitz.com/edit/taiga-issue-2346?file=src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.template.html

### Previous behavior

https://github.com/user-attachments/assets/18818ae6-1028-4c2d-b793-9610f674a51f

### New behavior

https://github.com/user-attachments/assets/de6f333f-6665-496d-9b25-8be65d9b0e92

### DIFF
--- a/projects/kit/pipes/filter-by-input/filter-by-input.pipe.ts
+++ b/projects/kit/pipes/filter-by-input/filter-by-input.pipe.ts
@@ -75,10 +75,9 @@ export class TuiFilterByInputPipe implements PipeTransform {
         query: string,
     ): readonly T[] {
         const match = this.getMatch(items, stringify, query);
+        const filtered = items.filter((item) => matcher(item, query, stringify));
 
-        return match != null
-            ? items
-            : items.filter((item) => matcher(item, query, stringify));
+        return match != null && filtered.length === 1 ? items : filtered;
     }
 
     private filter2d<T>(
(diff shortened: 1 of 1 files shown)


mdlufy (approved):

nsbarsukov:
If the value `Canada 0` is clicked, the user won’t see the list of other options when clicking on the combobox again — not until they delete it.

PR #11191, opened 2025-06-26
fix(kit): `InputFiles` make default title empty

vladimirpotekhin:
Fixes #


waterplea:
I think I already have it on main in some of my other PRs

PR #11201, opened 2025-06-27
fix(kit): `DataListWrapper` fails to emit `(itemClick)` event

### LINKED ISSUE #11197: 🐞 - `SelectOption` fails to handle `click` event listener
### Reproduction url

https://stackblitz.com/edit/tui-select-option-broken-click

### Description

```ts
import { TuiSelectOption } from '@taiga-ui/kit';
// [...]
providers: [tuiAsOptionContent(TuiSelectOption)],
```

```html
<tui-textfield [(open)]="open" (click)="open.set(true)">
  <input tuiTextfield />

  <tui-data-list *tuiTextfieldDropdown>
    <button tuiOption new (click)="log($event)" [value]="42">No logs</button>
    <button tuiOption new (click)="log($event)">Has logs</button>
  </tui-data-list>
</tui-textfield>
```

Removing `this.open.set(false)` from `handleOption` "fixes" bug. 
https://github.com/taiga-family/taiga-ui/blob/0c75661b12bd8eb202307a2ebcd5f023876eabb4/projects/core/components/textfield/textfield.component.ts#L182-L185

### Taiga UI version

4.42.0

### Browser Used

Chrome

### OS Used

macOS

nsbarsukov:
Fixes #11197

### DIFF
--- a/projects/core/components/textfield/textfield.component.ts
+++ b/projects/core/components/textfield/textfield.component.ts
@@ -181,9 +181,7 @@ export class TuiTextfieldComponent<T> implements TuiDataListHost<T>, AfterConten
 
     public handleOption(option: T): void {
         this.accessor?.setValue(option);
-        // Ensure all external (click) handlers are called before closing dropdown
-        // <button tuiOption (click)="..." />
-        setTimeout(() => this.open.set(false));
+        this.open.set(false);
     }
 
     protected get hasLabel(): boolean {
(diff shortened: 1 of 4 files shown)


projects/core/components/textfield/textfield.component.ts
```diff
@@ -181,7 +181,9 @@ export class TuiTextfieldComponent<T> implements TuiDataListHost<T>, AfterConten
 
     public handleOption(option: T): void {
         this.accessor?.setValue(option);
-        this.open.set(false);
+        // Ensure all external (click) handlers are called before closing dropdown
+        // <button tuiOption (click)="..." />
+        setTimeout(() => this.open.set(false));
```
splincode: what do you think? 

```ts
delay(0).pipe(takeUntilDestroyed(this.destroyRef)).subscribe(
 () => this.open.set(false)
);
```

waterplea: We decided it is not necessary for such cases

mdlufy (approved):

PR #11241, opened 2025-07-03
feat(cdk): improve error message for portals

splincode:


projects/cdk/classes/portals.ts
```diff
@@ -90,6 +90,10 @@ export function tuiAsPortal(portal: typeof TuiPortalService): Provider {
 
 export class TuiNoHostException extends Error {
     constructor() {
-        super(ngDevMode ? 'Portals cannot be used without TuiPortalHostComponent' : '');
+        super(
+            ngDevMode
+                ? 'Portals cannot be used without TuiPortalHostComponent; perhaps you forgot to wrap your component with tui-root.'
```
nsbarsukov: nitpicking

```suggestion
                ? 'Portals cannot be used without TuiPortalHostComponent; perhaps you forgot to wrap your application with tui-root.'
```

PR #11245, opened 2025-07-03
feat(kit): `Shimmer` add directive

mdlufy:
Fixes #

### DIFF
--- a/projects/demo/src/modules/directives/shimmer/examples/1/index.html
+++ b/projects/demo/src/modules/directives/shimmer/examples/1/index.html
@@ -32,7 +32,7 @@
     <tui-avatar-stack [tuiShimmer]="loading">
         <tui-avatar
             *ngFor="let avatar of avatars"
-            [src]="avatar.src"
+            [src]="avatar"
         />
     </tui-avatar-stack>
     <footer>
@@ -84,7 +84,7 @@
     <tui-avatar-stack>
         <tui-avatar
             *ngFor="let avatar of avatars"
-            [src]="avatar.src"
+            [src]="avatar"
             [tuiSkeleton]="loading"
         />
     </tui-avatar-stack>
--- a/projects/demo/src/modules/directives/shimmer/examples/1/index.html
+++ b/projects/demo/src/modules/directives/shimmer/examples/1/index.html
@@ -1,0 +1,129 @@
+<label tuiLabel>
+    <input
+        tuiSwitch
+        type="checkbox"
+        [(ngModel)]="loading"
+    />
+    Loading
+</label>
+
+<span [style.font]="'var(--tui-font-text-xl)'">Shimmer</span>
+
+<section
+    tuiAppearance="floating"
+    tuiCardLarge
+>
+    <header tuiHeader>
+        <hgroup
+            tuiTitle
+            [tuiShimmer]="loading"
+        >
+            <h5>You got $237&nbsp;000,42 left</h5>
+            <p tuiSubtitle>Where's the money, Lebowski?</p>
+        </hgroup>
+
+        <aside tuiAccessories>
+            <tui-avatar
+                appearance="primary"
+                src="@tui.star"
+            />
+        </aside>
+    </header>
+    <tui-avatar-stack [tuiShimmer]="loading">
+        <tui-avatar src="https://avatars.githubusercontent.com/u/87331898" />
+        <tui-avatar src="https://avatars.githubusercontent.com/u/12021443" />
+        <tui-avatar src="https://avatars.githubusercontent.com/u/35179038" />
+        <tui-avatar src="https://avatars.githubusercontent.com/u/46284632" />
+        <tui-avatar src="https://avatars.githubusercontent.com/u/10106368" />
+        <tui-avatar src="https://avatars.githubusercontent.com/u/11832552" />
+    </tui-avatar-stack>
+    <footer>
+        <button
+            appearance="secondary"
+            size="m"
+            tuiButton
+            type="button"
+        >
+            It's down there
+        </button>
+        <button
+            size="m"
+            tuiButton
+            type="button"
+        >
+            Take another look
+        </button>
+    </footer>
+</section>
+
+<span [style.font]="'var(--tui-font-text-xl)'">Skeleton</span>
+
+<section
+    tuiAppearance="floating"
+    tuiCardLarge
+>
+    <header tuiHeader>
+        <hgroup tuiTitle>
+            <h5 [tuiSkeleton]="loading && 5">
+                {{ loading ? '' : 'You got $237&nbsp;000,42 left' }}
+            </h5>
+            <p
+                tuiSubtitle
+                [tuiSkeleton]="loading && 6"
+            >
+                {{ loading ? '' : "Where's the money, Lebowski?" }}
+            </p>
+        </hgroup>
+
+        <aside tuiAccessories>
+            <tui-avatar
+                appearance="primary"
+                src="@tui.star"
+                [tuiSkeleton]="loading"
+            />
+        </aside>
+    </header>
+    <tui-avatar-stack>
+        <tui-avatar
+            src="https://avatars.githubusercontent.com/u/87331898"
+            [tuiSkeleton]="loading"
+        />
+        <tui-avatar
+            src="https://avatars.githubusercontent.com/u/12021443"
+            [tuiSkeleton]="loading"
+        />
+        <tui-avatar
+            src="https://avatars.githubusercontent.com/u/35179038"
+            [tuiSkeleton]="loading"
+        />
+        <tui-avatar
+            src="https://avatars.githubusercontent.com/u/46284632"
+            [tuiSkeleton]="loading"
+        />
+        <tui-avatar
+            src="https://avatars.githubusercontent.com/u/10106368"
+            [tuiSkeleton]="loading"
+        />
+        <tui-avatar
+            src="https://avatars.githubusercontent.com/u/11832552"
+            [tuiSkeleton]="loading"
+        />
+    </tui-avatar-stack>
+    <footer>
+        <button
+            appearance="secondary"
+            size="m"
+            tuiButton
+            type="button"
+        >
+            It's down there
+        </button>
+        <button
+            size="m"
+            tuiButton
+            type="button"
+        >
+            Take another look
+        </button>
+    </footer>
+</section>
--- a/projects/demo/src/modules/directives/shimmer/examples/1/index.html
+++ b/projects/demo/src/modules/directives/shimmer/examples/1/index.html
@@ -30,12 +30,10 @@
         </aside>
     </header>
     <tui-avatar-stack [tuiShimmer]="loading">
-        <tui-avatar src="https://avatars.githubusercontent.com/u/87331898" />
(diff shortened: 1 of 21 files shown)


projects/kit/directives/shimmer/shimmer.directive.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    @Input('tuiShimmer')
+    public set shimmerValue(cache: boolean) {
+        this.shimmer.set(cache);
+    }
+
+    public ngOnChanges(): void {
+        if (!this.isBrowser) {
+            return;
+        }
```
mdlufy, line 52: Need to prevent the error `this.el.animate is not a function` on prerender build

projects/kit/directives/shimmer/shimmer.directive.ts
```diff
@@ -0,0 +1,73 @@
...
+
+        this.animation?.commitStyles();
+        this.animation?.cancel();
```
mdlufy, line 63: Fast switching `tuiShimmer` state force an Error: Uncaught (in promise): AbortError: The user aborted a request.

https://github.com/user-attachments/assets/f9b519df-5da3-4f2c-88da-6a7904779cfe

mdlufy: Standart behaviour due to .cancel() call on current finished promise
https://developer.mozilla.org/en-US/docs/Web/API/Animation/cancel#exceptions

projects/kit/directives/shimmer/shimmer.directive.ts
```diff
@@ -0,0 +1,83 @@
...
+                })
+                .catch((error: unknown) => {
+                    // fast switching animation state force AbortError
+                    // https://developer.mozilla.org/en-US/docs/Web/API/Animation/cancel#exceptions
+                    if (error instanceof DOMException && error.name === 'AbortError') {
+                        return;
+                    }
+
+                    console.error(error);
+                });
```
mdlufy: For fix error 
https://github.com/taiga-family/taiga-ui/pull/11245/files#r2183956998

projects/kit/directives/shimmer/shimmer.directive.ts
```diff
@@ -0,0 +1,83 @@
...
+    },
+})
+export class TuiShimmer implements OnChanges {
+    private readonly isBrowser = isPlatformBrowser(inject(PLATFORM_ID));
+    private readonly el = tuiInjectElement();
+    private animation?: Animation;
+
+    protected readonly nothing = tuiWithStyles(TuiShimmerStyles);
+
+    public shimmer = signal(false);
```
waterplea: No need for it to be a signal

projects/demo/src/modules/directives/shimmer/index.html
```diff
@@ -0,0 +1,104 @@
...
+        </p>
+
+        <h3>When to Use</h3>
+
+        <p>
+            To indicate loading states in cards, headers, lists, avatars, and other UI elements. When you want to
+            visually communicate that content is loading, rather than missing or frozen.
+        </p>
+
+        <h3>How It Works</h3>
```
waterplea: How it works is unnecessary. Instead let's add to the previous bullet that Shimmer is used when you have cached data that is currently being refreshed and if you have no data at all — a better choice would be Skeleton.

waterplea (changes requested):
Disabled animations example is wrong: what I meant was we need to react to --tui-duration. If it is set to 0 — we need to not apply the animations (both programmatical one and the CSS one) and instead just set disabled opacity. Also we need to disable pointer events, I think, and add attr.inert (nevermind browser support, it would be graceful degradation).

projects/demo/src/modules/directives/shimmer/examples/import/template.md
```diff
@@ -0,0 +1,13 @@
+```html
+<!-- Usual HTML element -->
+<hgroup
```
waterplea: Let's make this simpler, with no comments, just:
```html
<div [tuiShimmer]="shimmer">
  I'm shimmering
</div>
```

mdlufy: Okay, updated example on usual HTML element.

I think it's important to show that shimmer can be used not only on regular elements, but also on other components of the Taiga UI. Therefore, I suggest leaving the second example with the calendar. 

P.S. this file will be included to llms-full.txt and the model will be able to explicitly answer: "Is it possible to use shimmer on other Taiga UI components, not only usual HTML elements?" and show a minimal example.

projects/demo/src/modules/directives/shimmer/index.html
mdlufy: Done

projects/kit/directives/shimmer/shimmer.directive.ts
mdlufy: Recoded to usual input

mdlufy:
> Disabled animations example is wrong: what I meant was we need to react to --tui-duration. If it is set to 0 — we need to not apply the animations (both programmatical one and the CSS one) and instead just set disabled opacity. Also we need to disable pointer events, I think, and add attr.inert when shimmering (never mind browser support, it would be graceful degradation).

Updated, added react to `--tui-duration` value. @waterplea, will that be enough or we need to dynamically track `--tui-duration` value if it changes (via mutation observer, e.g.)?

projects/kit/directives/shimmer/shimmer.directive.ts
```diff
@@ -0,0 +1,89 @@
...
+                    }
+
+                    console.error(error);
+                });
+        }
+    }
+
+    protected get disabled(): boolean {
+        return (
+            this.isBrowser &&
```
mdlufy: Prevent error: `getComputedStyle is not defined` on prerender build

projects/demo/src/modules/directives/shimmer/examples/2/index.html
```diff
@@ -0,0 +1,59 @@
...
+
+        <aside tuiAccessories>
+            <tui-avatar
+                appearance="primary"
+                src="@tui.star"
+            />
+        </aside>
+    </header>
+    <tui-avatar-stack
+        [style.--tui-duration]="0"
```
waterplea: Just add it to :host of this example

projects/kit/directives/shimmer/shimmer.directive.ts
waterplea: No need for a getter, update the state in ngOnChanges, otherwise it will do this expensive calculation each change detection cycle.

projects/kit/directives/shimmer/shimmer.style.less
```diff
@@ -0,0 +1,37 @@
...
+        mask-repeat: no-repeat;
+        mask-composite: subtract;
+        animation: tuiShimmer 8s infinite linear;
+    }
+
+    &._disabled {
+        animation: none;
+        mask: none;
+        pointer-events: none;
+        cursor: default;
```
waterplea: Does it matter if there's pointer-events: none? I think we should move those to _shimmer class since we do not want people clicking shimmering UI elements either.

projects/kit/directives/shimmer/shimmer.directive.ts
waterplea: Let's just change it to empty `catch`, no need to dig into the error.

projects/kit/directives/shimmer/shimmer.style.less
mdlufy: Yes, cursor: default is unnesessary

projects/demo/src/modules/directives/shimmer/examples/1/index.html
```diff
@@ -0,0 +1,129 @@
...
+            />
+        </aside>
+    </header>
+    <tui-avatar-stack [tuiShimmer]="loading">
+        <tui-avatar src="https://avatars.githubusercontent.com/u/87331898" />
+        <tui-avatar src="https://avatars.githubusercontent.com/u/12021443" />
+        <tui-avatar src="https://avatars.githubusercontent.com/u/35179038" />
+        <tui-avatar src="https://avatars.githubusercontent.com/u/46284632" />
+        <tui-avatar src="https://avatars.githubusercontent.com/u/10106368" />
+        <tui-avatar src="https://avatars.githubusercontent.com/u/11832552" />
```
nsbarsukov: links repeat twice in this file.
Let's create array `users`/`avatars` with these links + use `*ngFor`

nsbarsukov: btw

https://avatars.githubusercontent.com/u/87331898
equals to
https://avatars.githubusercontent.com/mdlufy

projects/kit/directives/shimmer/shimmer.directive.ts
```diff
@@ -0,0 +1,83 @@
...
+        } else {
+            this.animation = this.el.animate([{opacity: 0, offset: 0.5}, {opacity: 1}], {
+                duration: 800,
+            });
+
+            this.animation.finished
+                .then(() => {
+                    this.el.style.opacity = '';
+                })
+                .catch(() => {});
```
nsbarsukov, line 79: `noop` ?

mdlufy: Yes 
https://github.com/taiga-family/taiga-ui/pull/11245#discussion_r2189642956

projects/demo/src/modules/directives/shimmer/examples/2/index.ts
```diff
@@ -0,0 +1,40 @@
...
+    ],
+    templateUrl: './index.html',
+    styleUrls: ['./index.less'],
+    changeDetection,
+})
+export default class Example {
+    protected loading = true;
+
+    public avatars = [
+        {src: 'https://avatars.githubusercontent.com/mdlufy'},
```
waterplea: Why not just strings here and in the first example?

projects/demo/src/modules/directives/shimmer/examples/1/index.ts
```diff
@@ -0,0 +1,47 @@
...
+        TuiTitle,
+    ],
+    templateUrl: './index.html',
+    styleUrls: ['./index.less'],
+    changeDetection,
+})
+export default class Example {
+    protected loading = true;
+
+    public avatars = [
```
waterplea: protected readonly?

projects/kit/directives/shimmer/shimmer.directive.ts
```diff
@@ -0,0 +1,83 @@
...
+    @Input('tuiShimmer')
+    public shimmer = false;
+
+    public ngOnChanges(): void {
+        if (!this.isBrowser) {
+            return;
+        }
+
+        this.disabled =
+            parseFloat(
```
waterplea: Just `!parseFloat(...)`

projects/kit/directives/shimmer/shimmer.directive.ts
```diff
@@ -0,0 +1,83 @@
...
+                getComputedStyle(this.el).getPropertyValue('--tui-duration').trim(),
```
waterplea: `trim()` is unnecessary, `parseFloat('     200ms   ')` works fine.

waterplea (approved):

PR #11252, opened 2025-07-04
feat(kit): `InputTime` supports `[prefix]` / `[postfix]`

### LINKED ISSUE #11199: 🚀 - `InputTime` supports `[prefix]` / `[postfix]`
### Description

https://taiga-ui.dev/next/components/input-time

Image

nsbarsukov:
Fixes #11199

### DIFF
--- a/projects/kit/components/input-time/input-time.component.ts
+++ b/projects/kit/components/input-time/input-time.component.ts
@@ -57,10 +57,6 @@ export class TuiInputTimeComponent {
     });
 
     protected setValue(value: string): void {
-        const mode = this.host.timeMode();
-        const time = TuiTime.fromString(value);
-
-        this.control.onChange(time);
-        this.textfield.value.set(time.toString(mode));
+        this.host.setValue(TuiTime.fromString(value));
     }
 }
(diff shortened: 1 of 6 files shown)


projects/kit/components/input-time/input-time.component.ts
```diff
@@ -57,10 +57,6 @@ export class TuiInputTimeComponent {
     });
 
     protected setValue(value: string): void {
-        const mode = this.host.timeMode();
```
waterplea, line 60: Why all this is no longer needed?

waterplea (approved):
I wonder if there's a way to generalize pre/postfixes for other controls as well with a host directive or something... 🤔

projects/kit/components/input-time/input-time.component.ts
nsbarsukov: `TuiInputTimeDirective.setValue` implements the same logic (as all deleted lines)

https://github.com/taiga-family/taiga-ui/blob/c283c56f7d10caea1d443563a27f15233eacc883/projects/kit/components/input-time/input-time.directive.ts#L121-L123

PR #11296, opened 2025-07-11
feat(kit): `Drawer` support safe-area top/bottom

mdlufy:
Partially solved #10963

### DIFF
--- a/projects/kit/components/drawer/drawer.style.less
+++ b/projects/kit/components/drawer/drawer.style.less
@@ -2,8 +2,8 @@
 
 :host {
     position: fixed;
-    top: 3rem;
-    bottom: 0;
+    top: ~'max(3rem, env(safe-area-inset-top))';
+    bottom: env(safe-area-inset-bottom);
     inline-size: 36.25rem;
     max-inline-size: calc(100vw - 3rem);
     background: var(--tui-background-elevation-1);


splincode (approved):

projects/kit/components/drawer/drawer.style.less
```diff
@@ -2,8 +2,8 @@
 
 :host {
     position: fixed;
-    top: 3rem;
-    bottom: 0;
+    top: ~'max(3rem, env(safe-area-inset-top))';
+    bottom: env(safe-area-inset-bottom);
```
waterplea: This would cause a hole underneath, should we add box-shadow? Also there's a demo where top is 0, please change it to env(...) and add shadow in both directions there

mdlufy: For default it's okay, we need to add border-bottom-left-radius

image

For modal, yes, need to add

image

mdlufy: UPD: done
https://github.com/taiga-family/taiga-ui/pull/11296#issue-3221684257

waterplea: No, for non-modal we also need it. It is designed to come from the edge of the screen, no need for the bottom radius.

waterplea: Actually, it's probably better to add padding, rather than changing bottom

mdlufy: Updated:

Image

Image

projects/kit/components/drawer/drawer.style.less
```diff
@@ -11,6 +11,14 @@
     border-top-left-radius: 1.25rem;
     border-top-right-radius: 1.25rem;
 
+    &[style*='top: 0'] {
+        border-radius: 0;
+
+        .t-aside {
+            padding-block-start: env(safe-area-inset-top);
+        }
+    }
```
mdlufy: Padding on `:host` not correctly working due to backdrop.. in `:before`, so we can workaround through inline style `top: 0`, or to use `::ng-deep` in the example component:

```
.drawer {
    top: 0;
    border-radius: 0;
    inline-size: 20rem;

    &::ng-deep > aside {
        padding-block-start: env(safe-area-inset-top);
    }
}
```

projects/kit/components/drawer/drawer.style.less
```diff
@@ -11,6 +11,14 @@
     border-top-left-radius: 1.25rem;
     border-top-right-radius: 1.25rem;
 
+    &[style*='top: 0'] {
```
waterplea: This is breaking change and also bad. We cannot do that. Let's just change the example from `top: 0` to `top: env(safe-area-inset-top)` and add white shadow

PR #11297, opened 2025-07-11
fix(core): support invalid highlight inside `tuiTable`

### LINKED ISSUE #11281: 🐞 - TuiInputChip invalid highlight inside TuiTable
### Reproduction url

https://stackblitz.com/edit/angular-eb6xsshs?file=src%2Fapp%2Fapp.template.html

### Description

TuiInputChip inside TuiTable does not have a highlight for an invalid state of control

### Taiga UI version

4.44.0

### Browser Used

Chrome

### OS Used

macOS

splincode:
Fixes #11281

image

### DIFF
--- a/projects/core/styles/theme/appearance/table.less
+++ b/projects/core/styles/theme/appearance/table.less
@@ -34,8 +34,9 @@ TODO: remove :not([tuiWrapper]) after legacy controls deletion
         outline: 0.125rem solid var(--tui-border-focus);
     });
 
-    &:invalid,
-    &[data-mode~='invalid'] {
+    &:invalid:not([data-mode]),
+    &[data-mode~='invalid'],
+    &.tui-invalid:not([data-mode]) {
         background-origin: border-box;
         background-position: top right;
         background-repeat: no-repeat;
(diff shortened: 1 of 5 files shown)


projects/core/styles/theme/appearance/table.less
```diff
@@ -35,7 +35,8 @@ TODO: remove :not([tuiWrapper]) after legacy controls deletion
     });
 
     &:invalid,
```
waterplea: Should we add :not([data-mode]) here as well? We should be able to set [invalid]="false" to hide native :invalid state for textfields.

projects/demo/src/modules/components/input-chip/examples/9/index.html
```diff
@@ -0,0 +1,53 @@
...
+>
+    <thead>
+        <tr tuiThGroup>
+            <th tuiTh>Options</th>
+        </tr>
+    </thead>
+    <tbody tuiTbody>
+        <tr tuiTr>
+            <td
+                *tuiCell="'column'"
```
waterplea: Let's remove dynamic table and have a simple static one so to not overcomplicate the example

splincode: fixed

projects/core/styles/theme/appearance/table.less
splincode: fixed

vladimirpotekhin (approved):

Issue #10938, opened 2025-04-29
🚀 - Add support for icon fonts


FeironoX5:
## Which @taiga-ui/* package(s) are relevant/related to the feature request?
`core`

## Description
Lots of popular icon sets use icon fonts instead of SVGs to display icons (i.e. Material Icons to Material Symbols transition), as they load faster and provide better customization options, such as specifiyng weight. TaigaUI doesn't have any way to use ones yet, which would be a great addition without changing the already implemented approaches (such as using `:before`&`:after` in `iconStart/End`).

waterplea:
We should add mode resolver by icon name and use following:
- `@image.xxx` for the same mode as now but with `background` rather than `mask` for colored icons
- `@font.xxx` for using icon name (with `@font.` removed) as `content` inside pseudo elements instead of using it as mask/background
- `@icon.xxx` current mode (should be default fallback if doesn't match the ones above so `@icon` is just convenience)

FeironoX5:
I have encountered a problem while making changes. `TuiIconPipe` assumes that the resolver will return the result only as an image URL, without the added `"url()"` wrapper, but logic here is dependant on icon type, so, to extract adding the `"url()"` logic from icon resource resolvers, I have to sacrifice code cleanliness.

Drawbacks:  
- The logic for adding the `"url()"` wrapper has to be moved to the places where resolvers are used => code duplication in all places except the pipe  
- Some resource resolvers lose their meaning  

Before:  
```typescript
export const TUI_ICON_IMAGE_RESOURCE_RESOLVER = new InjectionToken<TuiStringHandler<string>>(
    ngDevMode ? 'TUI_ICON_IMAGE_RESOURCE_RESOLVER' : '',
    {
        factory: () => {
            return (icon) => `url(${icon.replace('@image.', '')})`;
        },
    },
);

export const TUI_ICON_FONT_RESOURCE_RESOLVER = new InjectionToken<TuiStringHandler<string>>(
    ngDevMode ? 'TUI_ICON_FONT_RESOURCE_RESOLVER' : '',
    {
        factory: () => {
            return (icon) => `${icon.replace('@font.', '')}`;
        },
    },
);
```  

After:  
```typescript
export const TUI_ICON_IMAGE_RESOURCE_RESOLVER = new InjectionToken<TuiStringHandler<string>>(
    ngDevMode ? 'TUI_ICON_IMAGE_RESOURCE_RESOLVER' : '',
    {
        factory: () => {
[...]

PR #11311, opened 2025-07-13
feat(core, kit): implement icon mode resolution system (#10938)

FeironoX5:
Refs #10938

- Icon resolution is now handled by TUI_ICON_MODE_RESOLVER for icon mode resolution (@font/@image(.data/.mask)/@icon) and icon resource resolvers: TUI_ICON_SVG_RESOURCE_RESOLVER, TUI_ICON_IMAGE_RESOURCE_RESOLVER, TUI_ICON_FONT_RESOURCE_RESOLVER
- Updated affected components, styles, and some examples accordingly

---

### Next steps:
- [ ] Set default --t-font-icon CSS variable
- [ ] Implement resource resolvers override mechanism
- [ ] Update demo, provide more examples
- [ ] Replace all TUI_ICON_RESOLVER usages


waterplea (commented):
1. Let'd do [data-icon-start/end="icon/img/font"] instead of classes.
2. I don't think we need the icon mode resolver, I think it's ok to hardcode `@font.`/`@img.`/`@icon.` prefixes. We will just use last one as a default to maintain backwards compatibility with `@tui.` or any custom icons people might have.
3. I don't want to complicate the resolver, I think in case of `@font.` we can just put whatever's after it to --t-icon-xxx CSS variable as is, no need to resolve it, and in all other cases just proceed to pass the string to the resolver as we do now.

erperejildo:
@FeironoX5 why was this closed? Any plans to work on this again soon?

waterplea:
@FeironoX5 it was opened from a different branch: https://github.com/taiga-family/taiga-ui/pull/11592

PR #11322, opened 2025-07-15
feat(core): support nullable icon in notification

splincode:


projects/demo/src/modules/components/alert/index.html
```diff
@@ -110,7 +110,7 @@
             <tr
                 name="[icon]"
                 tuiDocAPIItem
-                type="PolymorpheusContent"
+                type="TuiStringHandler<string> | string"
```
mdlufy, line 113: Why remove PolymorpheusContent?

splincode: the signature does not match reality

image

waterplea: Why should it support null then? Why not empty string?

PR #11325, opened 2025-07-15
fix(kit): correct open dropdown in textfield

### LINKED ISSUE #11275: 🐞 - `SelectLike` does not open dropdown on `Chevron` click (directly on icon)
### Description

https://taiga-ui.dev/components/textfield#dropdown

https://github.com/user-attachments/assets/5ad77efd-3118-489d-8345-5183f2b473ed

BUT (without bug)
https://taiga-ui.dev/components/select/API

https://github.com/user-attachments/assets/adf10514-e248-40b7-959e-344c86f72e13

### Taiga UI version

4.44.0

### Browser Used

Chrome

### OS Used

macOS

splincode:
Fixes #11275

### DIFF
--- a/projects/core/components/textfield/textfield.component.ts
+++ b/projects/core/components/textfield/textfield.component.ts
@@ -171,7 +171,6 @@ export class TuiTextfieldBaseComponent<T>
 
         if (
             !this.dropdownOpen.tuiDropdownEnabled ||
-            !this.cva?.interactive() ||
             this.input?.nativeElement.matches('input:read-only,textarea:read-only')
         ) {
             return;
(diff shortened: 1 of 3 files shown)


mdlufy (approved):

projects/core/components/textfield/textfield.component.ts
```diff
@@ -171,7 +171,7 @@ export class TuiTextfieldBaseComponent<T>
 
         if (
             !this.dropdownOpen.tuiDropdownEnabled ||
-            !this.cva?.interactive() ||
+            !(this.cva?.interactive() ?? this.control?.enabled) ||
```
waterplea: Do we even need this line? Wouldn't the next line cover this case? 🤔

splincode: Before:

`this.cva` is undefined -> `this.cva?.interactive()` ->  `!(undefined)` -> if (true)


After:

`this.cva ?? this.control` -> is (undefined ?? {enabled: true}) ->  `!(undefined ?? true)` -> if (false)

waterplea: I mean if it's not interactive then next line would match anyway, I think we can just remove this line altogether.

splincode: However it's interactive 🤔 or I don't understand what is mean interactive mode

image
image

waterplea (approved):

PR #11326, opened 2025-07-15
fix(kit): `ComboBox` / `InputTime` has broken cleaner

### LINKED ISSUE #11323: 🐞 - `ComboBox` / `InputTime` has broken cleaner
### Reproduction url

https://taiga-ui.dev/components/combo-box#basic
https://taiga-ui.dev/components/input-time#mode

### Description

Press cleaner icon. Value is not cleaned

Relates:
* https://github.com/taiga-family/taiga-ui/pull/11313

`update` function is executed too early (before `textfield.value.set` inside `setValue` updated DOM)
https://github.com/taiga-family/taiga-ui/blob/6dbfd964d721717b64f6366899fd08b0fd792a97/projects/cdk/utils/dom/value.ts#L43-L45
### Taiga UI version

4.45.0

### Browser Used

Chrome

### OS Used

macOS

nsbarsukov:
Fixes #11323

### DIFF
--- a/projects/kit/components/combo-box/combo-box.directive.ts
+++ b/projects/kit/components/combo-box/combo-box.directive.ts
@@ -11,7 +11,6 @@ import {
 import {tuiAsControl, TuiControl} from '@taiga-ui/cdk/classes';
 import {TUI_ALLOW_SIGNAL_WRITES, TUI_STRICT_MATCHER} from '@taiga-ui/cdk/constants';
 import type {TuiStringMatcher} from '@taiga-ui/cdk/types';
-import {tuiInjectElement} from '@taiga-ui/cdk/utils/dom';
 import type {TuiDataListAccessor} from '@taiga-ui/core/components/data-list';
 import {tuiAsOptionContent} from '@taiga-ui/core/components/data-list';
 import type {TuiTextfieldAccessor} from '@taiga-ui/core/components/textfield';
@@ -51,7 +50,6 @@ export class TuiComboBox<T>
     extends TuiControl<T | string | null>
     implements TuiTextfieldAccessor<T>
 {
-    private readonly el = tuiInjectElement<HTMLInputElement>();
     private readonly host: TuiTextfieldComponent<T> = inject(TuiTextfieldComponent);
     private readonly textfield: TuiTextfieldDirective<T> = inject(TuiTextfieldDirective);
     private readonly open = tuiDropdownOpen();
--- a/projects/kit/components/combo-box/combo-box.directive.ts
+++ b/projects/kit/components/combo-box/combo-box.directive.ts
@@ -60,6 +60,9 @@ export class TuiComboBox<T>
     private readonly itemsHandlers: TuiItemsHandlers<T | string> =
         inject(TUI_ITEMS_HANDLERS);
 
+    private readonly defaultAccessor: TuiTextfieldAccessor<T> =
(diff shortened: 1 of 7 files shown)


splincode (approved):

projects/kit/components/combo-box/combo-box.directive.ts
```diff
@@ -127,13 +130,8 @@ export class TuiComboBox<T>
...
     public setValue(value: T | null): void {
-        const stringified = this.stringify(value);
-
-        if (stringified !== this.textfield.value()) {
-            this.textfield.value.set(stringified);
-            setTimeout((end = stringified.length) => this.el.setSelectionRange(end, end));
-        }
-
+        this.defaultAccessor.setValue(value);
+        setTimeout((end = this.el.value.length) => this.el.setSelectionRange(end, end));
```
waterplea: Isn't this a default behavior when input value is programmatically updated

projects/kit/components/input-time/input-time.directive.ts
```diff
@@ -65,6 +65,8 @@ export class TuiInputTimeDirective
     private readonly fillers = toSignal(inject(TUI_TIME_TEXTS));
     private readonly prefix = signal('');
     private readonly postfix = signal('');
+    private readonly defaultAccessor: TuiTextfieldAccessor<TuiTime | null> =
```
waterplea: We already have `private readonly textfield = inject(TuiTextfieldDirective);`

projects/kit/components/input-time/input-time.directive.ts
```diff
@@ -120,7 +122,12 @@ export class TuiInputTimeDirective
 
     public setValue(value: TuiTime | null): void {
         this.onChange(value);
-        this.textfield.value.set(this.stringify(value));
+
+        if (value) {
+            this.textfield.value.set(this.stringify(value));
+        } else {
+            this.defaultAccessor.setValue(value);
```
waterplea: What if we instead update the stringify method on tui-textfield? Then we wouldn't need branching here?

projects/kit/components/combo-box/combo-box.directive.ts
```diff
@@ -60,6 +60,9 @@ export class TuiComboBox<T>
     private readonly itemsHandlers: TuiItemsHandlers<T | string> =
         inject(TUI_ITEMS_HANDLERS);
 
+    private readonly defaultAccessor: TuiTextfieldAccessor<T> =
```
waterplea: We already have `private readonly textfield: TuiTextfieldDirective<T> = inject(TuiTextfieldDirective);`

projects/kit/components/input-time/input-time.directive.ts
nsbarsukov: We have the such case (see providers of the component)
https://taiga-ui.dev/components/input-time#dropdown-datalist

Should we remove `stringify` from the example and always set `stringify` internally ? 🤔

waterplea (approved):

projects/kit/components/input-time/input-time.directive.ts
nsbarsukov: Discussed with @waterplea personally.
We decided not to overwrite `stringify` inside `TuiInputTimeDirective` because prefix/postfix can be not required for datalist

PR #11327, opened 2025-07-16
feat(addon-mobile): add support for required option in sheet dialog

### LINKED ISSUE #11082: 🚀 - Add param `required` to the `TuiSheetDialogService` like the `TuiDialogService`
### Description

Need to add a `required` parameter to the `TuiSheetDialogService` like the `TuiDialogService` with exactly the same functionality because there is no way to intercept a cross click, overlay click or Escape click with emits an error into Observable

rashiddok:
Description:
- add **required** parameter to sheet dialog options
- add support **required** option in  TuiSheetDialogComponent
- sheet dialog demo update with required option

Closes #11082

### DIFF
--- a/projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
+++ b/projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
@@ -19,7 +19,7 @@ import {exhaustMap, filter, isObservable, merge, of, Subject, take} from 'rxjs';
 
 import type {TuiSheetDialogOptions} from './sheet-dialog.options';
 
-const REQUIRED_ERROR = new Error('Required dialog was dismissed');
+const REQUIRED_ERROR = new Error(ngDevMode ? 'Required dialog was dismissed' : '');
 
 @Component({
     standalone: true,
--- a/projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
+++ b/projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
@@ -19,6 +19,8 @@ import {exhaustMap, filter, isObservable, merge, of, Subject, take} from 'rxjs';
 
 import type {TuiSheetDialogOptions} from './sheet-dialog.options';
 
+const REQUIRED_ERROR = new Error('Required dialog was dismissed');
+
 @Component({
     standalone: true,
     selector: 'tui-sheet-dialog',
@@ -67,7 +69,7 @@ export class TuiSheetDialogComponent<I> implements AfterViewInit {
             filter(Boolean),
             takeUntilDestroyed(),
         )
-        .subscribe(() => this.context.$implicit.complete());
+        .subscribe(() => this.close());
 
     public ngAfterViewInit(): void {
         this.el.scrollTop = this.initial;
@@ -91,5 +93,13 @@ export class TuiSheetDialogComponent<I> implements AfterViewInit {
                 .map((e) => e.nativeElement.offsetTop - this.context.offset)
                 .concat(this.el.clientHeight ?? Infinity)[this.context.initial] ?? 0
         );
+    }
+
+    private close(): void {
+        if (this.context.required) {
+            this.context.$implicit.error(REQUIRED_ERROR);
+        } else {
+            this.context.$implicit.complete();
+        }
     }
 }
--- a/projects/demo/src/modules/components/sheet-dialog/index.html
+++ b/projects/demo/src/modules/components/sheet-dialog/index.html
@@ -157,7 +157,7 @@
                 type="boolean"
                 [(value)]="required"
             >
-                Cross click, overlay click or Escape click emits an error into
+                Dismissing dialog emits an error into
                 <code>Observable</code>
                 (you can catch it with "catch" operator or onError handler)
             </tr>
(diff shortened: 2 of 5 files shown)


projects/demo/src/modules/components/sheet-dialog/index.html
```diff
@@ -131,6 +143,15 @@
                 <code>--tui-offset</code>
                 )
             </ng-template>
+            <ng-template
+                documentationPropertyName="required"
+                documentationPropertyType="boolean"
+                [(documentationPropertyValue)]="required"
+            >
+                Cross click, overlay click or Escape click emits an error into
```
waterplea: Let's just say "Dismissing dialog ..." since there's no cross here and no escape on mobile

waterplea (approved):

projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
```diff
@@ -19,6 +19,8 @@ import {exhaustMap, filter, isObservable, merge, of, Subject, take} from 'rxjs';
 
 import type {TuiSheetDialogOptions} from './sheet-dialog.options';
 
+const REQUIRED_ERROR = new Error('Required dialog was dismissed');
```
splincode:
```ts
const REQUIRED_ERROR = new Error(ngDevMode ? 'Required dialog was dismissed' : '');
```

projects/demo/src/modules/components/sheet-dialog/index.html
```diff
@@ -42,30 +44,40 @@
...
+                #sheetTemplate
+                let-completeWith="completeWith"
+                let-observer
             >
                 <p [style.flex-grow]="1">
                     Karl Gambolputty de von
                     Ausfern-schplenden-schlitter-crasscrenbon-fried-digger-dingle-dangle-dongle-dungle-burstein-von-knacker-thrasher-apple-banger-horowitz-ticolensic-grander-knotty-spelltinkle-grandlich-grumblemeyer-spelterwasser-kurstlich-himbleeisen-bahnwagen-gutenabend-bitte-ein-nürnburger-bratwustle-gerspurten-mitzweimache-luber
                     von Hautkopft of Ulm was the last-surviving relative of Johann Gambolputty de von.
                 </p>
-                <footer tuiFloatingContainer>
```
mdlufy: Why remove tuiFloatingContainer?

projects/demo/src/modules/components/sheet-dialog/index.html
```diff
@@ -131,6 +143,15 @@
...
             </ng-template>
+            <ng-template
+                documentationPropertyName="required"
+                documentationPropertyType="boolean"
+                [(documentationPropertyValue)]="required"
+            >
+                Dismissing dialog emits an error into
+                <code>Observable</code>
+                (you can catch it with "catch" operator or onError handler)
+            </ng-template>
```
mdlufy: Let's update to new approach:

```
<tr
    name="[closeable]"
    tuiDocAPIItem
    type="boolean"
    [(value)]="required"
 >
     Dismissing dialog emits an error into
     <code>Observable</code>
     (you can catch it with "catch" operator or onError handler)
</tr>
```

projects/demo/src/modules/components/sheet-dialog/index.html
rashiddok: tuiFloatingContainer adds `flex-direction: column` that breaks buttons view

rashiddok: returned `tuiFloatingContainer` directive and added styles for buttons container

projects/demo/src/modules/components/sheet-dialog/index.html
rashiddok: done

projects/demo/src/modules/components/sheet-dialog/index.html
mdlufy: > tuiFloatingContainer adds `flex-direction: column` that breaks buttons view

Hmm... It looks like it's fine. Can you show me a screenshot/video, how does it break?

image

rashiddok: It's fine until there is only one button. When more buttons added they take most of screen space and when dialog gets dissmised with swipe down buttons stays on place while dialog content goes behind buttons.

https://github.com/user-attachments/assets/c286cd4d-618c-427e-8add-a7290a456e26

mdlufy: > When more buttons added they take most of screen space and when dialog gets dissmised with swipe down buttons stays on place while dialog content goes behind buttons.

`When more buttons added` It's designed to using only with 2 buttons max)

`when dialog gets dissmised with swipe down buttons stays on place while dialog content goes behind buttons.` 
That's because it's based on sticky + and bottom. Would you prefer that when the SheetDialog is rolled down, the button immediately goes down too?

In result, I don't think it's worth abandoning the tuiFloatingContainer in the SheetDialog example or changing it

@waterplea, what do you think about such a case? Should we rework the logic of the tuiFloatingContainer, so that when scrolling down from the moment when the SheetDialog itself begins to collapse, the button would also go down?

waterplea: I think floating container is working as expected, that's the point of it, to keep buttons visible, there's no meaningful distinction between scrolling the sheet and swiping it down to close.

rashiddok: Removed "Close" button and and styles from tuiFloatingContainer.

mdlufy: Okay, let's merge it

PR #11328, opened 2025-07-16
fix(core): add fine-grained reactivity update in textfield

### LINKED ISSUE #11041: 🐞 - `TuiTextfieldComponent` change detection issue with `content` input
### Reproduction url

https://stackblitz.com/edit/angular-exwxdhb6?file=src%2Fapp%2Fapp.template.html

### Description

Try to set value from http response. Example imitates http request with `timer(500)`. Textfield does not display any content until you click on it

https://github.com/taiga-family/taiga-ui/blob/a5f4df7f12a50084b357059723e2b822a1af2521/projects/core/components/textfield/textfield.template.html#L29

### Taiga UI version

4.39.2

### Browser Used

Chrome

### OS Used

Windows

### LINKED ISSUE #11324: 🐞 - `TuiSelect` not showing selected value when patched and disabled
### Reproduction url

https://stackblitz.com/edit/angular-uwp9lxbo?file=src%2Fapp%2Fparent%2Fparent.component.ts

### Description

A bug occurs when the `TuiSelect` component is bound to a form. The issue arises under the following conditions:
- the data used to populate the select options is loaded asynchronously (with a delay),
- after the data is fetched, the form is prefilled using patchValue(),
- immediately after patching the value, the form is disabled (not happening if form stays enabled).

The `TuiSelect` field appears empty, even though a valid value was patched into the form. The correct value only becomes visible once the user interacts with the form (focuses or clicks the select field).

### Taiga UI version

4.45.0

### Browser Used

Chrome

### OS Used

Windows

splincode:
Fixes #11324
Fixes #11041

### DIFF
--- a/projects/core/components/textfield/textfield-multi/textfield-multi.template.html
+++ b/projects/core/components/textfield/textfield-multi/textfield-multi.template.html
@@ -1,3 +1,4 @@
+<ng-container *ngIf="value()" />
 <tui-scroll-controls
     *ngIf="rows > 1"
     class="t-scrollbar"
--- a/projects/core/components/textfield/textfield.template.html
+++ b/projects/core/components/textfield/textfield.template.html
@@ -1,3 +1,4 @@
+<ng-container *ngIf="value()" />
 <ng-content select="input" />
 <ng-content select="select" />
 <ng-content select="textarea" />
--- a/projects/demo-cypress/src/tests/textfield-disabled-select.cy.ts
+++ b/projects/demo-cypress/src/tests/textfield-disabled-select.cy.ts
@@ -1,0 +1,68 @@
+import type {OnInit} from '@angular/core';
+import {ChangeDetectionStrategy, Component} from '@angular/core';
+import {FormControl, FormGroup, FormsModule, ReactiveFormsModule} from '@angular/forms';
+import {TUI_ANIMATIONS_SPEED, TuiRoot, TuiTextfield} from '@taiga-ui/core';
+import {TuiChevron, TuiSelect} from '@taiga-ui/kit';
+import {delay, of} from 'rxjs';
+
+@Component({
+    standalone: true,
+    imports: [
+        FormsModule,
+        ReactiveFormsModule,
+        TuiChevron,
+        TuiRoot,
+        TuiSelect,
+        TuiTextfield,
+    ],
+    template: `
+        <tui-root>
+            <tui-textfield
+                tuiChevron
+                [content]="content"
+                [formGroup]="form"
+            >
+                <input
+                    formControlName="test"
+                    placeholder="Placeholder"
+                    tuiSelect
+                />
+
+                <ng-template
+                    #content
+                    let-data
+                >
+                    {{ data }}
+                </ng-template>
+            </tui-textfield>
+        </tui-root>
+    `,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    providers: [{provide: TUI_ANIMATIONS_SPEED, useValue: 0}],
+})
+export class TestTextfield implements OnInit {
+    protected readonly form = new FormGroup({
+        test: new FormControl(''),
+    });
+
+    public ngOnInit(): void {
+        of('Text')
+            .pipe(delay(1_000))
+            .subscribe((text) => {
+                const control = this.form.get('test');
+
+                control?.setValue(text);
+                control?.disable();
+            });
+    }
+}
+
+describe('Textfield and tuiSelect', () => {
+    beforeEach(() => cy.viewport(200, 150));
+
+    it('show selected value after disabled', () => {
+        cy.mount(TestTextfield);
+        cy.wait(1_000);
+        cy.get('tui-textfield').compareSnapshot('tui-textfield-disabled');
+    });
+});


projects/demo/src/modules/components/select/examples/11/index.html
```diff
@@ -0,0 +1,47 @@
+<tui-textfield
+    tuiChevron
+    [content]="content"
+    [formGroup]="form"
+    [tuiTextfieldCleaner]="true"
+>
+    <input
+        formControlName="division"
+        placeholder="Division *"
```
nsbarsukov: Cypress CT

projects/core/components/textfield/textfield.template.html
```diff
@@ -26,10 +26,10 @@
     <ng-content select="tui-icon" />
 </span>
 <span
-    *ngIf="control?.value != null"
+    *ngIf="control?.value != null || cva?.value() != null"
```
nsbarsukov: Will it fix issue only for our custom controls `class Tui*** extends TuiControl` ?
What's is about basic `Textfield` usage ?
https://taiga-ui.dev/components/textfield#basic

Should we use `toSignal` + `control.valueChanges` instead ?

nsbarsukov: Check this case
https://github.com/taiga-family/taiga-ui/issues/11324#issuecomment-3077643102

It is still reproducible even with your new changes

projects/core/components/textfield/textfield-multi/textfield-multi.template.html
```diff
@@ -12,7 +12,7 @@
 >
     <ng-content select="label" />
     <ng-template
-        *ngFor="let item of control?.value; let index = index"
+        *ngFor="let item of control?.value || cva?.value(); let index = index"
```
waterplea: This way if there is control.value it will never fall through to cva.value() and will not listen to the signal. Let's just add
```html
<ng-container *ngIf="cva?.value()" /> 
```
somewhere in template to kick change detection if value changes.

projects/core/components/textfield/textfield-multi/textfield-multi.template.html
```diff
@@ -58,10 +58,10 @@
 </span>
 
 <span
-    *ngIf="control?.value != null"
+    *ngIf="control?.value != null || cva?.value() != null"
```
waterplea: These changes are unnecessary, value is the same, we only need change detection.

projects/core/components/textfield/textfield.template.html
waterplea: We can also add:
```html
<ng-container *ngIf="value()" />
```
So that change detection will react to string representation changes. That would probably cover most of the cases.

projects/demo/src/modules/components/select/examples/11/index.html
splincode: fixed

projects/core/components/textfield/textfield.template.html
splincode: fixed

projects/core/components/textfield/textfield-multi/textfield-multi.template.html
splincode: fixed

projects/core/components/textfield/textfield-multi/textfield-multi.template.html
splincode: fixed

waterplea (approved):

projects/demo-cypress/src/tests/textfield-disabled-select.cy.ts
```diff
@@ -0,0 +1,142 @@
...
+
+            <ng-template
+                #itemContent
+                let-data
+            >
+                <h1 tuiTitle>
+                    {{ data.name }}
+
+                    <ng-container *ngIf="data.description">
+                        <span tuiSubtitle>{{ data.description }}</span>
```
nsbarsukov: Do we really need all this staff for CT ?
Could we simplify it just the following lines ?

```html
<ng-template #content let-data>
    {{data}}
</ng-template>

```

projects/demo-cypress/src/tests/textfield-disabled-select.cy.ts
```diff
@@ -0,0 +1,142 @@
...
+            >
+                <input
+                    formControlName="division"
+                    placeholder="Division *"
+                    tuiSelect
+                />
+                <tui-data-list-wrapper
+                    *tuiTextfieldDropdown
+                    new
+                    [itemContent]="itemContent"
```
nsbarsukov: Redundant for CT?

projects/demo-cypress/src/tests/textfield-disabled-select.cy.ts
```diff
@@ -0,0 +1,142 @@
...
+        <tui-root>
+            <tui-textfield
+                tuiChevron
+                [content]="content"
+                [formGroup]="form"
+                [tuiTextfieldCleaner]="true"
+            >
+                <input
+                    formControlName="division"
+                    placeholder="Division *"
```
nsbarsukov: Could it be deleted ?

projects/demo-cypress/src/tests/textfield-disabled-select.cy.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    public ngOnInit(): void {
+        of({
+            name: 'Name',
+            description: 'Description',
+            division: {
+                id: 1,
+                name: 'Second division',
+                description: 'Test description...',
+            },
```
nsbarsukov: Probably, it is could be simplified

projects/demo-cypress/src/tests/textfield-disabled-select.cy.ts
splincode: fixed

projects/demo-cypress/src/tests/textfield-disabled-select.cy.ts
splincode: fixed

projects/demo-cypress/src/tests/textfield-disabled-select.cy.ts
splincode: fixed

projects/demo-cypress/src/tests/textfield-disabled-select.cy.ts
splincode: fixed

PR #11329, opened 2025-07-16
refactor(kit): skip `writeValue` for textfield if new value already equals to current control value

nsbarsukov:
## Why ?

1. Open https://taiga-ui.dev/components/input-time
2. Textfield is empty & control is `null`
3. Textfield contains `12` & control is still `null`
4. Form control is patched by `null`

**Previous behavior:** textfield is empty
**New behavior:** textfield contains `12`


waterplea (approved):

projects/kit/components/combo-box/combo-box.directive.ts
```diff
@@ -134,8 +134,10 @@ export class TuiComboBox<T>
     }
 
     public override writeValue(value: T | string | null): void {
-        super.writeValue(value);
-        this.textfield.value.set(this.stringify(value));
+        if (value !== this.value()) {
+            super.writeValue(value);
+            this.textfield.value.set(this.stringify(value));
```
nsbarsukov: But what is about `formControl.reset` ?

splincode: Is this covered by some cypress tests to make sure everything works better than before?

nsbarsukov: - https://github.com/taiga-family/taiga-ui/pull/11366

**TODO**: rebase current branch onto `main` after its merge

PR #11412, opened 2025-07-24
test: add component test for check compatibility tuiTable with tuiTextfield

splincode:


projects/demo-cypress/src/tests/textfield-in-table.cy.ts
```diff
@@ -0,0 +1,104 @@
...
+    class Test {
+        protected readonly input = new PolymorpheusComponent(MyInputTextfield);
+        protected readonly select = new PolymorpheusComponent(MySelectTextfield);
+    }
+
+    beforeEach(() => {
+        cy.mount(Test);
+    });
+
+    it('inherit table appearance inside tuiTable for textfield', () => {
```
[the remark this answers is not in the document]
splincode: fixed

mdlufy (approved):

projects/demo-cypress/src/tests/textfield-in-table.cy.ts
```diff
@@ -0,0 +1,91 @@
...
+})
+class Test {
+    protected readonly input = new PolymorpheusComponent(MyInputTextfield);
+    protected readonly select = new PolymorpheusComponent(MySelectTextfield);
+}
+
+describe('TuiTable and TuiTextfield', () => {
+    beforeEach(() => cy.mount(Test));
+
+    it('should inherit table appearance inside tuiTable for textfield', () => {
```
waterplea, line 88: Can we also add other textfield-like components such as textarea and input-chip? InputCardGroup? Although I'm not sure about the last one, it doesn't make much sense to put in a table.

splincode: sure, I can open separate PR

PR #11425, opened 2025-07-25
refactor(addon-doc): drop `any` keyword

splincode:


projects/addon-doc/utils/clean-object.ts
```diff
@@ -5,9 +5,9 @@ export type TuiDeepPartial<T> = {
 type EmptyValue = '' | null | undefined;
 
 function checkValueIsEmpty<T>(value: EmptyValue | T): value is EmptyValue {
-    const nextValue: any = typeof value === 'string' ? value.trim() : value;
-
-    return ['', NaN, null, undefined].includes(nextValue);
+    return typeof value === 'string'
```
mdlufy: may be use tuiIsString or it's deprecated approach?

mdlufy (approved):

projects/addon-doc/utils/clean-object.ts
splincode: fixed

PR #11427, opened 2025-07-25
refactor(cdk): do not use `EMPTY_QUERY` in the future

splincode:


mdlufy (approved):

projects/cdk/constants/empty.ts
```diff
@@ -15,6 +15,7 @@ const rect = {
  * For type safety when using @ContentChildren and @ViewChildren
  *
  * NOTE: Be careful subscribing to 'changes'
+ * @deprecated use viewChild(), viewChildren(), contentChild() and contentChildren() for strict type checking
```
mdlufy, line 18: What's problem with typing, provide negative example, please

splincode: Image

waterplea (approved):

PR #11436, opened 2025-07-28
feat(kit): supports configuration of initial zoom coefficient in preview

### LINKED ISSUE #11132: 🚀 - `Preview` supports configuration of initial zoom coefficient
### Description

Now, `Preview` is always initially opened as 80% height of full page size
https://github.com/taiga-family/taiga-ui/blob/507d44da0071fc32101b9ca193e2beeb9cd87f6e/projects/kit/components/preview/preview.component.ts#L27

Size of the content (when it takes 80% height of the full page size) becomes min possible scale coefficient for zoom controller.

**The task – possibility to open content with initial zoom more than possible one.**

__

For example, open this example
https://taiga-ui.dev/components/preview#full-preview

A4 page takes 80% height of the page and zoom controller is set to 55% (min possible one).
We need to have possibility to open A4 page with following condition: A4 page content takes 100% page and zoom is initially set to 69% (55% is still min possible coefficient).

Image

Farbaks:
Fixes #11132 

### Description
- Support configuration of initial zoom coefficient value

### DIFF
--- a/projects/kit/components/preview/preview.component.ts
+++ b/projects/kit/components/preview/preview.component.ts
@@ -92,7 +92,7 @@ export class TuiPreviewComponent {
     public rotatable = false;
 
     @Input()
-    public initialScaleCoef = 0.8;
+    public initialScale = 0.8;
 
     protected rotate(): void {
         this.rotation$.next(this.rotation$.value - ROTATION_ANGLE);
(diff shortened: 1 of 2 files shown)


Farbaks:
PS: Went with the Input() implementation over the DI, because the DI was going to be an unnecessarily more complex implementation due to how the Preview component is current being used through the Preview Dialog component.

splincode (approved):

projects/kit/components/preview/preview.component.ts
```diff
@@ -92,6 +91,9 @@ export class TuiPreviewComponent {
     @Input()
     public rotatable = false;
 
+    @Input()
+    public initialScaleCoef = 0.8;
```
waterplea: Let's drop Coef from the name

waterplea (approved):

PR #11438, opened 2025-07-28
chore: format by html-eslint plugin

splincode:


projects/demo/src/modules/pipes/format-date/index.html
```diff
@@ -19,13 +19,15 @@
     </ng-template>
 
     <tui-setup *pageTab="'Setup'">
-        <li>
-            <p>Provide custom service for formatting</p>
+        <ul>
```
waterplea: tui-setup already has ol inside

splincode: fixed

waterplea (approved):

waterplea:
Hopefully this doesn't break any :empty CSS rules or adds unexpected whitespace around tags.

PR #11467, opened 2025-07-29
ci: add html-eslint

splincode:


projects/addon-doc/tokens/doc-icons.ts
```diff
@@ -12,12 +12,17 @@ export interface TuiDocIcons {
     readonly copy: string;
     readonly check: string;
     readonly link: string;
+    readonly externalLink: string;
```
waterplea: Should be optional to not have breaking change until v5

projects/legacy/components/color-selector/color-selector.options.ts
```diff
@@ -63,6 +64,9 @@ export const TUI_COLOR_SELECTOR_DEFAULT_OPTIONS: TuiColorSelectorOptions = {
     mode: TuiColorSelectorMode.SolidColor,
     color: [0, 0, 0, 1],
     colors: new Map<string, string>(),
+    icons: {
+        checkmark: '@tui.check',
```
waterplea: I think we have it in common icons token

splincode: fixed

projects/addon-doc/tokens/doc-icons.ts
splincode: fixed

projects/addon-doc/tokens/doc-icons.ts
```diff
@@ -13,11 +13,19 @@ export interface TuiDocIcons {
...
+    /**
+     * TODO: drop optional in v5
+     */
+    readonly externalLink?: string;
     /**
      * TODO: drop optional in v5
      */
     readonly shrink?: string;
     readonly expand?: string;
+    readonly ltr: {
```
waterplea: Same here. Please be smarter about such things, it is exactly the same thing in exactly the same place, we shouldn't have to point out every line on reviews.

splincode: image

nsbarsukov (approved):

Issue #11264, opened 2025-07-08
🚀 - Full RTL support


waterplea:
### Description

Below is the list of issues I've spotted while navigating demo portal in RTL mode. Once they are all resolved we can say we fully support RTL mode and all further issues will be resolved as separate issues whenever spotted:

- [x] Fade
- [x] Group + border-radius
- [x] Calendar chevrons, Preview arrows, Pagination, TablePagination
- [x] AppBar back chevron
- [x] BottomSheet bar
- [x] Rating + half a star
- [x] Breadcrumbs chevron
- [x] Scrollbar horizontal
- [x] InputPhoneInternational
- [x] InputNumber + step
- [x] File + close
- [x] Push + close
- [x] Link + icons
- [x] Mobile Switch
- [x] InputCardGroup
- [x] Connected
- [x] Navigation
- [x] Dropdown/Hint direction correction
- [x] AvatarStack
- [x] SwipeActions
- [x] Pager scrolling
- [x] Range
- [ ] Table + sticky
- [ ] Carousel broken

waterplea:
We should also take a look at how filler works, particularly in date inputs.

waterplea:
I think we can close this as resolved, all other fixes will be coming to v5 due to bumped browser support and ability to use `inset` styles. @nsbarsukov keep RTL in mind when implementing new `InputRange` please.

PR #11474, opened 2025-07-30
feat(kit): `Range` supports RTL

nsbarsukov:
Relates:
* https://github.com/taiga-family/taiga-ui/issues/11264


projects/kit/components/range/range.style.less
```diff
@@ -56,17 +58,25 @@
     &::after {
         .fullsize(absolute, inset);
 
+        --t-direction: right;
```
waterplea: We already have this:
https://github.com/taiga-family/taiga-ui/blob/main/projects/core/components/root/root.style.less#L22

projects/kit/components/range/range.style.less
```diff
@@ -137,14 +138,26 @@
     left: 0;
     right: 0;
     z-index: 1;
-    transform: translate(0, -50%);
+    transform: translateY(-50%);
 
-    &:last-of-type {
-        --tui-slider-thumb-transform: translateX(50%) translateX(@thumb-shift);
-    }
+    --t-direction: 1;
```
waterplea: Same

waterplea (approved):

projects/kit/components/range/range.style.less
nsbarsukov: https://github.com/taiga-family/taiga-ui/pull/11474/commits/81e4d55589804103ee964a91ee1e726e21aa8ae9

projects/kit/components/range/range.style.less
nsbarsukov: https://github.com/taiga-family/taiga-ui/pull/11474/commits/81e4d55589804103ee964a91ee1e726e21aa8ae9

PR #11479, opened 2025-07-30
chore(core): `Textfield` remove unnecessary explicit font size

waterplea:
Fixes #

### DIFF
--- a/projects/core/styles/components/textfield.less
+++ b/projects/core/styles/components/textfield.less
@@ -72,7 +72,7 @@ tui-textfield {
     input,
     select,
     textarea {
-        font: var(--tui-font-text-ui-m);
+        font: inherit;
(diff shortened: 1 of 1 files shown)


splincode (approved):

projects/core/styles/components/textfield.less
```diff
@@ -72,7 +72,7 @@ tui-textfield {
     input,
     select,
     textarea {
-        font: var(--tui-font-text-ui-m);
+        font: inherit;
```
nsbarsukov, line 75: Is it expected behavior that placeholder inherits font from `[tuiFluidTypography]` ?

image

waterplea: Checked with designers and they said yeah

PR #11480, opened 2025-07-30
feat(schematics): skip install addons

splincode:


projects/cdk/schematics/ng-add/schema.json
```diff
@@ -42,6 +42,10 @@
                     {
                         "value": "layout",
                         "label": "layout            Layout components"
+                    },
+                    {
+                        "value": "",
+                        "label": "Not needed"
```
vladimirpotekhin, line 48: this is toggle list with instructions 

`Press <space> to select, <a> to toggle all, <i> to invert selection, and <enter> to process` 

Why do we need an empty option ?

splincode: How do I exit if I've entered select mode?

PR #11483, opened 2025-07-30
feat(addon-doc): ignore padding when header is not defined

splincode:


waterplea (approved):

nsbarsukov (approved):
Should it be `fix` instead of `feat` ?

PR #11484, opened 2025-07-30
chore(demo): move documentation page of legacy `InputRange`

nsbarsukov:


projects/demo/src/modules/components/input-range-legacy/index.html
```diff
@@ -1,6 +1,8 @@
 <tui-doc-page
-    header="InputRange"
+    deprecated
+    header="InputRangeLegacy"
     package="LEGACY"
+    path="legacy/components/input-slider"
```
nsbarsukov:
```suggestion
    path="legacy/components/input-range"
```

splincode (approved):

waterplea (approved):

PR #11513, opened 2025-08-02
Feat(core): Implement `Hint` hover state event emitter

### LINKED ISSUE #11443: 🚀 - add open hint event
### Description

hi, please add the ability to track the opening and closing of a hint

I imagine it something like this:

```html
<tui-icon
  tuiHintDirection="right"
  tuiTooltip="Supports basic <strong>HTML</strong>"
  (tuiHintOpen)="onOpenHint($event)"
/>
```

Farbaks:
Fixes #11443


projects/core/directives/hint/hint-hover.directive.ts
```diff
@@ -76,6 +76,9 @@ export class TuiHintHover extends TuiDriver {
     @Input()
     public tuiHintHideDelay: TuiHintOptions['hideDelay'] = this.options.hideDelay;
 
+    @Output('tuiHintHoverChange')
+    public readonly tuiHintHoverChange = new EventEmitter<boolean>();
```
waterplea: I think this should be called `tuiHintVisible` and it should go into main hint directive, somewhere here:
https://github.com/taiga-family/taiga-ui/blob/main/projects/core/directives/hint/hint.directive.ts#L77

projects/core/directives/hint/hint.directive.ts
```diff
@@ -51,6 +59,9 @@ export class TuiHintDirective<C>
     @Input('tuiHintAppearance')
     public appearance = inject(TUI_HINT_OPTIONS).appearance;
 
+    @Output('tuiHintVisible')
+    public readonly tuiHintVisible = new EventEmitter<boolean>();
```
waterplea: Let's name it `visible` for brevity since it already has an alias

PR #11523, opened 2025-08-04
feat(cdk): detect webdriver by default in e2e token

splincode:
Common standard: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/webdriver

Playwright, like other browser automation tools, sets navigator.webdriver to true in the browser environment it controls. This is a standard property indicating that the browser is under automated control.

While navigator.webdriver is the most reliable, in some scenarios, Playwright might inject or modify other global variables or functions. This is less standardized and depends on Playwright's internal implementation, but you could look for:
Specific properties on window or document that are unique to Playwright's setup.
The presence of certain event listeners or functions that Playwright might add for its internal workings.

This approach is generally discouraged due to its reliance on internal Playwright details that could change in future versions.

### DIFF
--- a/projects/cdk/tokens/environment.ts
+++ b/projects/cdk/tokens/environment.ts
@@ -53,15 +53,19 @@ export const TUI_IS_TOUCH = new InjectionToken(ngDevMode ? 'TUI_IS_TOUCH' : '', 
 });
 
 /**
+ * @deprecated: use only {@link TUI_IS_E2E}
  * Detect if app is running under Cypress
  * {@link https://docs.cypress.io/faq/questions/using-cypress-faq#Is-there-any-way-to-detect-if-my-app-is-running-under-Cypress Cypress docs}
+ * TODO: remove in v5
  */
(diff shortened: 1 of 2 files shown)


projects/cdk/tokens/environment.ts
```diff
@@ -74,5 +78,8 @@ export const TUI_IS_PLAYWRIGHT = new InjectionToken<boolean>(
  * Detect if app is running under any of test frameworks
  */
 export const TUI_IS_E2E = new InjectionToken(ngDevMode ? 'TUI_IS_E2E' : '', {
-    factory: () => inject(TUI_IS_CYPRESS) || inject(TUI_IS_PLAYWRIGHT),
+    factory: () =>
+        inject(TUI_IS_CYPRESS) ||
+        inject(TUI_IS_PLAYWRIGHT) ||
+        inject(WA_NAVIGATOR).webdriver,
```
waterplea, line 84: `!!`?

splincode: this API is in every runtime

image
image

waterplea: Oh, sorry, it's already boolean. Didn't check the type.

PR #11525, opened 2025-08-04
chore(demo): fix broken source code link for `Copy`

nsbarsukov:


splincode (approved):

projects/demo/src/modules/components/copy/index.html
```diff
@@ -1,7 +1,7 @@
 <tui-doc-page
     header="Copy"
     package="KIT"
-    type="components"
```
waterplea, line 4: This way all examples are broken. Better to provide a manual link, in v5 it would move to components anyway

PR #11560, opened 2025-08-08
fix(core): `DropdownSelection` fix position inside new textarea

### LINKED ISSUE #11071: 🐞 - [DropdownSelection] proper dropdown positioning issue when resizing textarea height
### Reproduction url

https://stackblitz.com/edit/haewjlmq?file=src%2Fapp%2Fapp.template.html

### Description

Issue:
When increasing the height of the textarea (either when it's empty or contains some text), the dropdown becomes misplaced.

Expected Result:
The dropdown should appear near the '@' symbol.

Actual Result:
The dropdown appears outside the textarea, misaligned with the '@'.

Image

### Taiga UI version

4.38.0

### Browser Used

Chrome

### OS Used

macOS

### LINKED ISSUE #11286: 🚀 - Allow using up and down arrows to navigate dropdown selection input text
### Description

When using a dropdown selection (https://taiga-ui.dev/directives/dropdown-selection#textarea), up and down arrows default behaviour is prevented to handle dropdown trigger. 
This makes sense when one wants to be able to visualize options, but in the case of the aforementioned feature   where the dropdown can be triggered externally, we just lose the ability to navigate the input text (especially true on textarea).

In the dropdown selection case, it feels like the up and down arrows should keep their default behaviour.

vladimirpotekhin:
To fix the issue while maintaining compatibility with both the new and old textarea implementations, we now render the `ghost` inside the textarea's `ghost` element — if it exists.

Additionally, this PR fixes a bug where the dropdown wouldn’t appear after a newline character.

Fixes #11071 
Fixes #11286

### DIFF
--- a/projects/core/directives/dropdown/dropdown-selection.directive.ts
+++ b/projects/core/directives/dropdown/dropdown-selection.directive.ts
@@ -111,8 +111,12 @@ export class TuiDropdownSelection
 
     public ngOnDestroy(): void {
         if (this.ghost) {
-            this.vcr.element.nativeElement.removeChild(this.ghost);
+            this.ghostHost.removeChild(this.ghost);
         }
+    }
+
+    private get ghostHost(): HTMLElement {
+        return this.el.querySelector('tui-textfield .t-ghost') || this.el;
     }
 
     private getRange(): Range {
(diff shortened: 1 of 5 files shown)


waterplea:
Would it also fix this?
https://github.com/taiga-family/taiga-ui/issues/11286

waterplea (approved):

vladimirpotekhin:
> Would it also fix this? #11286

Initially no, but I fixed this case in the example in the [second commit ](https://github.com/taiga-family/taiga-ui/pull/11560/commits/189d6197899b9f099ed8dcf82d2f12033fea4aa6)

PR #11561, opened 2025-08-08
feat(core): introduce `--tui-appearance` to support binding appearance from CSS

splincode:
### Benefits

- You can assign custom appearance from any css theme to your components if you don't have access to HTML
- `@property` Next-gen CSS variables

image


waterplea:
I don't understand, appearance is not a variable value, it's a selector. Kinda like CSS class. It makes no sense to me to use it as a variable just to trigger getComputedStyle to get it in JS where it belongs.

PR #11566, opened 2025-08-08
refactor: unnecessary optional chain or conditional

splincode:
When
`@typescript-eslint/no-unnecessary-condition': 'error'`

### DIFF
--- a/scripts/shared/execute.ts
+++ b/scripts/shared/execute.ts
@@ -5,13 +5,13 @@ import {infoLog} from 'ng-morph';
 export function execute(shell: string, options?: Partial<CommonExecOptions>): string {
     infoLog(`ᐅ ${shell}`);
 
-    return execSync(
-        shell,
-        options ?? {
-            stdio: 'inherit',
-            encoding: 'utf8',
-        },
-    )
-        ?.toString()
-        .trim();
+    return String(
+        execSync(
+            shell,
+            options ?? {
+                stdio: 'inherit',
+                encoding: 'utf8',
+            },
+        ),
+    ).trim();
 }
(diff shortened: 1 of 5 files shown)


scripts/shared/is-tui-package-name.ts
```diff
@@ -1,5 +1,5 @@
-export function isTuiPackageName(name = '', ignores: readonly string[] = []): boolean {
+export function isTuiPackageName(name?: string, ignore: readonly string[] = []): boolean {
```
waterplea: This is unnecessary, it was fine as it was, `?.` could just be removed as `name` is always a `string`

scripts/shared/get-all-tags.ts
```diff
@@ -1,6 +1,6 @@
 import {execute} from './execute';
 
-export function getAllTags(name: string): Record<string, string> {
+export function getAllTags(name: string): Record<string, string> | undefined {
```
waterplea: How does it return `undefined`?

scripts/shared/execute.ts
```diff
@@ -5,13 +5,13 @@ import {infoLog} from 'ng-morph';
...
+    const std = execSync(
         shell,
         options ?? {
             stdio: 'inherit',
             encoding: 'utf8',
         },
-    )
-        ?.toString()
-        .trim();
+    ) as Buffer | undefined;
```
waterplea: How does it return `undefined`? Types say it returns `Buffer | string`. Instead of `.toString()` if you are worried maybe it can just be `String(...)`?

scripts/shared/is-tui-package-name.ts
splincode: fixed

scripts/shared/get-all-tags.ts
splincode: fixed

scripts/shared/execute.ts
splincode: fixed

PR #11576, opened 2025-08-10
 feat(core, kit): implement icon mode resolution system (#10938)

FeironoX5:
Refs #10938

- Changes in `TUI_ICON_RESOLVER` for icon mode resolution (`@font/@image(.mask)/@icon`)
- Updated affected components, styles, and some examples accordingly
- `url()` wrapping logic is now handled by `tuiInjectIconResolver()`

---

### Next steps:
- [ ] Set default `--tui-font-icon` CSS variable in core theme
- [ ] Update docs with new icon syntax examples

### Previous PR comment:

> We should add mode resolver by icon name and use following:
> 
> - `@image.xxx` for the same mode as now but with `background` rather than `mask` for colored icons
> - `@font.xxx` for using icon name (with `@font.` removed) as `content` inside pseudo elements instead of using it as mask/background
> - `@icon.xxx` current mode (should be default fallback if doesn't match the ones above so `@icon` is just convenience)

### DIFF
--- a/projects/demo/src/modules/app/pages.ts
+++ b/projects/demo/src/modules/app/pages.ts
@@ -196,7 +196,7 @@ export const pages: DocRoutePages = [
             },
             {
                 name: 'tui-avatar-labelled',
-                figmaVersion: '1.0.1',
+                figmaVersion: '1.3.0',
             },
             {
                 name: 'tui-avatar-stack',
@@ -217,14 +217,14 @@ export const pages: DocRoutePages = [
                 title: 'Badge',
                 keywords: 'бэдж, овал, badge',
                 route: DemoRoute.Badge,
-                meta: {},
+                meta: {figmaVersion: '1.4.1'},
             },
             {
                 section: 'Components',
                 title: 'BadgedContent',
                 keywords: 'бэдж, бейдж, circle, овал, круг, badge, нотификация',
                 route: DemoRoute.BadgedContent,
-                meta: {},
+                meta: {figmaVersion: '1.4.1'},
             },
             {
                 section: 'Components',
@@ -240,7 +240,10 @@ export const pages: DocRoutePages = [
         title: 'Block',
         keywords: 'кнопка, чек, форма, form, ввод, checkbox, radio, радио, label',
         route: DemoRoute.Block,
-        meta: [{name: 'tui-checkbox-block'}, {name: 'tui-radio-block'}],
+        meta: [
+            {name: 'tui-checkbox-block', figmaVersion: '1.2.0'},
+            {name: 'tui-radio-block', figmaVersion: '1.2.0'},
+        ],
     },
     {
         section: 'Layout',
@@ -258,7 +261,10 @@ export const pages: DocRoutePages = [
                 title: 'Button',
                 keywords: 'кнопка, button, icon-button, иконка',
                 route: DemoRoute.Button,
-                meta: [{name: 'tui-button'}, {name: 'tui-button-vertical'}],
+                meta: [
+                    {name: 'tui-button', figmaVersion: '1.9.1'},
+                    {name: 'tui-button-vertical', figmaVersion: '1.0.1'},
+                ],
             },
             {
                 section: 'Components',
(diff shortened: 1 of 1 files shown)


projects/core/styles/components/icon.less
```diff
@@ -41,7 +42,25 @@ tui-icon {
...
     }
+
+    &[data-icon='image']::after {
+        background: var(--t-icon) no-repeat center/contain;
+        mask: none;
+    }
+
+    &[data-icon='font']::after {
+        content: var(--t-icon);
+        font-family: var(--tui-font-icon, inherit);
```
waterplea: Let's use a shorthand:
```css
font: var(--tui-font-icon, inherit) 1em/1;
text-align: center;
text-transform: none;
```

projects/core/styles/components/icon.less
```diff
@@ -41,7 +42,25 @@ tui-icon {
...
+    }
+
+    &[data-icon='svg']::after {
         mask: var(--t-icon) no-repeat center/contain;
         background: currentColor;
     }
+
+    &[data-icon='image']::after {
+        background: var(--t-icon) no-repeat center/contain;
+        mask: none;
```
waterplea: Unnecessary

projects/core/styles/components/icons.less
```diff
@@ -22,19 +24,60 @@
...
 
-    &::after {
-        display: var(--t-icon-end);
+    &[data-icon-end='svg']::after {
         mask: var(--t-icon-end) no-repeat center/contain padding-box;
     }
+
+    &[data-icon-start='image']::before {
+        background: var(--t-icon-start) no-repeat center/contain padding-box;
+        mask: none;
```
waterplea: Unnecessary since mask is only set for SVGs

projects/core/styles/components/icons.less
```diff
@@ -22,19 +24,60 @@
...
+    }
+
+    &[data-icon-end='image']::after {
+        background: var(--t-icon-end) no-repeat center/contain padding-box;
+        mask: none;
```
waterplea: Unnecessary

projects/core/styles/components/icons.less
```diff
@@ -22,19 +24,60 @@
...
+    }
+
+    &[data-icon-start='font']::before,
+    &[data-icon-start='font']::after {
+        font-family: var(--tui-font-icon, inherit);
```
waterplea: Same comment about shorthand

projects/core/tokens/icon-resolver.ts
```diff
@@ -1,32 +1,94 @@
...
-import {type TuiStringHandler} from '@taiga-ui/cdk/types';
+import type {TuiHandler, TuiStringHandler} from '@taiga-ui/cdk/types';
 
 import {TUI_ASSETS_PATH} from './assets-path';
 import {TUI_ICON_REGISTRY} from './icons';
 
+const ICON_MODE_PREFIXES = {
+    font: '@font.',
+    image: '@img.',
+    imageMask: '@img.mask.',
```
waterplea: What's the point of img.mask?

projects/core/tokens/icon-resolver.ts
```diff
@@ -1,32 +1,94 @@
...
+};
+
+type ICON_TARGET_TAG = 'raw' | 'url';
+
+export interface TuiResolvedIcon {
```
waterplea: Let's not change the resolver API. Let's just add `tuiGetIconMode(icon: string): 'img' | 'icon' | 'font'` and use it both in the resolver and in host binding for `data-` type:
```ts
return icon => tuiGetIconMode(icon) === 'font' ? icon.replace('@font.', '') : ...
```

projects/core/tokens/icon-resolver.ts
FeironoX5: The original purpose was to allow changing the color of an SVG or PNG icon, but since `icon.includes("/")` produces the same effect by enabling SVG icon mode, it now only enhances readability and could be removed.

PR #11592, opened 2025-08-11
feat(core): implement icon mode resolution system (#10938)

FeironoX5:
Refs #10938

This PR includes fixes for the [previous one](https://github.com/taiga-family/taiga-ui/pull/11576).

### DIFF
--- a/projects/core/styles/components/icons.less
+++ b/projects/core/styles/components/icons.less
@@ -30,40 +30,34 @@
         font-size: 1.5rem;
         flex-shrink: 0;
         box-sizing: content-box;
+        background: currentColor;
     }
 
     &::before {
         display: var(--t-icon-start, none);
+        mask: var(--t-icon-start) no-repeat center/contain padding-box;
     }
 
     &::after {
         display: var(--t-icon-end, none);
-    }
-
-    &[data-icon-start='svg']::before,
-    &[data-icon-end='svg']::after {
-        background: currentColor;
-    }
-
-    &[data-icon-start='svg']::before {
-        mask: var(--t-icon-start) no-repeat center/contain padding-box;
-    }
-
-    &[data-icon-end='svg']::after {
         mask: var(--t-icon-end) no-repeat center/contain padding-box;
     }
 
     &[data-icon-start='image']::before {
+        mask: none;
         background: var(--t-icon-start) no-repeat center/contain padding-box;
     }
 
     &[data-icon-end='image']::after {
+        mask: none;
         background: var(--t-icon-end) no-repeat center/contain padding-box;
     }
 
     &[data-icon-start='font']::before,
     &[data-icon-end='font']::after {
         display: grid;
+        mask: none;
+        background: none;
         font: 1.3em/1 var(--tui-font-icon, inherit);
         text-align: center;
         place-content: center;
--- a/projects/core/styles/components/icon.less
+++ b/projects/core/styles/components/icon.less
@@ -42,19 +42,19 @@ tui-icon {
 
         content: '';
         display: block;
-    }
-
-    &[data-icon='svg']::after {
         mask: var(--t-icon) no-repeat center/contain;
         background: currentColor;
     }
 
     &[data-icon='image']::after {
+        mask: none;
         background: var(--t-icon) no-repeat center/contain;
     }
 
     &[data-icon='font']::after {
         content: var(--t-icon);
+        mask: none;
+        background: none;
         font: 1em/1 var(--tui-font-icon, inherit);
         text-align: center;
         text-transform: none;


FeironoX5:
Next steps:
- Fix imports to pass lint tests
- Update docs
- Set global font icon variable

projects/core/directives/icons/icons.directive.ts
```diff
@@ -34,22 +39,31 @@ class TuiIconsStyles {}
...
+    protected readonly iconEndMode = computed(() =>
+        tuiGetIconMode(this.iconEnd()?.toString()),
+    );
 
     public readonly iconStart = signal<Icon>(
         inject(TUI_ICON_START, {self: true, optional: true}) || '',
     );
 
-    public readonly iconEnd = signal<string>(
+    public readonly iconEnd = signal<Icon>(
```
waterplea: It is supposed to be a string, the `Icon` mode is a hack to make this directive compatible with `Avatar`

projects/core/tokens/icon-resolver.ts
```diff
@@ -21,12 +35,48 @@ export const TUI_ICON_RESOLVER = new InjectionToken<TuiStringHandler<string>>(
...
+
+    if (icon.startsWith(ICON_MODE_PREFIXES.font)) {
+        return 'font';
+    }
+
+    return 'svg';
+}
+
+export function tuiInjectIconResolver(
+    tag: ICON_TARGET_TAG = 'url',
```
waterplea: There shouldn't be any modes for injecting the resolver, it would be a breaking change if it starts adding `url(`. Instead if the name starts with `'@font.` it must remove the prefix and return the rest, otherwise it should behave the same way it behaved before. And in every place where it is used we need to take care of adding `url(` or not adding it, if it's a font.

projects/addon-commerce/components/input-card/input-card.component.ts
```diff
@@ -67,7 +67,8 @@ export class TuiInputCard implements OnInit {
     protected readonly mask = tuiMaskito(TUI_MASK_CARD);
     protected readonly image = computed(() => {
         const system = this.options.paymentSystemHandler(this.value());
-        const icon = system && this.icons[system] && this.resolver(this.icons[system]);
+        const iconValue = system && this.icons[system];
```
waterplea: Is this change relevant?

projects/core/tokens/icon-resolver.ts
```diff
@@ -4,14 +4,26 @@ import {type TuiStringHandler} from '@taiga-ui/cdk/types';
...
 export const TUI_ICON_RESOLVER = new InjectionToken<TuiStringHandler<string>>(
     ngDevMode ? 'TUI_ICON_RESOLVER' : '',
     {
         factory: () => {
             const path = inject(TUI_ASSETS_PATH);
 
-            return (icon) =>
-                `${path}/${icon.replace('@tui.', '').split('.').join('/')}.svg`;
+            return (icon) => {
+                // regex for @{any text}. used as a default fallback, returns path to svg from icon pack
```
waterplea: Let's move the comment one line higher, so that there's no need for `{}` and `return`

projects/core/tokens/icon-resolver.ts
```diff
@@ -21,12 +33,45 @@ export const TUI_ICON_RESOLVER = new InjectionToken<TuiStringHandler<string>>(
...
 
-    return (icon) =>
-        !icon || icon.includes('/') ? icon : (icons[icon] ?? resolver(icon));
+    return (icon) => {
+        if (!icon) {
+            return '';
+        }
+
+        if (tuiGetIconMode(icon) === 'image') {
+            return icon.slice(ICON_MODE_PREFIXES.image.length);
```
waterplea: Resolver already removes the prefix, no need to treat images differently here

projects/core/tokens/icon-resolver.ts
```diff
@@ -21,12 +33,45 @@ export const TUI_ICON_RESOLVER = new InjectionToken<TuiStringHandler<string>>(
...
+}
+
 export function tuiInjectIconResolver(): TuiStringHandler<string> {
     const icons = inject(TUI_ICON_REGISTRY);
     const resolver = inject(TUI_ICON_RESOLVER);
 
-    return (icon) =>
-        !icon || icon.includes('/') ? icon : (icons[icon] ?? resolver(icon));
+    return (icon) => {
+        if (!icon) {
```
waterplea:
```
if (!icon || icon.includes('/') {
  return icon
}
```

projects/core/styles/components/icons.less
```diff
@@ -22,19 +24,57 @@
...
         inline-size: 1em;
         block-size: 1em;
         line-height: 1em;
         font-size: 1.5rem;
         flex-shrink: 0;
         box-sizing: content-box;
+    }
+
+    &::before {
+        display: var(--t-icon-start, none);
```
waterplea, line 37: Why fallback to none? Previously it worked without it, just worried something might be affected by the change.

projects/core/tokens/icon-resolver.ts
```diff
@@ -4,14 +4,25 @@ import {type TuiStringHandler} from '@taiga-ui/cdk/types';
 import {TUI_ASSETS_PATH} from './assets-path';
 import {TUI_ICON_REGISTRY} from './icons';
 
+const ICON_MODE_PREFIXES = {
+    font: '@font.',
+    image: '@img.',
+};
```
splincode: as const;

projects/core/tokens/icon-resolver.ts
```diff
@@ -4,14 +4,25 @@ import {type TuiStringHandler} from '@taiga-ui/cdk/types';
...
+
+type TUI_ICON_MODE = 'font' | 'image' | 'svg';
```
splincode: `type IconMode`?

projects/core/styles/components/icons.less
FeironoX5: The difference only occurs if the directive sets an empty string, then var(--t-icon-start) becomes inline, while var(--t-icon-start, none) becomes none, both are equally supported by all browsers from [https://github.com/orgs/taiga-family/discussions/3232](https://github.com/orgs/taiga-family/discussions/3232), it just makes the code bit more explicit.

projects/core/tokens/icon-resolver.ts
FeironoX5: Not sure if this needs to be exported as it is only used as the return type of the `tuiGetIconMode`, and TS infers it automatically where the function is called.

waterplea: Sure, no need to import it, but then no need for TUI_ either and the code style for types is PascalCase. That's what @splincode meant :)

waterplea: I'm not even sure we need it at all, can just inline it in return type of `tuiGetIconMode`

vladimirpotekhin (approved):

PR #11606, opened 2025-08-12
feat(core): use design-tokens import by default

splincode:


waterplea:
These are proprietary colors and a bazillion of variables we do not need. We need to add duration and curve variables though.

PR #11614, opened 2025-08-12
chore(demo): use `ITEMS` token when working with mocked database items

splincode:


vladimirpotekhin (approved):

waterplea:
We specifically did it this way so that there's less dependencies in the examples.

PR #11627, opened 2025-08-13
fix(kit): `formControl.setValue()` inside `effect` should not produce circular dependency

### LINKED ISSUE #11551: 🐞 - circular dependency when using `tuiInputDateRange` and effects
### Reproduction url

https://stackblitz.com/edit/angular-sunytozn

### Description

There is a bug with `tuiInputDateRange` and effects. We use signals with effects to synchronize the work of the store and the date range form control. After updating to version 4.48.0, we encountered a cyclic dependency.

As far as I understand, the problem lies in this file: projects/kit/components/input-date/input-date.directive.ts, where the signal is also updated. 
https://github.com/taiga-family/taiga-ui/commit/e2835bbbfb4b3eafb02255d13e9cb3572fc860ff#diff-072a28555b55d1d7ceac779a2294a5ca1db47d67b313e761f22f76fed9a4c27bR124

In stackblitz, you can see that the effect, which includes a form update, is triggered several times, even though the value has not changed. In our project, this place is triggered constantly, which causes a cyclic dependency. Only wrapping it in `untracked` solves the problem.

### Taiga UI version

4.48.0

### Browser Used

Chrome

### OS Used

macOS

nsbarsukov:
Fixes #11551


projects/cdk/utils/miscellaneous/untracked.ts
```diff
@@ -0,0 +1,52 @@
+import {untracked} from '@angular/core';
+
+/**
+ * @deprecated "experimentalDecorators": false
```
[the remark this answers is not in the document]
nsbarsukov:
```suggestion
 * @deprecated use "experimentalDecorators": false
```

projects/cdk/utils/miscellaneous/pure.ts
```diff
@@ -47,6 +47,8 @@ function decorateGetter(
  * @throws error if used not on getter or function
  *
  * CAUTION: they must be pure.
+ * @deprecated "experimentalDecorators": false
```
nsbarsukov:
```suggestion
 * @deprecated use "experimentalDecorators": false
```

projects/cdk/utils/miscellaneous/pure.ts
```diff
@@ -47,6 +47,8 @@ function decorateGetter(
...
+ * TODO(v5): drop compatibility with legacy "experimentalDecorators": true
```
nsbarsukov:
```suggestion
 * TODO(v5): drop compatibility for legacy "experimentalDecorators": true
```

projects/cdk/utils/miscellaneous/pure.ts
```diff
@@ -90,6 +92,7 @@ export function tuiPure(
         throw new TuiPureException();
     }
 
+    // TODO(v5): drop compatibility with legacy "experimentalDecorators": true
```
nsbarsukov:
```suggestion
    // TODO(v5): drop compatibility for legacy "experimentalDecorators": true
```

PR #11649, opened 2025-08-14
chore: fix typography font size info

kenny-not-dead:
Сhanged the actual font size in the demo. Maybe this will help reduce the number of questions about the discrepancy, because I asked, and it turns out the question has already been raised =)


splincode (approved):

waterplea (approved):
We need some way to hack this back to 15px in proprietary theme because there it is 15px.

kenny-not-dead:
@waterplea do you mean the size change directly in the token? It's just that it appears 84 times in this repository, not counting the demo package

waterplea:
@kenny-not-dead maybe let's just use `getComputedStyle` to get its actual value? This way it will be automatically correct all the time in both public and proprietary themes.

kenny-not-dead:
@waterplea I tried it in my repository, indeed, this may be a solution, but I didn't quite understand at what level in the taiga this method should be applied, and how much worse, for example, could the method of redefining the token at the package level be in this case?

waterplea:
> @waterplea I tried it in my repository, indeed, this may be a solution, but I didn't quite understand at what level in the taiga this method should be applied, and how much worse, for example, could the method of redefining the token at the package level be in this case?

Just in the typography demo page create a dictionary property that computes values for all vars (with a fallback for SSR)

kenny-not-dead:
Just torn away from my main work. I hope I understood correctly, in the typography component, we create a method that takes a variable as input, and via getComputedStyle gets this value, after which we calculate the actual font size. Pseudocode might look something like this in this case? 
I apologize if I misunderstood how it should have been done

image

kenny-not-dead:
@waterplea Updated this pull request. Please clarify if I understood you correctly

waterplea:
@kenny-not-dead yes, something like that, please fix linting errors and maybe we can add some memoization? For example with `tuiMapper` pipe, so that this heavy operation is not executed on every change detection. Last one is optional since it's just a demo page anyway.

kenny-not-dead:
@waterplea  I fixed it, haven't used tuiMapper pipe yet

projects/demo/src/modules/markup/typography/index.ts
```diff
@@ -12,5 +13,18 @@ import {TuiDemo} from '@demo/utils';
...
+    private readonly isServer = isPlatformServer(inject(PLATFORM_ID));
+
     protected readonly routes = DemoRoute;
+
+    protected computedFontSize(className: string): string | null {
+        if (this.isServer) {
+            return null;
+        }
+
+        const element = this.doc.getElementsByClassName(className)?.item(0);
```
waterplea: It is preferable to use `querySelector`

kenny-not-dead:
@waterplea I corrected the code, according to the indicated comment. It really looks better this way

PR #11653, opened 2025-08-14
feat(demo-playwright): add scrollbar & dropdown performance tests 

mdlufy:
Failed screenshots:

**1. textfield-in-table.cy.ts-tui-table__1.diff.png**

If add an example like in the test to the main branch, the scroll will appear. 

This means that the scroll is not displayed in the test, although it should (it does not appear immediately). It didn't fire before, apparently because there was a [thtrottleTime](https://github.com/taiga-family/taiga-ui/blob/main/projects/core/components/scrollbar/scroll-controls.component.ts#L25) and the scroll didn't have time to appear.

Same behaviour on the main branch without throttle: https://github.com/taiga-family/taiga-ui/pull/11667

image

**2. tests-kit-dropdown-selecti-8e8ca-must-be-properly-positioned-chromium-retry2/05-dropdown-selection-scrolled.diff.png**

If open https://taiga-ui.dev/directives/dropdown-selection#textarea with `height: 400` in real browser the page will not scroll to the end, but will remain a short distance away. It looks like there is an outdated screenshot in the main branch.

image


splincode (approved):

projects/demo-playwright/tests/core/scrollbar/scrollbar.stress.pw.spec.ts
```diff
@@ -0,0 +1,594 @@
+import {DemoRoute} from '@demo/routes';
+import {
+    PerformanceCollector,
+    TuiDocumentationPagePO,
+    tuiGoto,
+} from '@demo-playwright/utils';
+import {expect, test} from '@playwright/test';
+
+test.describe('TuiScrollbar Stress Tests', () => {
```
vladimirpotekhin: let's move performance tests to a separate folder

mdlufy: done

PR #11663, opened 2025-08-15
chore(demo): normalize antialiased between chrome/safari/firefox

splincode:


projects/demo/src/styles/styles.less
```diff
@@ -5,6 +5,11 @@
 @import '@taiga-ui/styles/taiga-ui-global.less';
 @import '@docsearch/css/dist/style.css';
 
+* {
+    -webkit-font-smoothing: antialiased;
```
waterplea: `font-smoothing` is inherited, maybe it's better to move it down a few lines to `html` so it is not applied to everything via `*`?

projects/demo/src/styles/styles.less
```diff
@@ -4,13 +4,15 @@
 @import '@taiga-ui/addon-mobile/styles/taiga-ui-mobile.less';
 @import '@taiga-ui/styles/taiga-ui-global.less';
 @import '@docsearch/css/dist/style.css';
-
+q
```
waterplea: oops

splincode: I am sorry

PR #11677, opened 2025-08-18
fix(kit): switch check icon moved 2 pixels to the left

### LINKED ISSUE #11369: 🐞 - `Switch` check icon wrong padding
### Reproduction url

https://taiga-ui.dev/components/switch

### Description

check mark icon should be 2 pixels to the left according to the specs

### Taiga UI version

4.45.0

### Browser Used

Chrome

### OS Used

macOS

rashiddok:
Fixes #11369

### DIFF
--- a/projects/kit/styles/components/switch.less
+++ b/projects/kit/styles/components/switch.less
@@ -36,7 +36,7 @@
         inline-size: 2rem;
 
         &::before {
-            inline-size: 1rem;
+            inline-size: 1.125rem;
             transform: translateX(-1rem);
             mask-size: 0.75rem;
         }


projects/kit/styles/components/switch.less
```diff
@@ -82,6 +82,7 @@
         background: currentColor;
         mask: var(--t-checked-icon) no-repeat center;
         mask-size: 1rem;
+        inline-size: 1.45rem;
```
waterplea: 1.45rem === 23.2px

Are we sure we need fraction pixels?

rashiddok: Whoops

rashiddok: For sure 1.45 was incorrect changed to 1.375rem

waterplea: It is now broken for "s" size

rashiddok: > It is now broken for "s" size

Can I take a look at specs? Changes I've made does not affect s size but should them?

waterplea: Your change is applied to both sizes, because M is a default and S overrides defaults. Currently on `main` the icon is exactly left half of the component, but by specs it has to be 1px off to the right. We need size of 1.125rem for size S.

image

rashiddok: Currently size S have its its own `inline-size: 1rem;` so it overrides changes applied to defaults.
Changed inline-size to 1.175rem to correspond to the specs

PR #11695, opened 2025-08-19
chore(core): add tuiDropdownSided to documentation

### LINKED ISSUE #11492: 📚 - add tuiDropdownSided in api dropdown
### What is the affected URL?

https://taiga-ui.dev/directives/dropdown

### Description

Please add to the API dropdown display the tuiDropdownSided capability with a description of what it is needed for

### Browser Used

Chrome

### OS Used

macOS

Farbaks:
Fixes #11492


waterplea:
Dropdown docs are also used in other components, search code base for `tuiDocDropdown`. We need to either hide these options there if it makes no sense in that context, or make them work there as well. Alternatively, they can be hidden by default and only enabled on pages where it is needed.

Farbaks:
@waterplea I've made the requested changes, if you could please review it

waterplea (approved):

PR #11707, opened 2025-08-20
feat(experimental): `Dialog` add fullscreen appearance

waterplea:
Fixes #

### DIFF
--- a/projects/experimental/components/dialog/appearances/fullscreen.less
+++ b/projects/experimental/components/dialog/appearances/fullscreen.less
@@ -38,6 +38,11 @@
         display: none;
     }
 
+    &:not(._closable):not(.tui-app-bar)::before {
+        content: '';
+        block-size: 2rem;
+    }
+
     > tui-app-bar,
     > ng-component > tui-app-bar {
         position: sticky;
@@ -77,7 +82,8 @@
         margin: 0 calc(var(--tui-width) / 2 - 50vw) calc(-2rem - env(safe-area-inset-top));
 
         &::before {
-            display: none;
+            left: 0;
+            right: 0;
         }
     }
 
@@ -85,9 +91,12 @@
         > footer,
         > ng-component > footer {
             flex-direction: row;
-            background: var(--tui-background-base);
             background: color-mix(in hsl, var(--tui-background-base) 80%, transparent);
             backdrop-filter: blur(2rem);
+
+            &::before {
+                display: none;
+            }
 
             > button,
             > a {
--- a/projects/experimental/components/dialog/appearances/default.less
+++ b/projects/experimental/components/dialog/appearances/default.less
@@ -21,17 +21,17 @@
             margin-block-end: 1.25rem;
         }
 
+        /** TODO remove tui-confirm in v5 */
         > p,
         > ng-component > p,
-        /** TODO remove tui-confirm in v5 */
- > tui-confirm > p {
+        > tui-confirm > p {
             margin: -0.75rem 0 1.25rem;
         }
 
+        /** TODO remove tui-confirm in v5 */
         > footer,
         > ng-component > footer,
-        /** TODO remove tui-confirm in v5 */
- > tui-confirm > footer {
+        > tui-confirm > footer {
             margin-block-start: 1.25rem;
         }
     }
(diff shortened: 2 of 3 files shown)


projects/demo/src/modules/components/dialog-experimental/examples/6/index.html
```diff
@@ -1,34 +1,114 @@
...
-    (click)="custom = true"
+    (click)="scrollable = true"
 >
-    Show custom
+    Show scrollable
 </button>
 <ng-template
-    [tuiDialogOptions]="{size: 's', appearance: 'taiga compact'}"
-    [(tuiDialog)]="augmented"
+    [tuiDialogOptions]="{appearance: 'fullscreen', label: 'Fullscreen heading'}"
```
mdlufy, line 18: Will we migrate `tuiSheetDialogOptions` `fullscreen: true` flag to `appearance: 'fullscreen'` too in the future?

mdlufy: And what if the user want to use custom appearance with fullscreen? Is it a valid case?

projects/addon-mobile/components/sheet-dialog/sheet-dialog.style.less
```diff
@@ -48,6 +50,22 @@
             display: flex;
             flex-direction: column;
             flex-grow: 1;
+
+            ::ng-deep > header,
+            ::ng-deep > [tuiSlides] > * > header,
```
mdlufy, line 55: Are `tuiSlided` already created or it's for future support?

projects/experimental/components/dialog/appearances/fullscreen.less
```diff
@@ -0,0 +1,139 @@
...
+    &.tui-leave {
+        animation-name: tuiFade, tuiSlide;
+    }
+
+    > .t-close {
+        position: static;
+        block-size: 4rem;
+        inline-size: auto;
+        align-self: flex-end;
+        margin: env(safe-area-inset-top) 0 2rem;
```
mdlufy: It seems that it will be better to set safe-area for top, but not for margin:

With margin:

image

With top (`top: ~'max(1rem, env(safe-area-inset-top))'`)

image


Note: need to change top only for the fullscreen to prevent such cases -

image

projects/experimental/components/dialog/appearances/fullscreen.less
```diff
@@ -0,0 +1,139 @@
...
+    // Replace with :has when it's supported
+    //&:has(tui-app-bar) .t-close {
+    &.tui-app-bar .t-close {
+        display: none;
+    }
+
+    > tui-app-bar,
+    > ng-component > tui-app-bar {
+        position: sticky;
+        top: env(safe-area-inset-top);
```
mdlufy: Seems like we need to set `top: 0`, because safe-area already included in margin

Now:

image


If set `top: 0`:

image

projects/addon-mobile/styles/taiga-ui-mobile.less
```diff
@@ -12,7 +12,6 @@
 }
 
 [data-platform='ios'] {
-    @import 'ios/app-bar.less';
```
mdlufy, line 15: breaking change?

projects/addon-mobile/components/sheet-dialog/sheet-dialog.style.less
waterplea: I will make a PR right after this one, it's almost done

projects/addon-mobile/styles/taiga-ui-mobile.less
waterplea: Fixing a mistake. We have size of 48px that is never used, iOS was overriding it to 44px, Android to 56px and desktop always had 64px.

projects/demo/src/modules/components/dialog-experimental/examples/6/index.html
waterplea: Yes, you can use "fullscreen custom", we use ~= to match individual words, not = to match whole data-appearance

nsbarsukov (approved):

Issue #11665, opened 2025-08-15
🐞 - `TuiInputChip` moves long values to a new string


Stivin:
### Reproduction url

https://stackblitz.com/edit/angular-yyevxj6w?file=src%2Fapp%2Fapp.template.html

### Description

Hi! It was found that without using chip, long values are moved to a new line and thus expand the field vertically.

### Taiga UI version

4.49.0

### Browser Used

Chrome

### OS Used

macOS

waterplea:
Thank you! We need `white-space: nowrap` here if the mode is horizontal:
https://github.com/taiga-family/taiga-ui/blob/main/projects/core/components/textfield/textfield-multi/textfield-item.style.less#L4

galanggg:
Hi @waterplea , i create the PR here #11708 Thanks.

PR #11708, opened 2025-08-20
fix(styles): change white-space property to nowrap

galanggg:
This pull request makes a small update to the styling of the `textfield-item` component to improve its text display behavior. Refer this issue TuiInputChip moves long values to a new string (https://github.com/taiga-family/taiga-ui/issues/11665)

* Changed the `white-space` property from `pre-wrap` to `nowrap` in `textfield-item.style.less`, causing text to remain on a single line and overflow with an ellipsis instead of wrapping.

### DIFF
--- a/projects/core/components/textfield/textfield-multi/textfield-item.style.less
+++ b/projects/core/components/textfield/textfield-multi/textfield-item.style.less
@@ -12,7 +12,7 @@
         }
 
         &::after {
-            content: ', ';
+            content: ',\00a0';
         }
     }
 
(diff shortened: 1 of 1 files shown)


projects/core/components/textfield/textfield-multi/textfield-item.style.less
```diff
@@ -1,7 +1,7 @@
 :host {
     max-inline-size: 100%;
     flex-shrink: 0;
-    white-space: pre-wrap;
+    white-space: nowrap;
```
waterplea, line 4: This breaks `', '`, the space is now collapsed. You need to change regular space to non-breaking on line 15: `",\00a0"`

galanggg: Noted, i put the update in a new commit [2e616e5](https://github.com/taiga-family/taiga-ui/pull/11708/commits/2e616e5d7e382de497b8498690e17500d37c7bdb). Thank you

waterplea (approved):

Issue #11116, opened 2025-06-17
🚀 - `Transition` directive


waterplea:
### Description

Some elements have CSS transitions and when during creation something triggers reflow before host bindings are applied we sometimes get transition to initial state:
https://stackblitz.com/edit/angular-nhb47vcp-h7ydnaff

We need to make have a universal solution to fix that.

The only idea that comes to mind is a host directive `Transition` that would have a `tui-transition` class that has all of it's children set to `transition: none !important` and a renderer hook that would remove this class. It's not that handy to add it all the time, but maybe we can at least start with textfields and some other components that behave like that, for example, appearance directive. If we have a better idea – let's hear it. Added problem is that element might be created but not yet added to DOM, and if we remove the class before we add it to DOM – the transition might still happen. Anyway, this issue needs investigating.

waterplea:
Here's the basic idea for the directive:
```ts
@Directive({
  standalone: true,
  host: { '[style.transition]': 'transition()' }
})
export class TuiTransition {
  protected readonly transition = toSignal(timer(0).pipe(
    map(() => ''),
    startWith('none')
  ));
}
```
We need it on `Fade` and `Appearance` directives as host directive. And we also need to make sure textfields have transitions on internal elements, such as label, disabled this way too.

KuzGruz:
Hello, @waterplea !
The issue with transition is quite noticeable in the new `tuiTextfield`. When the component appears, an animation may play on the label (as if the value disappears and the label moves back to its normal position, i.e., it grows), or the size of the field may change.

I'm attaching an example of how the field height changes from the initial state during routing, when the component appears (I applied the directive `tuiTextfieldSize="m"`).

![Image](https://github.com/user-attachments/assets/3ec646fe-cab1-4990-b198-0e2f2902e344)

waterplea:
Since this is a nested element and not host the directive I suggested would not fix that, but we can use `transition-duration: inherit` in label styles to remedy that. Here's [another issue](https://github.com/taiga-family/taiga-ui/issues/11640) for the same thing, I'm closing it as duplicate and bumping priority of this issue and dropping contributions label, I'll get to it after I finish what I'm working on now.

waterplea:
Issue from a linked ticket is still reproduced on the latest version:
https://stackblitz.com/edit/angular-r3z4cbs6-abvrevsn

vladimirpotekhin:
[...]

PR #11720, opened 2025-08-20
fix(core): remove initial transition in textfield

### LINKED ISSUE #11640: 🐞 - `InputChip` - problem with sizing when placed next to `BadgeContent`
### Reproduction url

https://stackblitz.com/edit/angular-r3z4cbs6-dxgxy77f?file=src%2Fapp%2Ffilter%2Ffilter.component.ts

### Description

Hi! Previously found a similar problem with Textfield + BadgeContent when they are in the Search component, but the Search component will be wrapped as a separate custom component. Then, when opening/reloading the page, there was a strange behavior of the animation, when the input fields first get larger than desired in height and immediately start to shrink to the normal size.

https://github.com/user-attachments/assets/4d7354f3-09de-455e-b96c-4c3eb13f7b13

### Taiga UI version

4.49.0

### Browser Used

Chrome

### OS Used

macOS

splincode:
Fixes #11640

### DIFF
--- a/projects/core/components/textfield/textfield.component.ts
+++ b/projects/core/components/textfield/textfield.component.ts
@@ -1,6 +1,7 @@
 import {AsyncPipe, NgIf} from '@angular/common';
 import {
     type AfterContentInit,
+    afterNextRender,
     ChangeDetectionStrategy,
     Component,
     computed,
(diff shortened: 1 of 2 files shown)


waterplea:
We need to solve this globally:
https://github.com/taiga-family/taiga-ui/issues/11116

PR #11761, opened 2025-08-25
feat(kit): `Slides` add new component

waterplea:
Fixes #


splincode (approved):

projects/kit/components/slides/slides.style.less
```diff
@@ -0,0 +1,35 @@
+[tuiSlides] {
+    display: grid;
+    isolation: isolate;
+
+    &[data-direction='-1'] > * {
+        --tui-from: translateX(-2rem);
+
+        &.tui-leave {
+            --tui-from: translateX(1rem);
```
mdlufy, line 9: Can we animate full width, not only fixed? It could be usefull for onboarding

projects/demo/src/modules/components/slides/index.less
```diff
@@ -0,0 +1,7 @@
+[tuiSlides] ng-component {
+    display: block;
+
+    &.tui-slides {
+        display: none;
+    }
+}
```
mdlufy, line 7: Looks some new approach, I haven't seen the ng-component tag in user side styles before. Why?

projects/demo/src/modules/components/slides/index.html
```diff
@@ -0,0 +1,54 @@
...
+                    </a>
+                    to host directives of route component is required.
+                </ng-container>
+                <ng-container *ngSwitchCase="3">Using slides inside a dialog.</ng-container>
+            </ng-template>
+        </tui-doc-example>
+    </ng-template>
+
+    <tui-setup *pageTab="'Setup'" />
+</tui-doc-page>
```
mdlufy, line 85: Maybe add API page?

projects/demo/src/modules/components/slides/index.less
waterplea: ng-component is created when dynamic component doesn't have a selector

projects/kit/components/slides/slides.style.less
waterplea: We can create a CSS var for it later

projects/demo/src/modules/components/slides/index.less
mdlufy: Is it possible to make some kind of selector? Otherwise it's not obvious. Or would it be unnecessary?

mdlufy (approved):

projects/demo/src/modules/components/slides/index.less
waterplea: It is unnecessary

vladimirpotekhin (approved):

PR #11781, opened 2025-08-27
ci: prevents schematic tests from running when they aren't needed

splincode:


mdlufy (approved):

.github/workflows/schematic.yml
```diff
@@ -1,6 +1,10 @@
 name: ⚙️ Schematics
 on:
+  workflow_dispatch:
```
vladimirpotekhin, line 3: Is this for manual trigger of workflow?

splincode: Yes, manual triggering of the action is available from any branch

vladimirpotekhin (approved):

PR #11800, opened 2025-08-30
feat(core): migrate from Angular events to observables in `DropdownContext`

Markiewic:
Remove host-listeners from `TuiDropdownContext` that work outside the zone.

Zoneless host-listeners still force marking `lView` as dirty, and do this on every click on absolutely any place. On the next tick, all this will be processed, and in most cases there will be no need for this.

Replace host-listeners with subscription via RxJS using `tuiZonefree`, which will not lead to marking `lView` as dirty.

### DIFF
--- a/projects/core/directives/dropdown/dropdown-context.directive.ts
+++ b/projects/core/directives/dropdown/dropdown-context.directive.ts
@@ -1,5 +1,5 @@
 import {DOCUMENT} from '@angular/common';
-import {computed, DestroyRef, Directive, inject, NgZone} from '@angular/core';
+import {computed, Directive, inject} from '@angular/core';
 import {takeUntilDestroyed} from '@angular/core/rxjs-interop';
 import {EMPTY_CLIENT_RECT} from '@taiga-ui/cdk/constants';
 import {TuiActiveZone} from '@taiga-ui/cdk/directives/active-zone';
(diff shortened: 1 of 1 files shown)


projects/core/directives/dropdown/dropdown-context.directive.ts
```diff
@@ -41,6 +43,16 @@ export class TuiDropdownContext extends TuiRectAccessor {
...
     protected readonly driver = inject(TuiDropdownDriver);
+    protected readonly doc = inject(DOCUMENT);
+    protected readonly zone = inject(NgZone);
+    protected readonly destroyRef = inject(DestroyRef);
+
+    protected readonly sub = merge(
+        tuiTypedFromEvent(this.doc, 'pointerdown'),
+        tuiTypedFromEvent(this.doc, 'contextmenu', {capture: true}),
+    )
+        .pipe(tuiZonefree(this.zone), takeUntilDestroyed(this.destroyRef))
```
splincode: Could you check if it works without injectors?

```ts
.pipe(
   tuiZonefree(),
   takeUntilDestroyed()
)
```

splincode (approved):

projects/core/directives/dropdown/dropdown-context.directive.ts
Markiewic: Yes, it works. Updated my branch.

PR #11824, opened 2025-09-03
Updating Angular

waterplea:
Fixes #

### DIFF
--- a/projects/demo/src/modules/components/calendar/examples/4/index.html
+++ b/projects/demo/src/modules/components/calendar/examples/4/index.html
@@ -1,1 +1,1 @@
-<tui-calendar />
+<tui-calendar [value]="value" />
--- a/projects/demo/src/modules/components/calendar/examples/4/index.ts
+++ b/projects/demo/src/modules/components/calendar/examples/4/index.ts
@@ -1,7 +1,7 @@
 import {Component} from '@angular/core';
 import {changeDetection} from '@demo/emulate/change-detection';
 import {encapsulation} from '@demo/emulate/encapsulation';
-import {TuiDayOfWeek} from '@taiga-ui/cdk';
+import {TuiDay, TuiDayOfWeek} from '@taiga-ui/cdk';
 import {TUI_FIRST_DAY_OF_WEEK, TuiCalendar} from '@taiga-ui/core';
 
 @Component({
@@ -17,4 +17,6 @@ import {TUI_FIRST_DAY_OF_WEEK, TuiCalendar} from '@taiga-ui/core';
         },
     ],
 })
-export default class Example {}
+export default class Example {
+    protected value = new TuiDay(2025, 6, 4);
+}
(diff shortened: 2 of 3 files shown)


projects/demo/src/main.server.ts
```diff
@@ -11,11 +11,13 @@ import {UNIVERSAL_PROVIDERS} from '@ng-web-apis/universal';
 import {App} from './modules/app/app.component';
 import {config} from './modules/app/app.config';
 import {ServerErrorHandler} from './modules/app/server-error-handler';
+import {provideAnimations} from '@angular/platform-browser/animations';
 
 const serverConfig = mergeApplicationConfig(config, {
     providers: [
         importProvidersFrom(ServerModule),
         provideServerRendering(),
+        provideAnimations(),
```
mdlufy, line 20: To fix error in ssr:

```
/components/expand
NG05105: Unexpected synthetic property @tuiParentAnimation found. Please make sure that:

Make sure provideAnimationsAsync(), provideAnimations() or provideNoopAnimations() call was added to a list of providers used to bootstrap an application.
```

image

mdlufy: @nsbarsukov what do you think why it's not enough provideAnimations from app.config.ts?

waterplea: We will remove animations before v5 release anyway so it doesn't matter, we just need to satisfy it for now so it compiles.

vladimirpotekhin (approved):

projects/demo/src/modules/app/app.config.ts
```diff
@@ -79,7 +79,7 @@ export const config: ApplicationConfig = {
         NG_EVENT_PLUGINS,
         tuiEnableFontScaling(),
         tuiNotificationOptionsProvider({size: 'm'}),
-        provideHttpClient(),
+        provideHttpClient(withFetch()),
```
mdlufy, line 82: To fix:

`NG02801: Angular detected that `HttpClient` is not configured to use `fetch` APIs. It's strongly recommended to enable `fetch` for applications that use Server-Side Rendering for better performance and compatibility. To enable `fetch`, add the `withFetch()` to the `provideHttpClient()` call at the root of the application.`

tsconfig.json
```diff
@@ -46,5 +47,12 @@
             "@taiga-ui/testing": ["projects/testing/index"],
             "@taiga-ui/testing/*": ["projects/testing/*"]
         }
+    },
+    "angularCompilerOptions": {
+        "extendedDiagnostics": {
+            "checks": {
+                "unusedStandaloneImports": "suppress"
```
mdlufy, line 54: Maybe set "error" to be consistence with https://github.com/taiga-family/taiga-ui/pull/11824/files#diff-3ae20d611c1c7263a9c1bce0449291ebfea1c5d578b3fcdbb596353ad885db25R43 to both local and prod build?

mdlufy (approved):

Issue #11776, opened 2025-08-26
🐞 - slider do not working after pathing value


TatyanaKonina:
### Reproduction url

https://stackblitz.com/edit/angular-vvkhzdut?file=src%2Fapp%2Fapp.component.ts

### Description

The slider do not working propertly after updating from 4.48 to 4.50 version. After pathing value in form it displays wrong range.

how to reproduce

1) click the button on the reproduction example

current behavior - slider range is 1
expected behavior - slider range is 0.5

### Taiga UI version

4.50

### Browser Used

Chrome

### OS Used

macOS

splincode:
If I add quantum, everything works 

```
  <input
    tuiInputSlider
    [quantum]="0.25"
    [max]="max"
    [min]="min"
    [formControl]="value"
  />
```

PR #11878, opened 2025-09-09
chore(demo): add patch method in example to input-slider page

splincode:
Part of https://github.com/taiga-family/taiga-ui/issues/11776


projects/demo/src/modules/components/input-slider/examples/5/index.html
```diff
@@ -15,4 +15,14 @@
...
 <p><strong>Control value:</strong></p>
+
+<button
+    size="xs"
+    tuiButton
+    type="button"
+    (click)="value = 0.75"
+>
+    Set 0.75
+</button>
```
nsbarsukov, line 26: I am not sure that it is required. We need Cypress CT with bug fix

PR #11931, opened 2025-09-15
refactor(kit)!: `Avatar` drop images in favor of nested `img` tag, drop `FallbackSrc` pipe

waterplea:
Fixes #


projects/core/directives/icons/icons.directive.ts
```diff
@@ -18,10 +16,7 @@ import {
     tuiInjectIconResolver,
 } from '@taiga-ui/core/tokens';
 
-/**
- * Workaround for {@link TuiAvatar} to properly handle icon color in {@link TuiAppearance}
- */
-type Icon = SafeResourceUrl | string | null | undefined;
+const OPT = {self: true, optional: true} as const;
```
mdlufy, line 19: Maybe without it? It seems to me that readability is suffering

projects/core/tokens/icon-resolver.ts
```diff
@@ -23,7 +23,7 @@ export const TUI_ICON_RESOLVER = new InjectionToken<TuiStringHandler<string>>(
 export const TUI_ICON_START_RESOLVER = TUI_ICON_RESOLVER;
 
 export function tuiGetIconMode(icon?: string | null): string | undefined {
-    return icon?.slice(icon.indexOf('@') + 1, icon.indexOf('.')) || undefined;
+    return icon?.match(/@([^.]*)\./)?.[1] || icon || undefined;
```
mdlufy, line 26:
```suggestion
   // example: @tui.some - valid, @tui-some - invalid
    const regexp = /@([^.]*)\./;
   
    return icon?.match(regexp)?.[1] || icon || undefined;
```

projects/kit/components/avatar/avatar.directive.ts
```diff
@@ -0,0 +1,63 @@
...
+        {
+            directive: TuiIcons,
+            inputs: ['iconStart: tuiAvatar'],
+        },
+    ],
+    host: {
+        tuiAvatar: '',
+        '[attr.data-size]': 'size',
+        '[class._round]': 'round',
+        '[class._initials]': 'icons.iconStart()?.length < 3',
```
mdlufy, line 45: Why < 3 ?
Maybe add some explanation?

waterplea: It's initials, either 1 or 2 chars. I think it's fine without a comment.

projects/core/directives/icons/icons.directive.ts
waterplea: I prefer that actual code in input is in one line without options noise. They are not ever gonna change anyway

Issue #11481, opened 2025-07-30
🚀 - `Tabs` performance improvement


waterplea:
### Description

1. We need to play with CSS containment to improve tabs underline performance
2. We need to enable [CSS anchoring](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_anchor_positioning) instead of JS on browsers that do support it

Chiorufarewerin:
I will try it

PR #11932, opened 2025-09-15
perf(kit): improve Tabs underline performance

Chiorufarewerin:
Fixes https://github.com/taiga-family/taiga-ui/issues/11481

The main improvement comes from CSS anchoring. During testing CSS containment (strict/layout/paint/etc...) I didn't notice any improvement benefits nor degradation, therefore I decided to remain `strict`.

Before:
Screenshot 2025-09-15 15-04-52

After:

Screenshot 2025-09-15 15-08-22


mdlufy (changes requested):
Don't forget to check why e2e screenshots failed

Chiorufarewerin:
> Don't forget to check why e2e screenshots failed

Currently Screenshot tests show out of bound index as valid. https://taiga-ui.dev/navigation/tabs/API?activeItemIndex=-2
I suggest use the new behavior, because current behavior missinterprets what is actually "active". Tab doesn't have the active state, but has an underline. 

Also it fixes rtl. I mean, during the switch to rtl, the original behavior doesn't change it immediatly. 

So, do I need to change the current behavior or remain the same?

Chiorufarewerin:
The PR was updated in order to have the same behavior for anchoring and without it.

It changes 2 things:
1. The unactive tab is not longer has underline when out of bound
2. During switch to RTL mode, underline also updates

projects/kit/components/tabs/tabs.style.less
```diff
@@ -107,6 +109,13 @@ tui-tabs {
         inline-size: var(--t-width);
         background: var(--t-color);
         animation: tuiPresent 1ms;
+        contain: strict;
+
+        @supports (anchor-name: --tui-tab-active) {
+            left: anchor(left);
```
waterplea, line 113: inset-inline-start?

projects/kit/components/tabs/tabs-with-more.component.ts
```diff
@@ -127,7 +130,12 @@ export class TuiTabsWithMore implements AfterViewChecked, AfterViewInit {
     }
 
     public ngAfterViewChecked(): void {
-        this.refresh();
+        afterNextRender(
```
waterplea: Why do we need this change?

Chiorufarewerin: I remember it was fixing the old behavior for changing to RTL. Here are the screenshots to understand (before and after switching)

Image
Image

But I notice, this line doesn't fix it now. Maybe something changed. I guess I'll remove it.

The new behavior with anchor works fine.

projects/kit/components/tabs/tabs.style.less
Chiorufarewerin: In my case inset-inline-start doesn't work.

If I set it as property instead of left, there will be the incorrect behavior for RTL

Image

But if I set for anchor fn - `anchor(inset-inline-start);` - the browser says "incorrect property name"

I decided to remain left - works fine

PR #11940, opened 2025-09-16
refactor(addon-commerce)!: `InputCard` drop options and `icon` input

waterplea:
Fixes #


splincode (approved):

projects/kit/components/input-color/input-color.style.less
```diff
@@ -11,6 +11,7 @@
         padding: 0 !important;
         overflow: hidden;
         cursor: pointer;
+        order: 100;
```
mdlufy, line 14: Looks some strange

mdlufy (approved):

projects/kit/components/input-color/input-color.style.less
waterplea: 1 would also work, but just in case if we need to place something in between cleaner and picker.

PR #11959, opened 2025-09-17
feat(kit): add toast service

splincode:
Fixes TUIPRJ-32

https://github.com/user-attachments/assets/14f40f17-76a0-4252-ab2c-8a19c5da6d63

### DIFF
--- a/projects/demo/src/modules/components/toast/examples/2/index.ts
+++ b/projects/demo/src/modules/components/toast/examples/2/index.ts
@@ -1,16 +1,27 @@
-import {Component} from '@angular/core';
+import {isPlatformServer} from '@angular/common';
+import {Component, inject, PLATFORM_ID} from '@angular/core';
+import {toSignal} from '@angular/core/rxjs-interop';
 import {changeDetection} from '@demo/emulate/change-detection';
 import {encapsulation} from '@demo/emulate/encapsulation';
-import {TuiPlatform} from '@taiga-ui/cdk';
+import {TUI_IS_E2E, TuiPlatform} from '@taiga-ui/cdk';
 import {TuiButton, TuiLoader} from '@taiga-ui/core';
-import {TuiToast} from '@taiga-ui/kit';
+import {TuiProgressCircle, TuiToast} from '@taiga-ui/kit';
+import {BehaviorSubject, of, switchMap, take, timer} from 'rxjs';
 
 @Component({
     standalone: true,
-    imports: [TuiButton, TuiLoader, TuiPlatform, TuiToast],
+    imports: [TuiButton, TuiLoader, TuiPlatform, TuiProgressCircle, TuiToast],
     templateUrl: './index.html',
     styleUrls: ['./index.less'],
     encapsulation,
     changeDetection,
 })
-export default class Example {}
+export default class Example {
+    protected readonly trigger$ = new BehaviorSubject(0);
+    protected readonly value = toSignal(
+        inject(TUI_IS_E2E) || isPlatformServer(inject(PLATFORM_ID))
+            ? of(30)
+            : this.trigger$.pipe(switchMap(() => timer(0, 200).pipe(take(100)))),
+        {initialValue: 0},
+    );
+}
--- a/projects/demo/src/modules/components/toast/examples/3/index.html
+++ b/projects/demo/src/modules/components/toast/examples/3/index.html
@@ -1,0 +1,37 @@
+<button
+    tuiButton
+    type="button"
+    (click)="primitive()"
+>
+    String
+</button>
+<button
+    tuiButton
+    type="button"
+    (click)="template.set(true)"
+>
+    Template
+</button>
+<button
+    tuiButton
+    type="button"
+    (click)="component()"
+>
+    Component
+</button>
+<ng-template
+    [tuiToast]="template()"
+    [tuiToastOptions]="{closable: false}"
+    (tuiToastChange)="template.set($event)"
+>
+    <a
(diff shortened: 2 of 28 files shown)


projects/demo/src/modules/components/toast/examples/2/index.ts
```diff
@@ -1,16 +1,29 @@
...
     changeDetection,
 })
-export default class Example {}
+export default class Example {
+    protected readonly trigger$ = new BehaviorSubject(0);
+    protected readonly value = toSignal(
+        inject(TUI_IS_E2E) || isPlatformServer(inject(PLATFORM_ID))
+            ? of(30)
+            : this.trigger$.pipe(
+                  switchMap(() => timer(0, 200).pipe(map(identity), take(100))),
```
nsbarsukov: Why `map(identity)` ?

projects/demo/src/modules/components/toast/examples/3/index.html
```diff
@@ -0,0 +1,37 @@
...
+>
+    <a
+        href="https://github.com/taiga-family/taiga-ui"
+        iconEnd="@tui.external-link"
+        iconStart="@tui.github"
+        rel="noreferrer noopener"
+        target="_blank"
+        tuiToast
+    >
+        Check out sourcecode
```
nsbarsukov:
```suggestion
        Check out source code
```

projects/kit/components/multi-select/multi-select-group/multi-select-group.component.ts
```diff
@@ -63,9 +63,8 @@ export class TuiMultiSelectGroupComponent<T> {
     protected toggle(): void {
         const values = this.values();
         const matcher = this.handlers.identityMatcher();
-        const others = this.value().filter((current) =>
-            values.every((item) => !matcher(current, item)),
-        );
+        const value = this.value() || [];
+        const others = value.filter((a) => values.every((b) => !matcher(a, b)));
```
nsbarsukov: Why not separate PR ?

projects/kit/components/toast/toast.component.ts
```diff
@@ -0,0 +1,67 @@
...
+            switchMap((autoClose) => (autoClose ? timer(autoClose) : EMPTY)),
+            takeUntil(fromEvent(this.el, 'mouseenter')),
+            repeat({delay: () => fromEvent(this.el, 'mouseleave')}),
+            takeUntilDestroyed(),
+        )
+        .subscribe(() => this.context.$implicit.complete());
+
+    public ngAfterViewInit(): void {
+        if (this.button) {
+            this.el.querySelector('[tuiToast]')?.appendChild(this.button.nativeElement);
```
nsbarsukov, line 58: Could you explain what happens here and why we do it, please ?

projects/kit/components/toast/toast.service.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import {TuiToastComponent} from './toast.component';
+import {TUI_TOAST_OPTIONS, type TuiToastOptions} from './toast.options';
+
+const TOASTS = tuiCreateToken(
+    new BehaviorSubject<ReadonlyArray<TuiPopover<TuiToastOptions<unknown>, any>>>([]),
+);
+
+export const TUI_TOASTS_CONCURRENCY = tuiCreateTokenFromFactory<number>(() =>
+    inject(TUI_IS_MOBILE) ? 1 : 2,
```
nsbarsukov, line 21: Should we add documentation example to demonstrate how to configure this behavior ?

projects/kit/components/toast/toast.service.ts
```diff
@@ -0,0 +1,51 @@
...
+import {BehaviorSubject, pairwise} from 'rxjs';
+
+import {TuiToastComponent} from './toast.component';
+import {TUI_TOAST_OPTIONS, type TuiToastOptions} from './toast.options';
+
+const TOASTS = tuiCreateToken(
+    new BehaviorSubject<ReadonlyArray<TuiPopover<TuiToastOptions<unknown>, any>>>([]),
+);
+
+export const TUI_TOASTS_CONCURRENCY = tuiCreateTokenFromFactory<number>(() =>
```
nsbarsukov, line 20: Token or just rxjs power (similar to `Alerts`) ? 🤔 

https://taiga-ui.dev/components/alert#concurrency-limit

@taiga-family/core-team what do you think ?

nsbarsukov (approved):

projects/kit/components/toast/toast.component.ts
waterplea: So that X button to close toast is automatically added for closable toasts if you pass them as template/component, not just strings.

projects/demo/src/modules/components/toast/examples/2/index.ts
waterplea: Stupid me

projects/kit/components/toast/toast.service.ts
waterplea: The specs say it has to be 1 on mobile and 2 on desktop so we need it baked in, but overridable as usual.

projects/kit/components/multi-select/multi-select-group/multi-select-group.component.ts
waterplea: It's a small issue that I ran into when updating proprietary demo, value can be `null` initially so I just fixed the type so that it is more correct. Yes, I guess it's better to move it to a different PR, rolled back.

PR #11966, opened 2025-09-17
refactor(kit): refactor all toggles to use `Icons` directive

waterplea:
Fixes #


projects/kit/components/input-chip/input-chip.directive.ts
```diff
@@ -76,6 +76,7 @@ export class TuiInputChipBaseDirective<T>
 
     public setValue(value: T[]): void {
         this.textfield.value.set('');
+        new KeyboardEvent('keydown', {});
```
mdlufy: Why?

mdlufy (approved):

projects/kit/components/input-chip/input-chip.directive.ts
waterplea: Forgot to remove, was testing something

PR #11977, opened 2025-09-17
chore(demo): `InputTag` migrate to `InputChip`

mdlufy:
Fixes #


projects/demo/src/modules/components/table/examples/8/index.html
```diff
@@ -9,6 +9,7 @@
             </th>
             <th
                 tuiTh
+                [minWidth]="100"
```
mdlufy, line 12: Looks too narrow without it:

IMG_20250918_014339_061

splincode (approved):

PR #11995, opened 2025-09-18
feat(core): `Dialogs` auto attach to current active zone

### LINKED ISSUE #10137: 🚀 - Не изменять активную зону при открытии модального окна из нее
### Description

Есть sidebar (1) он закрывается когда я кликаю куда-нибудь за его пределы, но когда я жму кнопку добавить, должна появиться модалка (2) и собственно она открывается, но шторка закрывается. 

Хочется что бы сайдбар не закрывался (активная зона не менялась) при открытии модального окна и взаимодействии с ним

waterplea:
Fixes #10137

### DIFF
--- a/projects/addon-mobile/components/sheet-dialog/sheet-dialog.style.less
+++ b/projects/addon-mobile/components/sheet-dialog/sheet-dialog.style.less
@@ -21,7 +21,7 @@
     }
 
     &::before {
-        .fullsize();
+        .fullsize(fixed);
 
         content: '';
         z-index: -1;
--- a/projects/core/components/fullscreen/fullscreen.component.ts
+++ b/projects/core/components/fullscreen/fullscreen.component.ts
@@ -10,6 +10,7 @@ import {
     signal,
     ViewChild,
 } from '@angular/core';
+import {TuiActiveZone} from '@taiga-ui/cdk/directives/active-zone';
 import {TuiRoot} from '@taiga-ui/core/components/root';
 
 @Component({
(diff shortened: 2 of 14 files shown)


projects/demo/src/modules/directives/active-zone/examples/2/index.ts
```diff
@@ -0,0 +1,33 @@
...
+    encapsulation,
+    changeDetection,
+})
+export default class Example {
+    private readonly dialog = inject(TuiDialogService);
+
+    protected active = false;
+
+    protected onZone(active: boolean): void {
+        console.log(active);
```
splincode: console.info

splincode (approved):

projects/core/components/fullscreen/fullscreen.component.ts
```diff
@@ -19,6 +20,12 @@ import {TuiRoot} from '@taiga-ui/core/components/root';
     template: '<tui-root><ng-content /></tui-root>',
     styleUrls: ['./fullscreen.style.less'],
     changeDetection: ChangeDetectionStrategy.OnPush,
+    viewProviders: [
+        {
+            provide: TuiActiveZone,
+            useValue: null,
```
mdlufy, line 26: null to prevent di error or for another one? 
if so, maybe optional in inject will help?

mdlufy (approved):

projects/addon-mobile/components/sheet-dialog/sheet-dialog.style.less
```diff
@@ -21,7 +21,7 @@
     }
 
     &::before {
-        .fullsize();
```
mdlufy, line 24: maybe add some comment to avoid breaking? for example

// don't remove `fixed` due to...

waterplea: Pretty sure it can break due to all sorts of things, it's just a coincidence that this thing is also needed

projects/core/components/fullscreen/fullscreen.component.ts
waterplea: null is so that tui-root in the template does not attach itself to the parent zone. This creates two parallel branches which makes it hard to find the actual nested zone:
```
parent-zone
 |      |
 |   tui-root
dropdown zone
       |
focused element
```
focused element is both in tui-root zone and in parent zone, but since tui-root is in the view - dropdown zone is not under tui-root, DI from view does not propagate to content.

PR #12019, opened 2025-09-20
fix(experimental): `Dialog` fix `label`

waterplea:
Fixes #

### DIFF
--- a/projects/kit/components/confirm/confirm.style.less
+++ b/projects/kit/components/confirm/confirm.style.less
@@ -40,5 +40,9 @@
         margin-block-end: 0;
         margin-inline-end: 0;
         margin-inline-start: 0;
+
+        &:first-child {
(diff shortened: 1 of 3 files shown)


projects/kit/components/confirm/confirm.style.less
```diff
@@ -40,5 +40,9 @@
         margin-block-end: 0;
         margin-inline-end: 0;
         margin-inline-start: 0;
+
+        &:first-child {
+            margin-block-start: 2rem;
+        }
```
mdlufy, line 46: image

In my opinion, it looked visually better before that

mdlufy (approved):

projects/kit/components/confirm/confirm.style.less
waterplea: Yes, it's a bug, of course )

PR #12041, opened 2025-09-23
refactor(core)!: `Error` automatically show control errors, drop `FieldError` pipe

waterplea:
Fixes #

### DIFF
--- a/projects/demo/src/modules/components/error/examples/2/index.ts
+++ b/projects/demo/src/modules/components/error/examples/2/index.ts
@@ -1,3 +1,4 @@
+import {isPlatformBrowser} from '@angular/common';
 import {Component, inject, PLATFORM_ID, signal} from '@angular/core';
 import {toSignal} from '@angular/core/rxjs-interop';
 import {FormControl, FormGroup, ReactiveFormsModule, Validators} from '@angular/forms';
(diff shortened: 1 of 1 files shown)


projects/core/components/error/error.directive.ts
```diff
@@ -0,0 +1,123 @@
...
+        computed(
+            (errors = this.errors() || null) =>
+                errors && this.getError(errors[this.key()], this.content[this.key()]),
+        ),
+        {self: true, optional: true},
+    );
+
+    public registerOnChange(): void {}
+    public registerOnTouched(): void {}
+    public writeValue(): void {}
```
nsbarsukov, line 77: Why not `extends TuiControl` ?

projects/demo/src/modules/components/error/examples/1/index.html
```diff
@@ -4,6 +4,53 @@
...
+    <tui-error formControlName="password" />
+
+    <p>
+        If you want to show a validation message as soon as a user starts typing, subscribe on form value changes and
+        call markAsTouched on control on first value change. You can also use
+        <a
+            fragment="mark-control-as-touched-and-validate"
+            href="/utils/miscellaneous#mark-control-as-touched-and-validate"
+            tuiLink
+            [routerLink]="routes.Miscellaneous"
```
nsbarsukov: Should we drop `href` if we already have `routerLink` ?

nsbarsukov (approved):

projects/core/components/error/error.directive.ts
waterplea: Because it's not meant to be an actual control, just satisfy Angular to not throw "ControlValueAccessor not found" error.

Issue #11826, opened 2025-09-04
🐞 - `InputRange` has specific proprietary styles for readonly/disabled state


waterplea:
### Reproduction url

https://taiga-ui.dev/components/input-range

### Description

In proprietary designs we have lock icon on disabled/readonly states, in that case value must be displayed in left input

<img width="388" height="105" alt="Image" src="https://github.com/user-attachments/assets/0045abf6-64a9-4460-9904-667e155bcf45" />

<br />

<img width="388" height="105" alt="Image" src="https://github.com/user-attachments/assets/1dceea64-e46d-48d9-a374-369d86f7a71d" />

nsbarsukov:
The first point duplicates already existing issue:
* https://github.com/taiga-family/taiga-ui/issues/11759

PR #12061, opened 2025-09-25
feat(kit): `InputSlider` hides slider for non-interactive state

nsbarsukov:
Update according to the latest design specs

Partially solves #11826


waterplea (approved):

waterplea:
We also need it in v4

PR #12079, opened 2025-09-29
fix(kit): `MultiSelectGroup` fix empty view

wtorso:


splincode (approved):

projects/kit/components/multi-select/multi-select-group/multi-select-group.template.html
```diff
@@ -1,12 +1,14 @@
...
-    <button
-        tuiLink
-        type="button"
-        class="t-button"
-        (click)="toggle()"
-    >
-        {{ texts()?.[checked() ? 'none' : 'all'] }}
-    </button>
-</span>
+@if (values().length) {
```
splincode: `ngIf`

PR #12091, opened 2025-09-30
feat(cdk)!: `provideStyles`/`WithStyles` add new directive styles API, drop old one

waterplea:
Fixes #


splincode (approved):

projects/layout/components/card/collapsed.directive.ts
```diff
@@ -2,35 +2,30 @@ import {
...
+        '[style.margin-block-end.rem]': 'tuiCardCollapsed() ? 0.75 : 0',
+        '[style.clip-path]': 'tuiCardCollapsed() ? "inset(-0.75rem)" : "inset(0)"',
     },
 })
 export class TuiCardCollapsed {
-    protected readonly nothing = tuiWithStyles(TuiCardCollapsedStyles);
-
-    @Input('tuiCardCollapsed')
-    public collapsed = false;
+    public readonly tuiCardCollapsed = input(false);
```
vladimirpotekhin: Maybe use aliases like `public readonly collapsed = input(false, {alias: 'tuiCardCollapsed');`  ?
That way, we can continue to use the shorthand `collapsed` in `host`.

vladimirpotekhin (approved):

PR #12093, opened 2025-09-30
chore: `layout` use signal inputs

vladimirpotekhin:
Fixes #

### DIFF
--- a/projects/layout/components/app-bar/app-bar.component.ts
+++ b/projects/layout/components/app-bar/app-bar.component.ts
@@ -5,14 +5,12 @@ import {
     Component,
     type ElementRef,
     inject,
-    Input,
-    type QueryList,
-    ViewChildren,
+    input,
+    viewChildren,
     ViewEncapsulation,
 } from '@angular/core';
 import {MutationObserverService} from '@ng-web-apis/mutation-observer';
 import {ResizeObserverService} from '@ng-web-apis/resize-observer';
-import {EMPTY_QUERY} from '@taiga-ui/cdk/constants';
 import {tuiZonefull} from '@taiga-ui/cdk/observables';
 import {tuiInjectElement} from '@taiga-ui/cdk/utils/dom';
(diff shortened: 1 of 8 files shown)


splincode (approved):

projects/layout/components/app-bar/app-bar.component.ts
```diff
@@ -48,14 +45,13 @@ export class TuiAppBarComponent implements AfterViewInit {
             () =>
                 2 *
                 Math.max(
-                    this.side.first?.nativeElement.clientWidth,
-                    this.side.last?.nativeElement.clientWidth,
+                    this.side().at(0)?.nativeElement.clientWidth ?? 0,
```
waterplea: Are we allowed to use Array.at?

projects/layout/components/card/large.directive.ts
```diff
@@ -24,12 +24,13 @@ class TuiCardLargeStyles {}
...
+        '[attr.data-space]': 'space() || "normal"',
     },
 })
 export class TuiCardLarge {
     protected readonly nothing = tuiWithStyles(TuiCardLargeStyles);
 
-    @Input('tuiCardLarge')
-    public space: '' | 'compact' | 'normal' = 'normal';
+    public readonly space = input<'' | 'compact' | 'normal'>('normal', {
+        alias: 'tuiCardLarge',
```
waterplea: I think it is now harder to read with explicit alias. Maybe we should use actual input name? That's how I handled it when I was removing `protected readonly nothing = tuiWithStyles(TuiCardLargeStyles);` in [my PR](https://github.com/taiga-family/taiga-ui/pull/12091).

vladimirpotekhin: rolled back this file to avoid conflict with your PR

projects/layout/components/app-bar/app-bar.component.ts
vladimirpotekhin: Rewritten with a array[array.length - 1]

PR #12122, opened 2025-10-02
feat(kit): `InputRange` has new appearance for readonly/disabled state

nsbarsukov:
Fixes #11826

PR #12136, opened 2025-10-03
feat(kit): `InputRange` has new appearance for readonly/disabled state

nsbarsukov:
Cherry picked:
* https://github.com/taiga-family/taiga-ui/pull/12122
* https://github.com/taiga-family/taiga-ui/pull/12140

### DIFF
--- a/projects/kit/components/input-range/input-range.component.ts
+++ b/projects/kit/components/input-range/input-range.component.ts
@@ -59,6 +59,7 @@ import {
         new: '', // TODO(v5): remove after deletion of legacy control
         // TODO: use css :host:has(tui-textfield[data-size]) after browser bump
         '[attr.data-size]': 'size()',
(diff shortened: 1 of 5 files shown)


projects/kit/components/input-range/input-range.component.ts
```diff
@@ -205,4 +243,8 @@ export class TuiInputRangeComponent
 
         return [Math.min(start, prevEnd), Math.max(end, prevStart)];
     }
+
+    private isPrimitive(x: PolymorpheusContent): x is PolymorpheusPrimitive {
```
waterplea, line 246: In Polymorpheus we check for primitives this way:
```js
Object(value) === value
```
Maybe it would work here as well?

nsbarsukov: Done in https://github.com/taiga-family/taiga-ui/pull/12136/commits/f22072c23e47bd4447314c82b0dd96aef3ecccab

PR #12141, opened 2025-10-04
fix(kit): ensure cursor is visible on focus in single-row input-chip

### LINKED ISSUE #11889: 🐞 - InputChip - Autoscrolling to the cursor doesn't work in some cases
### Reproduction url

-

### Description

1. https://taiga-ui.dev/components/input-chip/API?rows=1&sandboxWidth=252 (it's important to have this exact sandboxWidth for this case)
2. Type 3 different words: one, two, three
3. Click outside of the area. Now you see 3 chips in the input
4. Click on any chip

Expected: the list is scrolled to the right and the cursor is visible
Actual: nothing happens. The list is scrolled and the cursor is visible only when you start typing

The bug cannot be reproduced, if you add more than 3 chips on the 2nd step.

I think the reason of such behaviour might be the fact, that the cursor is actually beside the "clearAll" icon

### Taiga UI version

4.53.0

### Browser Used

Chrome

### OS Used

macOS

akramcodez:
Fixes #11889

### Solution Summary
I fixed the InputChip autoscroll issue (#11889) where the cursor wasn’t visible when I focused existing chips in single-row inputs. The problem happened because the browser thought the cursor was already visible, and the scroll method ran too early, before Angular fully rendered the placeholder.

### Here’s what I did:
- Added a 10ms delay in scrollTo inside textfield-multi.component.ts.
- Created an onFocus() handler to make sure scrolling happens at the right time.
- Fixed a TypeScript error in input-date-multi.directive.ts by replacing this.scrollTo() with this.textfield.scrollTo().

Now, the cursor is visible immediately when I click on chips. The fix works for both single- and multi-row inputs, and it keeps everything compatible with Taiga UI’s existing architecture.

### DIFF
--- a/projects/core/components/textfield/textfield-multi/textfield-multi.component.ts
+++ b/projects/core/components/textfield/textfield-multi/textfield-multi.component.ts
@@ -177,4 +177,21 @@ export class TuiTextfieldMultiComponent<T> extends TuiTextfieldBaseComponent<T> 
             // Empty catch block - silently ignore showPicker errors
         }
     }
+
+    protected onFocus(): void {
+        setTimeout(() => this.scrollTo(true), 10);
+    }
+
+    public scrollTo(end = false): void {
+        const sign = this.el.matches('[dir="rtl"] :scope') ? -1 : 1;
+        const scrollLeft = end ? sign * Number.MAX_SAFE_INTEGER : 0;
+
+        // Allow change detection to run and placeholder to be rendered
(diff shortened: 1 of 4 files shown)


akramcodez:
PTAL @waterplea, thanks.

waterplea:
I believe this is already fixed in main so I'm closing this PR, thank you for your time and sorry we didn't get to it earlier.

PR #12147, opened 2025-10-06
chore(kit)!: `InputPhoneInternational` graduate from experimental

waterplea:
Fixes #


splincode (approved):

projects/demo/src/modules/components/input-phone-international/examples/5/index.ts
```diff
@@ -5,8 +5,8 @@ import {encapsulation} from '@demo/emulate/encapsulation';
 import {TuiTextfield} from '@taiga-ui/core';
 import {type TuiCountryIsoCode} from '@taiga-ui/i18n';
 import {
-    tuiInputPhoneInternationalOptionsProvider,
     TuiInputPhoneInternational,
+    tuiInputPhoneInternationalOptionsProvider,
```
splincode, line 9: what about `provideInputPhoneIntlOptions`

waterplea: Let's think about it before release, maybe rename component too

PR #12160, opened 2025-10-07
chore(demo): use chevron button instead of icon in table expand demo

hakimio:
Might be more appropriate to use a button for toggling expandable table rows instead of clickable icon.

### DIFF
--- a/projects/demo/src/modules/components/table/examples/9/index.html
+++ b/projects/demo/src/modules/components/table/examples/9/index.html
@@ -53,10 +53,16 @@
                 *tuiCell="'action'"
                 tuiTd
             >
-                <tui-icon
-                    class="t-chevron"
+                <button
+                    appearance="flat-grayscale"
+                    size="xs"
+                    tuiIconButton
+                    type="button"
+                    [style.border-radius]="'100%'"
                     [tuiChevron]="expand.expanded()"
-                />
+                >
+                    Toggle
+                </button>
             </td>
             <td
                 *tuiCell="'firstName'"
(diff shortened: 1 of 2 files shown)


hakimio:
Might be also useful to add another demo showing how to use the new `tui-table-expand` in a normal table with 10-20 rows. I think it's not really obvious.

projects/demo/src/modules/components/table/examples/9/index.html
```diff
@@ -53,10 +53,16 @@
                 *tuiCell="'action'"
                 tuiTd
             >
-                <tui-icon
-                    class="t-chevron"
+                <button
+                    appearance="flat-grayscale"
```
waterplea: Since the whole row is clickable in this example this change is accessibility only, therefore I guess maybe it's better to use appearance="icon" and no border-radius change? Just a thought though, up to you, thank you for the contribution.

hakimio: @waterplea I have changed those rows. 
What do you think about my previous suggestion about having a demo of normal 10-20 row table with expandable rows? Do you want to have a demo like that?

waterplea: Sure, if you feel like a more representative demo can be made go ahead and improve it 👍

hakimio: Ok, I will open a new PR for that. This one is ready for merging.

splincode:
@hakimio Let's not forget to cherry-pick this to main

projects/demo/src/modules/components/table/examples/9/index.html
waterplea: That new one can go directly to main, v5 is coming next month already

hakimio: Glad to hear `v5` is coming so soon.
For now I have created a PR for `v4`: #12285
If that one gets merged, I'll open another one for `v5`.

PR #12187, opened 2025-10-09
feat(cdk): `Portal` add new service

waterplea:
Fixes #

### DIFF
--- a/projects/core/directives/dropdown/dropdowns.component.ts
+++ b/projects/core/directives/dropdown/dropdowns.component.ts
@@ -9,7 +9,7 @@ import {TuiDropdownService} from './dropdown.service';
 @Component({
     standalone: true,
     selector: 'tui-dropdowns',
-    template: '<ng-container #viewContainer />',
+    template: '<ng-container #vcr />',
     changeDetection: ChangeDetectionStrategy.OnPush,
     providers: [tuiAsPortal(TuiDropdownService)],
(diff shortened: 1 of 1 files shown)


projects/cdk/directives/repeat-times/repeat-times.directive.ts
```diff
@@ -20,14 +20,14 @@ export class TuiRepeatTimesContext implements TuiContext<number> {
     selector: '[tuiRepeatTimes][tuiRepeatTimesOf]',
 })
 export class TuiRepeatTimes {
-    private readonly viewContainer = inject(ViewContainerRef);
-    private readonly templateRef = inject(TemplateRef<TuiRepeatTimesContext>);
+    private readonly vcr = inject(ViewContainerRef);
```
nsbarsukov, line 23: Should we expand our eslint rule with `vcr` ? 🤔 

https://github.com/taiga-family/toolkit/blob/86caf023ba6dbdfa00bfeeadba423aacccf60580/projects/eslint-plugin-experience-next/configs/taiga-specific.ts#L38-L57

waterplea: Sounds good, @splincode what do you think?

splincode: Great idea

splincode: @copilot make issue please

PR #12195, opened 2025-10-09
refactor(core)!: `Popups` use single portal container for dropdowns, alerts and push notifications

waterplea:
Fixes #

### DIFF
--- a/projects/core/styles/mixins/mixins.less
+++ b/projects/core/styles/mixins/mixins.less
@@ -171,3 +171,11 @@
 .tui-mask(@x, @y, @r) {
     mask: radial-gradient(circle at @x @y, transparent @r, black @r);
 }
+
+.portal-loop (@i) when (@i > 0) {
+    .portal-loop(@i - 1);
+
+    &:nth-of-type(@{i}) {
+        grid-row: ~'1000@{i}';
+    }
+}
--- a/projects/core/styles/mixins/mixins.scss
+++ b/projects/core/styles/mixins/mixins.scss
@@ -164,3 +164,11 @@
 @mixin tui-mask($x, $y, $r) {
     mask: radial-gradient(circle at $x $y, transparent $r, black $r);
 }
+
+@mixin portal-loop($n) {
+    @for $i from 1 through $n {
+        &:nth-of-type(#{$i}) {
+            grid-row: #{$i + 10000};
+        }
+    }
+}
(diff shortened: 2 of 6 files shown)


projects/core/components/alert/alert.directive.ts
```diff
@@ -1,15 +1,20 @@
...
+        {
+            directive: TuiPortalDirective,
+            inputs: ['options: tuiAlertOptions', 'open: tuiAlert'],
+            outputs: ['openChange: tuiAlertChange'],
+        },
+    ],
 })
-export class TuiAlert<T> extends TuiPopoverDirective<TuiAlertOptions<T>> {}
+export class TuiAlert<T> {
+    public readonly tuiAlertOptions = input<Partial<TuiAlertOptions<T>>>({});
```
waterplea, line 19: Just so it's properly typed.

projects/kit/components/push/push.directive.ts
```diff
@@ -1,39 +1,22 @@
...
+            directive: TuiPortalDirective,
+            inputs: ['open: tuiPush'],
+        },
+    ],
 })
-export class TuiPushDirective extends PolymorpheusTemplate {
-    private readonly push: TuiPushService = inject(forwardRef(() => TuiPushService));
-    private readonly show$ = new Subject<boolean>();
+export class TuiPushDirective {
+    private readonly sub = inject(TuiPortalDirective).openChange.subscribe(() => {});
```
nsbarsukov, line 17: Why not just empty ?
```suggestion
    private readonly sub = inject(TuiPortalDirective).openChange.subscribe();
```

projects/kit/components/push/push.directive.ts
```diff
@@ -1,39 +1,22 @@
...
-                tuiIfMap(() => this.push.open(this)),
-                takeUntilDestroyed(),
-            )
-            .subscribe();
-    }
-
-    @Input()
-    public set tuiPush(show: boolean) {
-        this.show$.next(show);
+        inject(DestroyRef).onDestroy(() => this.sub.unsubscribe());
```
nsbarsukov, line 20: Why not `takeUntilDestroy` ?

projects/kit/components/push/push.directive.ts
waterplea: Stupid Angular output type requires a function there. It is no longer an Observable, it is [OutputRef](https://angular.dev/api/core/OutputRef)

projects/kit/components/push/push.directive.ts
waterplea: Again, stupid OutputRef does not have .pipe()

PR #12209, opened 2025-10-10
feat(core): `Alert` make position configurable

waterplea:
Fixes #


mdlufy:
Don't forget to fix position: "bottom" on iOS

https://github.com/user-attachments/assets/7dce7d26-3a78-43ea-8d0a-03e21c563ae9

PR #12213, opened 2025-10-10
refactor(core)!: `Alerts` drop dedicated portal and related infrastructure

waterplea:
Fixes #

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -9,6 +9,7 @@ env:
   PERF_COLLECTOR_DEBUG: 1
   PERFORMANCE_RUNS: 6 # total runs including warmup
   PERFORMANCE_WARMUP_SKIP: 1 # number of initial runs to discard as warmup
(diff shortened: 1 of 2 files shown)


projects/kit/components/toast/toast.options.ts
```diff
@@ -1,16 +1,20 @@
...
-    autoClose: number;
-    closable: boolean;
-    appearance: string;
-    data?: I;
+    readonly autoClose: number;
+    readonly closable: boolean;
+    readonly appearance: string;
+    readonly data?: I;
+    readonly position: 'bottom' | 'top';
+    readonly orientation: 'center' | 'end' | 'start';
```
nsbarsukov, line 9: Do we want to imitate to css logical properties ?

```suggestion
    readonly blockPosition: 'start' | 'end';
    readonly inlinePosition: 'center' | 'end' | 'start';
```

or even just short version ?

```suggestion
    readonly block: 'bottom' | 'top';
    readonly inline: 'center' | 'end' | 'start';
```

PR #12215, opened 2025-10-11
refactor(core)!: `tui-loader[showLoader]` renamed to `tui-loader[loading]`

medbenmakhlouf:
This pull request refactors the `TuiLoader` Angular component to use the new `input()` and `computed()` APIs instead of the traditional `@Input()` decorator and getter methods. This change modernizes the component, making its state management more reactive and consistent with the latest Angular patterns. Additionally, the template and tests are updated to accommodate these API changes.

### DIFF
--- a/projects/addon-doc/components/example/example.template.html
+++ b/projects/addon-doc/components/example/example.template.html
@@ -59,8 +59,8 @@
                     <tui-loader
                         size="xs"
                         class="t-code-editor"
+                        [loading]="loading()"
                         [overlay]="true"
-                        [showLoader]="loading()"
                         (click)="edit(processor())"
                     >
                         @if (codeEditor?.content; as content) {
--- a/projects/core/components/loader/test/loader.component.spec.ts
+++ b/projects/core/components/loader/test/loader.component.spec.ts
@@ -13,7 +13,7 @@ describe('Loader', () => {
         template: `
             @if (custom) {
                 <tui-loader
-                    [showLoader]="showLoader"
+                    [loading]="showLoader"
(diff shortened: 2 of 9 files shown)


projects/core/components/loader/loader.component.ts
```diff
@@ -14,32 +14,25 @@ import {TUI_LOADER_OPTIONS} from './loader.options';
...
-    // TODO: Drop alias in v5
-    @Input('showLoader')
-    public loading = true;
+    public textContent = input<PolymorpheusContent>();
 
-    protected get isHorizontal(): boolean {
-        return !tuiSizeBigger(this.size);
-    }
+    // TODO: Drop alias in v5
+    public loading = input(true, {alias: 'showLoader'});
```
nsbarsukov: readonly for all inputs

nsbarsukov (approved):

projects/core/components/loader/loader.component.ts
medbenmakhlouf: @nsbarsukov , should I drop the alias since we are targeting these changes in v5 ?

medbenmakhlouf: @nsbarsukov  done for  readonly for all inputs, just waiting for your response about the alias, thanks

nsbarsukov: > @nsbarsukov , should I drop the alias since we are targeting these changes in v5 ?

Yep, you can do it

medbenmakhlouf: @nsbarsukov , done

PR #12217, opened 2025-10-11
refactor(core): `Textfield` uses Signal Inputs 

medbenmakhlouf:
This pull request refactors several textfield-related components and directives to use Angular's new `input()` and `model()` signal-based APIs instead of the traditional `@Input()` decorator. The changes modernize state management, simplify the code, and improve reactivity throughout the textfield components, templates, and related directives.

### DIFF
--- a/projects/kit/components/multi-select/multi-select-native/multi-select-native.component.ts
+++ b/projects/kit/components/multi-select/multi-select-native/multi-select-native.component.ts
@@ -3,6 +3,7 @@ import {ChangeDetectionStrategy, Component, computed, inject, Input} from '@angu
 import {TuiControl} from '@taiga-ui/cdk/classes';
 import {TUI_IS_ANDROID, TUI_IS_IOS} from '@taiga-ui/cdk/tokens';
 import {tuiInjectElement, tuiIsPresent} from '@taiga-ui/cdk/utils';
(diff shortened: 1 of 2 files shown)


projects/kit/components/multi-select/multi-select-native/multi-select-native.component.ts
```diff
@@ -57,7 +57,7 @@ export class TuiMultiSelectNative<T> {
 
     @Input()
     public set placeholder(placeholder: string) {
-        this.textfield.fillerSetter = placeholder;
+        this.textfield.filler.set(placeholder);
```
nsbarsukov: Let use `tuiSetSignal` instead

```
import {tuiSetSignal} from '@taiga-ui/cdk/utils/miscellaneous';
```

Dont forget to update `projects/core/components/textfield/textfield.component.ts`

```diff
- public readonly filler = model('');
+ public readonly filler = input('');
```

nsbarsukov (changes requested):

projects/kit/components/multi-select/multi-select-native/multi-select-native.component.ts
medbenmakhlouf: Thanks for the tip! @nsbarsukov , it should be good now

PR #12226, opened 2025-10-12
refactor(kit): `NativeSelect` uses signal inputs

medbenmakhlouf:
Fixes #

### DIFF
--- a/projects/kit/components/select/native-select/native-select.template.html
+++ b/projects/kit/components/select/native-select/native-select.template.html
@@ -7,15 +7,16 @@
         {{ placeholder() }}
     </option>
 }
+@let elements = items();
 
-@if (items() && !isFlat(items() || [])) {
+@if (elements && !isFlat(elements)) {
     @for (group of labels(); track group) {
         <optgroup [label]="labels()[$index]">
-            <ng-container *ngTemplateOutlet="options; context: {$implicit: (items() || [])[$index]}" />
+            <ng-container *ngTemplateOutlet="options; context: {$implicit: elements[$index]}" />
         </optgroup>
     }
 } @else {
-    <ng-container *ngTemplateOutlet="options; context: {$implicit: items()}" />
+    <ng-container *ngTemplateOutlet="options; context: {$implicit: elements}" />
 }
 
 <ng-template


projects/kit/components/select/native-select/native-select.component.ts
```diff
@@ -64,24 +62,18 @@ export class TuiNativeSelect<T>
         this.textfield.value.set(this.stringified());
     }, TUI_ALLOW_SIGNAL_WRITES);
 
-    @Input()
-    public items: ReadonlyArray<readonly T[]> | readonly T[] | null = [];
+    public items = input<ReadonlyArray<readonly T[]> | readonly T[]>([]);
```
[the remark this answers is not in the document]
nsbarsukov: AI is right. `null` should be added

projects/kit/components/select/native-select/native-select.component.ts
```diff
@@ -64,24 +62,18 @@ export class TuiNativeSelect<T>
...
-    @Input()
-    public labels: readonly string[] = [];
+    public labels = input<readonly string[]>([]);
 
-    // TODO(v5): use signal inputs
-    @Input('placeholder')
-    public set placeholderSetter(x: string) {
-        this.placeholder.set(x);
-    }
+    public placeholder = input('');
```
nsbarsukov: Add `readonly` for signal inputs

projects/kit/components/select/native-select/native-select.template.html
```diff
@@ -8,14 +8,14 @@
...
-            <ng-container *ngTemplateOutlet="options; context: {$implicit: items[$index]}" />
+@if (items() && !isFlat(items() || [])) {
+    @for (group of labels(); track group) {
+        <optgroup [label]="labels()[$index]">
+            <ng-container *ngTemplateOutlet="options; context: {$implicit: (items() || [])[$index]}" />
         </optgroup>
     }
 } @else {
-    <ng-container *ngTemplateOutlet="options; context: {$implicit: items}" />
+    <ng-container *ngTemplateOutlet="options; context: {$implicit: items()}" />
```
nsbarsukov: Is it better to use local variable (created by `@let`) instead of repeating `|| []` fallback ?

```html
@let items = items();

@if (items && !isFlat(items)) {
    @for (group of labels(); track group) {
        <optgroup [label]="labels()[$index]">
            <ng-container *ngTemplateOutlet="options; context: {$implicit: (items)[$index]}" />
        </optgroup>
    }
} @else {
    <ng-container *ngTemplateOutlet="options; context: {$implicit: items}" />
}
```

medbenmakhlouf: @nsbarsukov Yes, you’re right! But I had to rename the variable (I chose elements) because @let items = items() creates a shadow variable, and the template would raise an error.

PR #12230, opened 2025-10-12
chore(kit): implement clear method

JonataB:
Fixes #


splincode:
Thanks for the contribution, I think there is no point in making changes to the tree component for now, perhaps in the future it will be redesigned or an experimental new one will be added

PR #12236, opened 2025-10-13
feat(addon-commerce): `InputCardGroup` add `compact` mode

### LINKED ISSUE #12204: 🐞 - Wrong padding for TuiInputCardGroup size=m
### Reproduction url

stackblitz.com

### Description

For desktop device
My template is `width: 320px` for TuiInputCardGroup and `tuiTextfieldSize="m"`
With custom labels provided with TuiInputCardGroup

but paddings for t-wrapper and t-wrapper_active looks wrong

### Taiga UI version

4.57.0

### Browser Used

Chrome

### OS Used

macOS

waterplea:
Fixes #12204

### DIFF
--- a/projects/demo/src/modules/components/input-card-group/index.html
+++ b/projects/demo/src/modules/components/input-card-group/index.html
@@ -23,6 +23,7 @@
             <ng-template>
                 <tui-input-card-group
                     [codeLength]="codeLength"
+                    [compact]="compact"
                     [formControl]="formControl"
                     [icon]="icon"
                     [id]="id"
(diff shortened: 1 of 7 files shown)


projects/demo/src/modules/components/input-card-group/index.html
```diff
@@ -69,6 +70,14 @@
             >
                 Code length
             </tr>
+            <tr
+                name="[compact]"
+                tuiDocAPIItem
+                type="boolean"
+                [(value)]="compact"
+            >
+                Manually set compact mode (forced on mobile resolution)
```
nsbarsukov, line 79: Why not DI-option ?

```ts
tuiInputCardGroupOptionsProvider({compact: true})
```

Is there any case to change it in runtime ?

waterplea: We need a quick fix for a particular project, I think the whole component should be refactored eventually as it grew in features mostly due to that one project and not always optimally.

PR #12241, opened 2025-10-13
fix(kit): select does not open when clicking near the border in TuiInputPhoneInternational

### LINKED ISSUE #12223: 🐞 - I can't choose a phone code
### Reproduction url

https://taiga-ui.dev/components/input-phone-international#old-syntax

### Description

Try clicking in the top-right/top-left corner

Image

### Taiga UI version

latest

### Browser Used

Chrome

### OS Used

macOS

splincode:
Fixes #12223

### DIFF
--- a/projects/kit/components/input-phone-international/input-phone-international.style.less
+++ b/projects/kit/components/input-phone-international/input-phone-international.style.less
@@ -11,6 +11,10 @@
     &[data-size='s'] {
         inline-size: 4rem;
     }
+
+    &::after {
+        pointer-events: none;
+    }
 }
 
 .t-flag {


projects/kit/components/input-phone-international/input-phone-international.style.less
```diff
@@ -11,6 +11,10 @@
     &[data-size='s'] {
         inline-size: 4rem;
     }
+
+    &::after {
+        pointer-events: none;
```
nsbarsukov, line 16: Should we just extend selector here ?

https://github.com/taiga-family/taiga-ui/blob/6d9963d34b6047fc3c8671a958ef50325cf21d3d/projects/core/components/textfield/textfield.component.ts#L173-L179

waterplea: This is old component, it's going away in v5, let's not overcomplicate the fix

PR #12250, opened 2025-10-13
fix: update arc chart value input handling

### LINKED ISSUE #12232: 🚀 - Hacktoberfest special
### Description

As we are approaching v5 we need to migrate all inputs and outputs to signals so that the public API is ready for the first alpha release. It would be a great help if community can assist us with this. It should be pretty straightforward but there's a lot of components to go through.

robmanganelly:
Fixes #12232 

PARTIAL FIX : https://github.com/taiga-family/taiga-ui/issues/12232

This PR contains the fix for arc-chart component
Subsequent PRs will address other files. I tried changing one component at a time to prevent large batches.

##  Changes in this PR

changed inputs and outputs to use signal api
changed viewChildren to use signal api
cleaned up some redundancies 
added documentation to make explicit to consumers that internal state of the component can change some values
moved signal reads to template variables on top (easier to spot, read once use many)

## Problems

ESLint rules sometimes are forcing incorrect patterns. I skipped lint of the affected line on each case and documented the reason

## Chance for improvement

using native `@for` and template variables + computations could make the component even more reactive. Did not pursue because it was outside the scope of the ticket. 

will be paying attention to comments  and resolution before starting the next component

### DIFF
--- a/projects/addon-charts/components/arc-chart/arc-chart.component.ts
+++ b/projects/addon-charts/components/arc-chart/arc-chart.component.ts
@@ -60,7 +60,6 @@ export class TuiArcChart {
             take(1),
             // The linter rule 'no-restricted-syntax' incorrectly flags 'map(() => true)' here,
             // because the literal 'true' has type 'true' (not 'boolean'), which is intentional for this signal initialization.
-            // eslint-disable-next-line no-restricted-syntax
             map(() => true),
         ),
         {initialValue: false},
--- a/projects/addon-charts/components/arc-chart/arc-chart.template.html
+++ b/projects/addon-charts/components/arc-chart/arc-chart.template.html
@@ -2,31 +2,31 @@
 @let min = minLabel();
 @let max = maxLabel();
 @for (_ of '-'.repeat(data.length); track $index) {
-<svg
-    focusable="false"
-    viewBox="-100 -100 200 200"
-    xmlns="http://www.w3.org/2000/svg"
-    class="t-svg"
-    [style.height.rem]="getDiameter($index)"
-    [style.inset-block-start.rem]="getInset($index)"
-    [style.inset-inline-end.rem]="getInset($index)"
-    [style.inset-inline-start.rem]="getInset($index)"
->
-    <path
-        d="M -70 70 A 100 100 0 1 1 70 70"
-        vector-effect="non-scaling-stroke"
-    />
-    <path
-        #arc
-        d="M -70 70 A 100 100 0 1 1 70 70"
-        vector-effect="non-scaling-stroke"
-        class="t-value"
-        [class.t-value_inactive]="isInactive($index)"
-        [style.stroke]="`var(--tui-chart-categorical-${$index.toString().padStart(2, '0')})`"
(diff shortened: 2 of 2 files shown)


projects/addon-charts/components/arc-chart/arc-chart.component.ts
```diff
@@ -44,75 +38,72 @@ function arcsToIndex(arcs: QueryList<ElementRef<SVGElement>>): Array<Observable<
...
-        .subscribe(() => this.initialized.set(true));
+    private readonly indexChange$ = toObservable(this.arcs).pipe(
+        takeUntilDestroyed(),
+        // keep spread to remove readonly constraint
+        switchMap((arcs) =>
+            merge(...arcsToIndex([...arcs])).pipe(
+                tap((i) => this.activeItemIndex.set(i)),
+            ),
+        ),
+    );
```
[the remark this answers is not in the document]
robmanganelly: replaced spread with type assertion , see fix commits

projects/addon-charts/components/arc-chart/arc-chart.component.ts
```diff
@@ -44,75 +38,72 @@ function arcsToIndex(arcs: QueryList<ElementRef<SVGElement>>): Array<Observable<
...
 
-    @Input()
-    public value: readonly number[] = [];
+    protected readonly initialized = toSignal(
+        timer(0).pipe(
+            tuiZonefree(),
+            take(1),
+            // The linter rule 'no-restricted-syntax' incorrectly flags 'map(() => true)' here,
```
splincode, line 61: you can use `TUI_TRUE_HANDLER`

projects/addon-charts/components/arc-chart/arc-chart.template.html
```diff
@@ -1,5 +1,8 @@
+@let v = value();
```
splincode: or `@let data = value();`

robmanganelly: feedback addressed in 
[b4a8c42](https://github.com/taiga-family/taiga-ui/pull/12250/commits/b4a8c424409e57c6d1f7d6936d0a4145f6b3241d)

projects/addon-charts/components/arc-chart/arc-chart.component.ts
robmanganelly: `TUI_TRUE_HANDLER` type is too narrow for this.
type `true` !== type `boolean`

I left a more direct comment initially but copilot review asked me to replace it with what you see now.
please see attached pictures. 

Screenshot 2025-10-14 at 10 21 34 AM

Screenshot 2025-10-14 at 10 21 52 AM

Screenshot 2025-10-14 at 10 24 09 AM

I could always hack the initial value's type, but that seems dirty, skipping es-lint seemed the preferred solution because it is both technically correct and cleaner.

Please let me know your preference

splincode: okey, no problem

PR #12282, opened 2025-10-16
refactor(cdk): remove `tuiIsApple` and `TuiiOS`

medbenmakhlouf:
remove these functions as they are no longer used in the projects

### DIFF
--- a/projects/cdk/schematics/ng-update/v5/steps/constants/functions.ts
+++ b/projects/cdk/schematics/ng-update/v5/steps/constants/functions.ts
@@ -31,4 +31,40 @@ export const REPLACE_FUNCTIONS: readonly ReplacementType[] = [
         to: 'tuiBlurFocused',
         moduleSpecifier: ['@taiga-ui/cdk'],
     },
+    {
+        from: 'tuiIsApple',
+        moduleSpecifier: ['@taiga-ui/cdk'],
+        to: 'isApple',
+        removeImport: true,
+        newImports: [
+            {
+                name: 'isApple',
+                moduleSpecifier: '@ng-web-apis/platform',
+            },
+        ],
+    },
+    {
+        from: 'tuiIsIos',
+        moduleSpecifier: ['@taiga-ui/cdk'],
+        to: 'isIos',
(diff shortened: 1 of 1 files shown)


projects/cdk/utils/browser/is-ios.ts
```diff
@@ -1,7 +0,0 @@
-import {isIos} from '@ng-web-apis/platform';
-
-/**
- * @deprecated: drop in v5, use import {isApple} from '@ng-web-apis/platform';
- * @param navigator
- */
-export const tuiIsIos = isIos;
```
nsbarsukov, line 7: Expand schematics migrations 
https://github.com/taiga-family/taiga-ui/blob/ee02e2e2412311769ba0be3456f92f000ca98bb7/projects/cdk/schematics/ng-update/v5/steps/constants/functions.ts#L3
with
```ts
{
    from: 'isIos',
    moduleSpecifier: ['@taiga-ui/cdk'],
    to: 'isApple',
    removeImport: true,
    newImports: [
        {
            name: 'isApple',
            moduleSpecifier: '@ng-web-apis/platform',
        },
    ],
},
{
    from: 'tuiIsIos',
    moduleSpecifier: ['@taiga-ui/cdk'],
    to: 'isApple',
    removeImport: true,
    newImports: [
        {
            name: 'isApple',
            moduleSpecifier: '@ng-web-apis/platform',
        },
    ],
},
```

projects/legacy/utils/is-apple.ts
```diff
@@ -1,6 +0,0 @@
-import {isApple} from '@ng-web-apis/platform';
-
-/**
- * @deprecated: drop in v5.0
- */
-export const tuiIsApple = isApple;
```
nsbarsukov, line 6: Add schematics for legacy utility too

projects/cdk/utils/browser/is-ios.ts
medbenmakhlouf: @nsbarsukov , done

splincode (approved):

PR #12301, opened 2025-10-16
feat(core): add `tuiInjectMobileRes` helper

splincode:


projects/core/services/breakpoint.service.ts
```diff
@@ -39,3 +39,9 @@ export class TuiBreakpointService extends Observable<TuiBreakpointMediaKey | nul
         super((subscriber) => this.stream$.subscribe(subscriber));
     }
 }
+
+export function tuiInjectMobileRes(): Observable<boolean> {
```
mdlufy, line 43: `..Res` is an abbreviation, what is full name?

splincode: I was inspired by that name

image

mdlufy: i more like full name ) for example `tuiInjectMobileResourse` / `tuiInjectIsMobileResource`
but current is also okay

Issue #11333, opened 2025-07-16
🐞 - ProgressSegmented – Color segments misaligned with visual segments on container resize


eliottlebot:
### Reproduction url

https://stackblitz.com/edit/angular-sxtbze92?file=src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.template.html

### Description

**Current behavior:**
When using ProgressSegmented with multiple segments, the colored areas do not always align with their corresponding segments. At specific container widths, a color can overflow into the next segment.

**Steps to reproduce:**
Render ProgressSegmented with multiple segments (see on StackBlitz)
Modify the page size

**Expected behavior:**
Each color should remain strictly within its assigned segment, regardless of container width.

**Screenshots:**
Window width : 708px (misaligned)

<img width="1414" height="162" alt="Image" src="https://github.com/user-attachments/assets/c4471af7-399e-4e8f-a0b9-3d5813434db9" />

Window width : 707px (aligned)

<img width="1400" height="140" alt="Image" src="https://github.com/user-attachments/assets/f2b27bc4-c98f-4193-af81-6bf5c8caba01" />

### Taiga UI version

4.36.0 (@taiga-ui/kit)

### Browser Used

Firefox

### OS Used

macOS

Chiorufarewerin:
I will try it

PR #12342, opened 2025-10-21
fix(kit): `ProgressColorSegments` color misaligned with `ProgressSegmented`

Chiorufarewerin:
Fixes https://github.com/taiga-family/taiga-ui/issues/11333

I changed input signal to setter+signal for v4 compatability


splincode:
@Chiorufarewerin 
image

Chiorufarewerin:
Probably in older version of Angular there behavior differs:
1. Added tuiWatch for MutationObserver
2. Inside tests idkw, but overrideProvider doesn't correctly overides MutationObserverService. So I added injecting for MutationObserverServiceMock. It works, but idk, probably it uses only one MutationObserverServiceMock. For this kind of tests it's ok.

PR #12347, opened 2025-10-21
feat(kit): align close button to top for closable preset in tuiCell

splincode:


projects/addon-mobile/styles/common/card-large.less
```diff
@@ -9,4 +9,8 @@
         --t-space: 1rem;
         --t-comp: -0.125rem;
     }
+
+    [tuiAccessories] > [tuiButtonClose] {
```
waterplea, line 13: We plan to drop tuiButtonClose directive in v5, if possible, let's not use it in v4 as well

splincode: What selector should be used in this case?

```
<aside tuiAccessories>
            <button
                tuiButtonClose
                tuiIconButton
                type="button"
            >
                Close
            </button>
        </aside>
```

PR #12335, opened 2025-10-20
refactor(addon-mobile): `components` & `directives` use signal input

medbenmakhlouf:
This PR migrates all components and directives to signal inputs, signal queries, and functional outputs.

PR #12356, opened 2025-10-23
refactor(addon-mobile): uses input signals

splincode:


medbenmakhlouf:
@splincode Thanks for making the changes! I actually implemented a similar update about three days ago including signal queries as well 😄 [12335](https://github.com/taiga-family/taiga-ui/pull/12335) I made sure to put in the effort as requested, and it looks like both versions are quite aligned.

To avoid duplicating work and losing valuable time, would it be possible to set up a shared sheet or tracker for these kinds of changes? That way, we can better coordinate and keep things streamlined.
I'm happy to continue with these refactors — it's been fun and I’ve really enjoyed working on them for a while now!

PR #12384, opened 2025-10-25
refactor(core): `directives` use signals

medbenmakhlouf:
Fixes #


projects/core/directives/dropdown/dropdown-manual.directive.ts
```diff
@@ -7,13 +7,12 @@ import {TuiDropdownDriver} from './dropdown.driver';
...
-    @Input()
-    public tuiDropdownManual: boolean | '' = false;
+    public readonly tuiDropdownManual = input<boolean | ''>(false);
 
-    public ngOnChanges(): void {
-        this.driver.next(!!this.tuiDropdownManual);
-    }
+    protected readonly driverEffect = effect(() =>
+        this.driver.next(!!this.tuiDropdownManual()),
+    );
```
waterplea, line 17: Are we ok switching all the ngOnChanges to effects? There are some timing differences between those, I'm a bit worried this might cause issues somewhere.

projects/core/directives/dropdown/dropdown-open.directive.ts
```diff
@@ -84,14 +86,11 @@ export class TuiDropdownOpen implements OnChanges {
...
-    public tuiDropdownEnabled = true;
+    public readonly tuiDropdownEnabled = input(true);
 
-    @Input()
-    public tuiDropdownOpen: boolean | '' = false;
+    public readonly tuiDropdownOpen = input<boolean | ''>(false);
 
-    @Output()
-    public readonly tuiDropdownOpenChange = new EventEmitter<boolean>();
+    public readonly tuiDropdownOpenChange = output<boolean>();
```
waterplea, line 93: I wonder if these could be somehow refactored to a model

medbenmakhlouf: @waterplea I actually thought about refactoring this to use a model instead of separate input/output bindings, but the main reason I didn’t is that the input type can be **boolean | ''**, while the output is always just **boolean**.

For example:

``` html
<button
    size="m"
    tuiButton
    tuiChevron
    tuiDropdownOpen
    type="button"
    [tuiDropdown]="dropdown"
>
    Columns
</button>
```

The compiler treats the directive as a string, but internally we always handle it as a boolean.

If we switched to a model, we’d need to either handle all output events explicitly to enforce strict booleans, or make it always `[tuiDropdownOpen]="true".`

That would work, but would also mean updating a lot of existing usages.

PR #12386, opened 2025-10-26
feat(kit): add `TUI_PAGINATION_OPTIONS` token to customize button appearance in pagination

### LINKED ISSUE #12258: 🚀 - Pagination appearance
### Description

We need to be able to customize buttons appearances in Pagination component

4ooper:
Fixes #12258

### DIFF
--- a/projects/kit/components/pagination/pagination.component.ts
+++ b/projects/kit/components/pagination/pagination.component.ts
@@ -46,7 +46,7 @@ export class TuiPagination {
 
     protected readonly texts = inject(TUI_PAGINATION_TEXTS);
     protected readonly icons = inject(TUI_SPIN_ICONS);
-    protected readonly paginationOptions = inject(TUI_PAGINATION_OPTIONS);
+    protected readonly options = inject(TUI_PAGINATION_OPTIONS);
 
     @Input()
     public length = 1;
@@ -55,7 +55,7 @@ export class TuiPagination {
     public focusable = true;
 
     @Input()
-    public size: TuiSizeL | TuiSizeS = this.paginationOptions.defaultSize;
+    public size: TuiSizeL | TuiSizeS = this.options.defaultSize;
 
     @Input()
     public readonly disabled = false;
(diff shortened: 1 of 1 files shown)


projects/kit/components/pagination/pagination.template.html
```diff
@@ -1,11 +1,11 @@
 <div class="t-content">
     @if (size !== 's') {
         <button
-            appearance="flat"
             tabIndex="-1"
             tuiIconButton
             type="button"
             class="t-button"
+            [appearance]="getElementMode(-1)"
```
splincode: What do you think, is it advisable to use a pipe or computed signal?

projects/kit/components/pagination/pagination.options.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export interface TuiPaginationOptions {
+    readonly appearance: TuiStringHandler<{
+        isActive: boolean;
+        size: TuiSizeL | TuiSizeS;
+    }>;
+}
+
+export const [TUI_PAGINATION_OPTIONS, tuiPaginationOptionsProvider] =
+    tuiCreateOptions<TuiPaginationOptions>({
```
vladimirpotekhin, line 14: Since we're adding options anyway, let's also add the default size "l" here so we can use it as the default value for the size input.

4ooper:
Hi! Thanks for the review. I'll take a look a bit later.

projects/kit/components/pagination/pagination.options.ts
4ooper: Sure, already added.

projects/kit/components/pagination/pagination.component.ts
```diff
@@ -44,6 +46,7 @@ export class TuiPagination {
 
     protected readonly texts = inject(TUI_PAGINATION_TEXTS);
     protected readonly icons = inject(TUI_SPIN_ICONS);
+    protected readonly paginationOptions = inject(TUI_PAGINATION_OPTIONS);
```
splincode: `protected readonly options = inject(TUI_PAGINATION_OPTIONS);`

splincode (approved):

splincode:
@4ooper Don't forget to cherry-pick to v4.x

PR #12397, opened 2025-10-28
refactor!: `TUI_NUMBER_FORMAT` refactor to signals

vladimirpotekhin:
Fixes #

### DIFF
--- a/projects/demo/src/modules/directives/copy-processor/examples/1/index.ts
+++ b/projects/demo/src/modules/directives/copy-processor/examples/1/index.ts
@@ -23,10 +23,9 @@ export default class Example {
     protected format = inject(TUI_NUMBER_FORMAT);
 
     protected numberProcessor = computed(
-        () => (text: string) =>
-            text
-                .replace(this.format().decimalSeparator, '.')
-                .replaceAll(this.format().thousandSeparator, ''),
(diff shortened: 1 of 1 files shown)


projects/demo/src/modules/directives/copy-processor/examples/1/index.ts
```diff
@@ -21,13 +20,13 @@ export default class Example {
...
-        map(
-            ({decimalSeparator, thousandSeparator}) =>
-                (text: string) =>
-                    text.replace(decimalSeparator, '.').replaceAll(thousandSeparator, ''),
-        ),
+    protected numberProcessor = computed(
+        () => (text: string) =>
+            text
+                .replace(this.format().decimalSeparator, '.')
+                .replaceAll(this.format().thousandSeparator, ''),
```
nsbarsukov: Will it recompute `numberProcessor` if `this.format()` changed but `(text: string) => {...}` is never called **yet**?
Should we read `this.format()` in the outer function (`computed` callback)?

```suggestion
        ({decimalSeparator, thousandSeparator} = this.format()) => (text: string) =>
            text
                .replace(decimalSeparator, '.')
                .replaceAll(thousandSeparator, ''),
```

PR #12442, opened 2025-10-31
fix(kit): broken routable dialogs

nsbarsukov:
### Reproduction

* https://taiga-ui.dev/dialog/routable/path/to/dialog => NO error ✅
* https://taiga-ui.dev/next/dialog/routable/path/to/dialog => console.error ❌

```
core.mjs:6720 ERROR
TuiNoHostException: Portals cannot be used without TuiPortals component;
perhaps you forgot to wrap your application with tui-root.
    at TuiPopupService.add (service.ts:27:19)
    at TuiDialogService.add (modal.service.ts:19:34)
    at Observable._subscribe (portal.ts:45:18)
    at Observable._trySubscribe (Observable.js:37:25)
    at Observable.js:31:30
    at errorContext (errorContext.js:19:9)
    at Observable.subscribe (Observable.js:22:21)
    at source.subscribe.isComplete (switchMap.js:14:51)
    at OperatorSubscriber._next (OperatorSubscriber.js:13:21)
    at OperatorSubscriber.next (Subscriber.js:31:18)
```

### Why?
https://github.com/taiga-family/taiga-ui/blob/b61d3d08d5235e86f84029f802749b69e384dd64/projects/core/components/root/root.template.html#L4-L12

### DIFF
--- a/projects/kit/components/routable-dialog/test/routable-dialog.component.spec.ts
+++ b/projects/kit/components/routable-dialog/test/routable-dialog.component.spec.ts
@@ -41,10 +41,10 @@ describe('TuiRoutableDialog', () => {
     let tuiDialogService: TuiDialogService;
     let router: Router;
 
-    function createComponent(
+    async function createComponent(
         activatedRoute?: Partial<ActivatedRoute>,
         closeDialogImmediately = true,
-    ): void {
+    ): Promise<void> {
         tuiDialogService = mock(TuiDialogService);
         router = mock(Router);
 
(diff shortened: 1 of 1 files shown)


splincode:
why `chore`?

waterplea (approved):

waterplea:
One concern - this would prevent SSR dialog routes to open with dialog already there.

nsbarsukov:
> One concern - this would prevent SSR dialog routes to open with dialog already there.

@waterplea 
You are right, `afterNextRender` will never fire during SSR and dialog will be opened only during CSR.
I can replace `afterNextRender` by `delay(0)`. Is it better ?

However, replacing it by `delay(0)` still produces css animation of appearance.
So, there is no difference for user if dialog is opened during SSR or client hydration. See these 2 videos:



https://github.com/user-attachments/assets/0467111a-c07b-40dc-bc93-9a1e38ee4422



https://github.com/user-attachments/assets/206b8968-4622-41d5-b53f-c39698deb198

nsbarsukov:
I've forgot about SEO.
Probably, `delay(0)` is better for it.

PR #12447, opened 2025-11-03
feat(kit): add TUI_PAGINATION_OPTIONS token to customize button appearance in pagination

### LINKED ISSUE #12258: 🚀 - Pagination appearance

4ooper:
This is cherry-pick of commit from main branch.
Fixes #12258

### DIFF
--- a/projects/kit/components/pagination/index.ts
+++ b/projects/kit/components/pagination/index.ts
@@ -1,1 +1,2 @@
 export * from './pagination.component';
+export * from './pagination.options';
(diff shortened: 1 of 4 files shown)


waterplea:
Let's not cherry pick this, in v5 the small dot size is going away (in favor of a separate `tui-pager` component) so there's no need to send size to appearance so there's no need to send object to appearance. I know your PR in main was already merged but we will change that API before v5 release and I don't want to introduce new feature to v4 with signature that is not 1 to 1 for v5 to avoid extra migration work.

PR #12464, opened 2025-11-06
refactor(core): drop dependency on `Header`

waterplea:
Fixes #

### DIFF
--- a/projects/core/components/dialog/appearances/default.less
+++ b/projects/core/components/dialog/appearances/default.less
@@ -15,9 +15,15 @@
         inline-size: 25rem;
         padding: 1.5rem;
 
-        > header:not(:last-child),
-        > ng-component > header:not(:last-child) {
-            margin-block-end: 1.25rem;
+        > header,
+        > [tuiSlides] > * > header,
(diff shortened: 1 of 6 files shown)


projects/core/components/dialog/appearances/default.less
```diff
@@ -26,19 +32,26 @@
...
 
-    &[data-size='m'] {
+    &[data-size='m'],
+    &[data-size='l'] {
         inline-size: 37.5rem;
+
+        > header,
+        > [tuiSlides] > * > header,
+        > ng-component > header,
+        > ng-component > [tuiSlides] > * > header {
```
mdlufy, line 42: duplicates with 

 https://github.com/taiga-family/taiga-ui/pull/12464/files#diff-c36ed4136dcbd594591862a9397b28dec31e7ed1b8e3d1f4eb6a5ee0bc2cbe4cR18-R21

waterplea: It sets different font size and different bottom margin

splincode (approved):

PR #12467, opened 2025-11-06
feat(cdk): add `tuiTruncate`

splincode:
Part of https://github.com/taiga-family/taiga-ui/issues/12832

UIKIT-7821

image

https://github.com/user-attachments/assets/fd25be33-4fa4-486d-966e-ca84a86a5c48

### DIFF
--- a/projects/cdk/directives/truncate/truncate.directive.ts
+++ b/projects/cdk/directives/truncate/truncate.directive.ts
@@ -1,0 +1,79 @@
+import {
+    ChangeDetectionStrategy,
+    Component,
+    computed,
+    Directive,
+    inject,
+    input,
+    ViewEncapsulation,
+} from '@angular/core';
+import {toSignal} from '@angular/core/rxjs-interop';
+import {
+    WA_MUTATION_OBSERVER_INIT,
+    WaMutationObserverService,
+} from '@ng-web-apis/mutation-observer';
+import {WaResizeObserverService} from '@ng-web-apis/resize-observer';
+import {tuiInjectElement} from '@taiga-ui/cdk/utils/dom';
+import {tuiWithStyles} from '@taiga-ui/cdk/utils/miscellaneous';
+import {map} from 'rxjs';
+
+import {TuiTruncateService} from './truncate.service';
+
+@Component({
+    template: '',
+    styleUrl: './truncate.styles.less',
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+class Styles {}
+
+@Directive({
+    selector: '[tuiTruncate]',
+    providers: [
+        TuiTruncateService,
+        WaResizeObserverService,
+        WaMutationObserverService,
+        {
+            provide: WA_MUTATION_OBSERVER_INIT,
+            useValue: {
+                attributes: true,
+                characterData: true,
+                subtree: true,
+            },
+        },
+    ],
+    host: {
+        tuiTruncate: '',
+        '[attr.data-lines]': 'lines()',
+        '[attr.data-text]': 'truncated()',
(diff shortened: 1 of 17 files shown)


projects/cdk/directives/truncate/truncate.directive.ts
```diff
@@ -0,0 +1,78 @@
...
+    );
+
+    private readonly text = toSignal(
+        inject(WaMutationObserverService).pipe(map(() => this.el.innerText.trim())),
+        {initialValue: this.el.innerText.trim()},
+    );
+
+    protected readonly nothing = tuiWithStyles(Styles);
+    protected readonly $ = effect(() => {
+        this.el.setAttribute('data-text', this.truncated());
```
waterplea: Why not binding?

projects/cdk/directives/truncate/truncate.directive.ts
```diff
@@ -0,0 +1,78 @@
...
+            useValue: {
+                attributes: true,
+                characterData: true,
+                subtree: true,
+            },
+        },
+    ],
+    host: {
+        tuiTruncate: '',
+        '[attr.data-lines]': 'lines()',
```
waterplea, line 47: I think it's no longer used

splincode: I use it here, otherwise sometimes the lines may jump

image

projects/cdk/directives/truncate/truncate.directive.ts
splincode: I will check

splincode: fixed

projects/cdk/directives/truncate/truncate.directive.ts
```diff
@@ -0,0 +1,80 @@
...
+            useValue: {
+                attributes: true,
+                characterData: true,
+                subtree: true,
+            },
+        },
+    ],
+    host: {
+        tuiTruncate: '',
+        '[attr.title]': 'text()',
```
waterplea: Let's move this to the example instead, so that people can decide if they want title, or hint or any other way of showing full content ONLY WHEN it is not equal to the original text.

splincode: fixed

projects/cdk/directives/truncate/truncate.directive.ts
```diff
@@ -0,0 +1,79 @@
...
+    protected readonly text = toSignal(
+        inject(WaMutationObserverService).pipe(map(() => this.innerText)),
+        {initialValue: this.innerText},
+    );
+
+    protected readonly nothing = tuiWithStyles(Styles);
+
+    public readonly lines = input(1, {
+        alias: 'tuiTruncate',
+        transform: (value: unknown) => Math.max(Number(value) || 1, 1),
```
nsbarsukov: `unknown` => `boolean | number` ?

projects/demo/src/pages/directives/truncate/index.less
```diff
@@ -0,0 +1,5 @@
+.demo {
+    background: #87ceeb;
+    overflow: hidden;
+    resize: horizontal;
```
nsbarsukov, line 4: Is it really required to duplicate resize handler ?
Could we just use default one ?

Image

nsbarsukov (approved):

projects/demo/src/pages/directives/truncate/index.less
splincode: fixed

projects/cdk/directives/truncate/truncate.directive.ts
splincode: fixed

PR #12514, opened 2025-11-11
fix(kit): `InputNumber` with long `[prefix]`/`[postfix]` has invalid input length restriction

### LINKED ISSUE #12337: 🐞 - `InputNumber` with long `[postfix]` causes input length restriction
### Reproduction url

https://taiga-ui.dev/components/input-number/API?postfix=The%20suffix%20long

### Description

Image
Currently, I’m encountering an issue when using InputNumber with [postfix]. When the [postfix] contains multiple characters, the input field becomes limited — I can only enter numbers up to a certain length. For example, as shown in the image, I can only input numbers with up to 7 digits. If the [postfix] is longer, I might not be able to enter any numbers at all.

In addition, there’s no spacing between the [postfix] and the input value. I first noticed this issue in version 4.56.0 and upgraded to version 4.59.0 in hopes of resolving it, but the problem still persists."

### Taiga UI version

4.56.0

### Browser Used

Chrome

### OS Used

macOS

nsbarsukov:
Fixes #12337

### DIFF
--- a/eslint.config.ts
+++ b/eslint.config.ts
@@ -41,4 +41,10 @@ export default process.env.HTML_ESLINT
                   '@angular-eslint/prefer-signals': 'off',
               },
           },
+          {
+              files: ['**/*.spec.ts', '**/*.cy.ts'],
+              rules: {
+                  'no-irregular-whitespace': 'off',
+              },
+          },
       ];


projects/demo-playwright/tests/kit/input-number/input-number.pw.spec.ts
```diff
@@ -743,6 +745,15 @@ describe('InputNumber', () => {
 
                         await expect(inputNumber.textfield).toHaveValue('');
                     });
+
+                    test('allows to enter huge number', async () => {
+                        await inputNumber.textfield.focus();
+                        await inputNumber.textfield.pressSequentially('1234567890123456');
+
+                        await expect(inputNumber.textfield).toHaveValue(
+                            `${prefix}1 234 567 890 123 456${postfix}`,
```
nsbarsukov, line 754: Without new fix it fails with the following error

Image

PR #12526, opened 2025-11-11
fix(kit): impossible to select the first option in native `Select` without placeholder

### LINKED ISSUE #12450: 🐞 - Unable to select first option in native select without placeholder
### Reproduction url

https://stackblitz.com/edit/angular-sdd8tufj

### Description

If you do not have a value and no placeholder in native select on mobile - the first item looks already selected making it impossible to select it.

### Taiga UI version

4

### Browser Used

Chrome

### OS Used

macOS

nsbarsukov:
Fixes #12450

### DIFF
--- a/projects/kit/components/select/native-select/native-select.component.ts
+++ b/projects/kit/components/select/native-select/native-select.component.ts
@@ -69,7 +69,7 @@ export class TuiNativeSelect<T>
 
     protected selectOption(index: number): void {
         const items = (this.items()?.flat() ?? []) as T[];
-        const placeholderOffset = this.value() ? 0 : 1;
+        const placeholderOffset = this.stringified() ? 0 : 1;
 
         this.onChange(items[index - placeholderOffset] ?? null);
     }
--- a/projects/kit/components/select/native-select/native-select.template.html
+++ b/projects/kit/components/select/native-select/native-select.template.html
@@ -1,4 +1,4 @@
-@if (!value()) {
+@if (!stringified()) {
     <option
         disabled
         selected


projects/demo-cypress/src/tests/select/native-select.cy.ts
```diff
@@ -0,0 +1,105 @@
...
+        beforeEach(() => {
+            cy.mount(Sandbox, {
+                componentProperties: {
+                    valueChanges: createOutputSpy('valueChanges'),
+                },
+            });
+        });
+
+        it('initial empty textfield', () => {
+            cy.get('select').should('have.value', null);
```
nsbarsukov, line 80: Without fix it fails with:
```
AssertionError: 
Timed out retrying after 10000ms: 
expected 
'<select#tui_01762864566291._empty.ng-untouched.ng-pristine.ng-valid>' 
to have value null, but the value was 'First Item'
```

projects/demo-cypress/src/tests/select/native-select.cy.ts
```diff
@@ -0,0 +1,105 @@
...
+
+            cy.compareSnapshot('initial-empty-textfield');
+        });
+
+        it('possible to select 1st option', () => {
+            cy.get('select').select('First Item').should('have.value', 'First Item');
+            cy.get('@valueChanges').should('have.been.calledOnceWith', {
+                name: 'First Item',
+                id: 1,
+            });
```
nsbarsukov, line 91: Without fix it fails with:

```
AssertionError: 
Timed out retrying after 10000ms: 
expected valueChanges to have been called exactly once with arguments 
{name: First Item, id: 1}, 
but it was never called
```

projects/demo-cypress/src/tests/select/native-select.cy.ts
```diff
@@ -0,0 +1,105 @@
...
+        {name: 'First Item', id: 1},
+        {name: '2nd Item', id: 2},
+        {name: 'DisabledItem', id: 3},
+        {name: '4th', id: 4},
+    ];
+
+    public readonly placeholder = input('');
+
+    @Output()
+    public readonly valueChanges = new EventEmitter<Item | null>();
```
splincode, line 56: output signal?

splincode (approved):

projects/demo-cypress/src/tests/select/native-select.cy.ts
nsbarsukov: Does not work now 😢 

* https://github.com/cypress-io/cypress/issues/32137#issuecomment-3207460304

I'll add TODO comment

nsbarsukov: https://github.com/taiga-family/taiga-ui/pull/12526/commits/c8763d8ce1ca79aa3c5736323b0cf56127606ef2

projects/kit/components/select/native-select/native-select.component.ts
```diff
@@ -73,7 +69,8 @@ export class TuiNativeSelect<T>
 
     protected selectOption(index: number): void {
         const items = (this.items()?.flat() ?? []) as T[];
+        const placeholderOffset = this.value() ? 0 : 1;
```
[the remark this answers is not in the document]
splincode: @nsbarsukov fyi

projects/kit/components/select/native-select/native-select.template.html
```diff
@@ -1,4 +1,4 @@
-@if (showPlaceholder()) {
+@if (!value()) {
```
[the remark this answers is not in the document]
nsbarsukov: Good catch, it should be replaced with `stringified()`

PR #12529, opened 2025-11-11
refactor(kit): `directives` use signals [final]

medbenmakhlouf:
Fixes #


medbenmakhlouf:
@waterplea just did a rebase and it should be good now

splincode (approved):

Issue #11163, opened 2025-06-25
🚀 - `ButtonSelect` component


waterplea:
### Description

Right now we have `ButtonSelect` component which is unfinished so it's not documented properly. We need to finish it:

1. It should have native validator host directive
2. It should provide select option component with a checkmark
3. It should work with calendar and calendar-range in a dropdown
4. Add documentation page

waterplea:
With new [DatePicker](https://taiga-ui.dev/components/date-picker#accessibility) component it is now complete.

PR #12530, opened 2025-11-11
feat(kit): `ButtonSelect` add ability to work with array

vladimirpotekhin:
part of #11163

### DIFF
--- a/projects/demo/src/modules/components/button-select/examples/2/index.ts
+++ b/projects/demo/src/modules/components/button-select/examples/2/index.ts
@@ -33,5 +33,5 @@ export default class Example {
         {id: 777, name: 'Franklin Clinton'},
     ];
 
-    protected item = this.items;
+    protected value = this.items;
 }
--- a/projects/demo/src/modules/components/button-select/examples/import/import.md
+++ b/projects/demo/src/modules/components/button-select/examples/import/import.md
@@ -1,13 +1,17 @@
 ```ts
 import {Component} from '@angular/core';
 import {TuiButtonSelect} from '@taiga-ui/kit';
+import {TuiButton, TuiDropdown} from '@taiga-ui/core';
+
 // ...
 
 @Component({
   standalone: true,
   imports: [
     // ...
+    TuiButton,
     TuiButtonSelect,
+    TuiDropdown,
   ],
 })
 export class Example {}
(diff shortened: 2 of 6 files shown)


projects/demo/src/modules/components/button-select/examples/2/index.ts
```diff
@@ -0,0 +1,37 @@
...
+    protected readonly items: User[] = [
+        {id: 42, name: 'Tommy Vercetti'},
+        {id: 237, name: 'Carl Johnson'},
+        {id: 666, name: 'Niko Bellic'},
+        {id: 999, name: 'Trevor Philips'},
+        {id: 123, name: 'Michael De Santa'},
+        {id: 777, name: 'Franklin Clinton'},
+    ];
+
+    protected item = this.items;
```
nsbarsukov: Is it better to call it as `value` ?


```suggestion
    protected value = this.items;
```

projects/demo/src/modules/components/button-select/examples/import/import.md
```diff
@@ -0,0 +1,14 @@
...
+import {Component} from '@angular/core';
+import {TuiButtonSelect} from '@taiga-ui/kit';
+// ...
+
+@Component({
+  standalone: true,
+  imports: [
+    // ...
+    TuiButtonSelect,
+  ],
```
nsbarsukov, line 15: Does all use-cases (including upcoming `Calendar` and `CalendarRange` cases) with `TuiButtonSelect`  also require `TuiDropdown` & `TuiButton` imports ?
Should we also mention them in `Setup` page ?

projects/demo/src/modules/app/pages.ts
```diff
@@ -267,6 +267,12 @@ export const pages: DocRoutePages = [
                     name: 'tui-button-group-card',
                 },
             },
+            {
+                section: 'Components',
+                title: 'ButtonSelect',
+                keywords: 'кнопка, button, select, multiselect',
+                route: DemoRoute.ButtonSelect,
```
nsbarsukov, line 274: Add it to this "See also" group too.

https://github.com/taiga-family/taiga-ui/blob/d7a78e370ac999a6019a1294a88c230af6ca2d43/projects/demo/src/modules/app/app.const.ts#L21

So, user will have two opportunities to find new component:
* `Buttons` sidebar navigation group
* `See also` group for select-like entities

PR #12539, opened 2025-11-12
refactor: move `TuiPreviewDialogDirective` to `TuiPreview`

splincode:


projects/kit/components/preview/preview.ts
```diff
@@ -1,11 +1,13 @@
 import {TuiPreviewAction} from './action/preview-action.directive';
+import {TuiPreviewDialogDirective} from './dialog/preview-dialog.directive';
```
nsbarsukov, line 2: Delete `projects/kit/components/preview/dialog/ng-package.json` too

nsbarsukov (approved):

projects/kit/components/preview/preview.ts
splincode: fixed

vladimirpotekhin (approved):

PR #12544, opened 2025-11-12
fix(layout): `CardLarge` is not compatible with `Header` (font has CSS specificity problem)

nsbarsukov:
Now `[tuiCardLarge][data-space]` has the same specificity as `[tuiHeader][data-size=h5]`.
And order of style loading decides.

**Webpack:**
next

**ESBuild:**
esbuild


splincode (approved):

splincode:
@nsbarsukov Do you need a cherry pick to v4.x?

PR #12556, opened 2025-11-13
refactor!: `Multiselect`, `InputTag`, `Tag`, `Textarea` delete legacy components

vladimirpotekhin:


splincode (approved):

projects/demo/src/modules/components/data-list/examples/4/custom-list/index.html
```diff
@@ -39,17 +43,20 @@
...
-                            tuiOption
-                            type="button"
-                            [value]="item"
-                        >
-                            {{ item }}
-                        </button>
-                    }
-                </tui-opt-group>
+                <div tuiMultiSelectGroup>
+                    <tui-opt-group>
```
nsbarsukov, line 47: Why we cannot use `tui-opt-group[tuiMultiSelectGroup]` ?

nsbarsukov (approved):

projects/demo/src/modules/components/data-list/examples/4/custom-list/index.html
vladimirpotekhin: Separate tag is needed to avoid redundancy of "select all" buttons

PR #12557, opened 2025-11-13
chore(demo): add link for line clamp example

splincode:


waterplea:
We don't need this because hovering the link would show the hint

PR #12558, opened 2025-11-13
fix(addon-table): remove generic type constraint from `TuiTableSortPipe`

Markiewic:
This PR updates the `TuiTableSortPipe` by removing the generic type parameter from the pipe class and applying it instead to the transform method.

### Why
Angular Language Service cannot infer a generic type placed on a pipe class.
Because the class-level generic cannot be resolved by the template type-checker, ALS falls back to `any` for both input and output types of the pipe.
This leads to incorrect or missing type checking inside Angular templates.

This issue is described in detail here: https://github.com/angular/angular/pull/40870#issuecomment-780882880

### What Changed

Removed the generic parameter from the pipe class declaration. As a result, ALS correctly infers the pipe's input/output types, and template type checking works as expected.

### DIFF
--- a/projects/addon-table/components/table/pipes/table-sort.pipe.ts
+++ b/projects/addon-table/components/table/pipes/table-sort.pipe.ts
@@ -9,15 +9,15 @@ import {type TuiSortDirection} from '../table.options';
     name: 'tuiTableSort',
     pure: false,
 })
-export class TuiTableSortPipe<K> implements PipeTransform {
-    private readonly table = inject(TuiTableDirective<K>);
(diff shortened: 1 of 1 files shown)


waterplea (approved):

splincode:
@Markiewic Please don't forget to cherry-pick this to the v4.x branch

Markiewic:
> @Markiewic Please don't forget to cherry-pick this to the v4.x branch

@splincode does this require making new same merge request to v4 branch?

splincode:
@Markiewic yep

PR #12562, opened 2025-11-14
fix(layout): support html in hint aside

splincode:
Fixes https://github.com/taiga-family/taiga-ui/issues/12564

## Before

image

## After

image


projects/layout/components/navigation/hint-aside.directive.ts
```diff
@@ -23,7 +25,7 @@ export class TuiHintAsideDirective {
         computed(() =>
             this.aside.expanded() || this.dropdown
                 ? ''
-                : () => this.el.textContent?.trim(),
+                : () => this.domSanitizer.bypassSecurityTrustHtml(this.el.innerHTML),
```
waterplea: I'm not sure I like this. Maybe we should make it possible to pass a template instead like we do in HintOverflow?

splincode: fixed

PR #12596, opened 2025-11-18
feat(core): beauty ios design for buttons

splincode:


projects/core/styles/components/button.less
```diff
@@ -35,6 +35,9 @@
     font: var(--tui-font-text-m);
     font-weight: bold;
 
+    // stylelint-disable-next-line
+    corner-shape: superellipse(1.333);
```
waterplea, line 39: This would cause inconsistencies in cards and cells. If we want to introduce this thing it has to be everywhere and calculated properly:

Image

Image

PR #12598, opened 2025-11-18
fix(core): set pointer-events none for disabled elements

### LINKED ISSUE #12567: 🐞 - Making a first click on a disabled checkbox closes a dropdown
### Reproduction url

https://stackblitz.com/edit/angular-jta4rncu-teqgsiml?file=src%2Fapp%2Fapp.template.html

### Description

If the first click you do in the dropdown is on a disabled checkbox then dropdown will close itself (maybe it counts as an outside click???).
However clicking on the inside of the dropdown and then on the checkbox does not close the dropdown.
Probably the same could happen with other disabled controls

https://github.com/user-attachments/assets/fcd0759e-366f-46ce-9a6d-85ad540788a4

### Taiga UI version

4.61.0

### Browser Used

Chrome

### OS Used

Windows

splincode:
Fixes #12567

### DIFF
--- a/projects/core/styles/components/appearance.less
+++ b/projects/core/styles/components/appearance.less
@@ -40,6 +40,7 @@
 
     .appearance-disabled({
         cursor: initial;
+        pointer-events: none;
         opacity: var(--tui-disabled-opacity);
     });
 }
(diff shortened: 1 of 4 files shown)


waterplea:
This would be a breaking change, since now all browsers support hover effects on disabled items we specifically removed this rule some time ago so people can use hints to show why, say, a button is disabled.

Issue #11917, opened 2025-09-12
🗂 - v5 migrations


waterplea:
- [x] `tuiIsNativeFocused` => `tuiIsFocused`  & `tuiIsNativeFocusedIn` => `tuiIsFocusedIn` 
- [x] Accordion and Expand migrate for API from experimental package
- [x] tuiCreateToken, tuiCreateTokenFromFactory -> InjectionToken @mdlufy 
https://github.com/taiga-family/taiga-ui/pull/11229
- [x] `<tui-icon tuiBadge icon="x" />` -> `<tui-icon tuiBadge iconStart="x" />`
- [x] `<tui-avatar src="x" />` -> `<span tuiAvatar="x"></span>`
- [x] `<a tuiAvatar src="x"></a>` -> `<a tuiAvatar="x"></a>`
- [x] `<a [tuiAvatar]="x | tuiFallbackSrc: '@tui.user' | async"></a>` -> `<a tuiAvatar="@tui.user"><img [src]="x" /></a>`
- [x] Drop `TuiIconBadge` import
- [x] `TuiPdfViewer` change import from `kit` to `legacy`
- [x] `provideEventPlugins()` -> `provideTaiga()` (and in `ng add taiga-ui`)
- [x] NG_EVENT_PLUGINS => provideTaiga
https://github.com/taiga-family/taiga-ui/pull/13824
- [x] Proprietary textfields use small icons in L size, we need to investigate how to cover that
- [x] `.tui-slider-ticks-labels(l)` => `.tui-slider-ticks-abels()` @mdlufy 
https://github.com/taiga-family/taiga-ui/pull/13319
- [x] tuiHeader deprecated sizes
- [x] tui-pin -> tuiPin
- [x] *tuiTextfieldDropdown -> *tuiDropdown
- [x] TuiSidebar => TuiDrawer: `<some-component *tuiSidebar="open()"></some-component>` => `<tui-drawer *tuiPopup="open()"> <some-component /> </tui-drawer>` @mdlufy 
https://github.com/taiga-family/taiga-ui/pull/13471
- [x] TuiInputYearModule => TuiInputYear:  `<tui-input-year> <input tuiTexfieldLegacy /> </tui-input-year>` => `<tui-textfield> <input tuiInputYear /> </tui-textfield>` 
https://github.com/taiga-family/taiga-ui/pull/12073
[...]

PR #12602, opened 2025-11-19
refactor!: drop the `s` size from `tui-pagination` in favor of using `tui-pager`

splincode:
Part of #11917

### DIFF
--- a/projects/demo/src/modules/components/carousel/examples/1/index.less
+++ b/projects/demo/src/modules/components/carousel/examples/1/index.less
@@ -2,8 +2,8 @@
     display: block;
 }
 
-.pagination {
-    margin: 1rem auto 0;
+tui-pager {
+    margin: 0 auto;
 }
 
 .item {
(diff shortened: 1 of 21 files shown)


projects/demo/src/modules/components/carousel/examples/1/index.less
```diff
@@ -2,8 +2,9 @@
     display: block;
 }
 
-.pagination {
+tui-pager[data-size] {
```
waterplea: Do we really need this? Maybe we can just not override the size?

projects/demo/src/modules/components/carousel/examples/4/index.less
```diff
@@ -22,3 +22,8 @@
     background: var(--tui-background-base) center;
     background-size: cover;
 }
+
+tui-pager[data-size] {
```
waterplea: Same here

projects/demo/src/modules/components/carousel/examples/5/index.less
```diff
@@ -8,3 +8,8 @@
 .title {
     margin: 0 0 1rem;
 }
+
+tui-pager[data-size] {
```
waterplea: And here

PR #12607, opened 2025-11-19
refactor!: drop `FormatDate` service and pipe in favour of angular `DatePipe`

vladimirpotekhin:
Fixes #


splincode (approved):

projects/core/services/format-date.service.ts
```diff
@@ -1,14 +1,13 @@
...
+import {inject, Injectable, LOCALE_ID, type Signal, signal} from '@angular/core';
 
 @Injectable({
     providedIn: 'root',
 })
 export class TuiFormatDateService {
     protected readonly locale = inject(LOCALE_ID);
 
-    public format(timestamp: number): Observable<string> {
-        return of(
```
waterplea, line 11: This signal is static, no point in returning a signal from the service method

vladimirpotekhin: This is necessary in order to be able to override the method so that it is dynamic. See [example](https://github.com/taiga-family/taiga-ui/pull/12607/files#diff-fadd4ccdd4a59ea0534a3935326a5e7e9da3e7aff9107f8f4854f666b0a7efd1L15)

PR #12608, opened 2025-11-19
chore: `extendedDiagnostics` uses `error` for prod build & `warning` – for dev build

nsbarsukov:


projects/demo/tsconfig.app.json
```diff
@@ -4,7 +4,13 @@
...
     "angularCompilerOptions": {
         "strictMetadataEmit": false,
-        "compilationMode": "full"
+        "compilationMode": "full",
+        "extendedDiagnostics": {
+            "checks": {
+                "unusedStandaloneImports": "warning"
+            },
+            "defaultCategory": "error"
+        }
```
splincode, line 10: what if?

```
"extendedDiagnostics": {
    "defaultCategory": "warning"
}
```

tsconfig.build.json
```diff
@@ -39,9 +39,7 @@
     },
     "angularCompilerOptions": {
```
splincode: Do we need `angularCompilerOptions`? We had all of them by default:

https://github.com/taiga-family/toolkit/blob/main/projects/tsconfig/tsconfig.json#L24

projects/demo/tsconfig.app.prod.json
```diff
@@ -0,0 +1,10 @@
+{
+    "extends": "./tsconfig.app.json",
+    "angularCompilerOptions": {
+        "extendedDiagnostics": {
```
splincode, line 4:
```
"extendedDiagnostics": {
    "defaultCategory": "error"
}
```

projects/demo/tsconfig.app.json
nsbarsukov: Should we delete all mentioned checks and left only `defaultCategory` ?

https://github.com/taiga-family/toolkit/blob/a7495b75cb85c163ca52f3e46f19f321c82b3c3d/projects/tsconfig/tsconfig.json#L26-L39

Or your propose changes will not work

nsbarsukov: My mistake, it works, I've checked it now

splincode (approved):

PR #12642, opened 2025-11-24
refactor!: add `TUI_BREAKPOINT` token, drop `TuiBreakpointService` service

vladimirpotekhin:
…

Fixes #

### DIFF
--- a/projects/addon-commerce/components/input-card-group/input-card-group.directive.ts
+++ b/projects/addon-commerce/components/input-card-group/input-card-group.directive.ts
@@ -1,17 +1,12 @@
-import {computed, Directive, inject, signal} from '@angular/core';
+import {computed, Directive, inject, input} from '@angular/core';
 import {TUI_BREAKPOINT} from '@taiga-ui/core/tokens';
 
 @Directive({
-    inputs: ['compactSetter: compact'],
-    host: {'[class._compact]': 'compact()'},
+    host: {'[class._compact]': 'c()'},
 })
 export class TuiInputCardGroupDirective {
     private readonly breakpoint = inject(TUI_BREAKPOINT);
-    private readonly c = signal(false);
+    public readonly compact = input(false);
 
-    public readonly compact = computed(() => this.c() || this.breakpoint() === 'mobile');
-
-    public set compactSetter(compact: boolean) {
-        this.c.set(compact);
-    }
+    public readonly c = computed(() => this.compact() || this.breakpoint() === 'mobile');
 }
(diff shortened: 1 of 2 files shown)


projects/core/tokens/breakpoint.ts
```diff
@@ -0,0 +1,38 @@
...
+import {toSignal} from '@angular/core/rxjs-interop';
+import {tuiZoneOptimized} from '@taiga-ui/cdk/observables';
+import {TUI_WINDOW_SIZE} from '@taiga-ui/cdk/tokens';
+import {distinctUntilChanged, map, shareReplay} from 'rxjs';
+
+import {TUI_MEDIA, type TuiMedia} from './media';
+
+export type TuiBreakpointMediaKey = keyof Omit<TuiMedia, 'tablet'>;
+
+export const TUI_BREAKPOINT = new InjectionToken<Signal<TuiBreakpointMediaKey | null>>(
```
waterplea: Maybe we can remove `null` from the type? Or is it for the SSR reasons?

vladimirpotekhin: Removed the null, no SSR errors occurred.

projects/core/tokens/breakpoint.ts
```diff
@@ -0,0 +1,40 @@
...
+import {toSignal} from '@angular/core/rxjs-interop';
+import {tuiZoneOptimized} from '@taiga-ui/cdk/observables';
+import {TUI_WINDOW_SIZE} from '@taiga-ui/cdk/tokens';
+import {distinctUntilChanged, map} from 'rxjs';
+
+import {TUI_MEDIA, type TuiMedia} from './media';
+
+export type TuiBreakpointMediaKey = keyof Omit<TuiMedia, 'tablet'>;
+
+export const TUI_BREAKPOINT = new InjectionToken<Signal<TuiBreakpointMediaKey>>(
```
splincode, line 11: @vladimirpotekhin what do you think?

```ts
const initialValue = 'desktopLarge' as const;

// ...
map((key) => invert[key || sorted[sorted.length - 1] || 0] ?? initialValue)


// ...
 return toSignal(stream$, {initialValue});
```

vladimirpotekhin: Thanks, but I don't think there's any point in this case. It'll just add an extra line.

projects/addon-commerce/components/input-card-group/input-card-group.directive.ts
```diff
@@ -1,23 +1,17 @@
...
 
 @Directive({
     inputs: ['compactSetter: compact'],
     host: {'[class._compact]': 'compact()'},
 })
 export class TuiInputCardGroupDirective {
-    private readonly c$ = new BehaviorSubject(false);
-    private readonly m$ = inject(TuiBreakpointService).pipe(map((b) => b === 'mobile'));
+    private readonly breakpoint = inject(TUI_BREAKPOINT);
+    private readonly c = signal(false);
```
waterplea: Should we now use signal input here instead?

PR #12652, opened 2025-11-25
fix(kit): use logical css properties for `tui-segmented`

splincode:


projects/kit/components/segmented/segmented.style.less
```diff
@@ -113,10 +113,10 @@ tui-segmented {
 
     & > label > input:not([tuiRadio]) {
         position: absolute;
-        top: -0.125rem;
-        left: -0.125rem;
-        right: -0.125rem;
-        bottom: -0.125rem;
+        inset-block-start: -0.125rem;
```
waterplea: inset: -0.125rem;

projects/kit/components/segmented/segmented.style.less
```diff
@@ -136,6 +134,16 @@ tui-segmented {
         box-sizing: border-box;
         filter: drop-shadow(0 0.25rem 1.25rem rgba(0, 0, 0, 0.1));
     }
+
+    &:dir(ltr)::before {
```
waterplea: We need attribute selector instead:
https://caniuse.com/css-dir-pseudo
```css
[dir='rtl'] &::before {
```

projects/kit/components/segmented/segmented.style.less
splincode: fixed

projects/kit/components/segmented/segmented.style.less
splincode: fixed

PR #12653, opened 2025-11-25
feat(addon-mobile): `Switch` use native on iOS when supported

waterplea:
Fixes #


projects/addon-mobile/styles/ios/switch.less
```diff
@@ -1,6 +1,28 @@
...
+    &:invalid {
+        accent-color: var(--tui-status-negative);
+    }
+
+    &::before,
+    &::after {
+        display: none;
+    }
+
+    @supports (anchor-name: --xxx) {
```
splincode: Image

vladimirpotekhin (approved):

PR #12666, opened 2025-11-26
refactor(core): `Cell` use in `Option` as host directive

waterplea:
Fixes #

### DIFF
--- a/projects/cdk/directives/animated/animated.directive.ts
+++ b/projects/cdk/directives/animated/animated.directive.ts
@@ -1,7 +1,6 @@
 import {isPlatformServer} from '@angular/common';
 import {
     afterNextRender,
-    ApplicationRef,
     Directive,
     inject,
     type OnDestroy,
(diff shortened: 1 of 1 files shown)


projects/cdk/directives/animated/animated.directive.ts
```diff
@@ -84,10 +84,12 @@ function wrap(renderer: Renderer2): Renderer2['removeChild'] {
...
         const last = animations[animations.length - 1];
         const finish = (): void => {
-            if (!parent || parent.contains(el)) {
-                remove();
-                app.tick();
-            }
+            try {
+                if (!parent || parent.contains(el)) {
+                    remove();
+                    app.tick();
```
splincode: why not

```ts
const finish = (): void => {
  if (!parent || parent.contains(el)) {
        remove();
        
        clearInterval(animationId);
        animationId = setTimeout(() => app.tick());
  }
}
```

waterplea: Looks like we do not need it at all, tested in both main and v4. I'm not sure why when I was playing with this idea initially element wasn't disappearing without the tick. It confused me back then since removeChild is supposed to just call native remove method on the DOM element but I didn't look into it. Looks like we can just remove it:
https://github.com/taiga-family/taiga-ui/pull/12670

PR #12693, opened 2025-11-28
fix(addon-table): `TableControl` fix empty list view

wtorso:


projects/addon-table/directives/table-control/table-control.directive.ts
```diff
@@ -12,11 +12,13 @@ import {type TuiCheckboxRowDirective} from './checkbox-row.directive';
 export class TuiTableControlDirective<T> extends TuiControl<readonly T[]> {
     private readonly children = signal<ReadonlyArray<TuiCheckboxRowDirective<T>>>([]);
 
-    public readonly checked: Signal<boolean> = computed(() =>
-        this.children().every((i) => this.value().includes(i.tuiCheckboxRow())),
+    public readonly checked: Signal<boolean> = computed(
```
splincode, line 15: @wtorso Could you add an E2E Playwright test or some Cypress component tests for this case?

PR #12701, opened 2025-12-01
feat(core): add `TUI_FONT_OFFSET` signal token

### LINKED ISSUE #11815: 🐞 - tuiFade, tui-line-clamp - height doesn’t adjust dynamically to font size changes
### Reproduction url

-

### Description

**tuiFade**

When a custom tuiFadeHeight is specified, the height does not scale with tui-font-offset changes, which results in text being cut off vertically. 
line-height isn't changed.

**tui-line-clamp** 

When a custom lineHeight is specified, the heigh does not scale with tui-font-offset changes, which results in text being cut off vertically. 
max-height isn't changed.

### Taiga UI version

4.52.0

### Browser Used

Chrome

### OS Used

macOS

splincode:
Fixes #11815

### DIFF
--- a/projects/demo/src/modules/directives/fade/examples/2/index.less
+++ b/projects/demo/src/modules/directives/fade/examples/2/index.less
@@ -10,7 +10,7 @@
     block-size: 3.75rem;
 
     &_expanded {
-        block-size: var(--t-fade-scroll-height);
+        block-size: 10rem;
     }
 }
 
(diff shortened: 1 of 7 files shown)


projects/demo/src/modules/directives/fade/examples/2/index.less
```diff
@@ -10,7 +10,7 @@
     block-size: 3.75rem;
 
     &_expanded {
-        block-size: 10rem;
+        block-size: var(--t-fade-scroll-height);
```
waterplea: This is a private CSS variable, demo examples should not depend on it.

splincode: @waterplea Could you check out the branch? I don't know how to make dynamic height. When the font gets bigger, 10rem is not enough and the content is hidden.

PR #12724, opened 2025-12-02
chore: install tailwindcss and remove tui-space styles

splincode:


waterplea:
We do not need tailwind

PR #12735, opened 2025-12-03
refactor(core): migrate notification markup to directive

splincode:


projects/core/styles/components/notification.less
```diff
@@ -11,24 +11,24 @@
...
  */
-tui-notification,
 [tuiNotification] {
     .button-clear();
 
     position: relative;
     display: block;
     max-block-size: 100%;
     color: var(--tui-text-primary);
+    cursor: pointer;
```
waterplea: We need to move it to &:is(button, a)

waterplea: Or rather .interactive({}) mixin

.cspell.json
```diff
@@ -48,7 +48,8 @@
         "styleguide",
         "Vercetti",
         "Bellic",
-        "Niko"
+        "Niko",
+        "tuinotification"
```
waterplea: Where is it used all lowercase? Would be great if we didn't have to extend spellcheck just for that

splincode: fixed

projects/core/styles/components/notification.less
splincode: fixed

PR #12753, opened 2025-12-05
feat(core): Hint make offset configurable (#12689)

### LINKED ISSUE #12689: 🚀 - `Hint` make offset configurable
### Description

Currently Hint offset is hardcoded to be 8 on desktop and 16 on mobile. We need to turn it into input here, using current value as default:
https://github.com/taiga-family/taiga-ui/blob/main/projects/core/portals/hint/hint-position.directive.ts#L27

and expose it here:
https://github.com/taiga-family/taiga-ui/blob/main/projects/core/portals/hint/hint.directive.ts#L44

romantalman:
Fixes #12689

### DIFF
--- a/projects/core/portals/hint/hint-position.directive.ts
+++ b/projects/core/portals/hint/hint-position.directive.ts
@@ -24,7 +24,6 @@ const LEFT = 0;
 @Directive()
 export class TuiHintPosition extends TuiPositionAccessor {
     private readonly el = tuiInjectElement();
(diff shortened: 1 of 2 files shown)


splincode:
@romantalman Please, add cherry pick to v4.x

romantalman:
@splincode, [done](https://github.com/taiga-family/taiga-ui/pull/12767)

PR #12761, opened 2025-12-05
refactor: `Stepper` use cell

vladimirpotekhin:
Fixes #

### DIFF
--- a/projects/kit/components/stepper/step.component.ts
+++ b/projects/kit/components/stepper/step.component.ts
@@ -18,7 +18,7 @@ import {TuiStepperComponent} from './stepper.component';
     templateUrl: './step.template.html',
     styleUrl: './step.style.less',
     changeDetection: ChangeDetectionStrategy.OnPush,
-    hostDirectives: [{directive: TuiCell, inputs: ['tuiCell: size']}, TuiWithAppearance],
+    hostDirectives: [TuiCell, TuiWithAppearance],
     host: {
         type: 'button',
         '[tabIndex]': 'isActive() ? 0 : -1',
@@ -50,16 +50,25 @@ export class TuiStep {
         () => this.stepper.activeItemIndex() === this.index,
     );
 
-    protected readonly avatarContent = computed(() => {
-        if (this.isActive() || (this.stepState() === 'normal' && !this.icon())) {
-            return `${this.index + 1}`;
-        } else {
-            return (
-                this.icon() ||
-                (this.stepState() === 'error' ? this.icons.error : this.icons.check)
-            );
+    protected avatarAppearance = computed(() => {
+        if (this.isActive()) {
+            return 'primary';
         }
+
+        return this.stepState() === 'error' ? 'negative' : 'secondary';
     });
+
+    protected readonly avatarContent = computed(() =>
+        this.isActive() || (this.stepState() === 'normal' && !this.icon())
+            ? `${this.index + 1}`
+            : this.avatarIcon(),
+    );
+
+    protected readonly avatarIcon = computed(
+        () =>
+            this.icon() ||
+            (this.stepState() === 'error' ? this.icons.error : this.icons.check),
+    );
 
     protected get index(): number {
         return this.stepper.indexOf(this.el);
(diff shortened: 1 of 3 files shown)


splincode (approved):

projects/kit/components/stepper/step.template.html
```diff
@@ -1,16 +1,5 @@
...
-    class="t-marker t-marker_error"
-    [icon]="icons.error"
-/>
-<tui-icon
-    class="t-marker t-marker_pass"
-    [icon]="icons.check"
-/>
-<div class="t-marker t-marker_index"></div>
+<div
+    [appearance]="isActive() ? 'primary' : stepState() === 'error' ? 'negative' : 'secondary'"
```
waterplea: Maybe move it to computed prop to avoid nested ternary?

projects/kit/components/stepper/step.component.ts
```diff
@@ -1,33 +1,41 @@
...
 
 @Component({
     selector:
         'button[tuiStep], a[tuiStep]:not([routerLink]), a[tuiStep][routerLink][routerLinkActive]',
-    imports: [TuiIcon],
+    imports: [TuiAvatar],
     templateUrl: './step.template.html',
     styleUrl: './step.style.less',
     changeDetection: ChangeDetectionStrategy.OnPush,
+    hostDirectives: [{directive: TuiCell, inputs: ['tuiCell: size']}, TuiWithAppearance],
```
waterplea: Why are we exposing size?

projects/kit/components/stepper/step.component.ts
```diff
@@ -38,9 +46,20 @@ export class TuiStep {
     public readonly stepState = input<'error' | 'normal' | 'pass'>('normal');
```
waterplea, line 46: Maybe we can change it to [invalid]="boolean | null" so it is the same as with textfields? I kinda hate that we call it **step**State just because routerLink directive does not follow Angular's own requirement of prefixing directive input names :(

projects/kit/components/stepper/step.component.ts
```diff
@@ -38,9 +46,20 @@ export class TuiStep {
...
+    protected readonly isActive = computed(
+        () => this.stepper.activeItemIndex() === this.index,
+    );
+
+    protected readonly avatarContent = computed(() => {
+        if (this.isActive() || (this.stepState() === 'normal' && !this.icon())) {
+            return `${this.index + 1}`;
+        } else {
+            return (
+                this.icon() ||
```
waterplea: Just a thought for readability: maybe we should have a separate icon computed and then turn `avatarContent` into a ternary?

projects/kit/components/stepper/step.style.less
```diff
@@ -1,128 +1,38 @@
...
-    align-self: flex-start;
-    justify-content: center;
-    background: var(--tui-background-neutral-1);
-    color: var(--tui-text-action);
-    box-sizing: border-box;
-
-    &::before {
-        font-size: 1rem;
+    &[tuiCell][tuiCell]:hover {
+        background: unset;
```
waterplea: Should we maybe just move it into :host and add !important?

projects/kit/components/stepper/step.style.less
```diff
@@ -1,128 +1,38 @@
...
     }
 
-    :host:disabled & {
-        background: var(--tui-background-neutral-1);
-        color: var(--tui-text-secondary);
+    &._active._active:not(:disabled) {
+        cursor: unset;
```
waterplea, line 11: What is it otherwise?

projects/kit/components/stepper/step.style.less
```diff
@@ -1,128 +1,38 @@
...
+}
 
-    &_pass {
-        :host:not([data-state='pass']) &,
-        :host._active & {
-            display: none;
-        }
+[tuiAvatar][data-size] {
+    &::before {
+        transition: none;
```
waterplea, line 32: How does it look otherwise?

vladimirpotekhin: https://github.com/user-attachments/assets/24dd102d-3087-40fe-b5f7-211ad0527426

PR #12792, opened 2025-12-09
chore(addon-doc): migrate to signals

mdlufy:
Fixes #

### DIFF
--- a/projects/addon-doc/components/demo/index.ts
+++ b/projects/addon-doc/components/demo/index.ts
@@ -8,6 +8,7 @@ import {
     ElementRef,
     inject,
     input,
+    type Signal,
     signal,
     TemplateRef,
     viewChild,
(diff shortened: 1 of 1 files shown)


projects/addon-doc/components/example/example.component.ts
```diff
@@ -115,9 +115,12 @@ export class TuiDocExample {
 
     public readonly component = input<Promise<Type<unknown>>>();
 
-    @Input()
-    public set content(content: Record<string, TuiRawLoaderContent>) {
-        this.rawLoader$$.next(content);
+    public readonly content = input<Record<string, TuiRawLoaderContent>>({});
+
+    constructor() {
+        effect(() => {
```
waterplea: Such effects (here and in code component) would trigger for initial value too. Might not be a problem, but might be a better idea to use ngOnChanges

projects/addon-doc/components/navigation/navigation.component.ts
```diff
@@ -84,8 +84,9 @@ function tuiUniqBy<T extends Record<string, any>>(
     },
 })
 export class TuiDocNavigation {
-    @ViewChild(TuiInputDirective, {read: ElementRef})
-    private readonly searchInput?: ElementRef<HTMLInputElement>;
+    private readonly searchInput = viewChild(TuiInputDirective, {
+        read: ElementRef<HTMLInputElement>,
```
waterplea: This will not set type for nativeElement, it will still be any.

projects/addon-doc/components/navigation/scroll-into-view.directive.ts
```diff
@@ -23,8 +23,11 @@ export class TuiDocScrollIntoViewLink {
...
         .subscribe(() => this.el.scrollIntoView());
 
-    @Input()
-    public set tuiDocScrollIntoViewLink(shallWe: boolean) {
-        this.scroll$.next(shallWe);
+    public readonly tuiDocScrollIntoViewLink = input<boolean>(false);
+
+    constructor() {
+        effect(() => {
+            this.scroll$.next(this.tuiDocScrollIntoViewLink());
```
waterplea: This will immediately exectute, not when value is passed to input like with setter. Use ngOnChanges

PR #12796, opened 2025-12-09
feat(kit)!: `InputRange` uses new approach to implement negative values without minus sign

nsbarsukov:
https://taiga-ui.dev/next/components/input-range#negative

### DIFF
--- a/projects/demo-cypress/src/tests/input-number/always-negative-with-hidden-minus.cy.ts
+++ b/projects/demo-cypress/src/tests/input-number/always-negative-with-hidden-minus.cy.ts
@@ -30,7 +30,7 @@ import {TuiInputNumber, tuiInputNumberOptionsProvider} from '@taiga-ui/kit';
     `,
     changeDetection: ChangeDetectionStrategy.OnPush,
     providers: [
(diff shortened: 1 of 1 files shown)


projects/demo-cypress/src/tests/input-number/abs-transformer.cy.ts
nsbarsukov, line 1: Actually, this file was not deleted.
It was just renamed to `always-negative-with-hidden-minus.cy.ts` + more tests added

projects/demo-cypress/src/tests/input-number/always-negative-with-hidden-minus.cy.ts
```diff
@@ -0,0 +1,256 @@
...
+        });
+    });
+
+    it('hidden minus is not visible in the textfield', () => {
+        cy.viewport(200, 200);
+
+        cy.get('[tuiInputNumber]')
+            .type('15')
+            .should('have.value', `${CHAR_ZERO_WIDTH_SPACE}15 days ago`)
+            .compareSnapshot('input-number-with-hidden-minus');
```
nsbarsukov, line 254: Image

PR #12799, opened 2025-12-09
refactor(kit): refactor children to signals

waterplea:
Fixes #

### DIFF
--- a/projects/core/styles/mixins/slider.less
+++ b/projects/core/styles/mixins/slider.less
@@ -15,7 +15,6 @@
     --t-offset: calc(@thumb / 2);
 
     display: flex;
-    align-items: center;
     font: var(--tui-font-body-s);
     color: var(--tui-text-secondary);
     margin: 0.25rem var(--t-offset) 0;
--- a/projects/core/styles/mixins/slider.scss
+++ b/projects/core/styles/mixins/slider.scss
@@ -17,7 +17,6 @@ $track-inset: (
     --t-offset: calc($thumb / 2);
 
     display: flex;
-    align-items: center;
     font: var(--tui-font-body-s);
     color: var(--tui-text-secondary);
     margin: 0.25rem var(--t-offset) 0;
--- a/projects/demo/src/modules/components/input-range/examples/3/index.html
+++ b/projects/demo/src/modules/components/input-range/examples/3/index.html
@@ -10,14 +10,10 @@
 </tui-input-range>
 
 <div class="ticks-labels">
-    <span>
-        <tui-icon icon="@tui.volume-x" />
-    </span>
+    <tui-icon icon="@tui.volume-x" />
     <span>20%</span>
     <span>40%</span>
     <span>60%</span>
     <span>80%</span>
-    <span>
-        <tui-icon icon="@tui.volume-2" />
-    </span>
+    <tui-icon icon="@tui.volume-2" />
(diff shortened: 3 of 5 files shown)


projects/kit/components/input-range/input-range.component.ts
```diff
@@ -33,6 +33,9 @@ import {
     type PolymorpheusPrimitive,
 } from '@taiga-ui/polymorpheus';
 
+const transform = (x?: readonly [string, string] | null): readonly [string, string] =>
+    x ?? ['', ''];
```
splincode, line 37: why not `[this.options.prefix, this.options.prefix]`?

waterplea: transform functions must be pure:
https://angular.dev/guide/components/inputs#input-transforms

projects/demo-cypress/src/tests/range/range.cy.ts
```diff
@@ -35,7 +35,7 @@ describe('TuiRange', () => {
         public min = 1;
         public segments = 10;
         public step = 1;
-        public keySteps: TuiKeySteps | null = null;
+        public keySteps: TuiKeySteps | undefined = undefined;
```
nsbarsukov, line 38: Why ?

projects/kit/components/input-range/input-range.component.ts
```diff
@@ -92,29 +84,19 @@ export class TuiInputRange extends TuiControl<readonly [number, number]> {
...
-    @Input()
-    public keySteps: TuiKeySteps | null = null;
-
-    @Input({transform: (x: readonly [string, string] | null) => x ?? ['', '']})
-    public prefix: readonly [string, string] = [this.options.prefix, this.options.prefix];
+    public readonly min = input(0);
+    public readonly max = input(100);
+    public readonly step = input(1);
+    public readonly segments = input(1);
+    public readonly keySteps = input<TuiKeySteps>();
```
nsbarsukov, line 91: Why not ` TuiKeySteps | null` (as it was before) ?

projects/kit/components/range/range.component.ts
```diff
@@ -60,40 +59,34 @@ export class TuiRange extends TuiControl<[number, number]> {
...
-
     public readonly step = input(1);
-
     public readonly size = input<TuiSizeS>(this.options.size);
-
     public readonly segments = input(1);
-
-    public readonly keySteps = input<TuiKeySteps | null>(null);
-
+    public readonly keySteps = input<TuiKeySteps>();
```
nsbarsukov, line 66: The same question for `Range`

nsbarsukov (approved):

projects/demo-cypress/src/tests/range/range.cy.ts
waterplea: So that it becomes just an optional input without explicit `null` as a value

projects/kit/components/input-range/input-range.component.ts
waterplea: So that it becomes just an optional input without explicit `null` as a value

projects/kit/components/range/range.component.ts
waterplea: So that it becomes just an optional input without explicit `null` as a value

PR #12800, opened 2025-12-09
 fix(cdk): `tuiValueChanges` should not work for same control value emission

chat-du-cheshire:
Fix same value control emission in tuiValueChanges directive

https://stackblitz.com/edit/angular-pyaxiwzf


projects/cdk/directives/value-changes/value-changes.directive.ts
```diff
@@ -15,13 +24,14 @@ export class TuiValueChanges<T> implements DoCheck {
...
 
     public readonly tuiValueChanges = outputFromObservable(this.tuiValueChanges$);
 
     public ngDoCheck(): void {
-        this.refresh$.next(
-            this.control?.valueChanges || this.container?.valueChanges || EMPTY,
-        );
+        const control = this.control || this.container;
+
+        this.refresh$.next(control ? tuiControlValue<T>(control) : EMPTY);
```
waterplea: This way you recreate `tuiControlValue` on each change detection cycle, that's suboptimal.

chat-du-cheshire: Moved into tuiValueChanges$ stream

waterplea (approved):
Thank you for the contribution! Please run lint fix and prettier so we can merge

projects/cdk/directives/value-changes/value-changes.directive.ts
```diff
@@ -1,27 +1,28 @@
 import {Directive, type DoCheck, inject} from '@angular/core';
 import {outputFromObservable} from '@angular/core/rxjs-interop';
 import {ControlContainer, NgControl} from '@angular/forms';
-import {distinctUntilChanged, EMPTY, type Observable, Subject, switchAll} from 'rxjs';
+import {distinctUntilChanged, skip, Subject, switchMap} from 'rxjs';
+
+import {tuiControlValue} from '../../observables';
```
splincode: change it to `import {tuiControlValue} from '@taiga-ui/cdk/observables';`

chat-du-cheshire: Done

splincode (approved):

splincode:
@chat-du-cheshire Thank you for your contribution. Would you add cherry pick to v4.x?

chat-du-cheshire:
> @chat-du-cheshire Thank you for your contribution. Would you add cherry pick to v4.x?

NP, what is 4.x?

splincode:
@chat-du-cheshire branch
image

chat-du-cheshire:
done https://github.com/taiga-family/taiga-ui/pull/12850

but too many lint errors, I don't know why

and as I can see `import {outputFromObservable} from '@angular/core/rxjs-interop';` doesn't exist

chat-du-cheshire:
I can adopt solution for ng 16.2 but still don't know what to do with lint

PR #12808, opened 2025-12-10
refactor: use css logical properties

splincode:


projects/addon-charts/components/bar-chart/bar-chart.style.less
```diff
@@ -33,7 +33,7 @@
 .t-host {
     position: absolute;
     left: 50%;
-    right: 50%;
+    inset-inline-end: 50%;
```
waterplea, line 36: inset-inline: 50%

projects/addon-charts/components/line-chart/line-chart.style.less
```diff
@@ -60,8 +60,8 @@
 
     &_vertical {
         top: 0;
-        bottom: 0;
         left: 50%;
+        inset-block-end: 0;
```
waterplea, line 64: inset: 0 auto 0 50%

projects/addon-charts/components/ring-chart/ring-chart.style.less
```diff
@@ -75,8 +75,8 @@
     position: absolute;
     top: 25%;
```
waterplea, line 76: inset: 25%

PR #12830, opened 2025-12-12
feat(kit): add `tuiList` directive

splincode:


projects/layout/directives/list/index.ts
waterplea, line 1: Let's move this to @layout/components. I know it technically is directive, but it feels like a component from user's perspective. And it's a great candidate for layout, rather than kit. Also the demo page should move to layout section from the markup.

projects/kit/directives/list/list.directive.ts
```diff
@@ -0,0 +1,44 @@
...
+        '[attr.data-size]': 'size()',
+    },
+})
+export class TuiList {
+    private readonly options = inject(TUI_LIST_OPTIONS);
+    protected readonly nothing = tuiWithStyles(Styles);
+
+    public readonly size = input(this.options.size, {
+        alias: 'tuiList',
+        transform: (size: TuiSizeL | TuiSizeS | '') => size || this.options.size,
```
waterplea: I'm not sure we can use `this` in transformer, even if it's technically never going to change. Let's male options protected and write "size() || options.size" in binding instead.

projects/demo/src/modules/markup/lists/examples/1/index.html
```diff
@@ -1,27 +1,50 @@
...
+    <div
+        class="wrapper"
+        [tuiPlatform]="platform"
+    >
+        <code>{{ platform === 'web' ? 'Web' : 'Android/iOS' }}</code>
+
+        <h2>Body L</h2>
+
+        <ul
+            aria-label="List of Continents"
```
waterplea: Not sure we want to support it as official API of lists, let's not add this behavior. While this is a good idea, it would only work with text and people might want to add tooltips or style it differently, it can also be problematic with screen readers and the text is not selectable. So let's leave title styling up to the user:

Image

projects/layout/directives/list/list.style.less
```diff
@@ -0,0 +1,166 @@
...
+        gap: 1rem 0.75rem;
+        align-items: start;
+
+        & > * {
+            grid-column: 2;
+            padding-inline-start: 0;
+        }
+    }
+
+    & > li::before {
```
waterplea, line 17: Can we do it with ::marker rather than ::before? I think it would be expected by users that they can change ::marker color

projects/demo/src/modules/markup/lists/index.html
waterplea, line 1: Let's add traditional Setup tab with imports instruction

projects/kit/directives/list/list.style.less
```diff
@@ -0,0 +1,166 @@
+[tuiList] {
+    display: grid;
+    padding-inline-start: 1rem;
+    list-style: none;
+    gap: 1rem;
+    padding-block-start: 0.25rem;
+    padding-block-end: 0.25rem;
+
+    & [tuiList] {
```
waterplea: Unnecessary `& `

projects/kit/directives/list/list.style.less
```diff
@@ -0,0 +1,166 @@
...
+            padding: 0;
+        }
+
+        &[data-size='l'] {
+            & > li {
+                line-height: 1.25rem;
+            }
+
+            &:is(ul) > li::before {
+                margin-block-start: calc(1.25rem / 2 - 0.1875rem);
```
waterplea: Seems like a useless calc, maybe we can add local var instead, write this formula once and just change that variable inside different data-size?

projects/layout/directives/list/list.style.less
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    &[aria-label]::before {
+        content: attr(aria-label);
+        display: flex;
+        align-items: center;
+    }
+
+    & > li {
+        display: grid;
```
waterplea: Can we keep original display for list items? Otherwise it would be difficult for people to use tags inside them like:
```html
<li>
  This item uses <strong>nested tags</strong> inside which will break
</li>
```

projects/layout/directives/list/list.style.less
```diff
@@ -0,0 +1,166 @@
+[tuiList] {
+    display: grid;
+    padding-inline-start: 1rem;
+    list-style: none;
+    gap: 1rem;
+    padding-block-start: 0.25rem;
```
waterplea: Let's write `padding: 0.25rem 0;`. There shouldn't be left padding. It is there in the spec but that's just wrong specs. Otherwise there will be double padding in cards, sheets etc.

projects/layout/directives/list/list.style.less
splincode: I tried this earlier, but it still doesn’t work in Safari.

projects/layout/directives/list/list.style.less
splincode: All my logic was deliberately built for nested lists so that the marker does not require absolute positioning. As an alternative, I can suggest the following:

```html
 <li>
        <span>
            This item uses
            <strong>nested tags</strong>
            inside which will break
        </span>
    </li>
```

projects/layout/directives/list/list.style.less
splincode: `There shouldn't be left padding.`

Why? we can see here:

image

projects/layout/directives/list/index.ts
splincode: fixed

projects/kit/directives/list/list.directive.ts
splincode: fixed

projects/demo/src/modules/markup/lists/examples/1/index.html
splincode: fixed

projects/kit/directives/list/list.style.less
splincode: fixed

projects/kit/directives/list/list.style.less
splincode: fixed

projects/layout/directives/list/list.style.less
waterplea: All those environments already have paddings:
image
It's just a problem of people making the spec not realizing how it is actually used in the end.

projects/layout/directives/list/list.style.less
waterplea: OMG, I see. Ok.

projects/layout/directives/list/list.style.less
waterplea: I understand the idea but this would piss people off. I think we need to use `font-variant-numeric: tabular-nums;` so digits have predictable width and add `ol:has(> li:nth-child(10))` so that we can extend the space for lists longer than 9 items. I'll check with design team regardless `font-variant-numeric: tabular-nums;`

projects/demo/src/modules/markup/lists/index.html
splincode: fixed

projects/layout/directives/list/list.style.less
splincode: fixed

projects/layout/directives/list/list.style.less
splincode: fixed

projects/demo/src/modules/app/app.routes.ts
```diff
@@ -194,6 +194,11 @@ export const ROUTES: Routes = [
         loadComponent: async () => import('../components/cell'),
         title: 'Cell',
     }),
+    route({
+        path: DemoRoute.Lists,
```
waterplea: Let's rename litstS to list, with no "S" as this is no longer a page about lists but a page about tuiList directive.

projects/layout/directives/list/list.style.less
```diff
@@ -0,0 +1,56 @@
+[tuiList] {
+    list-style: none;
+    padding-block-start: 0.25rem;
+    padding-block-end: 0.25rem;
```
waterplea: Why not `padding: 0.25rem 0;`?

splincode: fixed

projects/demo/src/modules/app/app.routes.ts
splincode: fixed

PR #12837, opened 2025-12-14
refactor(core): finalize core and kit to signal input

medbenmakhlouf:
Fixes #


waterplea:
@medbenmakhlouf thank you very much for your help getting the public API up to date! We have a whole new set of date picker components coming up once the designs are finalized so I think it's best not to try and move current ones to signals, we'd want to have as little of tuiSetSignal hacks as possible, so I'll close this PR. Looks like everything is finished now with the last few PRs coming on our part and I hope first v5 release candidate will be out this week! Once again, thank you for your help.

PR #12921, opened 2025-12-23
fix(kit): `Tabs` does not close dropdown if application does not contains `provideAnimations()`

nsbarsukov:
## Reproduction
```html
<tui-tabs-with-more class="tabs" [activeItemIndex]="activeItemIndex()">
    @for (tab of tabs; track tab) {
        <button 
            *tuiItem 
            tuiTab 
            type="button" 
            (click)="activeItemIndex.set($index)"
        >
            {{ tab }}
        </button>
    }
</tui-tabs-with-more>
```
Open `More` section and click on any tab.

https://github.com/user-attachments/assets/2df18584-0c96-4fc7-8fe3-5d18999ada7d

https://stackblitz.com/edit/tui-tabs-without-ng-animations-bug

## Problem 1
https://github.com/taiga-family/taiga-ui/blob/8b64ec1223e391be56961760d6878d3486428f09/projects/kit/components/tabs/tab.directive.ts#L33-L37

It will never dispatch `(tui-tab-activate)` for case above: `TabsWithMore` gets new `activeItemIndex` property and remove clicked tab from DOM before its parent gets bubbled click event.

## Problem 2
https://github.com/taiga-family/taiga-ui/blob/8b64ec1223e391be56961760d6878d3486428f09/projects/cdk/tokens/active-element.ts#L69-L76

**Without `provideAnimations`:**
```ts
// removedElement equals to the clicked tab
isValidFocusout(tuiGetActualTarget(event), removedElement), 
```

**With `provideAnimations`:**
```ts
// removedElement does NOT equal to the clicked tab
isValidFocusout(tuiGetActualTarget(event), removedElement), 
```

### DIFF
--- a/projects/demo-cypress/src/tests/input-phone-international.cy.ts
+++ b/projects/demo-cypress/src/tests/input-phone-international.cy.ts
@@ -104,8 +104,10 @@ describe('InputPhoneInternational', () => {
 
         it('Focused textfield with country calling code only => blur => empty textfield => no form control value changes', () => {
             cy.get('@input')
+                .should('have.value', '')
                 .focus()
                 .should('have.value', '+1 ')
+                .wait(0)
                 .blur()
                 .should('have.value', '');
 


projects/kit/components/tabs/tabs-with-more.component.ts
nsbarsukov, line 1: No logic is changed inside this file.
Just rewrite some getters to `computed(...)` (without changing its content)

nsbarsukov: The same for `projects/kit/components/tabs/tabs-with-more.template.html`

projects/cdk/tokens/removed-element.ts
```diff
@@ -24,6 +40,8 @@ export const TUI_REMOVED_ELEMENT = new InjectionToken(
                         startWith(element),
                     ),
                 ),
+                // Applications with / without `provideAnimations()` has different timing for DOM element removal – make consistent behavior.
+                ngAnimationEnabled ? identity : debounceTime(0),
```
waterplea, line 44: We cannot have debounceTime. The whole point of removed element tracking is so ActiveZone is synchronous, i.e:
```ts
el.blur();
// <-- already false here
```
If removed element is debounced then blur event caused by removal will be treated the same as `el.blur()`.

nsbarsukov: Okay, I have no idea how to fix it now.
I will close this PR and create issue to fix it later

nsbarsukov: - https://github.com/taiga-family/taiga-ui/issues/13005

PR #12983, opened 2026-01-14
fix(kit): `InputChip` make chip hoverable in readonly mode

vladimirpotekhin:
Fixes #


projects/kit/components/input-chip/input-chip.component.ts
```diff
@@ -98,7 +98,9 @@ export class TuiInputChipComponent<T> {
     }
 
     protected delete(): void {
-        this.textfield.cva?.onChange(this.value().filter((_, i) => i !== this.index));
+        if (this.textfield.cva?.interactive) {
```
splincode: Could you add e2e or cypress test, please

splincode (approved):

PR #13003, opened 2026-01-16
chore: add migration for some tokens

vladimirpotekhin:
Fixes #


splincode (approved):

projects/demo-cypress/src/tests/combo-box/combo-box-input-event.cy.ts
nsbarsukov, line 60: Rebase branch

PR #13008, opened 2026-01-16
fix(kit): `ComboBox` should emit `(input)` event on erasing unmatched value on blur

### LINKED ISSUE #12681: 🐞 - `ComboBox` does not emit `(input)` event when the field (with unmatched value) is cleared on blur
### Reproduction url

https://stackblitz.com/edit/angular-kwapbrg4?file=src%2Fapp%2Fdatabase.ts

### Description

In TuiComboBox, the input event is correctly emitted when typing in the search field.
It is also correctly emitted when clearing the value via the cleaner button.

However, if you:

1. Type any text into the TuiComboBox input
2. Do not choose anything from the dropdown list
3. Click outside the component (which closes the dropdown and resets the input)

The input value is cleared, but the input event is not emitted, even though the behavior should be consistent with clearing via the cleaner.

### Taiga UI version

4.62.0

### Browser Used

Chrome

### OS Used

macOS

nsbarsukov:
Fixes #12681

### DIFF
--- a/projects/kit/components/combo-box/combo-box.directive.ts
+++ b/projects/kit/components/combo-box/combo-box.directive.ts
@@ -10,7 +10,7 @@ import {
 import {tuiAsControl, TuiControl} from '@taiga-ui/cdk/classes';
 import {TUI_STRICT_MATCHER} from '@taiga-ui/cdk/constants';
 import {type TuiStringMatcher} from '@taiga-ui/cdk/types';
-import {tuiInjectElement} from '@taiga-ui/cdk/utils';
+import {tuiInjectElement} from '@taiga-ui/cdk/utils/dom';
 import {
     tuiAsOptionContent,
     type TuiDataListAccessor,


projects/kit/components/combo-box/combo-box.directive.ts
```diff
@@ -113,10 +115,14 @@ export class TuiComboBoxDirective<T>
...
 
     protected readonly blurEffect = effect(() => {
-        const incomplete = untracked(() => this.strict() && this.value() === null);
+        const incomplete = untracked(
+            () => this.strict() && this.input.value() && this.value() === null,
+        );
 
         if (!this.host.focused() && incomplete) {
-            this.input.value.set('');
+            this.el.value = '';
```
waterplea, line 123: Should we maybe use the same technique as in cleaner?
https://github.com/taiga-family/taiga-ui/blob/main/projects/core/components/input/input.directive.ts#L67

nsbarsukov: Cleaner technique includes 

https://github.com/taiga-family/taiga-ui/blob/1db2ca04a69d3eeb5a55b0a0db08a91fffaa53cd/projects/core/components/input/input.directive.ts#L68

Not expected behavior for `blur` handler

PR #13009, opened 2026-01-16
chore: add accordion migration

vladimirpotekhin:
Fixes #


projects/cdk/schematics/ng-update/utils/templates/index.ts
```diff
@@ -1,3 +1,4 @@
+export * from './has-ancestor';
```
nsbarsukov: `export function hasAncestor` + `export function hasChild` === `hierarchy.ts` / `dom-traversal.ts` ?

projects/demo-cypress/src/tests/combo-box/combo-box-input-event.cy.ts
nsbarsukov, line 60: Rebase branch

PR #13019, opened 2026-01-19
fix(kit): `InputChip` fix pasting with whitespace separator, fix caret position

### LINKED ISSUE #13016: 🐞 - `InputChip` copy-paste from exel issue
### Reproduction url

https://taiga-ui.dev/components/input-chip/API?separator=

### Description

[separator] = /[\s,;]/ does not work when copying values ​​from an Excel spreadsheet.

### Taiga UI version

4.67.0

### Browser Used

Chrome

### OS Used

macOS

### LINKED ISSUE #13017: 🐞 - InputChip sets the cursor to the beginning of the field when using copy-paste
### Reproduction url

https://taiga-ui.dev/components/input-chip#multi-select

### Description

InputChip sets the cursor to the beginning of the field when using copy-paste

### Taiga UI version

4.64.0

### Browser Used

Chrome

### OS Used

macOS

vladimirpotekhin:
Fixes #13017
Fixes #13016

### DIFF
--- a/projects/kit/components/input-chip/input-chip.directive.ts
+++ b/projects/kit/components/input-chip/input-chip.directive.ts
@@ -83,10 +83,13 @@ export class TuiInputChipBaseDirective<T>
 
     protected onEnter(): void {
         const value = this.textfield.value().trim();
-        const items: any[] = this.separator ? value.split(this.separator) : [value];
-        const valid = items.filter(
-            (item) => item && !this.handlers.disabledItemHandler()(item),
-        );
+        const items = this.separator ? value.split(this.separator) : [value];
+
+        const valid = items
+            .map(tuiSanitizeText)
+            .filter(
+                (item) => item && !this.handlers.disabledItemHandler()(item as T),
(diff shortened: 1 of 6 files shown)


projects/kit/components/input-chip/input-chip.directive.ts
```diff
@@ -112,7 +115,10 @@ export class TuiInputChipBaseDirective<T>
                 ? event.dataTransfer?.getData('text/plain') || ''
                 : tuiGetClipboardDataText(event);
 
-        this.textfield.value.set(tuiSanitizeText(value));
+        if (this.textfield.input?.nativeElement) {
```
waterplea, line 118: Why is this change necessary?

vladimirpotekhin: Yes, otherwise the carriage jumps ( [see](https://github.com/taiga-family/taiga-ui/blob/main/projects/cdk/utils/dom/value.ts#L87) )

PR #13051, opened 2026-01-22
chore: add migtaion for `Slides`, `ElasticContainer`, `FloatingContaner`, `textfieldDropdown`

vladimirpotekhin:
part of #11917


projects/cdk/schematics/ng-update/v5/steps/constants/identifiers-to-replace.ts
```diff
@@ -37,4 +37,34 @@ export const IDENTIFIERS_TO_REPLACE: ReplacementIdentifierMulti[] = [
             moduleSpecifier: '@taiga-ui/core',
         },
     },
+    {
```
splincode, line 40: Could you add simple test?

splincode (approved):

PR #13075, opened 2026-01-26
feat(core): implement version encapsulation for global styles

waterplea:
Fixes #

### DIFF
--- a/projects/demo/src/modules/directives/hint/examples/1/index.ts
+++ b/projects/demo/src/modules/directives/hint/examples/1/index.ts
@@ -1,11 +1,11 @@
 import {Component, signal} from '@angular/core';
 import {changeDetection} from '@demo/emulate/change-detection';
 import {encapsulation} from '@demo/emulate/encapsulation';
-import {TuiHint} from '@taiga-ui/core';
+import {TuiButton, TuiHint} from '@taiga-ui/core';
 import {TuiAutoColorPipe, TuiAvatar} from '@taiga-ui/kit';
 
 @Component({
-    imports: [TuiAutoColorPipe, TuiAvatar, TuiHint],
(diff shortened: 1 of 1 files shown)


projects/core/components/textfield/textfield-multi/textfield-multi.component.ts
```diff
@@ -39,7 +40,12 @@ import {TUI_TEXTFIELD_ITEM} from './textfield-item.component';
         WaResizeObserver,
     ],
     templateUrl: './textfield-multi.template.html',
-    styleUrl: './textfield-multi.style.less',
+    styles: `
+        [data-tui-version='${TUI_VERSION}'] {
+            @import '@taiga-ui/core/styles/components/textfield.less';
+            @import 'textfield-multi.style.less';
```
splincode: `@import './textfield-multi.style.less';` ?

projects/kit/components/avatar/avatar.directive.ts
```diff
@@ -18,7 +19,11 @@ import {TUI_AVATAR_OPTIONS} from './avatar.options';
 
 @Component({
     template: '',
-    styles: '@import "@taiga-ui/kit/styles/components/avatar.less";',
+    styles: `
+        [data-tui-version='${TUI_VERSION}'] {
```
nsbarsukov, line 23: Should we use additional selector for the such kind of conditional import?


```suggestion
        [tuiAvatar][data-tui-version='${TUI_VERSION}'] {
```

Otherwise, DevTools contain too many false-positive selectors

Image

nsbarsukov: \+ `:where` to prevent increasing specificity

waterplea: Those are empty and are removed in production, I don't think we should complicate the code because of those:

image

PR #13081, opened 2026-01-26
chore(demo): improve docs for llm

sayranovv:
Improve documentation for llms

1. Added a map of imports by package
2. Added a component generation checklist
3. Added a section defining CDK types
4. Reworked the table API to add an outputs section with event types
5. Added a Common Mistakes section


projects/demo/scripts/llms/llms-full-generate.ts
```diff
@@ -220,6 +221,34 @@ async function main(): Promise<void> {
...
+        process.cwd(),
+        'projects/demo/src/llms-header-sections',
+    );
+
+    const headerFiles = [
+        'import-map.md',
+        'checklist.md',
+        'cdk-types.md',
+        'common-mistakes.md',
+    ];
```
mdlufy: We have https://github.com/taiga-family/taiga-ui/blob/main/projects/demo/src/llms-config.json, may be expand it ?

projects/demo/scripts/llms/utils/generate-import-map.ts
```diff
@@ -0,0 +1,279 @@
...
+interface PackageExports {
+    components: Set<string>;
+    directives: Set<string>;
+    pipes: Set<string>;
+    services: Set<string>;
+    types: Set<string>;
+    tokens: Set<string>;
+    utils: Set<string>;
+    classes: Set<string>;
+}
```
mdlufy: Let's add yet another layer in import map, to make it: package -> entity -> export instead of package -> all exports

PR #13088, opened 2026-01-27
feat(kit): support configure copy hint inside hint

splincode:
image


waterplea:
Do we still need it, now that we have a dedicated ButtonCopy for hints?

splincode:
@waterplea should we support override `tuiHintAppearance="dark" tuiHintDirection="top"`?

PR #13096, opened 2026-01-27
feat(kit): add ability to use copy directive with textfield multi

vladimirpotekhin:


projects/kit/components/copy/copy.directive.ts
```diff
@@ -67,9 +72,24 @@ export class TuiCopyDirective {
...
+            ? !!this.textfieldMulti?.control()?.value?.length
+            : !!this.textfield.value();
+    }
+
     protected copy(): void {
-        this.textfield.input()?.nativeElement.select();
-        this.doc.execCommand('copy');
+        if (this.textfieldMulti) {
+            this.clipboard.copy(
+                Array.from(this.textfieldMulti.el.querySelectorAll('tui-textfield-item'))
```
waterplea: Maybe it makes more sense to use value + stringify, what do you think?

projects/kit/components/copy/copy.directive.ts
```diff
@@ -34,16 +35,20 @@ import {TUI_COPY_OPTIONS} from './copy.options';
...
         '[style.border-width.rem]': 'textfield.options.size() === "l" ? null : 0.25',
     },
 })
 export class TuiCopyDirective {
     private readonly copied$ = new Subject<void>();
-    private readonly doc = inject(DOCUMENT);
+    private readonly clipboard = inject(Clipboard);
 
     protected readonly textfield = inject(TuiTextfieldComponent);
+    protected readonly textfieldMulti = inject(TuiTextfieldMultiComponent, {
```
waterplea: Multi already provides itself as textfield, maybe we can somehow use that rather than making copy directive explicitly dependent on the class? You know, for bundle size reasons for those not using multi.

projects/kit/components/copy/copy.directive.ts
vladimirpotekhin: Yes, I was hesitant because of the custom chips, but this is probably a rare case. I rewrite it with stringify 👍

PR #13104, opened 2026-01-28
chore(demo): add vertical slider example

waterplea:
Fixes #

### DIFF
--- a/projects/demo/src/modules/components/slider/examples/7/index.less
+++ b/projects/demo/src/modules/components/slider/examples/7/index.less
@@ -10,5 +10,6 @@ tui-dropdown[data-appearance] {
         inline-size: 7rem;
         transform-origin: left;
         transform: rotate(-90deg) translate(-100%, 1rem);
+        cursor: ns-resize;
     }
 }


projects/demo/src/modules/components/slider/examples/7/index.less
```diff
@@ -0,0 +1,14 @@
+tui-dropdown[data-appearance] {
+    min-block-size: 8rem;
+    background: var(--tui-background-neutral-1);
+    box-shadow: none;
+    border: none;
+    backdrop-filter: blur(1rem);
+
+    [tuiSlider] {
```
nsbarsukov, line 8: Should we also set `cursor: ns-resize` ?

projects/demo/src/modules/components/slider/examples/7/index.less
```diff
@@ -0,0 +1,14 @@
...
+        position: absolute;
+        inline-size: 7rem;
+        transform-origin: left;
+        transform: rotate(-90deg) translate(-100%, 1rem);
```
nsbarsukov, line 12: Native `input[type="range"]` has non-standard attribure [orient="vertical"](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/range#orient).

Should we create directive `input[tuiSlider][orient]` + `tuiWithStyles` to imitate the such behavior ? Is it overkill ?

waterplea: I think it's an overkill

projects/demo/src/modules/components/slider/examples/7/index.less
nsbarsukov: https://github.com/user-attachments/assets/ad5b49b5-a2dc-43a8-a8e4-62215cced74c

PR #13112, opened 2026-01-29
chore(demo): improve example `Vertical` on `Slider` documentation page

nsbarsukov:
Relates:
* https://github.com/taiga-family/taiga-ui/pull/13104


waterplea:
Let's not introduce public API for this. Either let's drop <-> for the regular slider, as it's not in the specs or change the example since thumb is exposed without ::ng-deep.

nsbarsukov:
> Let's not introduce public API for this. Either let's drop <-> for the regular slider, as it's not in the specs or change the example since thumb is exposed without ::ng-deep.

@waterplea fixed

splincode (approved):

PR #13158, opened 2026-02-04
fix(kit): add correct size behaviour for TuiBlock

ivanobivan:
Add correct behaviour for size host attribute for TuiBlock
```
 tuiBlock: '',
        '[attr.data-size]': 'size || options.size || "l"',
        '[class._disabled]': '!!this.control?.disabled',

```


splincode (approved):

splincode:
@ivanobivan Could you add cherry pick to main, please

PR #13169, opened 2026-02-04
chore: stabilize input chip add

splincode:


projects/demo-playwright/utils/page-objects/input-chip.po.ts
```diff
@@ -13,9 +13,12 @@ export class TuiInputChipPO {
...
 
     public async addChip(value: string): Promise<void> {
+        const initialCount = await this.chips.count();
+
         await this.input.focus();
         await this.input.fill(value);
         await this.input.press('Enter');
-        await this.host.page().waitForTimeout(200);
+
+        await expect.poll(async () => this.chips.count()).toBeGreaterThan(initialCount);
```
nsbarsukov, line 22:
```ts
await expect(this.chips).toHaveCount(++initialCount);
```

splincode: fixed

splincode: image

PR #13170, opened 2026-02-04
feat(layout): use signals in `tui-elastic-container`

splincode:


waterplea:
Why v4? Seems to me like such improvements make sense in v5 since there's not much benefits of them to be cherry-picking them between two branches.

splincode:
Just expected height behavior between 4.x and 5.x

PR #13200, opened 2026-02-06
chore(demo): add picture in picture mode example

splincode:
image

### DIFF
--- a/projects/demo/src/pages/customization/portals/examples/3/index.html
+++ b/projects/demo/src/pages/customization/portals/examples/3/index.html
@@ -1,0 +1,27 @@
+@if (supported) {
+    <button
+        size="s"
+        tuiButton
+        type="button"
+        [disabled]="ref()"
+        (click)="openPopout()"
+    >
+        Open a PiP mode
+    </button>
+} @else {
+    <div
+        appearance="warning"
+        size="m"
+        tuiNotification
+    >
+        Document Picture-in-Picture is not supported in your browser. Check browser support
+        <a
+            href="https://developer.mozilla.org/en-US/docs/Web/API/DocumentPictureInPicture/requestWindow"
+            target="_blank"
+            tuiLink
+        >
+            here
+        </a>
+        .
+    </div>
+}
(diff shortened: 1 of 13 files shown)


projects/demo/src/pages/customization/portals/examples/3/index.html
```diff
@@ -0,0 +1,39 @@
...
+        size="s"
+        tuiButton
+        type="button"
+        [disabled]="ref()"
+        (click)="openPopout()"
+    >
+        Picture in Picture mode
+    </button>
+
+    @if (ref()) {
```
waterplea: Let's only show one button at a time, then we will not need styles file with flex and gap at all

projects/demo/src/pages/customization/portals/examples/3/popout.html
```diff
@@ -0,0 +1,26 @@
+<tui-root class="popout">
+    <div class="wrapper">
+        <tui-ring-chart
+            [value]="value"
+            [(activeItemIndex)]="activeItemIndex"
+        >
+            <span>{{ sum | tuiAmount: 'RUB' }}</span>
```
waterplea: Should we start using tuiTitle + tuiSubtitle for such cases?

projects/demo/src/pages/customization/portals/examples/3/popout.html
```diff
@@ -0,0 +1,26 @@
...
+            <div>Total</div>
+        </tui-ring-chart>
+
+        <div class="legend">
+            @for (label of labels; track label) {
```
waterplea: Let's use `track $index` unless more strict tracking is required.

waterplea (approved):

projects/demo/src/pages/customization/portals/examples/3/index.html
splincode: fixed

projects/demo/src/pages/customization/portals/examples/3/popout.html
splincode: fixed

PR #13216, opened 2026-02-09
fix(core): incorrect browser render on `tuiFade`

splincode:


projects/demo/src/pages/components/input-chip/examples/1/index.ts
```diff
@@ -13,5 +13,5 @@ import {TuiInputChip} from '@taiga-ui/kit';
 })
 export default class Example {
     protected value: string[] = inject('Pythons' as any);
-    protected readonly separator = /[\s\n,;]/;
+    protected readonly separator = /[\s,;]/;
```
waterplea, line 16: Is that intentional?

vladimirpotekhin (approved):

projects/demo/src/pages/components/input-chip/examples/1/index.ts
splincode: yes

PR #13309, opened 2026-02-18
fix(kit): close `TabsWithMore` dropdown when `activeItemIndex` is updated externally

### LINKED ISSUE #13005: 🐞 - `Tabs` does not close dropdown if application does not contains `provideAnimations()`
### Reproduction url

https://stackblitz.com/edit/tui-tabs-without-ng-animations-bug

### Reproduction
```html
<tui-tabs-with-more class="tabs" [activeItemIndex]="activeItemIndex()">
    @for (tab of tabs; track tab) {
        <button 
            *tuiItem 
            tuiTab 
            type="button" 
            (click)="activeItemIndex.set($index)"
        >
            {{ tab }}
        </button>
    }
</tui-tabs-with-more>
```
Open `More` section and click on any tab.

https://github.com/user-attachments/assets/2df18584-0c96-4fc7-8fe3-5d18999ada7d

### Problem 1
https://github.com/taiga-family/taiga-ui/blob/8b64ec1223e391be56961760d6878d3486428f09/projects/kit/components/tabs/tab.directive.ts#L33-L37

It will never dispatch `(tui-tab-activate)` for case above: `TabsWithMore` gets new `activeItemIndex` property and remove clicked tab from DOM before its parent gets bubbled click event.

### Problem 2
https://github.com/taiga-family/taiga-ui/blob/8b64ec1223e391be56961760d6878d3486428f09/projects/cdk/tokens/active-element.ts#L69-L76

**Without `provideAnimations`:**
```ts
// removedElement equals to the clicked tab
isValidFocusout(tuiGetActualTarget(event), removedElement), 
```

**With `provideAnimations`:**
```ts
// removedElement does NOT equal to the clicked tab
isValidFocusout(tuiGetActualTarget(event), removedElement), 
```

### Taiga UI version

5.0.0-rc.0

### Browser Used

Chrome

### OS Used

macOS

Chiorufarewerin:
Fixes #13005 

Fix: close `TabsWithMore` dropdown when `activeItemIndex` is updated externally (no provideAnimations())

TuiTab intentionally delays (tui-tab-activate) dispatch until the bubbled click reaches the parent (so all other click handlers run first). In a “controlled” setup (e.g. projected tab has (click)="activeItemIndex.set(i)"), the consumer updates activeItemIndex immediately. tui-tabs-with-more then re-renders synchronously and can remove/move the clicked tab from the dropdown before the parent receives the bubbled click — so tui-tab-activate is never dispatched and the internal handler that closes the dropdown is never executed. This is much harder to reproduce when animations are enabled because DOM updates may be deferred.

This PR closes the “More” dropdown whenever activeItemIndex changes.

### DIFF
--- a/projects/demo-playwright/tests/kit/tabs/tabs.pw.spec.ts
+++ b/projects/demo-playwright/tests/kit/tabs/tabs.pw.spec.ts
@@ -62,6 +62,9 @@ describe('Tabs', () => {
                 await page.locator('button:has-text("Collaborators")').nth(1).focus();
                 await page.keyboard.down('Enter');
 
+                // Flaky without this in CI (mostly under load)
+                await page.waitForTimeout(100);
+
                 // Still previous one
                 await expect(example).toContainText('Currently active: John Cleese');
                 await expect(page.locator('tui-dropdown')).toHaveCount(2);
@@ -71,10 +74,7 @@ describe('Tabs', () => {
                 await page.keyboard.down('Enter');
 
                 await expect(example).toContainText('Currently active: Neil Innes');
-                // TODO: https://github.com/taiga-family/taiga-ui/issues/13005
-                // await expect(page.locator('tui-dropdown')).not.toBeAttached();
-                await expect(page.locator('tui-dropdown')).toHaveCount(1);
-                await page.waitForTimeout(100);
+                await expect(page.locator('tui-dropdown')).not.toBeAttached();
                 await expect.soft(example).toHaveScreenshot('01-tabs-8.png');
             });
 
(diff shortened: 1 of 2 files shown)


Chiorufarewerin:
What is interesting is that: locally it works

```
npx playwright test --config projects/demo-playwright/playwright.config.ts --ignore-snapshots projects/demo-playwright/tests/kit/tabs/tabs.pw.spec.ts:27 
npm info using npm@10.9.4
npm info using node@v22.22.0

Running 1 test using 1 worker
  1 passed (3.7s)

To open last HTML report run:

  npx playwright show-report projects/demo-playwright/tests-report

npm info ok
```

Any ideas?

UPD: `waitForTimeout` seems to be helped. My assumption is the next focus+Enter can occasionally run before the overlay is fully interactive.

splincode (approved):

PR #13310, opened 2026-02-18
chore: migrate stepper and tabs selector

vladimirpotekhin:
Fixes #

### DIFF
--- a/projects/cdk/schematics/ng-update/v5/steps/constants/tags-to-replace.ts
+++ b/projects/cdk/schematics/ng-update/v5/steps/constants/tags-to-replace.ts
@@ -25,13 +25,11 @@ export const TAGS_TO_REPLACE: readonly ReplacementTag[] = [
     {
         from: 'nav',
         to: 'tui-stepper',
-        // cspell:disable-next-line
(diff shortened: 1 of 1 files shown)


projects/cdk/schematics/ng-update/v5/steps/constants/tags-to-replace.ts
```diff
@@ -21,4 +22,16 @@ export const TAGS_TO_REPLACE: readonly ReplacementTag[] = [
...
+        from: 'nav',
+        to: 'tui-stepper',
+        // cspell:disable-next-line
+        filterFn: (element) => hasElementAttribute(element, 'tuistepper'),
+    },
+    {
+        from: 'nav',
+        to: 'tui-tabs',
+        // cspell:disable-next-line
+        filterFn: (element) => hasElementAttribute(element, 'tuitabs'),
```
nsbarsukov: Is it better to disable the whole file ? 🤔 

https://github.com/taiga-family/taiga-ui/blob/419bd4099545c7fc6c473b71408641e965072f1d/.cspell.json#L4

nsbarsukov (approved):

PR #13325, opened 2026-02-18
refactor(core): use signal inputs in `CalendarYear`

nsbarsukov:


projects/cdk/utils/miscellaneous/set-signal.ts
```diff
@@ -1,8 +1,8 @@
-import {type InputSignal, type WritableSignal} from '@angular/core';
+import {type InputSignalWithTransform, type WritableSignal} from '@angular/core';
 import {SIGNAL} from '@angular/core/primitives/signals';
 
 export function tuiSetSignal<T>(
-    signal: InputSignal<T> | WritableSignal<T>,
+    signal: InputSignalWithTransform<any, T> | WritableSignal<T>,
```
nsbarsukov, line 5:
```ts
export interface InputSignal<T> extends InputSignalWithTransform<T, T> {}
```

https://github.com/angular/angular/blob/680e1c05c6442b95dcfe435be03317f862dd8793/packages/core/src/authoring/input/input_signal.ts#L106

PR #13326, opened 2026-02-18
feat(core): `DropdownA11y` add directive

waterplea:
Fixes #

### DIFF
--- a/projects/core/portals/dropdown/test/dropdown-a11y.directive.spec.ts
+++ b/projects/core/portals/dropdown/test/dropdown-a11y.directive.spec.ts
@@ -50,14 +50,18 @@ describe('TuiDropdownA11y', () => {
 
     it('input has implicit role without dropdown', () => {
         expect(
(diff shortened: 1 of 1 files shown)


projects/core/components/data-list/data-list.component.ts
```diff
@@ -102,8 +102,12 @@ export class TuiDataListComponent<T>
         }
     }
 
-    // TODO: Refactor to :has after Safari support bumped to 15
     public ngAfterContentChecked(): void {
+        if (this.el.parentElement?.closest('[role="listbox"]')) {
+            this.el.removeAttribute('role');
+        }
+
```
splincode: what is the difference if used

```ts
'[attr.role]': 'el.parentElement?.closest('[role="listbox"]') ? null : "listbox"'
```

waterplea: Somebody can override role from "listbox" to "menu", we do not want to erase that with "listbox" back again on change detection cycle.

PR #13380, opened 2026-02-27
chore(schematics): migrate `tui-axes` attributes

Chiorufarewerin:
Part of: #11917

axes verticalLines / horisontalLines => +1 (include X and Y), [axisX], [axisY] - delete inputs

### DIFF
--- a/projects/cdk/schematics/ng-update/v5/steps/templates/migrate-axes.ts
+++ b/projects/cdk/schematics/ng-update/v5/steps/templates/migrate-axes.ts
@@ -1,0 +1,60 @@
+import {type UpdateRecorder} from '@angular-devkit/schematics';
+import {type DevkitFileSystem} from 'ng-morph';
+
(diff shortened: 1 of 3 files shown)


projects/cdk/schematics/ng-update/v5/steps/templates/migrate-axes.ts
```diff
@@ -0,0 +1,58 @@
...
+
+interface Input {
+    fileSystem: DevkitFileSystem;
+    recorder: UpdateRecorder;
+    resource: TemplateResource;
+}
+
+// cspell:disable-next-line
+const REMOVE_ATTRIBUTES = ['axisx', 'axisy', '[axisx]', '[axisy]'];
+// cspell:disable-next-line
```
vladimirpotekhin: Let's use case-sensitive names here, remove disabling cspell comments , and just use toLowerCase

PR #13442, opened 2026-03-06
fix(kit): numeric names causes badge visibility with same number (#13388)

### LINKED ISSUE #13388: 🐞 - TuiFilter - numeric names causes badge visibility with same number
### Reproduction url

https://stackblitz.com/edit/angular-kjvw9ybq-gpzo9hmx?file=src%2Fapp%2Fapp.component.ts

### Description

- set numeric names in filter items
- badge is visible and duplicate filter item name

### Taiga UI version

latest

### Browser Used

Chrome

### OS Used

macOS

Ig0rABC:
Fixes #13388

### DIFF
--- a/projects/kit/components/filter/filter.component.ts
+++ b/projects/kit/components/filter/filter.component.ts
@@ -4,6 +4,7 @@ import {TuiControl} from '@taiga-ui/cdk/classes';
 import {TUI_STRINGIFY} from '@taiga-ui/cdk/constants';
 import {tuiFallbackValueProvider} from '@taiga-ui/cdk/tokens';
 import {type TuiHandler} from '@taiga-ui/cdk/types';
+import {tuiIsNumber, tuiIsString} from '@taiga-ui/cdk/utils/miscellaneous';
 import {
     TUI_ITEMS_HANDLERS,
     TuiWithItemsHandlers,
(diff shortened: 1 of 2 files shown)


waterplea:
Maybe just check if item !== handler(item)?

Ig0rABC:
> Maybe just check if item !== handler(item)?

Thats not covering cases of string containing numbers: `'10' !== Number('10')`

waterplea:
> > Maybe just check if item !== handler(item)?
> 
> Thats not covering cases of string containing numbers: `'10' !== Number('10')`

Ok. Adding a static method seems like an overkill to me, I would have just used it as default value for the input or do `==` to check.

Ig0rABC:
> > > Maybe just check if item !== handler(item)?
> > 
> > 
> > Thats not covering cases of string containing numbers: `'10' !== Number('10')`
> 
> Ok. Adding a static method seems like an overkill to me, I would have just used it as default value for the input or do `==` to check.

You're right in some extent but non-strict comparison don't display badges for objects because converting different type values to numbers: `Number(obj) == obj` -> `obj.valueOf() === obj.valueOf()`. Thats I converts bother to String at last commit for remove static method

I found static method approach here - https://github.com/taiga-family/taiga-ui/blob/main/projects/kit/components/tree/directives/tree-children.directive.ts#L10

waterplea:
We use static there because in another place we need to fallback to it so it has to be stored somewhere. Even then not sure why we did it that way instead of using a const. I think here it is fine to just store it as default value in `input(...)`

Ig0rABC:
> We use static there because in another place we need to fallback to it so it has to be stored somewhere. Even then not sure why we did it that way instead of using a const. I think here it is fine to just store it as default value in `input(...)`

Hi! Did as you say. Approve?

splincode:
@Ig0rABC Could you add cherry pick to 4.x please?

Ig0rABC:
@splincode #13636

PR #13443, opened 2026-03-07
fix(kit): incorrect positioning of the hint

### LINKED ISSUE #13053: 🐞 - Incorrect positioning of the hint
### Reproduction url

components/copy

### Description

https://github.com/user-attachments/assets/17ef629d-0c14-4701-8ce6-9fdea056e8fd

### Taiga UI version

latest

### Browser Used

Chrome

### OS Used

macOS

diwakar-2112:
## Description

Fix incorrect positioning of the hint component.

## Issue

Fixes #13053  

## Changes

- Corrected hint positioning logic
- Ensured proper alignment in the component layout

## Testing

- Verified behavior in the demo application
- Ran lint and type checks successfully

### DIFF
--- a/projects/kit/components/copy/copy.component.ts
+++ b/projects/kit/components/copy/copy.component.ts
@@ -1,6 +1,10 @@
 import {ClipboardModule} from '@angular/cdk/clipboard';
 import {ChangeDetectionStrategy, Component, inject, input} from '@angular/core';
 import {toSignal} from '@angular/core/rxjs-interop';
+import {
+    WaIntersectionObservee,
+    WaIntersectionObserverDirective,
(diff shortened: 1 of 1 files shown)


waterplea:
This should be fixed in the `Hint`. Hints should not stay visible if host disappears. You can reproduce the same issue with just regular hint:
https://taiga-ui.dev/directives/hint
Use keyboard to navigate on the HTML tab while pointer is on the element showing hint. Once you navigate to HTML hint will first move to the corner and then disappear. Because mouse is no longer on the element, not because the element is gone. It should disappear immediately when host is gone from DOM.

waterplea:
Closing this a per my previous comment

PR #13446, opened 2026-03-08
fix(core): prevent dropdown focus break with conditional @if/@else inputs

### LINKED ISSUE #13445: 🐞 - TuiTextfield: dropdown focus breaks when inputs inside the `TextField` are wrapped in @if/@else
### Reproduction url

https://stackblitz.com/edit/angular-rj9zxlhj?file=src%2Fapp%2Fmulti-select-input%2Fmulti-select-input.ts

### Description

`textfield` contains two inputs wrapped in `@if/@else`. If both fields are present in the same container (toggled with `@if`), the dropdown behavior breaks. If only one input is present, everything works correctly.

**Expected:**
Dropdown behavior should work the same with conditional input rendering.

**Actual:**
Dropdown behavior breaks when using conditional rendering for two inputs.

**Possible cause:**
Some `contentChild` queries in the `textfield` require `{ descendants: true }` for this case.

### Taiga UI version

4.72.0

### Browser Used

Chrome

### OS Used

macOS

nullawhale:
Fixes #13445

### DIFF
--- a/projects/demo/src/modules/components/input-chip/examples/4/index.html
+++ b/projects/demo/src/modules/components/input-chip/examples/4/index.html
@@ -69,45 +69,43 @@
     </tui-textfield>
 </label>
 
-<div>
-    <label tuiLabel>
-        Conditional input in textfield
-        <tui-textfield
-            multi
-            tuiChevron
-        >
-            <label tuiLabel>Multi Select with if/else</label>
-            <input
-                *ngIf="filter; else readonly"
-                tuiInputChip
-                [placeholder]="conditionalMulti.length ? '' : 'Pick from the list'"
-                [(ngModel)]="conditionalMulti"
-            />
-            <ng-template #readonly>
-                <input
-                    tuiInputChip
-                    tuiSelectLike
-                    [placeholder]="conditionalMulti.length ? '' : 'Pick from the list'"
-                    [(ngModel)]="multi"
-                />
-            </ng-template>
-            <tui-data-list-wrapper
-                *tuiTextfieldDropdown
-                new
-                tuiMultiSelectGroup
-                [items]="items"
-            />
-        </tui-textfield>
-    </label>
-    <label tuiLabel>
+<label tuiLabel>
+    Conditional input in textfield
+    <tui-textfield
+        multi
+        tuiChevron
+    >
+        <label tuiLabel>Multi Select with if/else</label>
         <input
-            tuiCheckbox
-            type="checkbox"
-            [(ngModel)]="filter"
+            *ngIf="filter; else readonly"
+            tuiInputChip
+            [placeholder]="conditionalMulti.length ? '' : 'Pick from the list'"
+            [(ngModel)]="conditionalMulti"
         />
-        Toggle filter
-    </label>
-</div>
+        <ng-template #readonly>
+            <input
+                tuiInputChip
+                tuiSelectLike
+                [placeholder]="conditionalMulti.length ? '' : 'Pick from the list'"
+                [(ngModel)]="multi"
+            />
+        </ng-template>
+        <tui-data-list-wrapper
+            *tuiTextfieldDropdown
+            new
+            tuiMultiSelectGroup
+            [items]="items"
+        />
+    </tui-textfield>
+</label>
+<label tuiLabel>
+    <input
+        tuiCheckbox
+        type="checkbox"
+        [(ngModel)]="filter"
+    />
+    Toggle filter
+</label>
 
 <label tuiLabel>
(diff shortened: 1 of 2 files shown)


projects/core/components/textfield/textfield.component.ts
```diff
@@ -79,7 +79,10 @@ export class TuiTextfieldBaseComponent<T>
     protected readonly auxiliaryQuery: QueryList<object> = EMPTY_QUERY;
 
     // TODO: Added just to avoid breaking anything until we refactor to signal queries
-    @ContentChild(forwardRef(() => TuiTextfieldBase), {read: ElementRef})
+    @ContentChild(forwardRef(() => TuiTextfieldBase), {
```
splincode, line 82: Hello, could you add a unit test or an e2e test (Cypress or Playwright) to prevent this issue in the future?

projects/core/components/textfield/textfield.component.ts
```diff
@@ -109,16 +112,16 @@ export class TuiTextfieldBaseComponent<T>
...
     public readonly control?: NgControl;
 
-    @ContentChild(TuiControl)
+    @ContentChild(TuiControl, {descendants: true})
     public readonly cva?: TuiControl<unknown>;
 
     // TODO: Replace with signal query when Angular is updated v5
     @ContentChild(forwardRef(() => TuiTextfieldBase), {
         read: ElementRef,
-        static: true,
```
waterplea, line 121: I have a feeling this was here for a reason 🤔

nullawhale: I couldn’t find a clear reason for using static, but it directly affected to search for input inside the @if wrapper

projects/core/components/textfield/textfield.component.ts
nullawhale: Could you please tell me what the error might be in "E2E testing / E2E result (pull_request)"?
Locally test passes successful, but here it fails for some unclear reason

projects/core/components/textfield/textfield.component.ts
waterplea: Without it there will be ExpressionChange error in v4. In main there's already a signal so that's ok, but here we probably need to add another non-static prop for the same thing to use it when focusing so that it fixes your issue and keeps the change detection working as expected:
https://github.com/taiga-family/taiga-ui/pull/8809

nullawhale: I restored `static: true` for the `input` and used `_input` with `descendants: true` for focus logic

projects/demo/src/modules/components/input-chip/examples/4/index.html
```diff
@@ -69,6 +69,50 @@
     </tui-textfield>
 </label>
 
+<div
+    *ngIf="isE2E"
```
waterplea: We shouldn't add examples just for tests. It's ok to show it on demo as an example of conditionally switching between editable and select-like behavior.

nullawhale: Made the example more beautiful and removed the check

projects/demo/src/modules/components/input-chip/examples/4/index.html
```diff
@@ -69,6 +69,46 @@
     </tui-textfield>
 </label>
 
+<div>
```
waterplea: nitpicking: unnecessary `div`?

projects/demo/src/modules/components/input-chip/examples/4/index.less
```diff
@@ -4,3 +4,9 @@
     flex-direction: column;
     gap: 1rem;
 }
+
+.conditional-input-row {
```
waterplea: Is this used somewhere?

projects/demo/src/modules/components/input-chip/examples/4/index.html
nullawhale: It was a visual wrapper. Removed

projects/demo/src/modules/components/input-chip/examples/4/index.less
nullawhale: Nowhere. Forgot to remove

nullawhale:
Looks like I ruined the branch with a rebase. Should I create a new branch? And should I squash all my commits?

waterplea:
@nullawhale up to you, commits are squashed when merged anyway, so whatever is easier for you

splincode:
@nullawhale Hello, could you add cherry pick to main, please

PR #13467, opened 2026-03-11
[WIP] refactor: replace `.fullsize()` mixin with `position` + `inset` properties

nsbarsukov:
https://github.com/taiga-family/taiga-ui/blob/b8e170393fabdfde82aae09e73b3f6a183dba93a/projects/styles/mixins/miscellaneous.less#L30-L31


waterplea:
When there is no inset mode it used to have width/height, not bottom/right, so we gotta be extra careful in those cases.

PR #13497, opened 2026-03-13
feat(kit): `InputDateMulti` add ability to use valueTransformer

### LINKED ISSUE #12348: 🚀 - `InputDateMulti` add ability to use valueTransformer
### Description

It needs to use a transformer from a InputDate, probably just in .map

nullawhale:
Fixes #12348

### DIFF
--- a/projects/addon-mobile/styles/common/card-large.less
+++ b/projects/addon-mobile/styles/common/card-large.less
@@ -10,7 +10,16 @@
         --t-comp: -0.125rem;
     }
 
-    [tuiAccessories] > [tuiButtonClose] {
-        --t-size: 1.5rem;
+    [tuiHeader] {
+        margin-block-end: -0.5rem;
+
+        & + [tuiCell] {
+            margin-block-start: 0;
+        }
+    }
+
+    [tuiHeader] [tuiTitle],
+    &[tuiHeader] [tuiTitle] {
+        gap: 0.5rem;
     }
 }
--- a/projects/addon-mobile/styles/common/header.less
+++ b/projects/addon-mobile/styles/common/header.less
@@ -19,6 +19,10 @@
 
     [tuiAccessories] {
         gap: 1rem;
+
+        [tuiLink] {
+            font: var(--tui-font-text-l);
+        }
     }
 
     &[data-size='xxs'],
--- a/projects/addon-mobile/styles/common/notification.less
+++ b/projects/addon-mobile/styles/common/notification.less
@@ -1,0 +1,30 @@
+tui-notification,
+[tuiNotification] {
+    &[data-size='l'],
+    &[data-size='l'] [tuiSubtitle] {
+        font: var(--tui-typography-ui-l);
+
+        &[data-icon-start] {
+            --t-start: 2.25rem;
+        }
+
+        &::before {
+            inset-inline-start: -1.25rem;
+        }
+    }
+
+    &[data-size='m'],
+    &[data-size='m'] [tuiSubtitle] {
+        font: var(--tui-typography-ui-m);
+
+        &[data-icon-start] {
+            --t-start: 2.25rem;
+        }
+
+        &::before {
+            font-size: 1.5rem;
+            inset-block-start: 0.625rem;
+            inset-inline-start: -1.5rem;
+        }
+    }
+}
--- a/projects/addon-mobile/styles/taiga-ui-mobile.less
+++ b/projects/addon-mobile/styles/taiga-ui-mobile.less
@@ -44,6 +44,7 @@
     @import 'common/segmented.less';
     @import 'common/header.less';
     @import 'common/message.less';
+    @import 'common/notification.less';
     @import 'common/notification-middle.less';
     @import 'common/toast.less';
 
--- a/projects/core/components/scrollbar/scrollbar.style.less
+++ b/projects/core/components/scrollbar/scrollbar.style.less
@@ -17,7 +17,7 @@
         opacity: 0;
     }
 
-    &:hover .t-hover-mode {
+    &:hover > .t-hover-mode {
         .transition(opacity);
 
         opacity: 1;
--- a/projects/core/directives/surface/surface.style.less
+++ b/projects/core/directives/surface/surface.style.less
@@ -14,10 +14,13 @@
     isolation: isolate;
     appearance: none;
     border: 0;
-    font-size: inherit;
-    line-height: inherit;
     text-decoration: none;
     transition-property: backdrop-filter, background, border-radius, box-shadow, mask, transform !important;
+
+    &:not([tuiCardLarge]) {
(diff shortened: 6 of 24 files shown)


projects/kit/components/input-date-multi/input-date-multi.options.ts
```diff
@@ -0,0 +1,62 @@
...
+                        return transformed.every((item): item is TuiDay => item !== null)
+                            ? transformed
+                            : null;
+                    },
+                    toControlValue: (value: TuiDay[] | null): unknown => {
+                        if (!value) {
+                            return null;
+                        }
+
+                        return value.map((item) =>
```
waterplea: Can be a ternary and also it should return empty array for `null`

projects/kit/components/input-date-multi/input-date-multi.options.ts
```diff
@@ -0,0 +1,62 @@
...
+export const TUI_INPUT_DATE_MULTI_OPTIONS = new InjectionToken<TuiInputDateMultiOptions>(
+    ngDevMode ? 'TUI_INPUT_DATE_MULTI_OPTIONS' : '',
+    {
+        factory: () => {
+            const options = inject(TUI_INPUT_DATE_OPTIONS_NEW);
+
+            return {
+                ...options,
+                valueTransformer: {
+                    fromControlValue: (value: unknown): TuiDay[] | null => {
```
waterplea: This control works with arrays, it should not have `null` internally. `null` in form control translates to empty array internally.

projects/kit/components/input-date-multi/input-date-multi.options.ts
nullawhale: Fixed

projects/kit/components/input-date-multi/input-date-multi.options.ts
nullawhale: Fixed

projects/kit/components/input-date-multi/input-date-multi.options.ts
```diff
@@ -0,0 +1,59 @@
...
```
waterplea: The types still include `null` across this file

projects/kit/components/input-date-multi/input-date-multi.options.ts
```diff
@@ -0,0 +1,59 @@
...
+                        if (!Array.isArray(value)) {
```
waterplea:
```ts
return Array.isArray(value)
  ? value.map(...).filter(Boolean)
  : []
```
Maybe just this? I'm not sure it makes sense to invalidate entire array and make function more complex than it needs to be just to satisfy types.

projects/kit/components/input-date-multi/input-date-multi.options.ts
```diff
@@ -0,0 +1,59 @@
...
+                        const transformed = value.map((item) =>
+                            options.valueTransformer.fromControlValue(item),
+                        );
+
+                        return transformed.every((item): item is TuiDay => item !== null)
+                            ? transformed
+                            : [];
+                    },
+                    toControlValue: (value: TuiDay[] | null): unknown =>
+                        value
```
waterplea: value cannot be falsy here

projects/kit/components/input-date-multi/input-date-multi.directive.ts
```diff
@@ -129,7 +138,24 @@ export class TuiInputDateMultiDirective extends TuiInputChipBaseDirective<TuiDay
         this.scrollTo();
     }
 
+    private getDay(item: unknown): TuiDay | null {
```
waterplea: Why do we need this? What is the case when `stringify` gets called not with `TuiDay`?

nullawhale: `stringify` is also used when rendering textfield chips, where items come from raw model value.  
With custom value transformers, chip items are not `TuiDay`, and `getDay` transform them

projects/kit/components/input-date-multi/input-date-multi.options.ts
nullawhale: Fixed

projects/kit/components/input-date-multi/input-date-multi.options.ts
nullawhale: Fixed

projects/kit/components/input-date-multi/input-date-multi.options.ts
nullawhale: Fixed

projects/kit/components/input-date-multi/input-date-multi.directive.ts
waterplea: Ok, good catch then. Probably we should simplify it to:
```ts
private getDate(item: unknown): TuiDay | null | undefined {
  return this.options.valueTransformer.fromControlValue([item])[0];
}
```
Maybe we do not even need a dedicated function for that and keep it in the stringify.

projects/kit/components/input-date-multi/input-date-multi.directive.ts
```diff
@@ -129,7 +138,24 @@ export class TuiInputDateMultiDirective extends TuiInputChipBaseDirective<TuiDay
...
+        }
+
+        const transformed = this.dateOption.valueTransformer.fromControlValue(item);
+
+        return transformed instanceof TuiDay ? transformed : null;
+    }
+
     private updateValue(day: TuiDay): void {
-        this.setValue(tuiArrayToggle(this.value(), day, (a, b) => a.daySame(b)));
+        this.setValue(tuiArrayToggle(this.value() ?? [], day, (a, b) => a.daySame(b)));
```
waterplea: Unnecessary fallback, it's never falsy

projects/kit/components/input-date-multi/input-date-multi.directive.ts
nullawhale: I think I implemented inherit InputDate provider incorrectly. I use InputDate transform from getDay, but it should be used inside tuiInputDateMultiOptionsProvider. Got it.

projects/kit/components/input-date-multi/input-date-multi.directive.ts
nullawhale: Fixed

projects/kit/components/input-date-multi/input-date-multi.directive.ts
nullawhale: Fixed. And moved logic from dedicated function `getDay` to `stringify`

waterplea (approved):
LGTM now, thank you for the contribution!

nullawhale:
Thank you

splincode:
@vladimirpotekhin @nsbarsukov please, review

PR #13516, opened 2026-03-14
chore(cdk): add TuiIsoToCountryCodePipe → getCountryCallingCode migration for v5

mdlufy:
Part of #11917

## Summary

- Add `TuiIsoToCountryCodePipe` (`@taiga-ui/legacy`) → `getCountryCallingCode` (`libphonenumber-js/core`) to identifier replacements

**Note:** only the TS identifier is replaced. The pipe name `tuiIsoToCountryCode` used in templates is not renamed automatically — update it manually to call `getCountryCallingCode` as a function.

### DIFF
--- a/projects/cdk/schematics/ng-update/v5/tests/__snapshots__/schematic-migrate-iso-to-country-code-pipe.spec.ts.snap
+++ b/projects/cdk/schematics/ng-update/v5/tests/__snapshots__/schematic-migrate-iso-to-country-code-pipe.spec.ts.snap
@@ -12,13 +12,13 @@ exports[`ng-update migrate TuiIsoToCountryCodePipe adds TODO comment for TuiIsoT
                 })
                 export class TestComponent {}
             ",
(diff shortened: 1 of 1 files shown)


projects/cdk/schematics/ng-update/v5/steps/constants/identifiers-to-replace.ts
```diff
@@ -675,4 +675,14 @@ export const IDENTIFIERS_TO_REPLACE: ReplacementIdentifierMulti[] = [
             },
         ],
     },
+    {
+        from: {
+            name: 'TuiIsoToCountryCodePipe',
+            moduleSpecifier: '@taiga-ui/legacy',
+        },
+        to: {
+            name: 'getCountryCallingCode',
```
nsbarsukov: `getCountryCallingCode` is just a function, not angular pipe.
Invalid migration

PR #13521, opened 2026-03-15
chore(cdk): add `TuiStatus` migration for v5

mdlufy:
Partially solved #11917 
`TuiStatus` add migration-warning


projects/cdk/schematics/ng-update/v5/tests/__snapshots__/schematic-migrate-identifiers.spec.ts.snap
```diff
@@ -74,6 +74,30 @@ exports[`ng-update identifiers migration moves TuiFlagPipe from core to kit: tes
 }
 `;
 
+exports[`ng-update identifiers migration moves TuiStatus from legacy to kit: test.ts 1`] = `
+{
+  "0. Before": "
+                import {TuiStatus} from '@taiga-ui/legacy';
```
vladimirpotekhin: Wrong migration. In version 4, the legacy package contains the TuiStatus type. We can't simply replace the type with a directive.

mdlufy: fixed

PR #13545, opened 2026-03-17
fix(kit): `FloatingContainer` fix jank when address bar showing in Firefox mobile

mdlufy:
Fixes #

### DIFF
--- a/projects/cdk/directives/visual-viewport/visual-viewport.directive.ts
+++ b/projects/cdk/directives/visual-viewport/visual-viewport.directive.ts
@@ -25,6 +25,9 @@ export class TuiVisualViewport {
             this.style.setProperty('--tui-viewport-scale', String(scale));
             this.style.setProperty('--tui-viewport-vh', tuiPx(this.w.innerHeight / 100));
             this.style.setProperty('--tui-viewport-vw', tuiPx(this.w.innerWidth / 100));
-            this.style.setProperty('--tui-viewport-svh', tuiPx(this.minInnerHeight));
+            this.style.setProperty(
+                '--tui-viewport-svh',
+                tuiPx(this.minInnerHeight / 100),
+            );
         });
 }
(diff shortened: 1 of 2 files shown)


projects/cdk/directives/visual-viewport/visual-viewport.directive.ts
```diff
@@ -11,16 +11,20 @@ import {tuiInjectElement, tuiPx} from '@taiga-ui/cdk/utils';
...
+            this.minInnerHeight = Math.min(this.minInnerHeight, this.w.innerHeight);
+
             this.style.setProperty('--tui-viewport-x', tuiPx(offsetLeft));
             this.style.setProperty('--tui-viewport-y', tuiPx(offsetTop));
             this.style.setProperty('--tui-viewport-height', tuiPx(height));
             this.style.setProperty('--tui-viewport-width', tuiPx(width));
             this.style.setProperty('--tui-viewport-scale', String(scale));
             this.style.setProperty('--tui-viewport-vh', tuiPx(this.w.innerHeight / 100));
             this.style.setProperty('--tui-viewport-vw', tuiPx(this.w.innerWidth / 100));
+            this.style.setProperty('--tui-viewport-svh', tuiPx(this.minInnerHeight));
```
waterplea: Shouldn't we add ` / 100`?

mdlufy: Now `--tui-viewport-svh` stores the full height and is used directly

If add / 100, need to update:
from `bottom: ~'max(0px, calc(var(--tui-viewport-svh) - ...))';` , 
to `bottom: ~'max(0px, calc(100 * var(--tui-viewport-svh) - ...))';`

As if it would be an extra division for no reason?

waterplea: But it would be a 1 to 1 polyfill, I think it makes sense

PR #13547, opened 2026-03-18
fix(kit): remove unnecessary blur on setDate when not in dropdown

JonataB:
Fixes #

### DIFF
--- a/projects/kit/components/input-date/input-date.directive.ts
+++ b/projects/kit/components/input-date/input-date.directive.ts
@@ -130,10 +130,6 @@ export abstract class TuiInputDateBase<
     public setDate(value: TuiDay | TuiDayRange | readonly TuiDay[]): void {
         this.onChange(value as T);
         this.open.set(false);
-
-        if (!this.el.closest('tui-dropdown')) {
-            this.el.blur();
-        }
     }
 
     protected processCalendar(calendar: TuiCalendar | TuiCalendarRange): void {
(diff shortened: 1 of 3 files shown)


projects/kit/components/input-date/input-date.directive.ts
```diff
@@ -130,10 +130,6 @@ export abstract class TuiInputDateBase<
     public setDate(value: TuiDay | TuiDayRange | readonly TuiDay[]): void {
         this.onChange(value as T);
         this.open.set(false);
-
-        if (!this.el.closest('tui-dropdown')) {
```
waterplea, line 134: We need to look into history why it was added. I think it might be needed on mobile to avoid screen keyboard when using picker or something else. It was definitely added for a reason.

JonataB: Ok i will check. 

Sorry. Never presume. My bad

waterplea:
I don't think this is the way to do it. Perhaps we can somehow just prevent focus from coming back into the input when the date is picked from the calendar?

JonataB:
> I don't think this is the way to do it. Perhaps we can somehow just prevent focus from coming back into the input when the date is picked from the calendar?

Hi @waterplea, i will check. Tks 

Instead of handling focus after you suggest to  prevent the focus from moving in the first place. If the input never loses focus we not need to do nothing.

Sorry my solution is 💩

Issue #13550, opened 2026-03-18
🐞 - `Textfield` supports RTL for `[filler]` property


nsbarsukov:
https://taiga-ui.dev/components/textfield/API?filler=HH:MM

<img width="591" height="376" alt="Image" src="https://github.com/user-attachments/assets/bc2e4e84-770f-4145-b272-b8d1c9f07e5a" />

PR #13556, opened 2026-03-19
fix(core): ensure textfield filler respects RTL mode

splincode:
Fixes https://github.com/taiga-family/taiga-ui/issues/13550

### DIFF
--- a/projects/core/components/textfield/textfield.component.ts
+++ b/projects/core/components/textfield/textfield.component.ts
@@ -105,9 +105,7 @@ export class TuiTextfieldComponent<T> implements TuiDataListHost<T> {
             return '';
         }
 
-        const input = this.input()?.nativeElement ?? this.el;
-
-        return input.matches('[dir="rtl"] :scope')
+        return this.input()?.nativeElement.matches('[dir="rtl"] :scope')
             ? filler.slice(0, filler.length - value.length) + value
             : value + filler.slice(value.length);
     });


projects/core/components/textfield/textfield.component.ts
```diff
@@ -99,9 +99,17 @@ export class TuiTextfieldComponent<T> implements TuiDataListHost<T> {
...
+        const filler = this.filler();
 
-        return filler.length > value.length ? filler : '';
+        if (filler.length <= value.length) {
+            return '';
+        }
+
+        const input = this.input()?.nativeElement ?? this.el;
+
+        return input.matches('[dir="rtl"] :scope')
```
waterplea: I think we can just use this.input()?.nativeElement.matches, it is always going to be there, we just do not use .required to avoid some race conditions that can throw.

projects/demo-playwright/tests/core/input/input-filler-rtl.pw.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+test.describe('Input filler in RTL mode', () => {
+    test('renders filler correctly in LTR and RTL', async ({page}) => {
+        await tuiGoto(page, `${DemoRoute.Input}/API?filler=HH:MM`);
+
+        const api = new TuiDocumentationApiPagePO(page);
+        const example = api.demo;
+        const input = example.locator('[tuiInput]');
+
+        await input.focus();
+        await input.pressSequentially('123');
```
nsbarsukov:
```suggestion
        await input.pressSequentially('12:3');
```

projects/demo-playwright/tests/core/input/input-filler-rtl.pw.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+        await expect.soft(example).toHaveScreenshot('input-filler-ltr.png');
+        await input.clear();
+        await example.evaluate((node) => node.setAttribute('dir', 'rtl'));
+        await input.focus();
+        await input.pressSequentially('123');
```
nsbarsukov:
```suggestion
        await input.pressSequentially('12:3');
```

projects/demo-playwright/tests/core/input/input-filler-rtl.pw.spec.ts
```diff
@@ -0,0 +1,22 @@
+import {DemoRoute} from '@demo/routes';
+import {TuiDocumentationApiPagePO, tuiGoto} from '@demo-playwright/utils';
+import {expect, test} from '@playwright/test';
+
+test.describe('Input filler in RTL mode', () => {
+    test('renders filler correctly in LTR and RTL', async ({page}) => {
+        await tuiGoto(page, `${DemoRoute.Input}/API?filler=HH:MM`);
+
+        const api = new TuiDocumentationApiPagePO(page);
+        const example = api.demo;
```
nsbarsukov:
```suggestion
        const example = new TuiDocumentationApiPagePO(page).demo;
```

projects/core/components/textfield/textfield.component.ts
splincode: fixed

PR #13557, opened 2026-03-19
fix(layout): `FloatingContainer` fix jank when address bar showing in Firefox mobile

mdlufy:
Fixes #

### DIFF
--- a/projects/layout/components/floating-container/floating-container.style.less
+++ b/projects/layout/components/floating-container/floating-container.style.less
@@ -26,17 +26,11 @@
     position: sticky;
     z-index: 1;
     display: grid;
-
-    --t-svh: ~'calc(100 * var(--tui-viewport-svh))';
-
-    inset-block-end: ~'max(0px, calc(var(--t-svh) - var(--tui-viewport-height) - var(--tui-viewport-y)))';
-
-    /* stylelint-disable plugin/no-unsupported-browser-features, plugin/use-baseline, unit-allowed-list */
-    @supports (height: 100svh) {
-        --t-svh: 100svh;
-    }
-    /* stylelint-enable plugin/no-unsupported-browser-features, plugin/use-baseline, unit-allowed-list */
-
+    // Fallback for browsers not supported svh unit
+    /* stylelint-disable-next-line */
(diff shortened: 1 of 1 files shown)


projects/cdk/directives/visual-viewport/visual-viewport.directive.ts
```diff
@@ -9,16 +9,23 @@ import {tuiPx} from '@taiga-ui/cdk/utils/miscellaneous';
 export class TuiVisualViewport {
     private readonly w = inject(WA_WINDOW);
     private readonly style = tuiInjectElement().style;
+    private minInnerHeight = Infinity;
```
splincode, line 12: why not `private minInnerHeight = this.w.innerHeight`?

projects/layout/components/floating-container/floating-container.style.less
```diff
@@ -26,7 +26,9 @@
     position: sticky;
     z-index: 1;
     display: grid;
-    inset-block-end: calc(100 * var(--tui-viewport-vh) - var(--tui-viewport-height) - var(--tui-viewport-y));
+    // Fallback for browsers not supported svh unit
+    inset-block-end: ~'max(0px, calc(100 * var(--tui-viewport-svh) - var(--tui-viewport-height) - var(--tui-viewport-y)))';
+    inset-block-end: ~'max(0px, calc(100svh - var(--tui-viewport-height) - var(--tui-viewport-y)))';
```
splincode: what do you think?

```
--t-svh: calc(100 * var(--tui-viewport-svh));

inset-block-end: ~'max(0px, calc(var(--t-svh) - var(--tui-viewport-height) - var(--tui-viewport-y)))';

@supports (height: 100svh) {
    --t-svh: 100svh;
}
```

projects/cdk/directives/visual-viewport/visual-viewport.directive.ts
mdlufy: Functionally, there is no difference — subscribing to WaViewportService is triggered immediately and the first call to Math.min(Infinity, innerHeight) will give the same result as Math.min(innerHeight, innerHeight)

Infinity is an idiom for the "tracking minimum" pattern: the algorithm does not depend on the initial value and is always correct, even if the order of initialization of fields someday changes

PR #13466, opened 2026-03-11
chore(schematics): migrations for outdated `success` & `error` appearances

nsbarsukov:
Relates:
* https://github.com/taiga-family/taiga-ui/issues/11917

PR #13558, opened 2026-03-19
chore: remove usages of glass appearance

waterplea:
Fixes #


splincode (approved):

projects/demo/src/pages/components/action-bar/examples/2/index.html
```diff
@@ -14,7 +14,7 @@
     <span [style.width.%]="isMobile() ? 100 : null">Table bar opened</span>
 
     <button
-        appearance="glass"
+        appearance="secondary-grayscale"
```
nsbarsukov, line 17: Could you also add schematic migration ?

Use these lines as reference

https://github.com/taiga-family/taiga-ui/pull/13466/changes#diff-3dd9913966b234bbcea4a38b34925db480ffa6fe19a3f0abb98443ca67f46173

PR #13560, opened 2026-03-19
fix(cdk): `Transitioned` fix the order of applying styles

vladimirpotekhin:
Fixes #


projects/cdk/directives/transitioned/transitioned.directive.ts
```diff
@@ -1,17 +1,19 @@
...
 @Directive({
     selector: '[tuiTransitioned]',
-    host: {'[style.transition]': '"none"'},
+    host: {style: 'transition: none'},
 })
 export class TuiTransitioned {
     constructor() {
         const el = tuiInjectElement();
 
         if (isPlatformBrowser(inject(PLATFORM_ID))) {
```
waterplea: You can drop this now, afterNextRender is already in browser only

PR #13564, opened 2026-03-19
refactor(addon-doc): drop optional

splincode:


projects/addon-doc/tokens/doc-icons.ts
```diff
@@ -12,10 +12,7 @@ export interface TuiDocIcons {
     readonly check: string;
     readonly link: string;
     readonly languages: string;
-    /**
-     * TODO: drop optional in v5
-     */
-    readonly shrink?: string;
+    readonly shrink: string;
     readonly expand?: string;
```
nsbarsukov: `expand` too ?

splincode: fixed

PR #13573, opened 2026-03-21
fix(kit): jittering segments and stop recomputation

JonataB:
Fixes #

### DIFF
--- a/projects/kit/components/progress/progress-bar/progress-color-segments.directive.ts
+++ b/projects/kit/components/progress/progress-bar/progress-color-segments.directive.ts
@@ -5,7 +5,7 @@ import {
     WaMutationObserverService,
 } from '@ng-web-apis/mutation-observer';
 import {tuiInjectElement} from '@taiga-ui/cdk/utils/dom';
-import {distinctUntilChanged, map, tap} from 'rxjs';
+import {distinctUntilChanged, map} from 'rxjs';
 
 @Directive({
     selector: 'progress[tuiProgressBar][tuiProgressColorSegments]',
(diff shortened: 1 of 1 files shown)


projects/kit/components/progress/progress-bar/progress-color-segments.directive.ts
```diff
@@ -5,23 +5,27 @@ import {
...
+import {distinctUntilChanged, map, tap} from 'rxjs';
 
 @Directive({
     selector: 'progress[tuiProgressBar][tuiProgressColorSegments]',
     providers: [
         WaMutationObserverService,
         {
             provide: WA_MUTATION_OBSERVER_INIT,
-            useValue: {attributeOldValue: true},
+            useValue: {attributeOldValue: true}, // used?
```
waterplea: Can it just be "attributes"?

projects/kit/components/progress/progress-bar/progress-color-segments.directive.ts
```diff
@@ -5,23 +5,27 @@ import {
...
-    host: {'[style.--tui-progress-color]': 'color()'},
+    host: {'[style.--tui-progress-color]': 'color()', '[style.--tui-duration]': '"0s"'},
 })
 export class TuiProgressColorSegments {
     private readonly el = tuiInjectElement<HTMLProgressElement>();
     private readonly position = toSignal(
-        inject(WaMutationObserverService, {self: true}).pipe(map(() => this.el.position)),
+        inject(WaMutationObserverService, {self: true}).pipe(
+            map(() => this.el.position),
+            distinctUntilChanged(),
```
waterplea, line 26: Shouldn't this already be the case with signals? I thought they do not trigger multiple times for the same value.

JonataB: Hi Alex, yes signals prevent computed() from rerun, but distinctUntilChanged guards earlier in the pipeline. Since WaMutationObserverService fires on any attribute mutation the two are complementary. But maybe i am wrong, you are the guru :)

sorry for english

projects/kit/components/progress/progress-bar/progress-color-segments.directive.ts
JonataB: obviously you are right 😅
so yes seems implicit

splincode:
Closed in favor of https://github.com/taiga-family/taiga-ui/pull/13626

PR #13574, opened 2026-03-23
chore: fix double matching error in surface example

vladimirpotekhin:


waterplea:
Already fixed it when removing glass appearance

PR #13576, opened 2026-03-23
refactor(legacy): move deleted `taiga-ui-global.less` to `@taiga-ui/legacy`

nsbarsukov:
Relates:
* https://github.com/taiga-family/taiga-ui/pull/12878


projects/cdk/schematics/ng-update/v5/tests/__snapshots__/schematic-migrate-mixins.spec.ts.snap
```diff
@@ -3,7 +3,7 @@
 exports[`ng-update v5: migrate mixins should migrate slider-ticks-labels mixin in LESS: test.less 1`] = `
 {
   "0. Before": "
-                @import '@taiga-ui/styles/utils';
+                @import '@taiga-ui/core/styles/taiga-ui-local';
```
nsbarsukov, line 6: Invalid before state.
`@taiga-ui/styles/utils` does not exists in `v4.x` branch

PR #13584, opened 2026-03-23
chore(addon-mobile): delete outdated styles in `button.less`

nsbarsukov:


projects/addon-mobile/styles/common/button.less
```diff
@@ -1,12 +1,7 @@
-[tuiCardLarge] [tuiButton] {
+[tuiCardLarge] > [tuiButton] {
     inline-size: 100%;
 }
 
-// TODO: Add > above and remove this rule in v5
```
nsbarsukov, line 5: @waterplea should we include `footer` too ?

```css
[tuiCardLarge] > [tuiButton],
[tuiCardLarge] > footer > [tuiButton] {
    inline-size: 100%;
}
```

Image

Image

Your PR title was only about header
* https://github.com/taiga-family/taiga-ui/pull/11151

waterplea: I guess so

PR #13586, opened 2026-03-23
feat(kit): update `tuiInputFiles` spec

splincode:
image

### DIFF
--- a/projects/kit/components/files/input-files/input-files.component.ts
+++ b/projects/kit/components/files/input-files/input-files.component.ts
@@ -4,12 +4,16 @@ import {
     ContentChild,
     TemplateRef,
     ViewEncapsulation,
+    inject,
+    Input,
 } from '@angular/core';
 import {type TuiContext} from '@taiga-ui/cdk/types';
 import {PolymorpheusComponent, PolymorpheusOutlet} from '@taiga-ui/polymorpheus';
+import {type TuiSizeL} from '@taiga-ui/core/types';
 
 import {TuiInputFilesContent} from './input-files.content';
 import {TuiInputFilesDirective} from './input-files.directive';
+import {TUI_INPUT_FILES_OPTIONS} from './input-files.options';
 
 @Component({
     standalone: true,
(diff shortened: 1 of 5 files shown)


projects/kit/components/files/input-files/input-files.style.less
```diff
@@ -3,14 +3,23 @@
...
     text-align: center;
-    border-radius: var(--tui-radius-m);
+    border-radius: var(--tui-radius-s);
     font: var(--tui-font-text-m);
     overflow-wrap: break-word;
-    padding: 0 0.5rem;
+    padding: 0.75rem 1rem;
+    gap: 0.5rem;
+
+    &[data-size='l'] {
```
waterplea, line 18: Font size is also different for L and M

projects/kit/components/files/input-files/input-files.component.ts
```diff
@@ -30,11 +34,13 @@ import {TuiInputFilesDirective} from './input-files.directive';
     encapsulation: ViewEncapsulation.None,
     changeDetection: ChangeDetectionStrategy.OnPush,
     host: {
+        tuiInputFiles: '',
         '(dragover.prevent.zoneless)': '0',
         '(drop.prevent)': 'onDropped($event)',
         '(dragenter)': 'onDrag($event.dataTransfer)',
         '(dragleave)': 'onDrag(null)',
         '[class._dragged]': 'fileDragged',
+        '[attr.data-size]': 'size || options.size',
```
waterplea: The default size for UI is set to be "L" in specs. Should we change the default in main and keep it `|| 'l'` in here?

projects/kit/components/files/input-files/input-files.style.less
splincode: fixed

projects/kit/components/files/input-files/input-files.component.ts
splincode: fixed

PR #13606, opened 2026-03-24
feat(core): add factory provider for tuiValidationErrorsProvider

m-chuguev:
Hey, what do you think about adding a factory to the error provider?

I used to do this in my projects:
```
providers: [{
    provide: TUI_VALIDATION_ERRORS,
    useFactory: () => ({
        required: inject(SomeService).getValue('value'),
    }),
}],
```

and I'd like to do something similar with tuiValidationErrorsProvider.

What do you think about this?


projects/cdk/utils/di/provide-factory.ts
```diff
@@ -0,0 +1,20 @@
+import {
+    type AbstractType,
+    type InjectionToken,
+    type Provider,
+    type Type,
+} from '@angular/core';
+
+export function tuiProvideFactory<T>(
```
waterplea: We have this, maybe it can be reused?
https://github.com/taiga-family/taiga-ui/blob/main/projects/cdk/utils/di/provide-options.ts

m-chuguev: Great solution.

But I'm a little confused by inheritance:
```
...inject(provide, {optional: true, skipSelf: true}) || fallback)
```

Is this okay to use in my case? Will there be breaking changes?

And what should I use as a 'fallback'?

waterplea: fallback would be empty dictionary. I think it's ok to use it, this way it would inherit error definitions from up the hierarchy if you did not override them.

m-chuguev: All done)

m-chuguev:
@waterplea How to fix tests? It doesn't look like the provider is to blame

PR #13613, opened 2026-03-25
feat(experimental): `File` add new component

splincode:


projects/experimental/components/file/file.ts
```diff
@@ -0,0 +1,42 @@
...
+        }
+    `,
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {
+        'data-tui-version': TUI_VERSION,
+        tuiFile: '',
+        '[attr.data-orientation]': 'orientation() || options.orientation',
+        '[attr.data-size]': 'size() || options.size',
+        '[tabIndex]': '0',
```
splincode: I am not sure about tabIndex

projects/experimental/components/file/file.styles.less
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    [data-platform='android'] & {
+        max-inline-size: 5.625rem;
+    }
+
+    [data-platform='ios'] &:active,
+    [data-platform='android'] &:active {
+        transform: scale(0.95);
+    }
```
splincode: Move to addon-mobile

projects/demo/src/pages/components/file/examples/2/index.html
```diff
@@ -0,0 +1,92 @@
+@for (platform of platforms; track $index) {
```
vladimirpotekhin: Since we now have a platform switcher in the demo, maybe it would be better to keep only one option in the example (without all platforms iterating) ?

projects/demo/src/pages/components/file/examples/1/index.less
```diff
@@ -0,0 +1,11 @@
+:host {
```
vladimirpotekhin: Let’s reduce space on the left in first example, it will look better on mobile.

Image

projects/experimental/components/file/file.styles.less
```diff
@@ -0,0 +1,185 @@
...
+
+    [tuiButtonX][data-appearance=''] {
+        color: var(--tui-text-secondary);
+    }
+
+    [tuiButtonX] {
+        margin: 0 1rem;
+    }
+
+    & > label > span > :first-child {
```
vladimirpotekhin: label and span seems look too general, maybe [tuiCell] is better?

projects/experimental/components/file/file.ts
```diff
@@ -0,0 +1,42 @@
...
+import {TUI_VERSION} from '@taiga-ui/cdk/constants';
+import {TuiCell} from '@taiga-ui/core/components/cell';
+
+import {TUI_FILE_OPTIONS_OPTIONS} from './file.options';
+
+@Component({
+    selector: '[tuiFile]',
+    imports: [TuiCell],
+    template: `
+        <label tuiCell>
```
vladimirpotekhin: Couldn't we make tuiCell a host directive? That would remove the extra nesting.

splincode: I did it that way initially, but then teamlead told me change it to
image

waterplea: Cell would still be a host directive, label will just be display: flex, gap: inherit. But I'll take it from here, I want to experiment with tuiSlides to accommodate all the animation transitions from the specs.

projects/demo/src/pages/components/file/examples/1/index.less
splincode: fixed

mdlufy (approved):

projects/demo/src/pages/components/file/examples/1/index.html
```diff
@@ -0,0 +1,60 @@
...
+            {{ item }}
+            <input
+                name="size"
+                type="radio"
+                [value]="item"
+                [(ngModel)]="size"
+            />
+        </label>
+    }
+</tui-segmented>
```
mdlufy, line 13: maybe see only default in this example and show different sises on API tab?

projects/demo/src/pages/components/file/examples/3/index.html
```diff
@@ -0,0 +1,29 @@
...
+    >
+        <tui-icon icon="@tui.image" />
+        <div tuiTitle>
+            <div class="clamp">A very very long file name that spans many many lines.png</div>
+            <div
+                tuiSubtitle
+                class="clamp"
+            >
+                You can use native
+                <code>line-clamp</code>
```
nsbarsukov, line 24: Image

vladimirpotekhin (approved):

nsbarsukov (approved):

PR #13615, opened 2026-03-25
feat(kit): update `tuiInputFiles` spec

splincode:


projects/kit/components/files/input-files/input-files.component.ts
```diff
@@ -29,6 +33,8 @@ import {TuiInputFilesDirective} from './input-files.directive';
     encapsulation: ViewEncapsulation.None,
     changeDetection: ChangeDetectionStrategy.OnPush,
     host: {
+        tuiInputFiles: '',
+        '[attr.data-size]': 'size() || options.size',
```
waterplea, line 43: Let's change default size to "l" so it matches other components like buttons and textifelds

splincode: fixed

PR #13634, opened 2026-03-26
feat: add version encapsulation to `cdk` and `core` 

vladimirpotekhin:
Fixes #


vladimirpotekhin:
> Hi! Reviewed the PR

No, in this proposal, two selectors increase the specificity of styles. We need to isolate them without increasing the specificity.

projects/cdk/directives/truncate/truncate.directive.ts
```diff
@@ -48,6 +49,7 @@ class Styles {}
     ],
     host: {
         tuiTruncate: '',
+        tuiTruncateV: `${TUI_VERSION}`,
```
splincode: `tuiTruncateV: TUI_VERSION,` ?

projects/core/styles/components/notification.less
```diff
@@ -22,7 +22,7 @@
  * Icons, Appearance
  */
 tui-notification,
```
splincode, line 24: `tui-notification[tuiNotificationV='@{tui-version}'],
[tuiNotificationV='@{tui-version}'] {` ?

projects/core/components/button/set-attributes.ts
```diff
@@ -0,0 +1,12 @@
+import {TUI_VERSION} from '@taiga-ui/cdk/constants';
+import {tuiInjectElement} from '@taiga-ui/cdk/utils/dom';
+
+export function tuiSetAttributes(selectors: string[]): undefined {
```
splincode:
```ts
export function tuiSetVersionedSelector(selectors: readonly string[]): undefined {

    for (const selector of selectors) {
        if (el.hasAttribute(selector)) {
            el.setAttribute(`${selector}V`, TUI_VERSION);
        }
    }

    return undefined;
}
```

splincode: what do you think?

projects/core/styles/components/notification.less
vladimirpotekhin: No, this will increase specificity. There is no `tui-notification` selector in 5.0, so isolation is not needed.

waterplea (approved):

projects/core/components/button/button.directive.ts
```diff
@@ -34,12 +34,15 @@ class TuiButtonStyles {}
     hostDirectives: [TuiWithAppearance, TuiWithIcons],
     host: {
         '[attr.data-size]': 'size',
+        '[attr.tuiButtonV]': "el.hasAttribute('tuibutton') ? version : null",
```
splincode: Image

`el.hasAttribute('tuiButton')`

projects/core/components/button/button.directive.ts
```diff
@@ -34,12 +34,15 @@ class TuiButtonStyles {}
...
+        '[attr.tuiIconButtonV]': "el.hasAttribute('tuiiconbutton') ? version : null",
```
splincode: `el.hasAttribute('tuiIconButton')`

projects/core/components/button/button.directive.ts
waterplea: I think Angular lowecases it.
image

splincode: In standard HTML documents, attribute names are case-insensitive.

projects/core/components/button/button.directive.ts
splincode: In standard HTML documents, attribute names are case-insensitive.

splincode (approved):

PR #13641, opened 2026-03-27
feat(i18n): add `addon-doc` translations to `@taiga-ui/i18n` package

splincode:


projects/addon-doc/components/example/example.component.ts
```diff
@@ -103,7 +104,7 @@ export class TuiDocExample implements OnChanges {
         );
 
     protected readonly route = inject(ActivatedRoute);
-    protected readonly defaultTab = this.texts[0];
+    protected readonly defaultTab = computed(() => this.texts());
```
nsbarsukov: Excessive `computed` ?

PR #13736, opened 2026-04-03
fix(addon-doc): `DocExample` has hydration problems

### LINKED ISSUE #13732: 🐞 - `DocExample` has hydration problems
### Reproduction url

https://taiga-ui.dev/components/select

### Description

Enable hydration and run 
```
nx serve demo -c production
```

Go to http://localhost:3333/components/select

```
ERROR RuntimeError: NG0500: During hydration Angular expected a comment node but found <p>.

Angular expected this DOM:

<hgroup _ngcontent-ng-c4199146294="" tuititle="" data-tui-version="5.0.0" ng-reflect-tui-title="">
  …
  <h3>…</h3>
  <!-- container -->  <-- AT THIS LOCATION
  …
</hgroup>

Actual DOM is:

<hgroup _ngcontent-ng-c4199146294="" tuititle="" data-tui-version="5.0.0" ng-reflect-tui-title="">
  …
  <p _ngcontent-ng-c4199146294="">…</p>
  <p>…</p>  <-- AT THIS LOCATION
  …
</hgroup>

Note: attributes are only displayed to better represent the DOM but have no effect on hydration mismatches.

To fix this problem:
  * check the "_TuiDocExample" component for hydration-related issues
  * check to see if your template has valid HTML structure
  * or skip hydration by adding the `ngSkipHydration` attribute to its host node in a template

 Find more at https://angular.dev/errors/NG0500
    at validateMatchingNode (core.mjs:14674:11)
    at locateOrCreateContainerAnchorImpl (core.mjs:20171:5)
    at declareTemplate (core.mjs:20081:19)
    at Module.ɵɵtemplate (core.mjs:20124:3)
    at TuiDocExample_Conditional_0_Template (example.template.html:19:13)
    at executeTemplate (core.mjs:12141:5)
    at renderView (core.mjs:12627:7)
    at createAndRenderEmbeddedLView (core.mjs:12694:5)
    at Module.ɵɵconditional (core.mjs:26922:31)
    at TuiDocExample_Template (example.template.html:1:1)
```

### Taiga UI version

5.0.0

### Browser Used

Chrome

### OS Used

macOS

nsbarsukov:
Fixes #13732

```
ERROR RuntimeError: NG0500: During hydration Angular expected a comment node but found <p>.

Angular expected this DOM:

<hgroup _ngcontent-ng-c4199146294="" tuititle="" data-tui-version="5.0.0" ng-reflect-tui-title="">
  …
  <h3>…</h3>
  <!-- container -->  <-- AT THIS LOCATION
  …
</hgroup>

Actual DOM is:

<hgroup _ngcontent-ng-c4199146294="" tuititle="" data-tui-version="5.0.0" ng-reflect-tui-title="">
  …
  <p _ngcontent-ng-c4199146294="">…</p>
  <p>…</p>  <-- AT THIS LOCATION
  …
</hgroup>

Note: attributes are only displayed to better represent the DOM but have no effect on hydration mismatches.

To fix this problem:
  * check the "_TuiDocExample" component for hydration-related issues
  * check to see if your template has valid HTML structure <--------------------------------- [our case]
  * or skip hydration by adding the `ngSkipHydration` attribute to its host node in a template
```

The root cause was invalid HTML nesting in
https://github.com/taiga-family/taiga-ui/blob/671f85792b7807489ddd2c06e7810feccd865cb7/projects/addon-doc/components/example/example.template.html#L20-L21

The `<p>` tag wrapped a `polymorpheusOutlet` that could render block-level elements (like `<div tuiNotification>`). Per HTML spec, `<p>` cannot contain block-level elements — the browser auto-closes the `<p>` when it encounters one, restructuring the DOM. This caused the server-rendered DOM to differ from the browser-parsed DOM, triggering Angular's NG0500 hydration mismatch error.

### DIFF
--- a/projects/addon-doc/components/example/example.style.less
+++ b/projects/addon-doc/components/example/example.style.less
@@ -43,10 +43,6 @@
     }
 }
 
-.t-description {
-    margin-block-start: 0.5rem;
-}
-
 .t-example {
     position: relative;
     border: 0.25rem solid var(--tui-background-base-alt);
--- a/projects/addon-doc/components/example/example.template.html
+++ b/projects/addon-doc/components/example/example.template.html
@@ -20,7 +20,7 @@
     </header>
 
     @if (description()) {
-        <div class="t-description">
+        <div tuiDescription>
             <ng-container *polymorpheusOutlet="description() as text">
                 {{ text }}
             </ng-container>


projects/addon-doc/components/example/example.style.less
```diff
@@ -43,6 +43,11 @@
     }
 }
 
+.t-description:not(:empty) {
+    margin: 0.25rem 0 0;
+    font: var(--tui-typography-body-m);
```
nsbarsukov: Or is it better to extends this selector with `<div>` too ? 🤔 

https://github.com/taiga-family/taiga-ui/blob/5b7867bec27aa084042f235fd5e90f0aba60b30e/projects/layout/components/header/header.styles.less#L13-L20

projects/addon-doc/components/example/example.template.html
```diff
@@ -17,11 +17,11 @@ <h3>
                 </a>
             </h3>
             @if (description()) {
-                <p>
+                <div class="t-description">
```
waterplea: `div` is technically invalid inside `hgroup`, that's why it was p for semantic HTML:
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/hgroup#technical_summary

We need to move it past `header`. Or we can drop all the divs from descriptions, I don't think we ever need them.

nsbarsukov: > Or we can drop all the divs from descriptions, I don't think we ever need them.

Multi-paragraph example description is common pattern inside our documentation.
Yep, we can make all break lines `<br />`. However, I hope, it is not the best option for us.

Let's use your another approach
> We need to move it past header.

projects/addon-doc/components/example/example.style.less
```diff
@@ -43,6 +43,10 @@
     }
 }
 
+.t-description:not(:empty) {
```
waterplea: Do we even need that? If it's empty it is not rendered under @if

nsbarsukov: My mistake, thanks

PR #13742, opened 2026-04-03
chore(kit): `Comment` add host attribute

mdlufy:
Fixes #


projects/demo/src/pages/directives/comment/index.html
```diff
@@ -1,7 +1,7 @@
 <tui-doc-page
     header="Comment"
     package="KIT"
-    path="kit/components/comment"
+    path="kit/components/comment/comment.directive.ts"
     type="directives"
```
waterplea: We should just move it to components folder in the demo and change type to components, then custom path should not be necessary

PR #13756, opened 2026-04-06
feat(kit): support disabled state for `tui-segmented`

splincode:
Fixes #13769

image

### DIFF
--- a/projects/kit/components/segmented/segmented.component.ts
+++ b/projects/kit/components/segmented/segmented.component.ts
@@ -1,7 +1,6 @@
 import {
     ChangeDetectionStrategy,
     Component,
-    effect,
     inject,
     input,
     model,
--- a/projects/kit/components/segmented/segmented.component.ts
+++ b/projects/kit/components/segmented/segmented.component.ts
@@ -1,6 +1,7 @@
 import {
     ChangeDetectionStrategy,
     Component,
+    effect,
     inject,
     input,
     model,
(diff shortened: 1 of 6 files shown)


projects/kit/components/segmented/segmented.style.less
```diff
@@ -48,13 +48,24 @@ tui-segmented:where(*&) {
         }
     }
 
+    &._disabled {
```
waterplea: Let's also add inert so that inputs and buttons are not keyboard focusable:
https://caniuse.com/wf-inert
I think pointer-events is good enough for older browsers

splincode: updated

waterplea: You forgot to apply the attribute. Also no need for ._disabled class, if we add attribute.

projects/kit/components/segmented/segmented.style.less
```diff
@@ -48,13 +48,26 @@ tui-segmented:where(*&) {
...
+
     & > a,
     & > button,
     & > label,
     & > label > input:not([tuiRadio]) {
         .transition(~'color, background, opacity, border-image');
         .button-clear();
 
+        &[inert],
+        &:disabled,
```
waterplea: Let's not add this, there's no spec for disabling individual items, links are not possible to disable and inputs are wrapped inside label, therefore the styles would have to be applied through :has(:disabled). Let's not introduce ._disabled style hacks to public API

splincode: fixed

projects/kit/components/segmented/segmented.style.less
splincode: fixed

projects/kit/components/segmented/segmented.component.ts
```diff
@@ -48,6 +52,11 @@ export class TuiSegmented implements OnChanges {
 
     public readonly size = input(inject(TUI_SEGMENTED_OPTIONS).size);
     public readonly activeItemIndex = model(0);
+    public readonly disabled = input(false);
+
+    protected readonly inert = effect(() => {
+        this.el.toggleAttribute('inert', this.disabled());
+    });
```
waterplea: Why not through binding?

splincode: sure, fixed

waterplea (approved):

vladimirpotekhin (approved):

PR #13766, opened 2026-04-07
feat(kit): remove empty labels

JonataB:
Fixes #


splincode (approved):

waterplea:
I think it's better to use CSS and hide :empty, because in first label, there's <ng-content /> that still needs to be styled as label if the custom text is inside.

JonataB:
new PR 
https://github.com/taiga-family/taiga-ui/pull/14000

tks1k: learned something new

PR #13815, opened 2026-04-10
fix(core): expose dropdown hover state via public getter

### LINKED ISSUE #13714: 🐞 - Cannot determine hover status in `TuiDropdownHover`
### Reproduction url

none

### Description

In the [previous version](https://github.com/taiga-family/taiga-ui/blob/v4.x/projects/core/directives/dropdown/dropdown-hover.directive.ts#L93) of `TuiDropdownHover` directive, the `hovered` property was public, allowing external components/directives to read the current hover state of the dropdown. In 5.0.0 the `hovered` property was [changed ](https://github.com/taiga-family/taiga-ui/blob/main/projects/core/portals/dropdown/dropdown-hover.directive.ts#L43) to `private`.

**Expected behavior:** The hover state should be accessible externally, either by making the property public again or by exposing a public getter/observable.

### Taiga UI version

5.0.0

### Browser Used

Chrome

### OS Used

Windows

diwakar-2112:
Fixes #13714 

 Exposed the hover state via a public getter to make it accessible externally.

### DIFF
--- a/projects/core/portals/dropdown/dropdown-hover.directive.ts
+++ b/projects/core/portals/dropdown/dropdown-hover.directive.ts
@@ -5,9 +5,9 @@ import {
     ElementRef,
     inject,
     input,
+    type Signal,
     signal,
-    Signal,
-    WritableSignal,
+    type WritableSignal,
 } from '@angular/core';
 import {toObservable} from '@angular/core/rxjs-interop';
 import {TuiActiveZone} from '@taiga-ui/cdk/directives/active-zone';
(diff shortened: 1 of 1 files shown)


projects/core/portals/dropdown/dropdown-hover.directive.ts
```diff
@@ -41,6 +40,9 @@ export class TuiDropdownHover extends TuiDriver {
     });
 
     private hovered = false;
```
waterplea: I think it's better to make it:
```ts
public readonly hovered: Signal<boolean> = signal(false);
```
and use it as:
```ts
(this.hovered as WritableSignal<boolean>).set(hovered);
```
on line 78

diwakar-2112: ok sure will update

diwakar-2112:
@waterplea Addressed your suggestion in [2c23dc8] . Please review.

waterplea (commented):
LGTM, just address the linting issues please

waterplea (approved):

splincode (approved):

PR #13839, opened 2026-04-11
chore(cdk): `TuiDirectionOrder` add todo migration for v5

mdlufy:
Partially solved #11917

### DIFF
--- a/projects/cdk/schematics/ng-update/v5/steps/constants/html-comments.ts
+++ b/projects/cdk/schematics/ng-update/v5/steps/constants/html-comments.ts
@@ -76,6 +76,12 @@ export const HTML_COMMENTS: HtmlComment[] = [
             'timeline-steps has been removed. Use TuiStepper instead. See example https://taiga-ui.dev/navigation/stepper',
     },
     {
+        tag: '*',
+        withAttrs: ['tuiDirectionOrder'],
+        comment:
+            'tuiDirectionOrder has been removed. Remove [directionOrder] and (directionOrderChange). Use [tuiSortBy] with (tuiSortChange) output and TuiSortDirection enum instead',
+    },
+    {
         tag: 'tui-tag',
         withAttrs: [],
         comment: 'tui-tag/tuiTag migrated to tuiChip. Check visuals and content manually',
--- a/projects/cdk/schematics/ng-update/v5/steps/constants/html-comments.ts
+++ b/projects/cdk/schematics/ng-update/v5/steps/constants/html-comments.ts
@@ -79,7 +79,7 @@ export const HTML_COMMENTS: HtmlComment[] = [
         tag: '*',
         withAttrs: ['tuiDirectionOrder'],
         comment:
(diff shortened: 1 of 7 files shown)


splincode (approved):

projects/cdk/schematics/ng-update/v5/tests/__snapshots__/schematic-migrate-direction-order.spec.ts.snap
```diff
@@ -0,0 +1,29 @@
...
+
+exports[`ng-update TuiTableDirectionOrder warning adds TODO for TuiTableDirectionOrder import: test.ts 1`] = `
+{
+  "0. Before": "
+                import {Component} from '@angular/core';
+                import {TuiTableDirectionOrder} from '@taiga-ui/addon-table';
+
+                @Component({
+                    standalone: true,
+                    imports: [TuiTableDirectionOrder],
```
hakimio: This will not work IRL. `TuiTableDirectionOrder` was included in `TuiTable` it wasn't imported separately as standalone directive but together with all the other table imports. You have to check the html template for its usage.

hakimio (changes requested):

projects/cdk/schematics/ng-update/v5/tests/__snapshots__/schematic-migrate-direction-order.spec.ts.snap
mdlufy: I've pushed updated version, you can see if everything suits you :)

projects/cdk/schematics/ng-update/v5/steps/constants/html-comments.ts
```diff
@@ -75,6 +75,12 @@ export const HTML_COMMENTS: HtmlComment[] = [
         comment:
             'timeline-steps has been removed. Use TuiStepper instead. See example https://taiga-ui.dev/navigation/stepper',
     },
+    {
+        tag: '*',
+        withAttrs: ['tuiDirectionOrder'],
+        comment:
+            'tuiDirectionOrder has been removed. Remove [directionOrder] and (directionOrderChange). Use [tuiSortBy] with (tuiSortChange) output and TuiSortDirection enum instead',
```
hakimio: `[direction]` is the equivalent for `[directionOrder]` not `[tuiSortBy]`. Take a look at [this example](https://taiga-ui.dev/components/table#sorting).
Also, why not just automatically replace `[directionOrder]` with `[direction]` and `(directionOrderChange)` with `(directionChange)`?

hakimio (changes requested):

projects/cdk/schematics/ng-update/v5/steps/constants/html-comments.ts
hakimio: And there is also `(directionChange)` event. `[directionOrder]` could be replaced by `[direction]` and `(directionOrderChange)` by `(directionChange)`.

mdlufy: Hmm, yeh, looks like you're right

hakimio (approved):

Issue #13784, opened 2026-04-08
🚀 - Create skills for Taiga (Editor, Maskito)


mdlufy:
### Description

Part of #11557 

https://skills.sh/docs

https://angular.dev/ai/agent-skills

https://github.com/angular/skills

Issue #11557, opened 2025-08-08
🗂 - Improve `AI support` for `Taiga UI` developers


mdlufy:
### Commencement

- [x] Add [`llms.txt`](https://taiga-ui.dev/llms.txt) and [`llms-full.txt`](https://taiga-ui.dev/llms-full.txt) files
- [x] Add [`AI support`](https://taiga-ui.dev/ai-support) documentation section

### Improvement

Improve `llms.txt` and `llms-full.txt` files content:
- [x] `llms.txt` autogeneration
- [x] `llms-config.json` to control files content
- [ ] Get rid of noise in `llms-full.txt` content

### Testing
Consider evaluating the quality of files:
- [ ] Learn approaches to monitor quality
- [ ] Consider quality assessment when using files as context (including in rag pipelines)

### Integrations

IDE & MCP-support:
- [ ] Rules files for different ides https://angular.dev/ai/develop-with-ai#rules-files
- [ ] Raise MCP-server https://angular.dev/ai/develop-with-ai#angular-cli-mcp-server-setup

Website (?):
- [ ] UI chat
- [ ] OpenAI API key support (and other providers)

splincode:
https://github.com/microsoft/playwright-mcp

mdlufy:
MCP Alpha version released
https://www.npmjs.com/package/@taiga-ui/mcp

PR #13843, opened 2026-04-12
chore: add WRITING_MIGRATION.md docs

mdlufy:
Partially solved #11557 

For #13784


.claude/skills/write-migration/SKILL.md
```diff
@@ -0,0 +1,456 @@
...
+
+**Option A — Local checkout** (fastest, works offline): Look for a sibling directory named `../taiga-ui-vN` (e.g.,
+`../taiga-ui-v4` when writing v5 migrations). If it exists, read files directly from there.
+
+**Option B — GitHub via `gh` CLI** (when local checkout is absent): The previous major is kept in the `vN.x` branch of
+the same repo (`taiga-family/taiga-ui`).
+
+```bash
+# Browse exports of a package in the previous major
+gh api repos/taiga-family/taiga-ui/contents/projects/kit/index.ts?ref=v4.x \
```
splincode: `v{N-1}.x`?

vladimirpotekhin (approved):

eslint.config.ts
```diff
@@ -19,6 +19,7 @@ export default process.env.HTML_ESLINT
           },
       ]
     : [
+          globalIgnores(['.claude/**']),
```
splincode: why?

.claude/skills/write-migration/SKILL.md
mdlufy: updated

eslint.config.ts
mdlufy: i firstly add .ts template to skill examples, so it's working bad due to eslint/ts errors and wrong paths, so it's not nesessary to have real .ts template, it would enough to have link to real migrations examples

mdlufy: removed

splincode (commented):
A few follow-up notes after reading the current PR state:

1. In `.claude/skills/write-migration/reference/utilities.md`, the import paths in the custom migration tips still look incorrect. The examples point to `../../utils/templates/elements` and `../utils/templates/dom-traversal`, but real migrations under `projects/cdk/schematics/ng-update/vN/steps/templates/` use deeper paths. Anyone copying those snippets from the skill will likely end up with broken imports.

2. The skill is written as a generic `vN` guide, but `reference/utilities.md` does not mention the attr-to-directive utility used in v4 (`replaceAttrsByDirective` / `ATTRS_TO_DIRECTIVE_REPLACE`). That makes the decision tree incomplete for older migrations and may push the model toward a custom migration where an existing mechanism already exists.

3. In `docs/WRITING_MIGRATIONS.md`, the "Unknown change" row says "then proceed as Scenario A", but the document now uses numbered scenarios. That looks like a stale reference and could confuse the reader.

The structure is much clearer now overall; these are the remaining things I would tighten up.

PR #13866, opened 2026-04-15
fix(core): `TuiNotification` revert using tui-lh from styles

### LINKED ISSUE #13862: 🐞 - Broken styles in tui-notification
### Reproduction url

...

### Description

[These changes](https://github.com/taiga-family/taiga-ui/commit/3c0e9b2acdad02c31c5523c725b3f6b7a25f18f3#diff-16756c9104f83637fddf905572779742ba2159358fa037a13df68dba4b9aaab5) has broken tui-notification's padding - `--t-offset ` is not defined now

### Taiga UI version

4.79.0

### Browser Used

Chrome

### OS Used

macOS

ivanobivan:
Fixes #13862
Styles don't work without explicit inclusion, this is clearly visible in the project
@import '@taiga-ui/core/styles/taiga-ui-theme';

However, if you include it, the palette changes significantly, and even modern styles start to look strange.

I simply removed this variable, and that solved the problem.

But maybe you can suggest a better solution.

### DIFF
--- a/projects/core/styles/components/notification.less
+++ b/projects/core/styles/components/notification.less
@@ -39,7 +39,7 @@ tui-notification,
     border-inline-start: var(--t-start) solid transparent;
     border-inline-end: var(--t-end) solid transparent;
 
-    --t-offset: calc((var(--t-height) - var(--tui-lh)) / 2);
+    --t-offset: calc((var(--t-height) - 1.4em) / 2);
     --t-height: var(--tui-height-l);
     --t-start: 0;
     --t-end: 0;
@@ -54,7 +54,7 @@ tui-notification,
 
     &::before {
         position: absolute;
-        top: calc(var(--t-offset) + var(--tui-lh));
+        top: calc(var(--t-offset) + 1.4em);
         left: -1rem;
         inset-inline-start: -1rem;
         transform: translateY(-100%);


projects/core/styles/components/notification.less
```diff
@@ -39,7 +39,7 @@ tui-notification,
     border-inline-start: var(--t-start) solid transparent;
     border-inline-end: var(--t-end) solid transparent;
 
-    --t-offset: calc((var(--t-height) - var(--tui-lh)) / 2);
+    --t-offset: calc((var(--t-height) - 1.4em) / 2);
```
splincode, line 42: what if `var(--tui-lh, 1.4em)`

waterplea: This is probably not the only variable we use like that. It's cumbersome to always add fallbacks, perhaps we can move those variables to tui-root?

ivanobivan: ok, you mean here? 
/taiga-ui/projects/core/components/root/root.style.less

tui-root {}

nsbarsukov: We already define this variables even in `proprietary/styles/theme/variables.less`.
Is this PR really required or should be just closed ?

nsbarsukov: @ivanobivan fixed this problem by syncing version of ALL taiga ui packages

ivanobivan: agree, did't notice theme.less in doc
close

PR #13875, opened 2026-04-16
chore(demo-playwright): use `globalSetup` to download fonts before tests

nsbarsukov:

PR #13887, opened 2026-04-17
chore(demo-playwright): use `globalSetup` to download fonts before tests

nsbarsukov:
Cherry-pick:
- https://github.com/taiga-family/taiga-ui/pull/13875


projects/demo-playwright/.gitignore
```diff
@@ -0,0 +1,8 @@
+tests-results/
```
splincode: Duplicate?
https://github.com/taiga-family/taiga-ui/blob/main/.gitignore#L35

splincode (approved):

projects/demo-playwright/.gitignore
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/13887/commits/1069e8a3d0b37144ace032b090863d2a27d1786b

projects/demo-playwright/.gitignore
```diff
@@ -1,6 +1,3 @@
-tests-results/
```
splincode: I meant remove from global gitignore

nsbarsukov: It is required for `demo-cypress` too

PR #13890, opened 2026-04-18
feat(cdk): `tuiSum` supports `bigint`

splincode:


projects/cdk/utils/math/sum.ts
```diff
@@ -1,11 +1,26 @@
...
  * Calculates sum of any number of passed arguments with proper floating point precision
  */
-export function tuiSum(...args: number[]): number {
-    const decimal = args.map((double) => (double.toString().split('.')[1] ?? '').length);
-    const maxDecimal = Math.max(0, ...decimal);
-    const precision = 10 ** maxDecimal;
-    const sumInt = args.reduce((acc, n) => acc + Math.round(n * precision), 0);
+export function tuiSum(): 0;
+export function tuiSum(...args: number[]): number;
+export function tuiSum(...args: bigint[]): bigint;
```
nsbarsukov, line 6: Could we now update this if-logic branching ?

https://github.com/taiga-family/taiga-ui/blob/3f84e273d4c0f40a3c735fafca1a5350a27b1f82/projects/kit/components/input-number/step/input-number-step.directive.ts#L36-L39

nsbarsukov (approved):

projects/cdk/utils/math/sum.ts
splincode: fixed

terminalOutput
```diff
@@ -0,0 +1,23 @@
+[34mBuilding Angular Package[39m
+[37m[39m
+[37m------------------------------------------------------------------------------[39m
```
vladimirpotekhin: delete

vladimirpotekhin (approved):

PR #13904, opened 2026-04-20
chore: potential fixes for code quality findings

splincode:


projects/addon-doc/components/example/example.template.html
```diff
@@ -65,6 +65,7 @@ <h3>
                                 size="s"
                                 tuiButton
                                 type="button"
+                                [attr.aria-label]="`Edit example on ${codeEditor?.name || 'external editor'}`"
```
vladimirpotekhin, line 68: The button contains text that screen readers should read, so is an aria label really needed here? Same for iconButton

splincode: false positive

PR #13915, opened 2026-04-21
fix(kit): no needed subscribe on `TuiPortalDirective`

splincode:


waterplea:
How is it going to show the push if you do not subscribe?
https://github.com/taiga-family/taiga-ui/blob/main/projects/cdk/portals/directive.ts#L19

splincode:
Yes, exactly, I'll add tests to ensure that screenshots catch the error, otherwise after deleting the code, the pipelines pass

PR #13916, opened 2026-04-21
fix(styles): revert fonts

splincode:


vladimirpotekhin (approved):

projects/demo-cypress/cypress.config.ts
```diff
@@ -7,6 +7,17 @@ const preset = {
...
 
+const imageDiffConfig = {
+    ROOT_DIR: 'tests-results',
+    SCREENSHOTS_DIR: 'snapshots',
+    REPORT_DIR: '.',
+    JSON_REPORT: {
+        FILENAME: 'report-summary',
+        OVERWRITE: true,
+    },
+    RETRY_OPTIONS: {doNotFail: process.env.CI !== 'true'},
```
nsbarsukov: Should it be `process.env.CI || false` (as it was before) ?

nsbarsukov (approved):

PR #13937, opened 2026-04-22
fix(core): `TuiDropdownA11y` sets `role="button"` on anchor elements

splincode:


waterplea:
This is an invalid use case, `a` should not act as a dropdown host, it's a link, not a button. It's not focusable without href. And with href it's no use as a dropdown host.

PR #13957, opened 2026-04-23
feat(addon-doc): unwrap default module for component outlet

splincode:
Before

```ts
protected readonly component1 = import('./examples/1').then((m) => m.default);
protected readonly component1 = import('./examples/2').then((m) => m.default);
protected readonly component1 = import('./examples/3').then((m) => m.default);
```

After

```ts
protected readonly component1 = import('./examples/1');
protected readonly component2 = import('./examples/2');
protected readonly component3 = import('./examples/3');
```

### DIFF
--- a/projects/addon-doc/components/example/example.component.ts
+++ b/projects/addon-doc/components/example/example.component.ts
@@ -1,5 +1,5 @@
 import {Clipboard} from '@angular/cdk/clipboard';
-import {AsyncPipe, DOCUMENT, NgComponentOutlet, NgTemplateOutlet} from '@angular/common';
+import {DOCUMENT, NgComponentOutlet, NgTemplateOutlet} from '@angular/common';
 import {
     ChangeDetectionStrategy,
     Component,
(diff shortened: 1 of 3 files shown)


projects/addon-doc/components/example/example.component.ts
```diff
@@ -121,10 +121,18 @@ export class TuiDocExample implements OnChanges {
...
+        input<Promise<Type<unknown> | {default: Type<unknown>}>>();
+
     public readonly content = input<Record<string, TuiRawLoaderContent>>({});
     public readonly preview = input(true);
 
+    protected readonly component$ = toObservable(this.component).pipe(
+        switchMap((component) => (component ? from(component) : of(null))),
+        map((c) => (c && 'default' in c ? c.default : c)),
+        takeUntilDestroyed(),
+    );
```
nsbarsukov: Could we just use `computed(...)` here ?

nsbarsukov: Or even `input(..., {transform})` ?

nsbarsukov (approved):

Issue #13914, opened 2026-04-21
🚀 - review styles and migrate to :where() to reduce CSS specificity


vladimirpotekhin:
### Description

Now that we have support for :where(), it makes sense to start using it more broadly across the codebase to avoid increasing CSS specificity unnecessarily.

The goal of this task is to review existing styles and identify places where selectors can be safely migrated to :where() in order to keep specificity as low as possible by default.

PR #13974, opened 2026-04-25
fix(kit): migrating to v5 breaks `TuiSwitch`

### LINKED ISSUE #13971: 🐞 - Migrating to v5 breaks TuiSwitch
### Reproduction url

https://stackblitz.com/edit/angular-nrpodpo1?file=src%2Fapp%2Fapp.html

### Description

I can't make `TuiSwitch` display correctly on version 5:

Image

Image

I think this is caused by the fact `tuiIcons` CSS comes after `tuiSwitch`. Because of that, `tuiIcons` takes precedence for `:after` and apply `display:none`.

Image

Forcing `display:block` in devtools fix the display:

Image

But the former doesn't restore the transition effect.

### Taiga UI version

5.3.0

### Browser Used

Chrome

### OS Used

Linux

reda-alaoui:
Fixes #13971

### DIFF
--- a/projects/styles/components/switch.less
+++ b/projects/styles/components/switch.less
@@ -19,7 +19,7 @@
  * @see-also
  * Checkbox, Radio
  */
-[tuiSwitch]:where(*&) {
+input[tuiSwitch]:where(*&) {
     .transition(~'background, box-shadow');
 
     display: inline-block;


projects/styles/components/switch.less
```diff
@@ -19,7 +19,7 @@
  * @see-also
  * Checkbox, Radio
  */
-[tuiSwitch]:where(*&) {
```
waterplea, line 22: Let's do `input[tuiSwitch]` instead, this way the specificity is bumped only a little bit to avoid unexpected conflicts.

waterplea (changes requested):

waterplea:
Thank you for your contribution. We decided to fix it as part of https://github.com/taiga-family/taiga-ui/issues/13914 by minimizing specificity of the icons directive.

PR #14001, opened 2026-04-29
fix(core): `Textfield` fix icons scaling

waterplea:
Fixes #

### DIFF
--- a/projects/styles/components/icons.less
+++ b/projects/styles/components/icons.less
@@ -20,6 +20,7 @@
 [tuiIcons]:where(*&) {
     --t-icon-start: none;
     --t-icon-end: none;
+    --t-zoom: ~'calc(clamp(0px, var(--tui-font-offset) - 10px, 1px) / 1px)';
 
     &::before,
     &::after {
@@ -31,7 +32,7 @@
         flex-shrink: 0;
         box-sizing: content-box;
         background: currentColor;
-        zoom: ~'calc(100% * (clamp(0px, var(--tui-font-offset) - 10px, 1px) / 0.8px))';
+        zoom: calc(100% + 25% * var(--t-zoom));
     }
 
     &::before {
--- a/projects/styles/components/textfield.less
+++ b/projects/styles/components/textfield.less
@@ -52,6 +52,7 @@ tui-textfield:where(*&) {
         &::before {
             z-index: 1;
             block-size: var(--t-height);
+            block-size: calc(var(--t-height) * (1 - 0.2 * var(--t-zoom)));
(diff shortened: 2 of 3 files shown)


projects/styles/components/icons.less
```diff
@@ -31,7 +32,7 @@
         flex-shrink: 0;
         box-sizing: content-box;
         background: currentColor;
-        zoom: ~'calc(100% * (clamp(0px, var(--tui-font-offset) - 10px, 1px) / 0.8px))';
+        zoom: calc(100% + 25% * var(--t-zoom));
```
splincode, line 35: **[Critical]** The `zoom` CSS property is non-standard and not supported in Firefox. Icons will not scale correctly in Firefox browsers.

```less
- zoom: calc(100% + 25% * var(--t-zoom));
+ transform: scale(calc(1 + 0.25 * var(--t-zoom)));
+ transform-origin: center;
```

projects/styles/components/icons.less
```diff
@@ -20,6 +20,7 @@
 [tuiIcons]:where(*&) {
     --t-icon-start: none;
     --t-icon-end: none;
+    --t-zoom: ~'calc(clamp(0px, var(--tui-font-offset) - 10px, 1px) / 1px)';
```
splincode, line 23: **[Suggestion]** The `10px` threshold in `clamp()` is a magic number without documentation. Consider adding a comment explaining the design rationale or extracting to a named variable for clarity.

projects/styles/components/textfield.less
```diff
@@ -52,6 +52,7 @@ tui-textfield:where(*&) {
         &::before {
             z-index: 1;
             block-size: var(--t-height);
+            block-size: calc(var(--t-height) * (1 - 0.2 * var(--t-zoom)));
```
splincode, line 55: **[Suggestion]** Duplicate `block-size` declarations (lines 54-55 and 63-64) without explanation. Either remove the redundant first declaration or add a comment like `// Fallback for browsers without calc()` to clarify the intent.

splincode (changes requested):

splincode (approved):

projects/styles/components/icons.less
waterplea: It only matters for mobile and OS font scaling. Nobody uses FF on mobile. We cannot use transform as it is reserved for rotation of chevrons etc + it does not make the icon bigger in the space it takes, just visually scales it. Let's consider it graceful degradation for FF not supporting icons scaling.

vladimirpotekhin (approved):

PR #14003, opened 2026-04-29
feat(layout):  add appearance to default options

### LINKED ISSUE #13990: 🚀 - Add default appearance option for `tuiCardLarge`
### Description

Hi! I'd like to propose adding support for setting a default appearance value for `tuiCardLarge` through the `tuiCardOptionsProvider`.

Currently, the appearance of `tuiCardLarge` must be set individually on each instance, but in many use cases, teams want a consistent design across the application.

It would be very helpful to allow configuring a default appearance globally via `tuiCardOptionsProvider`, similar to how other Taiga UI components support default options.

```ts
tuiCardOptionsProvider({
  appearance: 'outline',
}),
```

This would set "outline" as the default appearance for all `tuiCardLarge` components unless explicitly overridden.

Stivin:
Added appearance to default options and provider to override appearance settings globally or more locally. Thus, we can define the desired appearance once and not specify it each time in templates.

Fixes #13990

### DIFF
--- a/projects/layout/components/card/large.directive.ts
+++ b/projects/layout/components/card/large.directive.ts
@@ -7,14 +7,14 @@ import {
     ViewEncapsulation,
 } from '@angular/core';
 import {TUI_VERSION} from '@taiga-ui/cdk/constants';
-import {tuiCreateOptions} from '@taiga-ui/cdk/utils/di';
 import {tuiWithStyles} from '@taiga-ui/cdk/utils/miscellaneous';
-import {TuiWithAppearance} from '@taiga-ui/core/directives/appearance';
+import {
+    tuiAppearanceOptionsProvider,
+    TuiWithAppearance,
+} from '@taiga-ui/core/directives/appearance';
 import {TuiSurface} from '@taiga-ui/layout/components/surface';
 
-export const [TUI_CARD_OPTIONS, tuiCardOptionsProvider] = tuiCreateOptions({
-    space: 'normal',
-});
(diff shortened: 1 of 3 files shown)


waterplea (approved):

projects/layout/components/card/large.directive.ts
```diff
@@ -32,14 +32,15 @@ class Styles {}
...
     host: {
         tuiCardLarge: '',
-        '[attr.data-space]': 'tuiCardLarge() || this.options.space',
+        '[attr.data-space]': 'space() || this.options.space',
     },
 })
 export class TuiCardLarge {
     protected readonly options = inject(TUI_CARD_OPTIONS);
     protected readonly nothing = tuiWithStyles(Styles);
-    public readonly tuiCardLarge = input<'' | 'compact' | 'normal'>('');
```
splincode, line 44: When we rename a public field, isn't that breaking?

waterplea: Yes, let's roll that name change back

Stivin: Returned the name

splincode (approved):

splincode:
Could you add a cherry pick to v4.x branch?

Stivin:
> Could you add a cherry pick to v4.x branch?

But in library version 4, directive `TuiCardLarge` does not work with appearance.

splincode:
@Stivin Sorry, and thank you for clarifying that for me.

PR #14009, opened 2026-04-30
chore(experimental): `DatePicker` WIP implementation

waterplea:
Fixes #

### DIFF
--- a/projects/experimental/components/calendar/calendar.component.html
+++ b/projects/experimental/components/calendar/calendar.component.html
@@ -1,0 +1,16 @@
+@for (day of week(); track $index) {
+    <span>{{ day }}</span>
+}
+@for (week of month() | tuiCalendarSheet: true; track week) {
+    @for (day of week; track day) {
+        <button
+            type="button"
+            class="t-day"
+            [class.t-day_active]="value() && day.daySame(value()!)"
+            [class.t-day_adjacent]="day.month !== month().month"
+            (click)="pick.emit(day)"
+        >
+            {{ day.day }}
+        </button>
+    }
+}
--- a/projects/experimental/components/calendar/calendar.component.ts
+++ b/projects/experimental/components/calendar/calendar.component.ts
@@ -1,0 +1,42 @@
+import {
+    ChangeDetectionStrategy,
+    Component,
+    computed,
+    inject,
+    input,
+    output,
+} from '@angular/core';
+import {type TuiDay, TuiMonth} from '@taiga-ui/cdk/date-time';
+import {
+    TUI_CALENDAR_OPTIONS,
+    TuiCalendarSheetPipe,
+} from '@taiga-ui/core/components/calendar';
+import {TUI_SHORT_WEEK_DAYS} from '@taiga-ui/core/tokens';
+
+/**
+ * @deprecated: work in progress, do not use!
+ */
+@Component({
+    selector: 'tui-calendar',
+    imports: [TuiCalendarSheetPipe],
+    templateUrl: './calendar.component.html',
+    styleUrl: './calendar.component.less',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TuiCalendar {
+    private readonly options = inject(TUI_CALENDAR_OPTIONS);
+    private readonly days = inject(TUI_SHORT_WEEK_DAYS);
+
+    protected readonly week = computed((week = convert(this.days())) => [
+        ...week.slice(this.options.weekStart()),
+        ...week.slice(0, this.options.weekStart()),
+    ]);
+
+    public readonly month = input(TuiMonth.currentLocal());
+    public readonly value = input<TuiDay | null>(null);
+    public readonly pick = output<TuiDay>();
+}
+
+function convert(week: readonly string[]): readonly string[] {
+    return [week[week.length - 1] || '', ...week.slice(0, week.length - 1)];
+}
--- a/projects/experimental/components/data-grid/data-grid.component.html
+++ b/projects/experimental/components/data-grid/data-grid.component.html
@@ -1,0 +1,12 @@
+@for (item of items(); track item) {
+    <button
+        type="button"
+        class="t-item"
+        [class.t-item_active]="value().includes(item)"
+        (click)="pick.emit(item)"
+    >
+        <ng-container *polymorpheusOutlet="content() || `${item}` as text; context: {$implicit: item}">
+            {{ text }}
+        </ng-container>
+    </button>
+}
--- a/projects/experimental/components/date-picker/date-picker.component.html
+++ b/projects/experimental/components/date-picker/date-picker.component.html
@@ -1,0 +1,68 @@
+<header class="t-header">
+    @if (view() !== 'year') {
+        <button
+            appearance=""
+            tuiChevron
+            tuiLink
+            type="button"
+            class="t-button"
+            [textContent]="button()"
+            (click)="onView()"
+        ></button>
+    }
+    <button
+        tuiIconButton
+        type="button"
+        class="t-chevron"
+        [iconStart]="icons.decrement"
+        (click)="onSpin(-1)"
+    >
+        {{ texts()[0] }}
+    </button>
+    <button
+        tuiIconButton
+        type="button"
+        class="t-chevron"
+        [iconStart]="icons.increment"
+        (click)="onSpin(1)"
+    >
+        {{ texts()[1] }}
+    </button>
+</header>
+<section tuiSlides>
+    @switch (view()) {
+        @case ('day') {
+            <tui-carousel
+                [index]="current().year * 12 + current().month"
+                (indexChange)="current.set(getMonth($event))"
+            >
+                <tui-calendar
+                    *tuiItem="let index"
+                    [month]="index | tuiMapper: getMonth"
+                    [value]="value()"
+                    (pick)="onDay($event)"
+                />
+            </tui-carousel>
+        }
+
+        @case ('month') {
+            <tui-data-grid
+                [content]="month()"
+                [items]="0 | tuiMapper: getItems"
+                [value]="current().year === value()?.year ? current().month : []"
+                (pick)="onMonth($event)"
+            />
+        }
+
+        @case ('year') {
+            <tui-carousel>
+                <tui-data-grid
+                    *tuiItem="let offset"
+                    [items]="current().year + offset * 12 - 6 | tuiMapper: getItems"
+                    [value]="value()?.year || []"
+                    (pick)="onYear($event)"
+                />
+            </tui-carousel>
+        }
+    }
+</section>
--- a/projects/experimental/components/date-picker/date-picker.component.ts
+++ b/projects/experimental/components/date-picker/date-picker.component.ts
@@ -8,7 +8,7 @@ import {
     viewChild,
 } from '@angular/core';
 import {type TuiDay, TuiMonth} from '@taiga-ui/cdk/date-time';
-import {TuiMapperPipe} from '@taiga-ui/cdk/pipes';
+import {TuiMapperPipe} from '@taiga-ui/cdk/pipes/mapper';
 import {type TuiContext} from '@taiga-ui/cdk/types';
 import {TuiButton, tuiButtonOptionsProvider} from '@taiga-ui/core/components/button';
 import {TuiCarousel, TuiCarouselComponent} from '@taiga-ui/core/components/carousel';
@@ -42,10 +42,10 @@ import {type PolymorpheusContent} from '@taiga-ui/polymorpheus';
 })
 export class TuiDatePicker {
     private readonly carousel = viewChild(TuiCarouselComponent);
-
     protected readonly icons = inject(TUI_COMMON_ICONS);
     protected readonly texts = inject(TUI_SPIN_TEXTS);
     protected readonly months = inject(TUI_MONTHS);
+
     protected readonly button = computed(() =>
         this.view() === 'day'
             ? `${this.months()[this.current().month]} ${this.current().formattedYear}`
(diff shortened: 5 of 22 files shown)


vladimirpotekhin (approved):

projects/experimental/components/date-picker/date-picker.component.ts
```diff
@@ -0,0 +1,100 @@
...
+    );
+
+    public readonly view = model<'day' | 'month' | 'year'>('day');
+    public readonly value = model<TuiDay | null>(null);
+    public readonly current = model(TuiMonth.currentLocal());
+
+    public readonly month = input<PolymorpheusContent<TuiContext<number>>>(
+        ({$implicit}) => this.months()[$implicit],
+    );
+
```
splincode, line 62: `onView()` cannot exit year view. State cycles day→month→year→year... with no return path.

```suggestion
    protected onView(): void {
        this.view.update(v => v === 'day' ? 'month' : v === 'month' ? 'year' : 'day');
    }
```

projects/experimental/components/date-picker/date-picker.component.ts
```diff
@@ -0,0 +1,100 @@
...
+    protected getMonth(index: number): TuiMonth {
+        return new TuiMonth(Math.floor(index / 12), index % 12);
+    }
+
+    protected getItems(initial: number): readonly number[] {
+        return Array.from({length: 12}, (_, index) => initial + index);
+    }
+
+    protected onView(): void {
+        this.view.update((view) => (view === 'day' ? 'month' : 'year'));
```
splincode, line 72: `onSpin()` desynchronizes carousel index from `current` signal in day view. Carousel updates `current` via `(indexChange)`, but `onSpin()` also updates it conditionally, causing race conditions.

```suggestion
    protected onSpin(step: number): void {
        if (step > 0) {
            this.carousel()?.next();
        } else {
            this.carousel()?.prev();
        }
    }
```

Let the carousel's `indexChange` handler manage `current` exclusively.

projects/experimental/components/date-picker/date-picker.component.html
```diff
@@ -0,0 +1,68 @@
...
+        (click)="onSpin(-1)"
+    >
+        {{ texts()[0] }}
+    </button>
+    <button
+        tuiIconButton
+        type="button"
+        class="t-chevron"
+        [iconStart]="icons.increment"
+        (click)="onSpin(1)"
```
splincode, line 27: Type mismatch: `[value]` passes `current().month` (number) or `[]` (array), but DataGrid expects `readonly T[]`. When year doesn't match, active state breaks.

```suggestion
                [value]="value()?.year === current().year ? [current().month] : []"
```

projects/experimental/components/date-picker/date-picker.component.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    protected onSpin(step: number): void {
+        if (step > 0) {
+            this.carousel()?.next();
+        } else {
+            this.carousel()?.prev();
+        }
+
+        if (this.view() === 'month') {
```
splincode, line 84: `onMonth()` accepts invalid indices. Calling `onMonth(-1)` or `onMonth(13)` creates broken `TuiMonth` objects.

```suggestion
    protected onMonth(index: number): void {
        if (!tuiInRange(index, 0, 12)) return;
        this.current.update(({year}) => new TuiMonth(year, index));
        this.view.set('day');
    }
```

projects/experimental/components/data-grid/data-grid.component.html
```diff
@@ -0,0 +1,12 @@
+@for (item of items(); track item) {
```
splincode, line 1: `track item` uses object reference. For object arrays, this causes all items to re-render on every CD cycle.

```suggestion
@for (item of items(); track $index) {
```

Or accept a `trackBy` input for consumer-provided tracking.

projects/experimental/components/data-grid/data-grid.component.html
```diff
@@ -0,0 +1,12 @@
...
+    <button
+        type="button"
+        class="t-item"
```
splincode, line 4: `value().includes(item)` uses reference equality. Broken for object items.

```suggestion
protected readonly valueSet = computed(() => new Set(this.value()));
```

Then use `valueSet().has(item)` in template.

projects/experimental/components/calendar/calendar.component.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+@Component({
+    selector: 'tui-calendar',
+    imports: [TuiCalendarSheetPipe],
+    templateUrl: './calendar.component.html',
+    styleUrl: './calendar.component.less',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TuiCalendar {
+    private readonly options = inject(TUI_CALENDAR_OPTIONS);
```
splincode, line 27: Unconventional computed pattern with default parameter may break reactivity.

```suggestion
    protected readonly week = computed(() => {
        const week = convert(this.days());
        return [...week.slice(this.options.weekStart()), ...week.slice(0, this.options.weekStart())];
    });
```

projects/experimental/components/calendar/calendar.component.html
```diff
@@ -0,0 +1,16 @@
+@for (day of week(); track $index) {
+    <span>{{ day }}</span>
+}
+@for (week of month() | tuiCalendarSheet: true; track week) {
+    @for (day of week; track day) {
```
splincode, line 5: `track day` uses object reference. `TuiCalendarSheetPipe` creates new `TuiDay` instances each month change, causing full re-render of 42 cells.

```suggestion
@for (day of week; track day.day) {
```

projects/experimental/components/date-picker/date-picker.component.html
```diff
@@ -0,0 +1,68 @@
...
+    >
+        {{ texts()[1] }}
+    </button>
+</header>
+<section tuiSlides>
+    @switch (view()) {
+        @case ('day') {
+            <tui-carousel
```
splincode, line 35: `getItems()` allocates new array each CD cycle. Consider memoizing:

```suggestion
protected readonly monthItems = computed(() => Array.from({length: 12}, (_, i) => i));
```

Then use `monthItems()` in template.

projects/experimental/components/date-picker/date-picker.component.html
```diff
@@ -0,0 +1,68 @@
+<header class="t-header">
+    @if (view() !== 'year') {
+        <button
```
splincode, line 3: Missing `aria-label` on header button. Screen readers need context about the button's purpose.

```suggestion
            [attr.aria-label]="'Toggle calendar view. Current: ' + button()"
```

projects/experimental/components/calendar/calendar.component.html
```diff
@@ -0,0 +1,16 @@
...
+        <button
+            type="button"
+            class="t-day"
+            [class.t-day_active]="value() && day.daySame(value()!)"
```
splincode, line 9: Day buttons lack `aria-label`. Screen readers will only announce the day number without month/year context.

```suggestion
            [attr.aria-label]="day.fullDateString"
```

projects/experimental/components/data-grid/data-grid.component.html
```diff
@@ -0,0 +1,12 @@
+@for (item of items(); track item) {
+    <button
```
splincode, line 2: Item buttons lack `aria-label` for accessibility.

```suggestion
        [attr.aria-label]="content() | tuiMapper: ... || item"
```

projects/experimental/components/calendar/calendar.component.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @deprecated: work in progress, do not use!
+ */
+@Component({
+    selector: 'tui-calendar',
+    imports: [TuiCalendarSheetPipe],
+    templateUrl: './calendar.component.html',
+    styleUrl: './calendar.component.less',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TuiCalendar {
```
splincode, line 26: `convert()` duplicates logic from `TuiOrderWeekDaysPipe` in core calendar. Reuse existing pipe to avoid maintenance burden.

```suggestion
import {TuiOrderWeekDaysPipe} from '@taiga-ui/core/components/calendar';
// Use the pipe instead of local convert function
```

projects/experimental/components/date-picker/date-picker.component.ts
```diff
@@ -0,0 +1,100 @@
...
+        return Array.from({length: 12}, (_, index) => initial + index);
+    }
+
+    protected onView(): void {
+        this.view.update((view) => (view === 'day' ? 'month' : 'year'));
+    }
+
+    protected onSpin(step: number): void {
+        if (step > 0) {
+            this.carousel()?.next();
```
splincode: `onSpin()` triggers double state update when in month view: carousel's `indexChange` updates `current`, and explicit `current.update()` also runs.

Remove the conditional `current.update()` and rely solely on carousel's `indexChange` handler.

splincode (changes requested):

splincode (approved):

projects/experimental/components/date-picker/date-picker.component.ts
waterplea: That is expected, you only go deeper by onView, you go up when you pick the value at the current view.

projects/experimental/components/date-picker/date-picker.component.ts
waterplea: Carousel does not exist in month view so those ?. do nothing, and manual current update happens only if we are in the month view.

projects/experimental/components/data-grid/data-grid.component.html
waterplea: We need track item, otherwise we will see unwanted state transitions when we scroll due to DOM elements reuse.

projects/experimental/components/date-picker/date-picker.component.html
waterplea: ARIA/a11y will be handled separately

projects/experimental/components/data-grid/data-grid.component.html
waterplea: ARIA/a11y will be handled separately

projects/experimental/components/calendar/calendar.component.ts
waterplea: I think it's pretty conventional at this point already, we do this very often in the code base

projects/experimental/components/calendar/calendar.component.ts
waterplea: That pipe is getting deprecated after new date pickers are finished.

projects/experimental/components/calendar/calendar.component.html
waterplea: ARIA/a11y will be handled separately

projects/experimental/components/date-picker/date-picker.component.ts
waterplea: I will take care of boundaries in later PRs when I will add min/max/disabledItemHandler etc.

projects/experimental/components/date-picker/date-picker.component.ts
waterplea: Carousel does not exist in month view so those ?. do nothing, and manual current update happens only if we are in the month view.

projects/experimental/components/date-picker/date-picker.component.html
waterplea: getItems is only used in tuiMapper pipe, which takes care of the memoization.

projects/experimental/components/date-picker/date-picker.component.html
waterplea: It passes [current().month] if the year is the same or empty array if it is not the same, so the type is intact.

projects/experimental/components/data-grid/data-grid.component.html
waterplea: I don't see how set is any different but if we ever need this – we will rely on identityMatcher as in other similar cases.

projects/experimental/components/calendar/calendar.component.html
waterplea: We need track item, otherwise we will see unwanted state transitions when we scroll due to DOM elements reuse.

projects/experimental/components/date-picker/date-picker.component.html
```diff
@@ -0,0 +1,68 @@
...
+                [value]="current().year === value()?.year ? current().month : []"
+                (pick)="onMonth($event)"
+            />
+        }
+
+        @case ('year') {
+            <tui-carousel>
+                <tui-data-grid
+                    *tuiItem="let offset"
+                    [items]="current().year + offset * 12 - 6 | tuiMapper: getItems"
```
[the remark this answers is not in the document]
waterplea: I will take care of boundaries in later PRs when I will add min/max/disabledItemHandler etc.

projects/experimental/components/date-picker/date-picker.component.ts
```diff
@@ -0,0 +1,100 @@
...
+        } else {
+            this.carousel()?.prev();
+        }
+
+        if (this.view() === 'month') {
+            this.current.update((current) => current.append({year: step}));
+        }
```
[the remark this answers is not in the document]
waterplea: I will take care of boundaries in later PRs when I will add min/max/disabledItemHandler etc.

projects/experimental/components/date-picker/date-picker.component.html
```diff
@@ -0,0 +1,68 @@
...
+        (click)="onSpin(1)"
+    >
+        {{ texts()[1] }}
+    </button>
+</header>
+<section tuiSlides>
+    @switch (view()) {
+        @case ('day') {
+            <tui-carousel
+                [index]="current().year * 12 + current().month"
```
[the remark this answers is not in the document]
waterplea: I will take care of boundaries in later PRs when I will add min/max/disabledItemHandler etc.

PR #14022, opened 2026-05-01
fix(kit): use background-color instead of background

### LINKED ISSUE #13988: 🗂 - Replace the background property with background-color for tuiAppearance="floating".
### Description

The `background` property needs to be replaced with `background-color` for `tuiAppearance="floating"` to enable setting the element background similar to https://taiga-ui.dev/v4/layout/card-large#image

JonataB:
Fixes #13988


waterplea (approved):

splincode (approved):

splincode:
@JonataB Could you add cherry pick to v4.x?

PR #14047, opened 2026-05-06
chore(experimental): `DatePicker` add limits/disable

waterplea:
Fixes #

### DIFF
--- a/projects/experimental/components/date-picker/date-picker.component.html
+++ b/projects/experimental/components/date-picker/date-picker.component.html
@@ -62,8 +62,8 @@
 
         @case ('year') {
             <tui-carousel
-                [max]="(max().year - current().year) / 12"
-                [min]="(min().year - current().year) / 12"
+                [max]="yearMax()"
+                [min]="yearMin()"
             >
                 <tui-data-grid
                     *tuiItem="let offset"
--- a/projects/experimental/components/date-picker/date-picker.component.html
+++ b/projects/experimental/components/date-picker/date-picker.component.html
@@ -7,13 +7,14 @@
             type="button"
             class="t-button"
             [textContent]="button()"
-            (click)="onView()"
+            (click)="view.set(view() === 'day' ? 'month' : 'year')"
         ></button>
     }
     <button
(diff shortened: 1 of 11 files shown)


vladimirpotekhin (approved):

splincode (changes requested):
Thank you for your work, I have some questions. Full details will be added as inline comments.

projects/experimental/components/date-picker/date-picker.component.html
```diff
@@ -48,16 +53,21 @@
...
                 [value]="current().year === value()?.year ? current().month : []"
                 (pick)="onMonth($event)"
             />
         }
 
         @case ('year') {
-            <tui-carousel>
+            <tui-carousel
+                [max]="yearMax()"
+                [min]="yearMin()"
```
splincode, line 66: Year carousel min/max calculation creates moving target

The formula `(max().year - current().year) / 12` produces bounds that shift as `current().year` changes.

**Example:**
- min=2020, max=2030, current=2025
- Initial: min=(2020-2025)/12=-0.416, max=(2030-2025)/12=0.416
- After scrolling (current=2026): min=(2020-2026)/12=-0.5, max=(2030-2026)/12=0.333

**Problems:**
1. Bounds shift as user navigates
2. Fractional values (-0.416) instead of integers
3. Carousel index becomes invalid
4. Chevron disabled states incorrect

**Fix:** Use absolute year bounds instead of relative calculation:
```html
<tui-carousel
    [index]="current().year - 6"
    [min]="min().year + 6"
    [max]="max().year - 6"
    (indexChange)="current.update(m => new TuiMonth(m.year + $event, m.month))"
>
```

splincode (approved):

projects/experimental/components/date-picker/date-picker.component.html
waterplea: Looks like this comment is for initial commit, since then I fixed it. `current` does not change as you scroll year carousel, it only changes when you pick the year and then carousel is closed and there's rounding so there are no fractions

PR #14054, opened 2026-05-07
chore(experimental): `DatePicker` add mode: single/multi/range

waterplea:
Fixes #

### DIFF
--- a/projects/experimental/components/calendar/calendar.component.html
+++ b/projects/experimental/components/calendar/calendar.component.html
@@ -6,10 +6,12 @@
         <button
             type="button"
             class="t-day"
-            [class.t-day_active]="value() && day.daySame(value()!)"
+            [attr.data-range]="getRange(day)"
             [class.t-day_adjacent]="day.month !== month().month"
+            [class.t-day_today]="day.daySame(today)"
             [disabled]="disabledItemHandler()(day)"
             (click)="pick.emit(day)"
+            (mouseenter)="hovered.set(day)"
         >
             {{ day.day }}
         </button>
--- a/projects/experimental/components/calendar/calendar.component.ts
+++ b/projects/experimental/components/calendar/calendar.component.ts
@@ -31,9 +31,9 @@ import {TUI_SHORT_WEEK_DAYS} from '@taiga-ui/core/tokens';
 export class TuiCalendar {
     private readonly options = inject(TUI_CALENDAR_OPTIONS);
     private readonly days = inject(TUI_SHORT_WEEK_DAYS);
-
     protected readonly today = TuiDay.currentLocal();
     protected readonly hovered = signal<TuiDay | null>(null);
+
     protected readonly week = computed((week = convert(this.days())) => [
         ...week.slice(this.options.weekStart()),
         ...week.slice(0, this.options.weekStart()),
--- a/projects/experimental/components/calendar/calendar.component.ts
+++ b/projects/experimental/components/calendar/calendar.component.ts
@@ -53,8 +53,11 @@ export class TuiCalendar {
             return coerceArray(value).find((item) => day.daySame(item)) ? 'single' : null;
         }
 
-        const hov = this.hovered();
-        const range = value.isSingleDay && hov ? TuiDayRange.sort(hov, value.to) : value;
+        const hovered = this.hovered();
+        const range =
+            value.from === value.to && hovered
+                ? TuiDayRange.sort(hovered, value.to)
+                : value;
 
         if (range.isSingleDay && day.daySame(range.from)) {
             return 'single';
--- a/projects/experimental/components/calendar/calendar.component.ts
+++ b/projects/experimental/components/calendar/calendar.component.ts
@@ -54,6 +54,7 @@ export class TuiCalendar {
         }
 
         const hovered = this.hovered();
+
         const range =
             value.from === value.to && hovered
                 ? TuiDayRange.sort(hovered, value.to)
--- a/projects/experimental/components/calendar/calendar.component.ts
+++ b/projects/experimental/components/calendar/calendar.component.ts
@@ -1,3 +1,4 @@
+import {coerceArray} from '@angular/cdk/coercion';
 import {
     ChangeDetectionStrategy,
     Component,
@@ -5,9 +6,10 @@ import {
     inject,
     input,
     output,
+    signal,
 } from '@angular/core';
 import {TUI_FALSE_HANDLER} from '@taiga-ui/cdk/constants';
-import {type TuiDay, TuiMonth} from '@taiga-ui/cdk/date-time';
+import {TuiDay, TuiDayRange, TuiMonth} from '@taiga-ui/cdk/date-time';
 import {type TuiBooleanHandler} from '@taiga-ui/cdk/types';
 import {
     TUI_CALENDAR_OPTIONS,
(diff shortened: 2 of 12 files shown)


vladimirpotekhin (approved):

projects/experimental/components/date-picker/date-picker.component.ts
```diff
@@ -12,10 +13,12 @@ import {
     TUI_FIRST_DAY,
     TUI_LAST_DAY,
     type TuiDay,
+    TuiDayRange,
     TuiMonth,
 } from '@taiga-ui/cdk/date-time';
 import {TuiMapperPipe} from '@taiga-ui/cdk/pipes/mapper';
```
splincode, line 19: When switching from `range` to `multi` mode, the range value is not converted properly. This causes data loss.

```typescript
} else if (this.mode() === 'multi') {
    this.value.update((value) => {
        const initialValue = value instanceof TuiDayRange
            ? Array.from({length: value.to.day - value.from.day + 1}, (_, i) => value.from.append({day: i}))
            : coerceArray(value || []);
        return tuiArrayToggle(initialValue, day, (a, b) => a.daySame(b));
    });
}
```

projects/experimental/components/date-picker/date-picker.component.ts
```diff
@@ -12,10 +13,12 @@ import {
     TUI_FIRST_DAY,
     TUI_LAST_DAY,
     type TuiDay,
```
splincode, line 15: Consider adding an effect to handle mode changes and convert incompatible values automatically:

```typescript
constructor() {
    effect(() => {
        const mode = this.mode();
        const value = this.value();
        
        if (mode === 'single' && (Array.isArray(value) || value instanceof TuiDayRange)) {
            this.value.set(Array.isArray(value) ? value[0] || null : value.from);
        } else if (mode === 'multi' && value instanceof TuiDayRange) {
            this.value.set(Array.from({length: value.to.day - value.from.day + 1}, (_, i) => value.from.append({day: i})));
        } else if (mode === 'range' && Array.isArray(value)) {
            this.value.set(value.length >= 2 
                ? TuiDayRange.sort(value[0], value[value.length - 1])
                : value[0] ? new TuiDayRange(value[0], value[0]) : null);
        }
    });
}
```

projects/experimental/components/calendar/calendar.component.html
```diff
@@ -6,10 +6,12 @@
         <button
```
splincode, line 6: Add ARIA attributes for accessibility:

```html
<div 
    class="t-calendar"
    role="grid"
    [attr.aria-label]="'Calendar'"
>
    @for (day of week(); track $index) {
        <span role="columnheader">{{ day }}</span>
    }
    @for (week of month() | tuiCalendarSheet: true; track week) {
        <div role="row">
            @for (day of week; track day) {
                <button
                    type="button"
                    role="gridcell"
                    [attr.aria-selected]="getRange(day) !== null"
                    ...
                >
```

projects/experimental/components/calendar/calendar.component.ts
```diff
@@ -24,10 +26,13 @@ import {TUI_SHORT_WEEK_DAYS} from '@taiga-ui/core/tokens';
     templateUrl: './calendar.component.html',
     styleUrl: './calendar.component.less',
     changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {'(mouseleave)': 'hovered.set(null)'},
 })
 export class TuiCalendar {
     private readonly options = inject(TUI_CALENDAR_OPTIONS);
     private readonly days = inject(TUI_SHORT_WEEK_DAYS);
```
splincode, line 33: Add keyboard navigation support:

```typescript
protected onKeyDown(event: KeyboardEvent, day: TuiDay): void {
    let newDay: TuiDay | null = null;
    
    switch (event.key) {
        case 'ArrowRight': newDay = day.append({day: 1}); break;
        case 'ArrowLeft': newDay = day.append({day: -1}); break;
        case 'ArrowDown': newDay = day.append({day: 7}); break;
        case 'ArrowUp': newDay = day.append({day: -7}); break;
        case 'Home': newDay = day.append({day: -day.dayOfWeek()}); break;
        case 'End': newDay = day.append({day: 6 - day.dayOfWeek()}); break;
    }
    
    if (newDay) {
        event.preventDefault();
        this.pick.emit(newDay);
    }
}
```

projects/experimental/components/calendar/calendar.component.ts
```diff
@@ -1,13 +1,15 @@
...
     inject,
     input,
     output,
+    signal,
 } from '@angular/core';
 import {TUI_FALSE_HANDLER} from '@taiga-ui/cdk/constants';
-import {type TuiDay, TuiMonth} from '@taiga-ui/cdk/date-time';
+import {TuiDay, TuiDayRange, TuiMonth} from '@taiga-ui/cdk/date-time';
 import {type TuiBooleanHandler} from '@taiga-ui/cdk/types';
 import {
```
splincode, line 14: The `getRange` method doesn't visually distinguish multi-select state. Consider returning a distinct value for multi-selected days:

```typescript
protected getRange(day: TuiDay): 'end' | 'middle' | 'single' | 'start' | 'multi' | null {
    const value = this.value() || [];

    if (!(value instanceof TuiDayRange)) {
        const selectedDays = coerceArray(value);
        if (selectedDays.some(d => day.daySame(d))) {
            return selectedDays.length > 1 ? 'multi' : 'single';
        }
        return null;
    }
    
    // ... existing range logic
}
```

projects/experimental/components/date-picker/date-picker.component.ts
```diff
@@ -50,15 +53,34 @@ export class TuiDatePicker {
     protected readonly carousel = viewChild(TuiCarouselComponent);
     protected readonly icons = inject(TUI_COMMON_ICONS);
```
splincode, line 54: Consider removing duplicates from years array when multiple selected days fall in the same year:

```typescript
protected readonly years = computed(() => {
    const value = this.value() || [];
    const years = value instanceof TuiDayRange
        ? Array.from({length: value.to.year - value.from.year + 1}, (_, i) => value.from.year + i)
        : coerceArray(value).map(day => day.year);
    
    return Array.from(new Set(years)).sort();
});
```

splincode (changes requested):
Found several issues that need to be addressed before merge. See inline comments for details and code examples.

splincode (approved):

projects/experimental/components/date-picker/date-picker.component.ts
waterplea: This is expected, it's not supposed to work with different modes dynamically and retaining the value.

projects/experimental/components/date-picker/date-picker.component.ts
waterplea: There's no meaningful way to convert values when switching modes. Users should just reset value to `null` if they want to do it.

projects/experimental/components/calendar/calendar.component.html
waterplea: Like in last PR, a11y will be handled later separately.

projects/experimental/components/calendar/calendar.component.ts
waterplea: Like in last PR, a11y will be handled later separately.

projects/experimental/components/calendar/calendar.component.ts
waterplea: Multiple selected days looks like separate single days.

PR #13743, opened 2026-04-03
fix(kit): `InputNumber` fix hydration

waterplea:
Partial #13646

PR #14056, opened 2026-05-07
feat(kit): `Pincode` add component

mdlufy:
Add pincode component


.cspell.json
```diff
@@ -18,6 +18,8 @@
         "officedocument",
         "wordprocessingml",
         "longname",
-        "initialises"
+        "initialises",
+        "Pincode",
+        "pincode"
```
splincode: No needed

projects/kit/components/pincode/pincode.component.ts
```diff
@@ -0,0 +1,88 @@
...
+        computed(() => ({mask: this.mask(), overwriteMode: 'replace'})),
+    );
+
+    public readonly appearance = input<TuiPincodeAppearance>('numbers');
+    public readonly mode = input<TuiPincodeMode | null>(null);
+
+    public readonly mask = input(/^\d+$/, {
+        transform: (mask: MaskitoMask | string): MaskitoMask =>
+            tuiIsString(mask) ? new RegExp(mask) : mask,
+    });
```
splincode: Do you really need to use a mask?

.cspell.json
mdlufy: why? error without it

projects/kit/components/pincode/pincode.template.html
```diff
@@ -0,0 +1,17 @@
+<div
+    *tuiTextfieldContent
+    class="t-wrapper"
+>
+    @for (_ of '-'.repeat(el.maxLength); track $index) {
+        <label
+            class="t-item"
+            [attr.data-focus]="isFocused($index) || null"
```
waterplea: class.t-item_focus? What's the point of attribute if it's either true or missing?

projects/kit/components/pincode/pincode.template.html
```diff
@@ -0,0 +1,18 @@
+<div
+    *tuiTextfieldContent
+    class="t-wrapper"
+>
+    @for (_ of '-'.repeat(el.maxLength); track $index) {
```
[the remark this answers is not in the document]
waterplea: Default value is -1 which would crash at runtime, maybe we can add maxlength: 4 to host so it always has some value?

projects/kit/components/pincode/pincode.ts
```diff
@@ -0,0 +1,16 @@
...
+import {TuiDropdownContent} from '@taiga-ui/core/portals/dropdown';
+
+import {TuiPincodeComponent} from './pincode.component';
+
+export const TuiPincode = [
+    TuiPincodeComponent,
+    TuiLabel,
+    TuiTextfieldComponent,
+    TuiTextfieldOptionsDirective,
+    TuiDropdownContent,
```
waterplea: Why do we need dropdown here?

projects/kit/components/pincode/pincode.style.less
```diff
@@ -0,0 +1,375 @@
...
+    from {
+        scale: 1.1;
+    }
+
+    to {
+        scale: 0;
+    }
+}
+
+@keyframes tuiPincodeDotFade {
```
waterplea: We already have tuiFade for that

projects/kit/components/pincode/pincode.style.less
```diff
@@ -0,0 +1,375 @@
...
+    from {
+        opacity: 1;
+    }
+
+    to {
+        opacity: 0;
+    }
+}
+
+@keyframes tuiPincodeExitElement {
```
waterplea: We have tuiFade + tuiScale, can we just reuse them together instead of making a new one with both?

projects/kit/components/pincode/pincode.style.less
```diff
@@ -0,0 +1,375 @@
...
+        scale: 1.2;
+        animation-timing-function: linear;
+    }
+
+    100% {
+        scale: 1;
+    }
+}
+
+@keyframes tuiPincodeCaretBlink {
```
waterplea: tuiFade can be used instead as well

projects/kit/components/pincode/pincode.component.ts
mdlufy: yes, mask without input for digits

projects/kit/components/pincode/pincode.template.html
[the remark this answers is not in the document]
mdlufy: yes, add maxlength: 4 to host as default

projects/kit/components/pincode/pincode.ts
mdlufy: removed

projects/kit/components/pincode/pincode.template.html
mdlufy: Replaced with [class.t-item_focus]

.cspell.json
splincode: try `npm ci`

projects/kit/components/pincode/pincode.component.ts
```diff
@@ -0,0 +1,73 @@
...
+            tuiIsFocused(this.el) &&
+            index === Math.min(this.el.value.length, this.el.maxLength - 1)
+        );
+    }
+
+    protected onSelection(): void {
+        const pos = this.el.value.length;
+
+        if (this.el.selectionStart !== pos) {
+            this.el.setSelectionRange(pos, pos);
```
splincode: Can we clamp the caret the same way as `TuiInputPin` does when the value is already full? With `pos === maxLength`, typing another digit is blocked by native `maxlength`, so the user cannot replace the last digit by typing — only Backspace works. Something like `const pos = Math.min(this.el.value.length, this.el.maxLength - 1)` would keep overwrite behavior consistent.

projects/kit/components/pincode/pincode.template.html
```diff
@@ -0,0 +1,17 @@
...
+        <label
```
splincode: Can we either make these plain visual cells (`span`/`div`) or handle `pointerdown` here to focus the real input? These labels are not associated with the input, and if `.t-content` is above the transparent input, clicking a cell may not focus the control. `TuiInputPin` focuses on cell `pointerdown`, which also makes the visual cells behave like the actual field.

projects/demo/src/pages/components/pincode/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    ];
+
+    protected readonly typeVariants = ['text', 'password'] as const;
+    protected readonly mode = signal<TuiPincodeMode | null>(null);
+    protected readonly type = signal<'password' | 'text'>('text');
+    protected readonly control = new FormControl('');
+
+    constructor() {
+        effect(() => {
+            const mode = this.mode();
```
splincode: Can we clean up these timers inside the effect? If the user switches `mode` before the timeout fires, an old timer can reset a newer state back to `null`. `effect((onCleanup) => { const id = setTimeout(...); onCleanup(() => clearTimeout(id)); })` would make the demo deterministic.

splincode (commented):
Left a few small comments

projects/kit/components/pincode/pincode.component.ts
```diff
@@ -0,0 +1,73 @@
...
+        inputmode: 'numeric',
+        maxlength: '4',
+        ngSkipHydration: 'true',
+        spellcheck: 'false',
+        '[attr.data-mode]': 'mode()',
+        '(selectionchange)': 'onSelection()',
+    },
+})
+export class TuiPincodeComponent {
+    protected readonly el = tuiInjectElement<HTMLInputElement>();
```
splincode: Could we reduce divergence from `TuiInputPin` here? This component shares most of the low-level mechanics: `MaskitoDirective`, `inputmode`, `ngSkipHydration`, caret normalization and rendering cells from `maxlength`.

Maybe we can extract a small shared base/helper and keep only visual differences in each component:

```ts
abstract class TuiPinLike {
    protected readonly el = tuiInjectElement<HTMLInputElement>();
    protected readonly maskito = tuiMaskito({mask: /^\d+$/, overwriteMode: 'replace'});

    protected get length(): number {
        return this.el.maxLength > 0 ? this.el.maxLength : 4;
    }

    protected moveCaretToEnd(): void {
        const pos = Math.min(this.el.value.length, this.length - 1);
        this.el.setSelectionRange(pos, pos);
    }
}
```

Then `InputPin` and `Pincode` would differ mostly by template, styles, mode API, and future fixes around caret, mask, length would not drift between two PIN-like controls.

projects/kit/components/pincode/pincode.component.ts
```diff
@@ -0,0 +1,73 @@
...
+    protected readonly maskito = tuiMaskito({mask: /^\d+$/, overwriteMode: 'replace'});
+    public readonly mode = input<TuiPincodeMode | null>(null);
+
+    protected getStyle(index: number): Record<string, string> {
+        const n = this.el.maxLength;
```
splincode: Docs expose only `4`, `5` and `6` as supported lengths, but native `maxlength` can be any positive number. Since layout and animation math depend on this value, can we normalize it in one getter instead of using `this.el.maxLength` directly?

For example:

```ts
protected get length(): 4 | 5 | 6 {
    return this.el.maxLength === 5 || this.el.maxLength === 6 ? this.el.maxLength : 4;
}
```

This would keep runtime behavior aligned with the documented API and avoid surprising layouts for `maxlength="1"` or `maxlength="100"`.

projects/kit/components/pincode/pincode.style.less
```diff
@@ -0,0 +1,343 @@
...
```
splincode: Agree with teamlead here, reusing existing `tuiFade` and `tuiScale` is better than adding custom one-off keyframes when the motion is semantically the same. It keeps this new component closer to the shared Taiga animation vocabulary and reduces future maintenance.

splincode (commented):
Inline comments only too

projects/kit/components/pincode/pincode.style.less
mdlufy: yes

projects/kit/components/pincode/pincode.component.ts
mdlufy: yes, you are right

projects/demo/src/pages/components/pincode/index.ts
mdlufy: good idea, updated

projects/kit/components/pincode/pincode.component.ts
waterplea: This is unreasonable

projects/kit/components/pincode/pincode.template.html
waterplea: Yes, span makes more sense here in terms of valid html

projects/demo/src/pages/components/pincode/examples/3/index.html
```diff
@@ -0,0 +1,55 @@
...
+            [maxlength]="length"
+            [mode]="mode()"
+            [type]="type"
+        />
+    </tui-textfield>
+
+    <div class="controls">
+        <div class="row">
+            <span class="label">Type</span>
+            <tui-segmented size="s">
```
splincode: nit: Can we use a more descriptive loop variable here?

```html
@for (type of types; track type) {
```

It makes the example easier to read, especially because this is demo/documentation code.

projects/demo/src/pages/components/pincode/examples/3/index.html
```diff
@@ -0,0 +1,55 @@
...
+                            (ngModelChange)="reset()"
+                        />
+                        {{ t }}
+                    </label>
+                }
+            </tui-segmented>
+        </div>
+
+        <div class="row">
+            <span class="label">Length</span>
```
splincode: nit: Same here, `length` would read better than `l` in demo code.

```html
@for (length of lengths; track length) {
```

Examples are easier to scan when local variable names match the domain.

splincode (commented):
Nit

projects/demo/src/pages/components/pincode/examples/3/index.html
mdlufy: done

projects/demo/src/pages/components/pincode/examples/3/index.html
mdlufy: done

projects/kit/components/pincode/pincode.component.ts
mdlufy: It sounds good, but I'm not sure that at this stage it's worth spreading the logic through some kind of rough layer. As if this will only complicate the complexity of perception now, because there is little code right now and it is in one place. And perhaps, we'll deprecate InputPin ifTuiPincode could be customized so that it covers it

projects/kit/components/pincode/pincode.component.ts
mdlufy: done

mdlufy: oh, it's not working correctly with onSelection, when backspacing symbols
try to enter: 1,2,3,4 and try to backspace, it would be 1,2,4 instead 1,2,3

so we can use `Math.min(this.el.value.length, this.el.maxLength - 1)` only in `isFocused`, but not in `onSelection`

revert

splincode: Yes, makes sense. Forcing `selectionStart` to `maxLength - 1` inside `onSelection` breaks Backspace semantics.

Maybe we should keep `onSelection` as “move to the end”, and only use `Math.min(value.length, maxLength - 1)` for the visual focus state. This way Backspace still removes the last entered digit, while the visual caret does not point outside the rendered cells when the value is full.

projects/kit/components/pincode/pincode.component.ts
splincode: Makes sense. I agree that adding an extra shared layer right now can be premature if the logic is still small and the future API direction is not fully clear.

@waterplea My main concern was only about possible drift between two PIN-like controls. If the plan is to make `TuiPincode` flexible enough to cover `InputPin` later, then it is probably better to keep the implementation simple for now and revisit extraction, deprecation once the API stabilizes.

projects/kit/components/pincode/pincode.component.ts
```diff
@@ -0,0 +1,77 @@
...
+export class TuiPincodeComponent {
+    protected readonly el = tuiInjectElement<HTMLInputElement>();
+    protected readonly maskito = tuiMaskito({mask: /^\d+$/, overwriteMode: 'replace'});
+    public readonly mode = input<TuiPincodeMode | null>(null);
+
+    protected get length(): 4 | 5 | 6 {
+        return this.el.maxLength === 5 || this.el.maxLength === 6 ? this.el.maxLength : 4;
+    }
+
+    protected getStyle(index: number): Record<string, string> {
```
splincode: nit/perf: Can we avoid allocating a new style object from the template on every change detection pass?

`getStyle($index)` depends only on `index` and normalized `maxlength`, so we could make `maxlength` a signal input and precompute stable styles with `computed()`:

```ts
import {computed, input, numberAttribute} from '@angular/core';

type TuiPincodeLength = 4 | 5 | 6;

function tuiIsPincodeLength(length: number): length is TuiPincodeLength {
    return length === 4 || length === 5 || length === 6;
}

function tuiPincodeLength(value: unknown): TuiPincodeLength {
    const length = numberAttribute(value);

    return tuiIsPincodeLength(length) ? length : 4;
}
```

```ts
protected readonly length = input<TuiPincodeLength, unknown>(4, {
    alias: 'maxlength',
    transform: tuiPincodeLength,
});

protected readonly styles = computed(() => {
    const length = this.length();
    const cycle = 400 + (length - 1) * 100 + 300;

    return Array.from({length}, (_, index): Record<string, string> => {
        const offset = index - (length - 1) / 2;

        return {
            '--tui-animation-delay': `${index * 100}ms`,
            '--tui-animation-cycle': `${cycle}ms`,
            '--tui-spread-tx': `${offset * (PITCH / 3)}rem`,
            '--tui-collapse-tx': `${-offset * PITCH}rem`,
        };
    });
});
```

Then the template can iterate over stable values:

```html
@for (style of styles(); track $index) {
    <span [style]="style">...</span>
}
```

and host can keep native maxlength in sync:

```ts
host: {
    '[maxLength]': 'length()',
}
```

Not critical for 4–6 cells, but it keeps object allocation out of the template and makes `maxlength` normalization explicit.

splincode (commented):
What if...​

projects/kit/components/pincode/pincode.template.html
```diff
@@ -0,0 +1,18 @@
...
+        <span
+            class="t-item"
+            [class.t-item_filled]="!!el.value[$index]"
+            [class.t-item_focus]="isFocused($index)"
+            [style]="getStyle($index)"
+            (pointerdown.prevent)="el.focus()"
+        >
+            <span class="t-dot t-dot_placeholder"></span>
+            <span class="t-dot t-dot_value"></span>
+            <span class="t-value">{{ el.value[$index] }}</span>
```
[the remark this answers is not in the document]
mdlufy: fixed

projects/kit/components/pincode/pincode.component.ts
```diff
@@ -0,0 +1,77 @@
...
+    },
+})
+export class TuiPincodeComponent {
+    protected readonly el = tuiInjectElement<HTMLInputElement>();
+    protected readonly maskito = tuiMaskito({mask: /^\d+$/, overwriteMode: 'replace'});
+    public readonly mode = input<TuiPincodeMode | null>(null);
+
+    protected get length(): 4 | 5 | 6 {
+        return this.el.maxLength === 5 || this.el.maxLength === 6 ? this.el.maxLength : 4;
+    }
```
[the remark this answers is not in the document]
mdlufy: looks bad

projects/kit/components/pincode/pincode.template.html
```diff
@@ -0,0 +1,18 @@
+<div
```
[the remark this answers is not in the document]
mdlufy: added

projects/kit/components/pincode/pincode.component.ts
mdlufy: Oh, I don't even know.. It's a good idea, but it seems premature to me now, as if we need to stabilize the API first and optimize it later.

For me, for 4-6 cells, the performance gain will be near zero, and this is more of a "beautiful" refactor than a necessity, at the same time, the code will become noticeably more complicated

projects/kit/components/pincode/pincode.component.ts
```diff
@@ -0,0 +1,77 @@
...
+        ngSkipHydration: 'true',
+        spellcheck: 'false',
+        '[attr.data-mode]': 'mode()',
+        '(selectionchange)': 'onSelection()',
+    },
+})
+export class TuiPincodeComponent {
+    protected readonly el = tuiInjectElement<HTMLInputElement>();
+    protected readonly maskito = tuiMaskito({mask: /^\d+$/, overwriteMode: 'replace'});
+    public readonly mode = input<TuiPincodeMode | null>(null);
```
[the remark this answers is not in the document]
mdlufy: bad, mode is public api

projects/demo/src/pages/components/pincode/index.ts
```diff
@@ -38,7 +38,13 @@ export default class PageComponent {
     constructor() {
         effect((onCleanup) => {
             const mode = this.mode();
-            const delay = mode === 'dismissing' ? 800 : mode === 'submitting' ? 1900 : 0;
+
+            const delays: Partial<Record<string, number>> = {
+                dismissing: 800,
+                submitting: 1900,
+            };
```
splincode: maybe then we should move the constant to the file level?

projects/kit/components/pincode/pincode.component.ts
waterplea: I think it's OK to keep them separate for now, after all it's internal implementation and we can change it later on if we feel like it.

projects/kit/components/pincode/pincode.component.ts
waterplea: If we move to the last char we can make it always selected, this way you can both overwrite it and remove it with backspace.

projects/kit/components/pincode/pincode.component.ts
waterplea: Let's keep it for now. Just keep in mind we need --t- prefix for private CSS vars, not --tui-

projects/demo/src/pages/components/pincode/index.ts
waterplea: Seems unimportant for demo page

projects/kit/components/pincode/pincode.template.html
```diff
@@ -0,0 +1,19 @@
...
```
waterplea: Let's try to keep it all SPAN, DIV inside P are invalid HTML and if somebody puts that component inside P without knowing the internal structure it might cause HTML to become invalid.

projects/demo/src/pages/components/pincode/examples/4/index.html
```diff
@@ -0,0 +1,27 @@
+<div class="root">
+    <p class="hint">
+        Long loading example with loader.
+        <br />
+        Enter
+        <code>1234</code>
+        for success (submitting state), any other code for error
```
nsbarsukov: No need to duplicate this description twice
Image

Keep only one

nsbarsukov: The same comment for next example too

projects/demo/src/pages/components/pincode/index.html
```diff
@@ -0,0 +1,122 @@
...
+                [description]="description"
+                [heading]="example"
+            />
+            <ng-template #description>
+                @switch ($index) {
+                    @case (0) {
+                        Default with
+                        <code>[(ngModel)]</code>
+                        — works equivalently with
+                        <code>[formControl]</code>
```
nsbarsukov: Is it excessive to point out that CVA works with both `[ngModel]` and `formControl`? After all, `ngModel` actually uses `formControl` under the hood.

projects/demo/src/pages/components/pincode/examples/3/index.html
```diff
@@ -0,0 +1,15 @@
+<div class="root">
```
nsbarsukov: Could we just use `:host` to avoid extra html nesting ?

projects/demo/src/pages/components/pincode/examples/4/index.html
```diff
@@ -0,0 +1,27 @@
+<div class="root">
```
nsbarsukov: The same comment about `:host` for other examples too

projects/demo/src/pages/components/pincode/examples/4/index.ts
```diff
@@ -0,0 +1,75 @@
...
+        return value?.length === 4 ? this.verify(value) : of(null);
+    }
+
+    private verify(value: string): Observable<TuiPincodeMode | null> {
+        const result$ = from(fakeApiVerify(value)).pipe(shareReplay(1));
+
+        return merge(
+            timer(500).pipe(
+                takeUntil(result$),
+                map(() => 'pending' as const),
```
nsbarsukov: I wonder if we make this example more signal-friendly: `[(ngModel)]="signalBasedValue"` + Resource API

projects/demo/src/pages/components/pincode/index.html
```diff
@@ -0,0 +1,122 @@
...
+            </tr>
+            <tr
+                name="[mode]"
+                tuiDocAPIItem
+                type="TuiPincodeMode"
+                [items]="modeVariants"
+                [value]="mode()"
+                (valueChange)="mode.set($event ?? null)"
+            >
+                Visual state:
```
nsbarsukov: If this is purely about the visual state, wouldn't it be better to use `[appearance]` here and create a corresponding `projects/styles/theme/appearance/pincode.less` file?

@taiga-family/core-team What do you think?

For example:
```html
<tui-textfield>
    <input
        tuiPincode
        [formControl]="control"
        appearance="submitting"
    />
</tui-textfield>
```

projects/kit/components/pincode/pincode.component.ts
```diff
@@ -0,0 +1,74 @@
...
+    templateUrl: './pincode.template.html',
+    styleUrl: './pincode.style.less',
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    hostDirectives: [MaskitoDirective],
+    host: {
+        autocomplete: 'one-time-code',
+        inputmode: 'numeric',
+        maxlength: '4',
+        ngSkipHydration: 'true',
```
nsbarsukov: Could we use modern approach to avoid skipping hydration ?

- https://github.com/taiga-family/taiga-ui/pull/13743

projects/kit/components/pincode/pincode.template.html
mdlufy: Updated

projects/demo/src/pages/components/pincode/examples/4/index.html
mdlufy: Updated

projects/kit/components/pincode/pincode.component.ts
mdlufy: Updated

projects/demo/src/pages/components/pincode/index.html
mdlufy: Updated

projects/demo/src/pages/components/pincode/examples/3/index.html
mdlufy: yes, updated

projects/demo/src/pages/components/pincode/examples/4/index.html
mdlufy: Updated

projects/demo/src/pages/components/pincode/examples/4/index.ts
mdlufy: I tried it and it seems to work! Updated the code

projects/demo/src/pages/components/pincode/index.html
mdlufy: Yes, it may even go away in the end. So far, I've added an automatic bounce, as suggested by @waterplea 
We will continue to improve iteratively

projects/kit/components/pincode/pincode.component.ts
mdlufy: It seems yes, it will work, you can see the updated code

projects/demo/src/pages/components/pincode/examples/1/index.ts
```diff
@@ -0,0 +1,15 @@
...
+import {TuiPincode} from '@taiga-ui/kit';
+
+@Component({
+    imports: [FormsModule, TuiPincode],
+    templateUrl: './index.html',
+    encapsulation,
+    changeDetection,
+})
+export default class Example {
+    protected readonly value = signal('');
```
nsbarsukov: Excessive for this example

```suggestion
    protected readonly value = '';
```

In [my earlier comment](https://github.com/taiga-family/taiga-ui/pull/14056#discussion_r3287107432), I was referring only to the complex 3rd/4th/5th/6th examples, not all of them.

nsbarsukov: The same comment for 2nd example

projects/demo/src/pages/components/pincode/index.html
```diff
@@ -0,0 +1,123 @@
...
+                [content]="$index + 1 | tuiExample"
+                [description]="description"
+                [heading]="example"
+            />
+            <ng-template #description>
+                @switch ($index) {
+                    @case (0) {
+                        Basic usage with
+                        <code>[(ngModel)]</code>
+                    }
```
nsbarsukov:
```suggestion
```

projects/demo/src/pages/components/pincode/examples/3/index.ts
```diff
@@ -0,0 +1,37 @@
...
+const FAST_LATENCY = 300;
+
+async function fakeApiVerify(pin: string, abort: AbortSignal): Promise<boolean> {
+    return new Promise<boolean>((resolve, reject) => {
+        const id = setTimeout(() => resolve(pin === CORRECT), FAST_LATENCY);
+
+        abort.addEventListener('abort', () => {
+            clearTimeout(id);
+            reject(new DOMException('Aborted', 'AbortError'));
+        });
```
nsbarsukov: Is the abort logic really necessary in every example?

projects/kit/components/pincode/pincode.component.ts
```diff
@@ -0,0 +1,172 @@
...
+            this.focused() &&
+            index === Math.min(this.value().length, this.el.maxLength - 1)
+        );
+    }
+
+    protected onBeforeInput(event: InputEvent): void {
+        if (this.value().length >= this.el.maxLength) {
+            event.preventDefault();
+        }
+    }
```
nsbarsukov: Do we really need it if `<input>` already contains `maxlength`? It should prevent extra characters automatically

projects/kit/components/pincode/pincode.component.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    protected onSelection(): void {
+        const end = this.el.value.length;
+        const start = end === this.el.maxLength ? end - 1 : end;
+
+        if (this.el.selectionStart !== start || this.el.selectionEnd !== end) {
+            this.el.setSelectionRange(start, end);
+        }
+    }
```
nsbarsukov: https://maskito.dev/kit/plugins#caret-huard

projects/demo/src/pages/components/pincode/index.html
```diff
@@ -0,0 +1,123 @@
...
+            </tr>
+            <tr
+                name="maxlength"
+                tuiDocAPIItem
+                type="number"
+                [items]="lengths"
+                [(value)]="length"
+            >
+                Number of cells. Pending animation triggers once value reaches this length
+            </tr>
```
nsbarsukov, line 103: https://github.com/user-attachments/assets/36c8bbd2-ea2d-4d45-8be2-cdce58c506c2

projects/demo/src/pages/components/pincode/index.html
```diff
@@ -0,0 +1,123 @@
...
+            >
+                Fires after the success wave, before dots collapse. Use to reveal a loader for the next-step transition
+            </tr>
+            <tr
+                name="(finished)"
+                tuiDocAPIItem
+                type="void"
+            >
+                Fires at the end of the verification cycle — after success collapse or after invalid shake + auto-clear
+            </tr>
```
nsbarsukov, line 119: Let's use
https://github.com/taiga-family/taiga-ui/blob/b44ab898d8230cc0b0c6590a725c13ee5f73ebae/projects/addon-doc/components/api/api-item.component.ts#L83

projects/demo/src/pages/components/pincode/index.html
```diff
@@ -0,0 +1,123 @@
...
+                <code>true</code>
+                — success animation,
+                <code>false</code>
+                — shake and auto-clear
+            </tr>
+            <tr
+                name="maxlength"
+                tuiDocAPIItem
+                type="number"
+                [items]="lengths"
```
nsbarsukov: Could we omit `[items]` here to use number-type textfield ?

projects/demo/src/pages/components/pincode/examples/1/index.ts
mdlufy: Fixed

projects/demo/src/pages/components/pincode/index.html
mdlufy: Fixed

projects/demo/src/pages/components/pincode/examples/3/index.ts
mdlufy: No, removed

projects/demo/src/pages/components/pincode/index.html
mdlufy: Updated

projects/demo/src/pages/components/pincode/index.html
mdlufy: Yes, updated

projects/demo/src/pages/components/pincode/index.html
mdlufy: Fixed

projects/kit/components/pincode/pincode.component.ts
mdlufy: Updated

projects/kit/components/pincode/pincode.component.ts
mdlufy: Updated using postprocessor

nsbarsukov (approved):

projects/kit/components/pincode/pincode.style.less
```diff
@@ -0,0 +1,352 @@
...
+        + .t-content
+        .t-item_filled:last-of-type
+        .t-dot_value {
+        animation:
+            tuiPincodeInput 300ms cubic-bezier(0.34, 1.56, 0.64, 1),
+            @pending-wave-anim;
+        animation-delay: 0ms, calc(500ms + var(--t-animation-delay, 0ms)), calc(500ms + var(--t-animation-delay, 0ms));
+    }
+
+    &[data-mode='success']:not([type='password']) + .t-content .t-item_filled .t-value {
```
vladimirpotekhin: what do you think about grouping styles by [data-mode] state?

like 

```
&[data-mode='success'] {
  + .t-content ..... {}
 
   &[type='password'] .... {...}

   &:not([type='password']) ... {...}
}

&[data-mode='invalid'] {
  + .t-content .... {}

   &[type='password'] .... {...}

   &:not([type='password']) ... {...}
}

&[data-mode='pending' {
 ... similar structure...
}

File: projects/kit/components/pincode/pincode.component.ts [hunk 40]
```diff
@@ -0,0 +1,163 @@
...
+        if (isReject) {
+            this.el.value = '';
+            this.el.dispatchEvent(new Event('input'));
+        }
+
+        this.finished.emit();
+    }
+
+    public getStyle(index: number): Record<string, string> {
+        const n = this.el.maxLength;
```
vladimirpotekhin: It looks like all these calculations can be done with CSS

It will be enough to add a CSS index variable to the t-item and length

```
<span
            class="t-item"
            [style.--t-index]="$index"
```

```
host: {[style.--t-maxlength]="el.maxLength"}
```


```
--t-bounce: 400ms;
--t-stager: 100ms;
--t-tail: 300ms;


.t-item {
   --t-animation-delay: calc(-t-index * --t-stager);
  .... 
}
```

projects/kit/components/pincode/pincode.style.less
mdlufy: Updated

projects/kit/components/pincode/pincode.component.ts
mdlufy: Updated

projects/kit/components/pincode/pincode.component.ts
```diff
@@ -0,0 +1,103 @@
...
+    host: {
+        autocomplete: 'one-time-code',
+        inputmode: 'numeric',
+        spellcheck: 'false',
+        '[attr.data-state]': 'state()',
+        '[class._paste]': 'paste()',
+        '[maxLength]': 'maxLength()',
+        '[value]': 'value()',
+        '(input)': 'onInput($event.target.value)',
+    },
```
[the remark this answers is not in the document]
mdlufy: tuiMaskito use tuiDirectiveBinding, so no need to bind it

projects/cdk/classes/control.ts
```diff
@@ -51,7 +51,7 @@ export abstract class TuiControl<T> implements ControlValueAccessor {
 
     public readonly value = computed(() => this.internal() ?? this.fallback);
     public readonly readOnly = input(false);
-    public readonly pseudoInvalid = input<boolean | null>(null, {alias: 'invalid'});
+    public readonly pseudoInvalid = input<boolean | null>(undefined, {alias: 'invalid'});
```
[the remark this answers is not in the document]
waterplea: It's automatic

waterplea (approved):

vladimirpotekhin (approved):

projects/kit/components/pincode/pincode-content.component.ts
```diff
@@ -0,0 +1,24 @@
...
+    templateUrl: './pincode-content.template.html',
+    styleUrl: './pincode.style.less',
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {
+        '[style.--t-length]': 'pincode.maxLength()',
+        '[style.display]': '"contents"',
+        '(animationend)': 'pincode.onAnimationEnd($event)',
+        '(animationstart)': 'pincode.onAnimationStart($event)',
+    },
```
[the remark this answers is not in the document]
mdlufy: A rare case, let's not add it for the sake of it

PR #14088, opened 2026-05-12
feat(addon-mobile): `Keypad` add component

mdlufy:
Fixes #


splincode (approved):

projects/addon-mobile/components/keypad/keypad-input.directive.ts
```diff
@@ -0,0 +1,10 @@
+import {Directive} from '@angular/core';
+
+@Directive({
+    selector: 'input[tuiKeypadInput]',
+    host: {
+        inputmode: 'none',
+        '(keydown.zoneless.prevent)': '(0)',
```
splincode: Consider documenting why keyboard events are suppressed here. The directive sets `inputmode: 'none'` to prevent mobile keyboards, but the keydown handler might confuse developers.

```typescript
// Prevent default keyboard input to force keypad-only interaction
'(keydown.zoneless.prevent)': '(0)',
```

projects/addon-mobile/components/keypad/keypad.component.ts
```diff
@@ -0,0 +1,123 @@
...
+import {takeUntilDestroyed} from '@angular/core/rxjs-interop';
+import {tuiInjectElement} from '@taiga-ui/cdk/utils/dom';
+import {TuiIcon} from '@taiga-ui/core/components/icon';
+import {filter, fromEvent, race, switchMap, take, timer} from 'rxjs';
+
+import {TuiKeypadHostDirective} from './keypad-host.directive';
+
+export type TuiKeypadAction = 'backspace' | 'clear' | 'enter';
+
+// eslint-disable-next-line
```
splincode: The `(string & {})` pattern is used to allow string literals while preserving autocomplete. Consider adding a JSDoc comment to explain this TypeScript pattern.

```typescript
/**
 * Keypad key type. Allows predefined actions or any custom string value.
 * The `(string & {})` pattern preserves autocomplete while allowing extensibility.
 */
// eslint-disable-next-line
export type TuiKeypadKey = TuiKeypadAction | (string & {});
```

projects/addon-mobile/components/keypad/keypad.component.ts
```diff
@@ -0,0 +1,123 @@
...
+    protected readonly columns = computed(() =>
+        Math.max(1, ...this.keys().map((row) => row.length)),
+    );
+
+    protected readonly rows = computed(() => this.keys().length);
+
+    protected readonly templateMap = computed(
+        () => new Map(this.keyTemplates().map((d) => [d.tuiKeypadKey(), d.template])),
+    );
+
```
splincode: The RxJS pipeline for long-press clear functionality can be simplified. The `filter(() => false)` will never emit, which makes the `race` always wait for the timer.

```typescript
constructor() {

    // Long-press backspace clears entire value after 500ms
    fromEvent<PointerEvent>(el, 'pointerdown')
        .pipe(
            filter(({target}) =>
                (target as Element)?.closest('[data-key="backspace"]') !== null,
            ),
            switchMap(() => timer(500).pipe(takeUntilDestroyed())),
            takeUntil(fromEvent(el, 'pointerup')),
        )
        .subscribe(() => this.value.set(''));
}
```

This is clearer: start a 500ms timer on pointerdown, cancel if pointerup happens first.

projects/addon-mobile/components/keypad/keypad.component.ts
```diff
@@ -0,0 +1,123 @@
...
+    public readonly size = input<'fluid' | 'l' | 'm' | 's'>('m');
+
+    public readonly keys =
+        input<ReadonlyArray<readonly TuiKeypadCell[]>>(TUI_DEFAULT_KEYPAD);
+
+    public readonly disabled = input(false, {transform: booleanAttribute});
+    public readonly disabledKeys = input<readonly TuiKeypadKey[]>([]);
+    public readonly activeKey = input<TuiKeypadKey | null>(null);
+    public readonly value = model('');
+    public readonly key = output<TuiKeypadKey>();
```
splincode: The `open` computed signal depends on `TuiKeypadHostDirective` being present. When the directive is absent, it defaults to `true` (always open). This behavior should be documented.

```typescript
/**
 * Keypad component. Automatically opens when host element receives focus
 * (if tuiKeypadHost directive is present), or remains always visible otherwise.
 */
@Component({...})
export class TuiKeypadComponent {
```

projects/demo/src/pages/components/keypad/examples/5/index.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    private calculate(expr: string): string {
+        const sanitized = expr.replaceAll('×', '*').replaceAll('÷', '/');
+
+        if (!/^[\d+\-*/().\s]+$/.test(sanitized)) {
+            throw new Error('Invalid characters');
+        }
+
+        const open = (sanitized.match(/\(/g) ?? []).length;
+        const close = (sanitized.match(/\)/g) ?? []).length;
```
splincode: Using `new Function()` for calculation is a security risk even with sanitization. The regex can potentially be bypassed with certain Unicode characters or edge cases.

Consider using a proper math expression parser or implement a safe evaluator:

```typescript
private calculate(expr: string): string {
    // Use a safe math expression parser library like mathjs
    // or implement a recursive descent parser
    const tokens = this.tokenize(expr);
    const result = this.evaluate(tokens);
    return Number.isFinite(result) ? String(result) : 'Error';
}
```

projects/demo/src/pages/components/keypad/examples/5/index.ts
```diff
@@ -0,0 +1,126 @@
...
+            : `${expr} ${op} `;
+    }
+
+    private appendDot(expr: string): string {
+        const lastNumber = /[\d.]*$/.exec(expr)?.[0] ?? '';
+
+        return lastNumber.includes('.') ? expr : `${expr}.`;
+    }
+
+    private tryCalculate(expr: string): string {
```
splincode, line 108: The regex `/[\d.]*$/` could match an empty string if the expression ends with an operator. This might lead to unexpected behavior like `1+.` being allowed.

```typescript
private appendDot(expr: string): string {
    // Only allow dot if the last number token doesn't already have one
    const match = /(\d+\.?\d*)$/.exec(expr);
    const lastNumber = match?.[1] ?? '';
    
    if (lastNumber.includes('.')) {
        return expr;
    }
    
    return expr === '' ? '0.' : `${expr}.`;
}
```

projects/demo/src/pages/components/keypad/examples/6/index.html
```diff
@@ -0,0 +1,87 @@
...
+            <header tuiHeader="body-l">
+                <hgroup tuiTitle>
+                    <h1>{{ phoneNumber() }}</h1>
+                    <p tuiSubtitle>Calling...</p>
+                </hgroup>
+            </header>
+            <section>
+                <button
+                    appearance="warning"
+                    size="l"
```
splincode: Avatar buttons rely only on `title` attributes for accessibility. Consider adding `aria-label` for better screen reader support.

```html
<button
    appearance="negative"
    size="l"
    aria-label="Hang up call"
    tuiAvatar="@tui.phone-off"
    type="button"
    (click)="closeCall()"
></button>
```

projects/addon-mobile/components/keypad/keypad.component.ts
```diff
@@ -0,0 +1,123 @@
...
+                            take(1),
+                            filter(() => false),
+                        ),
+                    ),
+                ),
+                takeUntilDestroyed(),
+            )
+            .subscribe(() => this.value.set(''));
+    }
+
```
splincode: The method handles value updates but doesn't account for decimal separator localization. The default keypad uses `','` (European), but the calculator example uses `'.'`. Consider documenting this or providing a localization input.

```typescript
/** Decimal separator used in the keypad (default: comma for European format) */
public readonly decimalSeparator = input<string>(',');
```

projects/demo/src/pages/components/keypad/examples/6/index.html
```diff
@@ -0,0 +1,87 @@
+<button
+    iconStart="@tui.phone"
+    tuiButton
+    type="button"
+    (click)="open.set(true)"
+>
+    Open Phone Dialer
```
splincode: I wonder why there's empty space at the bottom

Image

projects/demo/src/pages/components/keypad/examples/5/index.html
```diff
@@ -0,0 +1,32 @@
+<button
+    iconStart="@tui.calculator"
+    tuiButton
+    type="button"
+    (click)="open.set(true)"
+>
+    Open Calculator
```
splincode, line 7: Chrome

Image

Firefox (white background)

Image

projects/addon-mobile/components/keypad/keypad.component.ts
```diff
@@ -0,0 +1,123 @@
...
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {
+        '[attr.data-open]': 'open()',
+        '[attr.data-size]': 'size()',
+        '[attr.disabled]': 'disabled() ? "" : null',
+        '[style.--tui-pad-columns]': 'columns()',
+        '[style.--tui-pad-rows]': 'rows()',
+    },
+})
+export class TuiKeypadComponent {
```
splincode: Consider extracting the size type to a named type for better reusability and documentation:

```typescript
export type TuiKeypadSize = 'fluid' | 'l' | 'm' | 's';

// Then use:
public readonly size = input<TuiKeypadSize>('m');
```

Benefits:
- Easier to reuse in demo pages and tests
- Better IDE autocomplete
- Single source of truth if you need to add/remove sizes
- Follows Taiga UI patterns (like `TuiSize`, `TuiAppearance`)

projects/addon-mobile/components/keypad/keypad.component.less
```diff
@@ -0,0 +1,118 @@
...
+        place-items: center;
+        block-size: 100%;
+        max-inline-size: unset;
+        /* stylelint-disable-next-line plugin/use-baseline -- container queries are intentionally used for fluid sizing */
+        container-type: size;
+
+        .t-key {
+            /* stylelint-disable-next-line plugin/use-baseline, unit-allowed-list, plugin/no-unsupported-browser-features -- container query units are required for fluid mode */
+            --t-size: min(calc(100cqi / var(--tui-pad-columns, 3)), calc(100cqb / var(--tui-pad-rows, 4)));
+
```
splincode: Container queries (`cqi`, `cqb`) have ~94% browser support. For `fluid` mode this is a good choice, but consider adding a fallback for older browsers:

```less
&[data-size='fluid'] {
    // Fallback for browsers without container query support
    --tui-pad-key-width: 5rem;
    --tui-pad-key-height: 4rem;
    
    container-type: size;
    
    .t-key {
        --t-size: min(calc(100cqi / var(--tui-pad-columns, 3)), calc(100cqb / var(--tui-pad-rows, 4)));
        
        inline-size: var(--t-size);
        block-size: var(--t-size);
        
        // Fallback when container queries are not supported
        @supports not (container-type: size) {
            inline-size: var(--tui-pad-key-width);
            block-size: var(--tui-pad-key-height);
            font-size: 1.5rem;
        }
    }
}
```

Alternatively, document that `fluid` size requires modern browsers (Safari 16+, Chrome 105+, Firefox 110+).

splincode: Or is it possible to write simpler and different code here?

projects/addon-mobile/components/keypad/keypad.component.html
```diff
@@ -0,0 +1,34 @@
...
+            } @else {
+                <button
+                    type="button"
+                    class="t-key"
+                    [attr.data-key]="key"
+                    [class.t-key_active]="activeKey() === key"
+                    [disabled]="isKeyDisabled(key)"
+                    (click)="onKeyClick(key)"
+                    (pointerdown.zoneless.prevent)="(0)"
+                >
```
splincode: Mixing old `*ngTemplateOutlet` with new control flow (`@if`, `@for`) creates inconsistency. In Angular 19+, consider using one of these approaches:

**Option 1: Use `@switch` (cleanest)**
```html
@switch (templateMap().get(key)) {
  @case (tpl) {
    <ng-container *ngTemplateOutlet="tpl; context: {$implicit: key, key: key}" />
  }
  @default {
    @if (key === 'backspace') {
      <tui-icon [icon]="'@tui.chevron-left'" />
    } @else if ...
  }
}
```

**Option 2: Extract to separate directive**
```typescript
@Directive({selector: '[tuiKeypadKeyContent]'})
export class TuiKeypadKeyContentDirective {
  public readonly key = input.required<TuiKeypadKey>();
}
```

**Option 3: Keep as-is but add TODO**
If this is temporary, add a comment:
```html
<!-- TODO: migrate to @switch when Angular 19+ control flow supports template context -->
```

This will improve consistency and make the template easier to maintain.

projects/demo/src/pages/components/keypad/index.ts
```diff
@@ -0,0 +1,65 @@
...
+    changeDetection,
+})
+export default class Page {
+    private readonly alerts = inject(TuiNotificationService);
+
+    protected readonly examples = [
+        'Basic keypad',
+        'Customize',
+        'Collapsible',
+        'Fluid',
```
splincode: Is it expected?

Image

projects/demo/src/pages/components/keypad/examples/5/index.ts
splincode: Using `new Function()` is dangerous even with regex validation, and this example is part of the public documentation.

**Why this is a problem:**
1. This example will be copied by users into production code
2. Regex validation can be bypassed with edge cases (Unicode characters, prototype pollution, etc.)
3. It sets a bad precedent for the community

**Recommended solutions:**

**Option 1: Use a safe math parser library**
```typescript
import { Parser } from 'expr-eval';
const parser = new Parser();
const result = parser.evaluate(expr);
```

**Option 2: Implement a simple recursive descent parser**
```typescript
private calculate(expr: string): string {
  // Tokenize → Parse → Evaluate
  // No eval, no Function, no security risks
}
```

**Option 3: At minimum, add a prominent warning**
```typescript
/**
 * WARNING: This is a demo example only.
 * Do NOT use new Function() in production code.
 * Use a proper math expression parser library instead.
 */
private calculate(expr: string): string {
  // ...
}
```

Please consider refactoring this example before merging, as it will be referenced by users building production applications.

projects/demo/src/pages/components/keypad/examples/3/index.html
```diff
@@ -0,0 +1,71 @@
...
+>
+    Open form
+</button>
+
+<ng-template
+    let-observer
+    [tuiSheetDialogOptions]="options"
+    [(tuiSheetDialog)]="open"
+>
+    <div
```
splincode: There's an inconsistency across demo examples regarding `tuiKeypadHost` usage:

- **Example 3**: Uses `tuiKeypadHost` (keypad auto-opens on focus)
- **Examples 1, 2, 4, 5, 6**: No `tuiKeypadHost` (keypad is always visible)

This creates confusion for users trying to understand when the directive is needed.

**Suggestion:**

1. **Add a note to the documentation** explaining the difference:
```html
<!-- Use tuiKeypadHost when you want the keypad to auto-open on input focus -->
<div tuiKeypadHost>
  <input tuiKeypadInput />
  <tui-keypad />
</div>

<!-- Omit tuiKeypadHost when you want the keypad to be always visible -->
<tui-keypad />
```

2. **Or add a comment in example 3**:
```html
<!-- tuiKeypadHost enables auto-open behavior when the input receives focus -->
<div tuiKeypadHost class="sheet">
```

This will help users understand the intended use case for each approach.

projects/demo/src/pages/components/keypad/examples/5/index.ts
splincode: But if you think it's excessive, forget it.

projects/addon-mobile/components/keypad/keypad.component.ts
```diff
@@ -0,0 +1,123 @@
...
+    private readonly host = inject(TuiKeypadHostDirective, {optional: true});
+    public readonly size = input<'fluid' | 'l' | 'm' | 's'>('m');
```
splincode: The `activeKey` input is only used for styling (`[class.t-key_active]`), but there's no logic to manage active key state (e.g., keyboard navigation, focus management).

**Questions:**
1. Is this a planned feature for keyboard accessibility?
2. Should arrow keys update `activeKey`?
3. Should Enter/Space trigger the active key?

**If this is intentional:** Add a comment explaining the future use case.

**If not needed:** Consider removing it to simplify the API:
```typescript
// Remove this input if not used for keyboard navigation
// public readonly activeKey = input<TuiKeypadKey | null>(null);
```

Leaving unused inputs in the public API can confuse users about the component's capabilities.

projects/addon-mobile/components/keypad/keypad.component.html
```diff
@@ -0,0 +1,34 @@
...
+        class="t-row"
+        [style.--i]="$index"
+    >
+        @for (key of row; track $index) {
+            @if (key === null) {
+                <div class="t-key t-key_spacer"></div>
+            } @else {
+                <button
+                    type="button"
+                    class="t-key"
```
splincode: The text "AC" for the clear button is hardcoded, which creates issues for localization:

```html
@else if (key === 'clear') {
    <span>AC</span>
}
```

**Suggestion:** Use an input or token for localization:

```typescript
// Component
public readonly clearLabel = input<string>('AC');
```

```html
@else if (key === 'clear') {
    <span>{{ clearLabel() }}</span>
}
```

**Alternative:** Use a data attribute and CSS content (better for i18n):
```html
<span data-key-label="clear"></span>
```

```less
[data-key-label='clear']::after {
    content: 'AC'; // Can be overridden via CSS custom properties
}
```

This allows users to customize the label for different locales without modifying the template.

projects/addon-mobile/components/keypad/keypad.component.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export type TuiKeypadAction = 'backspace' | 'clear' | 'enter';
+
+// eslint-disable-next-line
+export type TuiKeypadKey = TuiKeypadAction | (string & {});
+
+export type TuiKeypadCell = TuiKeypadKey | null;
+
+export const TUI_DEFAULT_KEYPAD: ReadonlyArray<readonly TuiKeypadCell[]> = [
+    ['1', '2', '3'],
```
splincode: This PR introduces a new component with non-trivial logic, but no test files are included.

**Critical test scenarios to cover:**

1. **Button clicks**
```typescript
it('should emit value on key click', () => {
  const fixture = TestBed.createComponent(TuiKeypadComponent);
  fixture.detectChanges();
  
  const button = fixture.nativeElement.querySelector('[data-key="1"]');
  button.click();
  
  expect(fixture.componentInstance.value()).toBe('1');
});
```

2. **Long-press backspace**
```typescript
it('should clear value on long-press backspace', fakeAsync(() => {
  const fixture = TestBed.createComponent(TuiKeypadComponent);
  fixture.componentInstance.value.set('123');
  fixture.detectChanges();
  
  const backspace = fixture.nativeElement.querySelector('[data-key="backspace"]');
  backspace.dispatchEvent(new PointerEvent('pointerdown'));
  tick(500); // Wait for long-press threshold
  
  expect(fixture.componentInstance.value()).toBe('');
}));
```

3. **Value updates via model**
```typescript
it('should update UI when value changes externally', () => {
  const fixture = TestBed.createComponent(TuiKeypadComponent);
  fixture.componentInstance.value.set('42');
  fixture.detectChanges();
  
  // Assert UI reflects the new value
});
```

4. **tuiKeypadInput directive**
```typescript
it('should set inputmode to none', () => {
  const fixture = TestBed.createComponent(TuiKeypadInputDirective);
  const input = fixture.nativeElement as HTMLInputElement;
  
  expect(input.inputMode).toBe('none');
});
```

Please add at least basic test coverage before merging to ensure the component behaves correctly across updates.

projects/addon-mobile/components/keypad/keypad.component.html
splincode: Or add something to `taiga-ui/i18n`

projects/addon-mobile/components/keypad/keypad.component.ts
splincode: Maybe we can add just e2e or cypress component tests?

projects/addon-mobile/components/keypad/keypad.component.ts
```diff
@@ -0,0 +1,123 @@
...
+
+@Directive({selector: 'ng-template[tuiKeypadKey]'})
+export class TuiKeypadKeyDirective {
+    public readonly tuiKeypadKey = input.required<TuiKeypadKey>();
+    public readonly template = inject<TemplateRef<unknown>>(TemplateRef);
+}
+
+@Component({
+    selector: 'tui-keypad',
+    imports: [NgTemplateOutlet, TuiIcon],
```
splincode: Question about file structure: `TuiKeypadKeyDirective` is defined in the same file as `TuiKeypadComponent`, while `TuiKeypadHostDirective` and `TuiKeypadInputDirective` are in separate files.

**Current structure:**
```
keypad.component.ts       → TuiKeypadComponent, TuiKeypadKeyDirective
keypad-host.directive.ts  → TuiKeypadHostDirective
keypad-input.directive.ts → TuiKeypadInputDirective
```

**Considerations:**

1. **Keep as-is** if `TuiKeypadKeyDirective` is tightly coupled to the component (only used internally for template customization)

2. **Extract to separate file** if:
   - It might be used independently
   - You want consistent file-per-directive structure
   - Easier to import individually in tests

**Recommendation:** For consistency with `TuiKeypadHostDirective` and `TuiKeypadInputDirective`, consider extracting `TuiKeypadKeyDirective` to `keypad-key.directive.ts`. This follows the single-responsibility principle and makes the codebase more predictable.

```typescript
// keypad-key.directive.ts
import {Directive, input, TemplateRef, inject} from '@angular/core';
import {TuiKeypadKey} from './keypad.component';

@Directive({selector: 'ng-template[tuiKeypadKey]'})
export class TuiKeypadKeyDirective {
    public readonly tuiKeypadKey = input.required<TuiKeypadKey>();
    public readonly template = inject<TemplateRef<unknown>>(TemplateRef);
}
```

This is a minor point, but consistency in file structure helps with discoverability and maintenance.

projects/experimental/components/keypad/ng-package.json
```diff
@@ -0,0 +1,8 @@
+{
+    "$schema": "../../../node_modules/ng-packagr/ng-package.schema.json",
+    "allowedNonPeerDependencies": ["."],
+    "dest": "../../../../dist/experimental/components/pad",
```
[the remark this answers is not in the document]
mdlufy: Not applicable anymore — the component moved to `addon-mobile`, whose `ng-package.json` has no `dest`.

projects/cdk/constants/used-icons.ts
```diff
@@ -30,6 +30,7 @@ export const TUI_USED_ICONS = [
     '@tui.circle-help',
     '@tui.circle-x',
     '@tui.clock',
+    '@tui.equal',
```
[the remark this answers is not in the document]
mdlufy: Done — added `@tui.delete` (and `@tui.equal`) to the used-icons lists.

projects/demo/src/pages/components/keypad/examples/2/index.less
```diff
@@ -0,0 +1,19 @@
+.container {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    gap: 1.5rem;
+
+    tui-pad {
```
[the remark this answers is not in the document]
mdlufy: Resolved — the example targets a `.custom` class now, not an element selector.

projects/addon-mobile/components/keypad/keypad-input.directive.ts
mdlufy: Done — added comments in `keypad-input.directive.ts` explaining both `inputmode: 'none'` (suppress the native mobile keyboard) and the suppressed `keydown` (the on-screen keypad is the single source of truth).

projects/addon-mobile/components/keypad/keypad.component.ts
mdlufy: Done — added a JSDoc explaining the `(string & {})` trick. The type now lives in `keypad.options.ts`.

projects/addon-mobile/components/keypad/keypad.component.ts
mdlufy: Reworked — dropped the RxJS pipeline entirely and use the `(longtap)` event plugin on the backspace key instead.

projects/addon-mobile/components/keypad/keypad.component.ts
mdlufy: Reworked per review — removed `open` and `TuiKeypadHostDirective`. Visibility is now consumer-driven via `@if (input.focused())`, and `focused()` is exposed on `tuiKeypadInput`.

projects/demo/src/pages/components/keypad/examples/5/index.ts
mdlufy: Done — replaced `new Function()` with a small safe shunting-yard evaluator (no `eval`/`Function`).

projects/demo/src/pages/components/keypad/examples/5/index.ts
mdlufy: Done — `appendDot` now only appends a dot to the current number segment (and starts `0.` when there is none).

projects/demo/src/pages/components/keypad/examples/6/index.html
mdlufy: Done — added `aria-label` to the icon-only avatar buttons.

projects/addon-mobile/components/keypad/keypad.component.ts
mdlufy: The keys (including the separator) are fully configurable via `[keys]` / `TUI_KEYPAD_OPTIONS`; the value logic just appends the pressed key, so the separator is whatever key you pass. Left out a dedicated input to keep the API small.

projects/demo/src/pages/components/keypad/examples/6/index.html
mdlufy: Should be resolved — the fixed `size` modes are gone; the pad is now fluid (fills its container, capped at 600px) with a fixed height, so there's no leftover space. Please re-check on the latest commit.

projects/demo/src/pages/components/keypad/examples/5/index.html
mdlufy: Styles were reworked (single fluid layout, hover via `@tui-mouse`, standard `tuiCollapse` + `tuiFade`). Please re-verify on the latest commit.

projects/addon-mobile/components/keypad/keypad.component.ts
mdlufy: `size` was removed entirely (the pad is fluid now), so this no longer applies.

projects/addon-mobile/components/keypad/keypad.component.less
mdlufy: Fluid is the only sizing mode now, so the code is simpler (single layout, no `s/m/l` variants). Container queries are within Taiga's supported baseline, so I kept them without a `@supports` fallback.

projects/addon-mobile/components/keypad/keypad.component.html
mdlufy: `*ngTemplateOutlet` is kept only to project the user's `tuiKeypadKey` template with context (built-in control flow can't render a `TemplateRef`); everything else uses `@if` / `@for` / `@let`.

projects/demo/src/pages/components/keypad/index.ts
mdlufy: This was tied to the `size` variants, which are removed now. Please re-check on the latest commit.

projects/demo/src/pages/components/keypad/examples/5/index.ts
mdlufy: Done — replaced with a safe shunting-yard evaluator, no `new Function()`.

projects/demo/src/pages/components/keypad/examples/3/index.html
mdlufy: Resolved — `tuiKeypadHost` is removed. All examples are consistent now; the collapsible one shows/hides with `@if (input.focused())` + `tuiCollapse`/`tuiFade`.

projects/addon-mobile/components/keypad/keypad.component.ts
mdlufy: `activeKey` is an intentional styling hook so consumers can highlight a key from the outside; kept minimal (no built-in navigation).

projects/addon-mobile/components/keypad/keypad.component.html
mdlufy: `AC` (and the backspace/enter icons) now come from `TUI_KEYPAD_OPTIONS`, so they're overridable per app; didn't wire a dedicated `@taiga-ui/i18n` token for now.

projects/addon-mobile/components/keypad/keypad.component.ts
mdlufy: Done — added `keypad.component.spec.ts`: creation, key click / backspace / clear / enter, long-press clear, plus a mobile-mode block (pointerdown prevented, `inputmode=none`). The component is also covered by the demo e2e (deep-select / a11y).

projects/addon-mobile/components/keypad/keypad.component.ts
mdlufy: Done — extracted `TuiKeypadKeyDirective` to `keypad-key.directive.ts`.

projects/addon-mobile/components/keypad/keypad-input.directive.ts
```diff
@@ -0,0 +1,30 @@
...
+        const mutatesValue =
+            !event.ctrlKey &&
+            !event.metaKey &&
+            !event.altKey &&
+            (event.key.length === 1 ||
+                event.key === 'Backspace' ||
+                event.key === 'Delete');
+
+        if (mutatesValue) {
+            event.preventDefault();
```
nsbarsukov: https://github.com/user-attachments/assets/7e4d6cff-b2e4-440e-b3d6-ebc1f44cc36a

projects/demo/src/pages/components/keypad/index.ts
```diff
@@ -0,0 +1,57 @@
...
+        'Calculator',
+        'Phone Dialer',
```
nsbarsukov: Unexpected zoom after "phone call is finished"

https://github.com/user-attachments/assets/5eb35554-d23b-4d4e-9e86-fc33dd7b6c1e

projects/demo/src/pages/components/keypad/examples/5/index.ts
```diff
@@ -0,0 +1,217 @@
...
+import {TuiKeypad, type TuiKeypadKey} from '@taiga-ui/addon-mobile';
+import {TuiAutoFocus} from '@taiga-ui/cdk';
+import {TuiButton} from '@taiga-ui/core';
+import {TuiPopout} from '@taiga-ui/experimental';
+
+const KEYBOARD_MAP: Record<string, TuiKeypadKey> = {
+    '+': '+',
+    '-': '-',
+    '*': '×',
+    '/': '÷',
```
nsbarsukov, line 13: Image

projects/demo/src/pages/components/keypad/examples/3/index.html
```diff
@@ -0,0 +1,79 @@
...
+<ng-template
+    let-observer
+    [tuiSheetDialogOptions]="options"
+    [(tuiSheetDialog)]="open"
+>
+    <div class="sheet">
+        <div class="top">
+            <input
+                #input="tuiKeypadInput"
+                readonly
```
nsbarsukov: Is it actually required ? Is it okay for a11y ?

projects/addon-mobile/components/keypad/keypad.options.ts
```diff
@@ -0,0 +1,35 @@
...
+    ['7', '8', '9'],
+    [',', '0', 'backspace'],
+];
+
+export interface TuiKeypadOptions {
+    readonly keys: ReadonlyArray<readonly TuiKeypadCell[]>;
+    readonly icons: Record<TuiKeypadAction, string>;
+}
+
+export const [TUI_KEYPAD_OPTIONS, tuiKeypadOptionsProvider] =
```
nsbarsukov: Do you have any documentation about it ?

projects/addon-mobile/components/keypad/keypad-input.directive.ts
```diff
@@ -0,0 +1,18 @@
...
+        // Suppress the native mobile keyboard: the on-screen keypad is the only value source
+        inputmode: 'none',
+        // Block every value mutation (typing, paste, IME, drop); beforeinput never fires for
+        // navigation/copy, so focus isn't trapped and shortcuts keep working
+        '(beforeinput.zoneless.prevent)': '(0)',
+    },
+})
+export class TuiKeypadInputDirective {
+    public readonly focused = tuiFocusedIn(tuiInjectElement());
+}
```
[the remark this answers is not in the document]
mdlufy: `tuiFocusedIn()` already returns a `Signal<boolean>` — it wraps `toSignal(...)` internally (see `projects/cdk/utils/focus/focused-in.ts`). So `focused` is a signal and calling `input.focused()` in the template is correct; there is no `Observable` here. Wrapping it again in `toSignal` would double-wrap a signal, so no change is needed.

projects/addon-mobile/components/keypad/keypad-input.directive.ts
mdlufy: Fixed

projects/demo/src/pages/components/keypad/index.ts
mdlufy: Fixed

projects/demo/src/pages/components/keypad/examples/5/index.ts
mdlufy: Fixed

projects/demo/src/pages/components/keypad/examples/3/index.html
mdlufy: No, removed

projects/addon-mobile/components/keypad/keypad.options.ts
mdlufy: Good catch, docs updated

projects/addon-mobile/components/keypad/keypad.options.ts
```diff
@@ -0,0 +1,30 @@
...
+    tuiCreateOptions<TuiKeypadOptions>({
+        keys: TUI_DEFAULT_KEYPAD,
+        icons: {
+            backspace: '@tui.delete',
+            clear: 'AC',
+            enter: '@tui.equal',
```
waterplea: These two seem calculator specific, while calculator is just a specific use case, maybe we shouldn't have those in default options?

projects/addon-mobile/components/keypad/keypad.component.less
```diff
@@ -0,0 +1,71 @@
...
+    inline-size: 100%;
+    block-size: 100%;
+    max-inline-size: 37.5rem;
+    margin-inline: auto;
+    aspect-ratio: var(--t-columns, 3) / var(--t-rows, 4);
+    /* stylelint-disable-next-line plugin/use-baseline -- container queries are used for fluid sizing */
+    container-type: size;
+}
+
+.t-row {
```
waterplea: We do not need this element, we can rely on $first in the second @for to determine first child in a row

projects/addon-mobile/components/keypad/keypad.component.less
```diff
@@ -0,0 +1,71 @@
...
+    & > :first-child {
+        grid-column-start: 1;
+    }
+}
+
+.t-key {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    aspect-ratio: 90 / 84;
```
waterplea: Why not 1?

projects/addon-mobile/components/keypad/keypad.component.less
```diff
@@ -0,0 +1,71 @@
...
+    /* stylelint-disable-next-line plugin/use-baseline, unit-allowed-list -- container query units size the oval to the cell */
+    inline-size: min(calc(90cqi / var(--t-columns, 3)), calc(96cqb / var(--t-rows, 4)));
+    block-size: auto;
+    border: none;
+    border-radius: 50%;
+    background: transparent;
+    font: var(--tui-keypad-font, 600 2rem/1 var(--tui-typography-family-display));
+    /* stylelint-disable-next-line plugin/use-baseline, unit-allowed-list, plugin/no-unsupported-browser-features -- container query units are required for fluid sizing */
+    font-size: clamp(1.5rem, min(7cqi, 7cqb), 3rem);
+    letter-spacing: var(--tui-keypad-letter-spacing, 0.0225rem);
```
waterplea: Do we really need this?

projects/addon-mobile/components/keypad/keypad.component.less
```diff
@@ -0,0 +1,71 @@
...
+    align-items: center;
+    justify-content: center;
+    aspect-ratio: 90 / 84;
+    /* stylelint-disable-next-line plugin/use-baseline, unit-allowed-list -- container query units size the oval to the cell */
+    inline-size: min(calc(90cqi / var(--t-columns, 3)), calc(96cqb / var(--t-rows, 4)));
+    block-size: auto;
+    border: none;
+    border-radius: 50%;
+    background: transparent;
+    font: var(--tui-keypad-font, 600 2rem/1 var(--tui-typography-family-display));
```
waterplea: Let's just use inherit and apply default value on the :host and then people can override it without extra variables in the public API

projects/addon-mobile/components/keypad/keypad.component.less
```diff
@@ -0,0 +1,71 @@
...
+    grid-template-columns: repeat(var(--t-columns, 3), 1fr);
+    grid-auto-rows: 1fr;
+    place-items: center;
+    inline-size: 100%;
+    block-size: 100%;
+    max-inline-size: 37.5rem;
+    margin-inline: auto;
+    aspect-ratio: var(--t-columns, 3) / var(--t-rows, 4);
+    /* stylelint-disable-next-line plugin/use-baseline -- container queries are used for fluid sizing */
+    container-type: size;
```
waterplea: Are we sure there's no way to avoid container queries here?

projects/addon-mobile/components/keypad/keypad.component.html
```diff
@@ -0,0 +1,33 @@
...
+                    [attr.aria-label]="ariaLabel(key)"
+                    [attr.data-key]="key"
+                    [class.t-key_active]="activeKey() === key"
+                    [disabled]="isKeyDisabled(key)"
+                    (click)="onKeyClick(key)"
+                    (longtap)="onLongTap(key)"
+                    (mousedown.zoneless.prevent)="(0)"
+                    (pointerdown.zoneless.prevent)="(0)"
```
waterplea: There's no need for both, also let's move it to host, otherwise when I tap between the keys the collapsable example collapses

projects/addon-mobile/components/keypad/keypad.component.less
```diff
@@ -0,0 +1,71 @@
...
+        transform: scale(1.14);
+    }
+
+    &:focus-visible {
+        outline: 0.125rem solid var(--tui-border-focus);
+        outline-offset: 0.125rem;
+    }
+
+    &:disabled {
+        opacity: 0.4;
```
waterplea: --tui-disabled-opacity?

projects/addon-mobile/components/keypad/keypad.component.ts
```diff
@@ -0,0 +1,91 @@
...
+import {TuiKeypadKeyDirective} from './keypad-key.directive';
+
+@Component({
+    selector: 'tui-keypad',
+    imports: [NgTemplateOutlet, TuiIcon],
+    templateUrl: './keypad.component.html',
+    styleUrl: './keypad.component.less',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {
+        '[attr.disabled]': 'disabled() ? "" : null',
```
waterplea: What's the point of global disable? Why even show the keypad if it's disabled? What's the point of disabled attr on host?

projects/addon-mobile/components/keypad/keypad.component.ts
```diff
@@ -0,0 +1,91 @@
...
+        return this.disabled() || this.disabledKeys().includes(key);
+    }
+
+    protected content(key: TuiKeypadKey): string {
+        const icons = this.options.icons as Record<string, string | undefined>;
+
+        return icons[key] ?? key;
+    }
+
+    protected ariaLabel(key: TuiKeypadKey): string | null {
```
waterplea: I'm skeptical about introducing ariaLabel as public API. We already have a way to customize content of a key, users can put accessible labels inside there themselves, wouldn't that be enough?

projects/addon-mobile/components/keypad/keypad.component.ts
```diff
@@ -0,0 +1,91 @@
...
+        '[style.--t-rows]': 'rows()',
+    },
+})
+export class TuiKeypadComponent {
+    protected readonly options = inject(TUI_KEYPAD_OPTIONS);
+
+    public readonly keys = input(this.options.keys);
+    public readonly disabled = input(false, {transform: booleanAttribute});
+    public readonly disabledKeys = input<readonly TuiKeypadKey[]>([]);
+    public readonly activeKey = input<TuiKeypadKey | null>(null);
```
waterplea: What's the point of activeKey?

projects/addon-mobile/components/keypad/keypad-input.directive.ts
```diff
@@ -0,0 +1,15 @@
+import {Directive} from '@angular/core';
+import {tuiInjectElement} from '@taiga-ui/cdk/utils/dom';
+import {tuiFocusedIn} from '@taiga-ui/cdk/utils/focus';
+
+@Directive({
+    selector: 'input[tuiKeypadInput]',
```
waterplea: Can people just use tuiSelectLike and (focus)/(blur) listeners to toggle open signal? Looks like a directive that's not crucial to have.

projects/addon-mobile/components/keypad/keypad.options.ts
mdlufy: Dropped the whole options token, so there are no default icons/layout to be calculator-specific anymore. The calculator wires its own keys.

projects/addon-mobile/components/keypad/keypad.component.less
mdlufy: No row wrappers left, it's a flat grid of projected buttons now.

projects/addon-mobile/components/keypad/keypad.component.less
mdlufy: Yeah, it's aspect-ratio: 1 now (moved onto the key component).

projects/addon-mobile/components/keypad/keypad.component.less
mdlufy: Nope, removed.

projects/addon-mobile/components/keypad/keypad.component.less
mdlufy: Went with that: default font/color on :host, the key just inherits, so you override font on the tui-keypad. No extra css vars.

projects/addon-mobile/components/keypad/keypad.component.less
mdlufy: Got rid of them. It's plain grid + aspect-ratio: 1 on the key now, scales fine in both dimensions.

projects/addon-mobile/components/keypad/keypad.component.html
mdlufy: Moved to a single pointerdown.prevent on the keypad host, so tapping the gaps between keys no longer blurs the field (fixes the collapse).

projects/addon-mobile/components/keypad/keypad.component.less
mdlufy: Yep, uses --tui-disabled-opacity now.

projects/addon-mobile/components/keypad/keypad.component.ts
mdlufy: Agreed, gone. No disabled input/attr on the component, you just disable the native buttons you project.

projects/addon-mobile/components/keypad/keypad.component.ts
mdlufy: Dropped it. The label lives on the projected button (aria-label or its content), nothing on the component.

projects/addon-mobile/components/keypad/keypad.component.ts
mdlufy: Removed.

projects/addon-mobile/components/keypad/keypad-input.directive.ts
mdlufy: The open toggling is already gone. What's left here is inputmode=none plus blocking beforeinput (paste/IME/hardware typing) so the pad stays the only input source, and it exposes focused() for the collapsible example. tuiSelectLike doesn't cover that part, so I kept it.

projects/addon-mobile/components/keypad/keypad-input.directive.ts
```diff
@@ -0,0 +1,19 @@
...
+        // stops typing/paste/IME from mutating the field so the on-screen keypad is
+        // the single source of truth. `keydown` is left alone to preserve navigation
+        // (Tab/arrows/Escape) and shortcuts — no keyboard trap (WCAG 2.1.2).
+        inputmode: 'none',
+        '(beforeinput.zoneless.prevent)': '(0)',
+    },
+})
+export class TuiKeypadInputDirective {
+    public readonly focused = tuiFocusedIn(tuiInjectElement());
+}
```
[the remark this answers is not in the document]
mdlufy: False positive: tuiFocusedIn already returns a Signal<boolean> (it wraps toSignal internally), so input.focused() in the template is correct. Wrapping it in toSignal again would double-wrap a signal. Verified the collapsible example works.

projects/demo/src/pages/components/keypad/examples/3/index.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    protected clear(): void {
+        this.pin.set('');
+        this.digits.set(this.shuffle());
+    }
+
+    private shuffle(): string[] {
+        return [...DIGITS].sort(() => Math.random() - 0.5);
+    }
```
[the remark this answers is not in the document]
mdlufy: Done in ccd16a21297 — switched to a uniform shuffle (random key per digit, then sort) instead of the biased comparator.

projects/addon-mobile/components/keypad/keypad.component.ts
```diff
@@ -0,0 +1,32 @@
...
+    },
+})
+export class TuiKeypadComponent {
+    protected readonly buttons = contentChildren(TuiKeypadButton);
+
+    public readonly columns = input(3);
+
+    protected readonly rows = computed(() =>
+        Math.max(1, Math.ceil(this.buttons().length / this.columns())),
+    );
```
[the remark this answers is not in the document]
mdlufy: Done in ccd16a21297 — clamped the divisor with Math.max(1, columns()) so columns=0 can't produce Infinity rows.

projects/addon-mobile/components/keypad/keypad.component.ts
```diff
@@ -0,0 +1,32 @@
...
+    host: {
+        '[style.--t-columns]': 'columns()',
+        '[style.--t-rows]': 'rows()',
+        // A tap anywhere on the pad (keys or the gaps between them) must not blur the
+        // field the keypad drives — prevent it once here rather than on each key
+        '(pointerdown.zoneless.prevent)': '(0)',
+    },
+})
+export class TuiKeypadComponent {
+    protected readonly buttons = contentChildren(TuiKeypadButton);
```
nsbarsukov: Private ?

projects/demo/src/pages/components/keypad/examples/3/index.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    protected clear(): void {
+        this.pin.set('');
+        this.digits.set(this.shuffle());
+    }
+
+    private shuffle(): string[] {
+        // uniform shuffle: independent random key per digit, then sort by it
+        return DIGITS.map((digit) => [Math.random(), digit] as const)
```
nsbarsukov, line 40: Flaky `demo.pw.spec.ts` ?

projects/demo/src/pages/components/keypad/index.less
```diff
@@ -0,0 +1,4 @@
+tui-keypad {
+    inline-size: 15rem;
+    block-size: 20rem;
```
nsbarsukov: With strictly defined height `columns=4` is not properly displayed on mobile devices

https://taiga-preview-demo--pr14088-keypad-demo-ou1uk910.web.app/components/keypad/API?columns=4

projects/demo/src/pages/components/keypad/examples/6/index.html
```diff
@@ -0,0 +1,180 @@
...
+                (click)="onKey('5')"
+            >
+                5
+            </button>
+            <button
+                tuiKeypadButton
+                type="button"
+                (click)="onKey('6')"
+            >
+                6
```
nsbarsukov:
```html
<button ...>
    <tui-icon icon="@tui.divide" />
</button>

@for (digit of [7, 8, 9]; track $index) {
    <button
        tuiKeypadButton
        type="button"
        (click)="onKey(digit)"
    >
        {{ digit }}
    </button>
}

<button ...>
    <tui-icon icon="@tui.x" />
</button>

@for (digit of [4, 5, 6]; track $index) {
    <button
        tuiKeypadButton
        type="button"
        (click)="onKey(digit)"
    >
        {{ digit }}
    </button>
}

// [...]
```

projects/demo/src/pages/components/keypad/examples/2/index.html
```diff
@@ -0,0 +1,42 @@
+<div class="container">
```
nsbarsukov: Is this required? It seems like `:host {...}` could be used here instead.

projects/demo/src/pages/components/keypad/examples/4/index.less
```diff
@@ -0,0 +1,81 @@
...
+.amount {
+    font: bold 2.5rem/1 var(--tui-typography-family-display);
+    text-align: center;
+    color: var(--tui-text-primary);
+    border: none;
+    outline: none;
+    background: transparent;
+    padding: 0;
+    inline-size: 100%;
+    caret-color: transparent; // keypad-driven display; avoid iOS caret glitches
```
nsbarsukov: Why `tuiKeypadInput` is not used here ?

projects/demo/src/pages/components/keypad/index.ts
```diff
@@ -0,0 +1,40 @@
...
+    styleUrl: './index.less',
+    changeDetection,
+})
+export default class Page {
+    protected readonly examples = [
+        'Basic keypad',
+        'Custom keys',
+        'Shuffled PIN',
+        'Collapsible',
+        'Fluid',
```
nsbarsukov: Should this case be handled ?

https://github.com/user-attachments/assets/cca4aa6e-ef4a-4432-9523-86d6c1aee61d

projects/addon-mobile/components/keypad/keypad.component.ts
mdlufy: Done, made it private.

projects/demo/src/pages/components/keypad/examples/3/index.ts
mdlufy: Good catch — replaced `Math.random()` with a seeded shuffle so the order is reproducible in snapshots, still reshuffles on clear.

projects/demo/src/pages/components/keypad/index.less
mdlufy: Fixed — keys were sized off the row height, so a fixed height made them overflow the column and clip the neighbours. They size from the column width now (with a max on height), and I dropped the hard-coded height here so it follows the width. columns=4 renders clean on mobile.

projects/demo/src/pages/components/keypad/examples/6/index.html
mdlufy: Done, the digit rows are `@for` loops now.

projects/demo/src/pages/components/keypad/examples/2/index.html
mdlufy: Yeah, dropped the wrapper div and moved the layout onto `:host`.

projects/demo/src/pages/components/keypad/examples/4/index.less
mdlufy: It is used — the field has `tuiKeypadInput`. This `caret-color: transparent` was a leftover local hack; I moved it into the directive itself, so any keypad-driven field hides the caret without extra CSS. Dropped it here.

projects/demo/src/pages/components/keypad/index.ts
mdlufy: Thanks for the video. Dragging the resizer past the demo bounds let the pad grow wider/taller than the viewport, so the right/bottom got clipped. Capped the resizable container at 100% in both axes so it cannot be pulled off-screen — it still scales down freely.

projects/demo/src/pages/components/keypad/examples/3/index.ts
```diff
@@ -0,0 +1,56 @@
...
+    imports: [TuiIcon, TuiKeypad],
+    templateUrl: './index.html',
+    styleUrl: './index.less',
+    encapsulation,
+    changeDetection,
+})
+export default class Example {
+    // Reshuffled on every clear — like PIN terminals that reorder the pad. Seeded (not
+    // Math.random) so the rendered order is reproducible and demo snapshots stay stable.
+    private seed = 1;
```
nsbarsukov: My mistake, `Math.random()` is okay for E2E test

https://github.com/taiga-family/taiga-ui/blob/fc100502feed5c63b11bb7c02d13957f59a34e17/projects/demo-playwright/utils/goto.ts#L33-L35

Let's use it again instead of overcomplicated (just for demo example) logic inside this file

nsbarsukov (approved):

projects/demo/src/pages/components/keypad/examples/3/index.ts
mdlufy: Reverted to `Math.random()` — thanks!

projects/addon-mobile/components/keypad/keypad-button.component.ts
```diff
@@ -0,0 +1,9 @@
+import {ChangeDetectionStrategy, Component} from '@angular/core';
+
+@Component({
+    selector: 'button[tuiKeypadButton], a[tuiKeypadButton]',
```
waterplea: Do we need this component? Maybe just apply the styles to any child of keypad, like we do in segmented without child driectives/components?

projects/addon-mobile/components/keypad/keypad-input.directive.ts
waterplea: I don't think it's worth it. tuiSelectLike is enough for everything, other than focus and if people need it – they can add it themselves by just listening to (focus)/(blur) or just using CSS

projects/demo/src/pages/components/keypad/index.html
```diff
@@ -0,0 +1,145 @@
...
+                            The keypad expands while the amount field is focused, inside a sheet.
+                        }
+                        @case (4) {
+                            Drag the corner — the keypad scales to fit its container.
+                        }
+                        @case (5) {
+                            A calculator in a picture-in-picture window — icon keys and hardware-keyboard mapping.
+                        }
+                        @case (6) {
+                            A phone dialer in a picture-in-picture window — a field driven by
```
waterplea: Let's drop this example, a calculator is already a cool one with PiP and fancy look, let's not overload demo.

projects/demo/src/pages/components/keypad/examples/4/index.html
```diff
@@ -0,0 +1,109 @@
...
+    [tuiSheetDialogOptions]="options"
+    [(tuiSheetDialog)]="open"
+>
+    <div class="sheet">
+        <div class="top">
+            <input
+                #input="tuiKeypadInput"
+                tuiKeypadInput
+                class="amount"
+                [ngModel]="value()"
```
waterplea: [(ngModel)]="value"

projects/demo/src/pages/components/keypad/examples/5/index.html
waterplea, line 1: Do we really need this example? You already can set dimensions of the keypad and it would adapt, I don't think we need a dynamic draggable example

projects/demo/src/pages/components/keypad/examples/2/index.html
```diff
@@ -0,0 +1,40 @@
...
+    </button>
+
+    <button
+        aria-label="Backspace"
+        tuiKeypadButton
+        type="button"
+        (click)="backspace()"
+        (longtap)="clear()"
+    >
+        @if (value()) {
```
waterplea: Button is still there and has press effect, we should move @if up to hide entire button

projects/addon-mobile/components/keypad/keypad-button.component.ts
mdlufy: Done — dropped `TuiKeypadButton`. `tui-keypad` now styles its projected `<button>`/`<a>` children directly (same approach as `tui-segmented`: `ViewEncapsulation.None` + version-scoped styles). Examples and docs updated to use plain native elements.

projects/demo/src/pages/components/keypad/index.html
mdlufy: Done — removed the Phone Dialer example. The Calculator already demonstrates the PiP case, no need to overload the demo.

projects/demo/src/pages/components/keypad/examples/4/index.html
mdlufy: Done — switched to `[(ngModel)]="value"`.

projects/demo/src/pages/components/keypad/examples/5/index.html
mdlufy: Done — removed the Fluid/draggable example (the pad already adapts to its container size).

projects/demo/src/pages/components/keypad/examples/2/index.html
mdlufy: Done — moved `@if (value())` up to wrap the whole backspace `<button>`, so the key is hidden entirely instead of leaving an empty pressable target.

projects/addon-mobile/components/keypad/keypad-input.directive.ts
mdlufy: Done — dropped the directive and switched the field to `tuiSelectLike`. It already sets `inputmode="none"` (so the native mobile keyboard never opens) and blocks typing on desktop; caret hidden via CSS. For the collapsible example the input toggles the pad via `(focus)/(blur)`, as you suggested. Reworked the keypad e2e to match.

mdlufy: Correction to my earlier reply: `tuiSelectLike` didn't hold up on iOS — a live centered `<input>` inside the sheet dialog hits Safari's composited-layer repaint/scroll bugs (Android was fine). Landed a simpler resolution instead: the keypad just drives a signal shown in a `<div>`, so there is no keypad-driven input at all and the `tuiKeypadInput` directive is removed. No new public API — which was your original point.

projects/addon-mobile/components/keypad/keypad.ts
```diff
@@ -0,0 +1,3 @@
+import {TuiKeypadComponent} from './keypad.component';
+
+export const TuiKeypad = [TuiKeypadComponent] as const;
```
waterplea: Unnecessary now, we can just name the component `TuiKeypad`

projects/addon-mobile/components/keypad/keypad.component.ts
```diff
@@ -0,0 +1,28 @@
...
+            @import './keypad.style.less';
+        }
+    `,
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {
+        'data-tui-version': TUI_VERSION,
+        '[style.--t-columns]': 'columns()',
+        '(mousedown.zoneless.prevent)': '(0)',
+        '(pointerdown.zoneless.prevent)': '(0)',
```
waterplea, line 24: No need for both

projects/addon-mobile/components/keypad/keypad.style.less
```diff
@@ -0,0 +1,58 @@
...
+        font: inherit;
+        color: inherit;
+        text-decoration: none;
+        cursor: pointer;
+        transition:
+            background var(--tui-duration) ease-out,
+            transform var(--tui-duration) var(--tui-curve-expressive-entrance);
+
+        @media @tui-mouse {
+            &:hover:not(:disabled):not([aria-disabled='true']) {
```
waterplea: Do we really need [aria-disabled] on each of those rules? We do not do that anywhere else in the library

waterplea (approved):

projects/addon-mobile/components/keypad/keypad.ts
mdlufy: Done — renamed to `TuiKeypad`, dropped the alias file.

projects/addon-mobile/components/keypad/keypad.component.ts
mdlufy: Done — kept only `pointerdown`.

projects/addon-mobile/components/keypad/keypad.style.less
mdlufy: Done — dropped `[aria-disabled]`, style only `:disabled` now.

PR #14119, opened 2026-05-15
fix(kit): `InputFiles` fix mime type normalize

Yanduz:
Fixes

### DIFF
--- a/projects/kit/components/files/files.validators.ts
+++ b/projects/kit/components/files/files.validators.ts
@@ -2,6 +2,8 @@ import {coerceArray} from '@angular/cdk/coercion';
 import {type ValidatorFn} from '@angular/forms';
 import {type TuiContext} from '@taiga-ui/cdk/types';
 
+import {type TuiFileLike} from './files.types';
+
 export const TUI_SIZE_ERROR = 'tuiSize';
 export const TUI_FORMAT_ERROR = 'tuiFormat';
 
@@ -32,16 +34,16 @@ export function tuiCreateFileFormatValidator(accept: string): ValidatorFn {
     };
 }
 
-function checkFormat({name, type}: File, formats: readonly string[]): boolean {
+function checkFormat({name, type}: TuiFileLike, formats: readonly string[]): boolean {
     const extension = `.${(name.split('.').pop() || '').toLowerCase()}`;
-    const normalizedType = type.toLowerCase();
+    const normalizedType = type?.toLowerCase() || '';
 
     return formats.some(
         (format) =>
             format === extension ||
             format === normalizedType ||
             (format.split('/')[1] === '*' &&
-                type?.split('/')[0] === format.split('/')[0]),
+                normalizedType.split('/')[0] === format.split('/')[0]),
     );
 }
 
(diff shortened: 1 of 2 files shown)


projects/kit/components/files/files.validators.ts
```diff
@@ -34,7 +34,7 @@ export function tuiCreateFileFormatValidator(accept: string): ValidatorFn {
 
 function checkFormat({name, type}: File, formats: readonly string[]): boolean {
     const extension = `.${(name.split('.').pop() || '').toLowerCase()}`;
-    const normalizedType = type.toLowerCase();
+    const normalizedType = type?.toLowerCase() || '';
```
splincode, line 39: Could you please add a regression test for this case?

The current fix protects `checkFormat` from files where `type` is missing/nullish, but this behavior is not covered by tests yet. It would be nice to ensure that a file-like object without `type` does not throw and still falls back to extension-based validation.

```ts
it('rejects file with missing type and non-matching extension', () => {
    const validator = tuiCreateFileFormatValidator('.jpg');
    const file = {name: 'photo.png'} as File;
    const control = new FormControl([file]);

    expect(validator(control)?.[TUI_FORMAT_ERROR].$implicit).toEqual([file]);
});
```

splincode: and

```ts
it('falls back to extension when MIME type is missing', () => {
    const validator = tuiCreateFileFormatValidator('.jpg');
    const file = {name: 'photo.jpg'} as File;
    const control = new FormControl([file]);

    expect(validator(control)).toBeNull();
});
```

projects/kit/components/files/files.validators.ts
```diff
@@ -34,7 +34,7 @@ export function tuiCreateFileFormatValidator(accept: string): ValidatorFn {
 
 function checkFormat({name, type}: File, formats: readonly string[]): boolean {
```
splincode: Maybe we can type this helper as `TuiFileLike` instead of `File` or `Partial<File>`?

`checkFormat` only needs `name` and `type`. `name` is required for extension-based validation, while `type` can be missing. This matches the existing `TuiFileLike` contract better and avoids using the too broad `Partial<File>` type.

```ts
function checkFormat({name, type}: TuiFileLike, formats: readonly string[]): boolean {
    const extension = `.${(name.split('.').pop() || '').toLowerCase()}`;
    const normalizedType = type?.toLowerCase() || '';

    return formats.some(
        (format) =>
            format === extension ||
            format === normalizedType ||
            (format.split('/')[1] === '*' &&
                normalizedType.split('/')[0] === format.split('/')[0]),
    );
}
```

Also, since we already normalize `type`, we can use `normalizedType` in the wildcard branch too:

```ts
normalizedType.split('/')[0] === format.split('/')[0]
```

waterplea (approved):

splincode (approved):

projects/kit/components/files/test/files.validators.spec.ts
```diff
@@ -193,5 +193,13 @@ describe('tuiCreateFileFormatValidator', () => {
 
             expect(validator(control)).toBeNull();
         });
+
+        it('rejects file with missing type and non-matching extension', () => {
```
splincode, line 197: Please add this case as well too:

```ts
it('falls back to extension when MIME type is missing', () => {
    const validator = tuiCreateFileFormatValidator('.jpg');
    const file = {name: 'photo.jpg'} as File;
    const control = new FormControl([file]);

    expect(validator(control)).toBeNull();
});
```

PR #14142, opened 2026-05-20
fix(kit): apply overscroll-behavior only when there is an extra element inside the t-ghost

vladimirpotekhin:
Fixes #13774

### DIFF
--- a/projects/kit/components/textarea/textarea.directive.ts
+++ b/projects/kit/components/textarea/textarea.directive.ts
@@ -47,7 +47,7 @@ class TuiTextareaCounter {
 @Component({
     template: `
         <ng-container>{{ context.$implicit.slice(0, limit()) }}</ng-container>
-        @if (limit() < context.$implicit.length) {
+        @if (limit() <= context.$implicit.length) {
             <span [textContent]="context.$implicit.slice(limit())"></span>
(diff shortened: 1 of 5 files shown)


projects/styles/components/textfield.less
```diff
@@ -243,6 +243,12 @@ tui-textfield:where(*&) {
         }
     }
 
+    &:has(textarea):not(:has(.t-ghost > *)) {
```
waterplea: Why here, why not in textarea? Why reverse condition?

projects/kit/components/textarea/textarea.directive.ts
```diff
@@ -46,8 +46,10 @@ class TuiTextareaCounter {
 
 @Component({
     template: `
-        <span [textContent]="context.$implicit.slice(0, limit())"></span>
-        <span [textContent]="context.$implicit.slice(limit())"></span>
+        <ng-container>{{ context.$implicit.slice(0, limit()) }}</ng-container>
+        @if (limit() < context.$implicit.length) {
```
waterplea: `<=`?

waterplea (approved):

nsbarsukov (approved):

PR #14143, opened 2026-05-20
fix(kit): apply overscroll-behavior only when there is an extra element inside the t-ghost

vladimirpotekhin:
…

Fixes #13774

### DIFF
--- a/projects/kit/components/textarea/textarea-limit.directive.ts
+++ b/projects/kit/components/textarea/textarea-limit.directive.ts
@@ -24,17 +24,20 @@ import {
 import {injectContext, PolymorpheusComponent} from '@taiga-ui/polymorpheus';
 
 import {tuiTextareaOptionsProvider} from './textarea.options';
+import {NgIf} from '@angular/common';
 
 @Component({
     standalone: true,
     template: `
-        <span [textContent]="context.$implicit.slice(0, limit())"></span>
+        <ng-container>{{ context.$implicit.slice(0, limit()) }}</ng-container>
         <span
+            *ngIf="context.$implicit.length > limit()"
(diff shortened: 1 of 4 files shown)


projects/kit/components/textarea/textarea-limit.directive.ts
```diff
@@ -24,14 +24,17 @@ import {
...
 
 import {tuiTextareaOptionsProvider} from './textarea.options';
+import {NgIf} from '@angular/common';
 
 @Component({
     standalone: true,
+    imports: [NgIf],
     template: `
-        <span [textContent]="context.$implicit.slice(0, limit())"></span>
+        <ng-container>{{ context.$implicit.slice(0, limit()) }}</ng-container>
```
nsbarsukov, line 33: Why `<ng-container>` is required here ?


```suggestion
        {{ context.$implicit.slice(0, limit()) }}
```

projects/kit/components/textarea/textarea-limit.directive.ts
```diff
@@ -24,14 +24,17 @@ import {
...
         <span
+            *ngIf="context.$implicit.length >= limit()"
             style="background: linear-gradient(transparent 0.25rem, var(--tui-status-negative-pale) 0.25rem, var(--tui-status-negative-pale) 100%)"
-            [textContent]="context.$implicit.slice(limit())"
+            [innerText]="context.$implicit.slice(limit())"
```
nsbarsukov: Revert

nsbarsukov (approved):

projects/kit/components/textarea/textarea-limit.directive.ts
vladimirpotekhin: to avoid extra spaces

waterplea (approved):

PR #14145, opened 2026-05-20
fix(core): `Carousel` fix for zooms outside 100%

waterplea:
Fixes #


projects/core/components/carousel/carousel.component.less
vladimirpotekhin: nit: shouldn't we set "overscroll-behavior-**x**: none" to not prevent regular vertical scrolling?

vladimirpotekhin (approved):

nsbarsukov (approved):

Issue #13646, opened 2026-03-30
🐞 - Hydration of inputs


waterplea:
### Reproduction url

https://taiga-ui.dev/next/components/input

### Description

Hydration of input-based components is currently disabled due to templates inside input tag. We need to move those templates into dynamic components so they do not create comment nodes that break SSR hydration.

### Taiga UI version

5

### Browser Used

Chrome

### OS Used

macOS

nsbarsukov:
@waterplea should we postpone this issue until `createComponent` allows to apply directives and specify bindings to dynamically created components ?

https://blog.angular.dev/announcing-angular-v20-b5c9c06cf301?gi=c8824f97e2f8#5775

PR #14204, opened 2026-06-02
fix(kit): `InputDate` hydration

mdlufy:
Fixes #13646


nsbarsukov (approved):

vladimirpotekhin (approved):

projects/kit/components/input-date/input-date-content.component.ts
```diff
@@ -0,0 +1,51 @@
...
+    selector: 'tui-input-date-content',
+    template: `
+        @if (host.native) {
+            <input
+                type="date"
+                [attr.list]="component.list()"
+                [max]="host.max().toJSON()"
+                [min]="host.min().toJSON()"
+                [value]="host.value()?.toJSON()"
+                (click.stop.zoneless)="(0)"
```
splincode: maybe we could move these serializations to computed signals?

```ts
protected readonly max = computed(() => this.host.max().toJSON());
protected readonly min = computed(() => this.host.min().toJSON());
protected readonly value = computed(() => this.host.value()?.toJSON());
```

```html
[max]="max()"
[min]="min()"
[value]="value()"
```

splincode: Computed makes sense here.
It is not required, but it can make the template cleaner and avoid repeated toJSON() calls.

mdlufy: Updated

splincode (approved):

projects/kit/components/input-date/input-date.component.ts
```diff
@@ -1,49 +1,23 @@
...
+    providers: [tuiAsTextfieldContent(TuiInputDateContent)],
+    hostDirectives: [TuiWithNativePicker, TuiTextfieldContent],
     host: {
         'data-tui-version': TUI_VERSION,
-        ngSkipHydration: 'true',
         '[attr.list]': 'null',
     },
 })
+// TODO(v6): rename to TuiInputDateDirective
 export class TuiInputDateComponent {
```
nsbarsukov: It already exists
https://github.com/taiga-family/taiga-ui/blob/ad35469298ed76d533df5c13209e062273772ce0/projects/kit/components/input-date/input-date.directive.ts#L186

mdlufy: Updated

PR #14210, opened 2026-06-02
feat(layout): `AppBar` support liquid glass

vladimirpotekhin:
Fixes #


splincode (approved):
A few moments before merge I see

splincode (approved):

projects/layout/components/dynamic-header/dynamic-header.style.less
```diff
@@ -21,6 +21,22 @@
...
+        --tui-from: translateY(calc(-50% * var(--t-dir)));
+    }
+
+    .tui-leave {
+        animation-name: tuiFade, tuiSlide, tuiBlur;
+
+        --tui-duration: 500ms;
+        --tui-from: translateY(calc(50% * var(--t-dir)));
+    }
+}
```
splincode, line 45: Add `._hidden` class definition. The `dynamic-header-anchor.directive.ts` toggles `._hidden` class but it is not defined here. Add:

```less
._hidden {
    display: none;
}
```

projects/core/components/root/root.component.ts
```diff
@@ -60,16 +61,24 @@ export class TuiRoot {
...
 
     protected readonly scrollbars =
         !inject(WA_IS_MOBILE) &&
         !this.child &&
         inject(TUI_SCROLLBAR_OPTIONS).mode !== 'native' &&
-        inject(TUI_OPTIONS).scrollbars !== 'native';
+        this.options.scrollbars !== 'native';
 
     protected get parent(): boolean {
         return this.doc.fullscreenElement
```
splincode, line 77: Add platform check. The liquidGlass getter doesn't check platform, but app-bar provider does. Add platform check for consistency.

projects/layout/components/dynamic-header/dynamic-header-container.directive.ts
```diff
@@ -34,6 +31,14 @@ export class TuiDynamicHeaderContainerDirective {
         ),
     );
 
+    public register(anchor: TuiDynamicHeaderAnchorDirective): void {
```
splincode, line 34: O(n²) complexity on register/unregister

Using spread operator `[...headers, anchor]` creates new array each time. For many headers, this becomes expensive.

Consider using `Set` or debouncing updates:
```ts
private readonly headersSet = new Set<TuiDynamicHeaderAnchorDirective>();

public register(anchor: TuiDynamicHeaderAnchorDirective): void {
    this.headersSet.add(anchor);
    this.scheduleUpdate();
}

private updateScheduled = false;
private scheduleUpdate(): void {
    if (!this.updateScheduled) {
        this.updateScheduled = true;
        requestAnimationFrame(() => {
            this.headers.set(Array.from(this.headersSet));
            this.update();
            this.updateScheduled = false;
        });
    }
}
```

projects/layout/components/app-bar/liquid-glass/progressive-blur.style.less
```diff
@@ -0,0 +1,51 @@
...
+            #000 100%
+        );
+    }
+}
+
+span {
+    position: absolute;
+    z-index: calc(-1 * var(--t-step));
+    inline-size: 100%;
+    inset-block-start: 0;
```
splincode, line 45: The current algorithm uses 20 DOM elements with `backdrop-filter`. Alternative: single element with CSS `backdrop-filter` + `mask-image` gradient could achieve similar effect with better performance.

```less
// instead of 20 span:
:host::after {
    content: "";
    position: absolute;
    inset: 0;
    backdrop-filter: blur(0.5rem);
    mask-image: linear-gradient(
        0deg,
        transparent 0%,
        #000 100%
    );
}
```

Also, max blur is only `0.00375rem * 19 = 0.07125rem` (~1px) - may be imperceptible on high-DPI displays. Consider exposing `--tui-liquid-glass-blur-intensity` CSS variable for tuning.

vladimirpotekhin: Unfortunately, the single element does not provide the required smoothness

projects/layout/components/dynamic-header/dynamic-header-container.directive.ts
vladimirpotekhin: I think there shouldn't be more than 20 headings on a page; this kind of optimization doesn't make sense.

projects/addon-mobile/styles/ios/app-bar.less
```diff
@@ -0,0 +1,15 @@
+._liquid-glass,
+&._liquid-glass {
```
splincode: maybe `.tui-liquid-glass` ?

projects/layout/components/dynamic-header/dynamic-header.style.less
```diff
@@ -1,43 +1,53 @@
...
+            --tui-duration: 500ms;
+            --tui-from: translateY(calc(50% * var(--t-dir)));
+        }
     }
 }
 
-.tui-enter:not(:first-of-type),
-.tui-leave:not(:first-of-type) {
-    position: absolute;
+[data-platform='ios']._liquid-glass {
```
splincode: `tui-`?

splincode (commented):
Inline comment.

splincode (commented):
A few inline thoughts before merge.

projects/demo/src/pages/components/app-bar/index.html
```diff
@@ -9,9 +9,30 @@
...
+                <ng-template #description>
+                    @switch ($index + 1) {
+                        @case (6) {
+                            To enable the liquid-glass mode, specify the option in
+                            <code>provideTaiga(&#123;apis: &#123;liquidGlass: signal(true)&#125;&#125;)</code>
+                        }
+                        @case (7) {
+                            On iOS, a progressive blur effect should be enabled when the container is scrolled. To
+                            activate it, apply the
+                            <code>_liquid-glass_blur</code>
```
splincode: If users are expected to apply this class manually, `_liquid-glass_blur` becomes part of the public API. Maybe we should avoid the private-looking underscore name and expose either a directive/input or a public-looking class like `tui-liquid-glass_blur`.

projects/layout/components/dynamic-header/dynamic-header-container.directive.ts
```diff
@@ -17,10 +17,7 @@ import {TuiDynamicHeaderAnchorDirective} from './dynamic-header-anchor.directive
 })
 export class TuiDynamicHeaderContainerDirective {
     private readonly el = tuiInjectElement();
-
-    protected readonly headers = contentChildren(TuiDynamicHeaderAnchorDirective, {
-        descendants: true,
-    });
+    private readonly headers = signal<TuiDynamicHeaderAnchorDirective[]>([]);
 
     public readonly hiddenHeaders = signal<Array<TemplateRef<unknown>>>([]);
```
splincode, line 22: Maybe `hiddenHeaders` should be a computed signal instead of an imperatively updated signal? 

`register()` / `unregister()` update `headers`, but they do not call `update()`. If a visible anchor is destroyed, `hiddenHeaders` can keep a stale `templateRef` until the next observer callback.

projects/layout/components/app-bar/app-bar.template.html
```diff
@@ -1,5 +1,9 @@
+<tui-progressive-blur />
```
splincode: I mean Angular still instantiates `TuiProgressiveBlur` and creates its 20 spans for every `tui-app-bar`, even when liquid glass is disabled. CSS hides it visually, but it does not remove the DOM/lifecycle cost. I think we should render it conditionally only for the liquid-glass mode.

```html
@if (liquidGlass) {
    <tui-progressive-blur />
}
```

projects/core/components/root/root.component.ts
```diff
@@ -60,12 +61,18 @@ export class TuiRoot {
...
         !inject(WA_IS_MOBILE) &&
         !this.child &&
         inject(TUI_SCROLLBAR_OPTIONS).mode !== 'native' &&
-        inject(TUI_OPTIONS).scrollbars !== 'native';
+        this.options.scrollbars !== 'native';
+
+    protected readonly liquidGlass =
+        inject(WA_IS_IOS) &&
+        this.options.apis !== 'stable' &&
+        (this.options.apis.all || !!this.options.apis.liquidGlass?.());
```
splincode: what if?

```ts
export const TUI_LIQUID_GLASS = new InjectionToken<Signal<boolean>>(
    ngDevMode ? 'TUI_LIQUID_GLASS' : '',
    {
        factory: () => {
            const options = inject(TUI_OPTIONS);
            const platform = inject(TUI_PLATFORM);

            return computed(() => {
                if (options.apis === 'stable') {
                    return false;
                }

                return (
                    platform === 'ios' &&
                    (options.apis.all || !!options.apis.liquidGlass?.())
                );
            });
        },
    },
);
```

and then 

```html
protected readonly liquidGlass = inject(TUI_LIQUID_GLASS);
```

```ts
host: {
    '[class.tui-liquid-glass]': 'liquidGlass()',
}
```

projects/layout/components/app-bar/liquid-glass/app-bar-button.style.less
```diff
@@ -0,0 +1,73 @@
+[data-platform='ios'].tui-liquid-glass [tuiAppBarButton]:where(*&) {
+    border-radius: 2.75rem;
+    margin-inline: 1rem;
+    backdrop-filter: blur(0.25rem) saturate(6) brightness(1.05);
+    box-shadow: 0 0.5rem 2rem rgba(0, 0, 0, 0.09);
+    overflow: hidden;
+    background: var(--t-appbar-button-bg);
+    color: var(--tui-text-primary);
+    transition:
+        all var(--tui-duration-fast) ease,
```
splincode: Could we avoid `transition: all` here and list only the properties we really want to animate? `all` can accidentally animate theme/background/filter changes and make the component less predictable.

projects/demo/src/pages/components/app-bar/examples/5/index.html
```diff
@@ -1,81 +1,25 @@
...
+</div>
 
 <div
+    #container
     tuiDynamicHeaderContainer
-    class="container"
-    [tuiPlatform]="activeItemIndex ? 'android' : 'ios'"
+    waIntersectionRootMargin="-120px 0px 1000000% 0px"
+    class="container tui-liquid-glass"
+    [class._liquid-glass_blur]="container.scrollTop > 0"
```
splincode: If users are expected to apply this class manually, `_liquid-glass_blur` becomes public API. Maybe we should avoid the private-looking underscore name and expose either a directive/input or a public-looking class like `tui-liquid-glass_blur`.

or 

```html
<tui-app-bar [tuiLiquidGlassBlur]="container.scrollTop > 0">
```

projects/layout/components/app-bar/liquid-glass/progressive-blur.style.less
```diff
@@ -0,0 +1,51 @@
+:host {
+    position: absolute;
+    inset: 0;
+    display: none;
```
splincode: `display: none` may prevent the opacity transition from working reliably because the element is removed from the render tree. If we keep this component mounted, maybe use `visibility` + `opacity`; if we render it conditionally, we probably do not need `display: none` here.

projects/layout/components/app-bar/liquid-glass/progressive-blur.style.less
```diff
@@ -0,0 +1,51 @@
...
+    :host-context(._liquid-glass_blur) & {
+        display: block;
+        opacity: 1;
+    }
+
+    &::before {
+        content: '';
+        position: absolute;
+        inset: 0;
+        z-index: -1;
```
splincode: These negative `z-index` layers look a bit fragile. Could we add an explicit stacking context, for example `isolation: isolate` / `z-index: 0`, so the blur layers cannot accidentally escape behind parent backgrounds?

projects/layout/components/app-bar/liquid-glass/app-bar-button.style.less
```diff
@@ -0,0 +1,73 @@
...
+    &.tui-enter {
+        --tui-blur: 1rem;
+
+        animation-name: tuiFade, tuiBlur;
+    }
+}
+
+[data-platform='ios'].tui-liquid-glass [tuiTheme='dark'] [tuiAppBarButton]:where(*&),
+[tuiTheme='dark'] [data-platform='ios'].tui-liquid-glass [tuiAppBarButton]:where(*&),
+[tuiTheme='dark'].tui-liquid-glass [tuiAppBarButton]:where(*&) {
```
splincode: This dark-theme selector is quite hard to reason about. Could we simplify the nesting or add a short comment explaining which theme-placement cases each selector covers?

```less
[data-platform='ios'].tui-liquid-glass {
    [tuiAppBarButton] {
        ...
    }

    [tuiTheme='dark'] [tuiAppBarButton],
    tui-app-bar[tuiTheme='dark'] [tuiAppBarButton] {
        ...
    }
}
```

projects/layout/components/app-bar/liquid-glass/app-bar-button.style.less
```diff
@@ -0,0 +1,73 @@
+[data-platform='ios'].tui-liquid-glass [tuiAppBarButton]:where(*&) {
+    border-radius: 2.75rem;
+    margin-inline: 1rem;
+    backdrop-filter: blur(0.25rem) saturate(6) brightness(1.05);
```
splincode, line 4: what if

```less
--t-appbar-button-blur: 0.25rem;
--t-appbar-button-saturation: 6;
--t-appbar-button-brightness: 1.05;
--t-appbar-button-active-brightness: 3;

backdrop-filter:
    blur(var(--t-appbar-button-blur))
    saturate(var(--t-appbar-button-saturation))
    brightness(var(--t-appbar-button-brightness));

&:active {
    filter: brightness(var(--t-appbar-button-active-brightness));
}
```

projects/addon-mobile/styles/ios/app-bar.less
```diff
@@ -0,0 +1,15 @@
+.tui-liquid-glass,
+&.tui-liquid-glass {
+    tui-app-bar {
+        --t-padding: 1rem;
```
splincode, line 3: Maybe use a more specific local variable name here? `--t-padding` is quite generic for app-bar styles. Something like `--t-appbar-padding` would be easier to maintain.

projects/core/utils/miscellaneous/provide-taiga.ts
```diff
@@ -16,6 +17,7 @@ import {tuiEnableFontScaling} from './font-scaling';
 
 interface ExperimentalAPIs {
     readonly all: boolean;
+    readonly liquidGlass: WritableSignal<boolean>;
```
splincode: Should `liquidGlass` be optional in `ExperimentalAPIs`? Runtime code already uses `liquidGlass?.()`, but the interface makes it required. This can be a breaking change for custom `apis` configurations.

projects/layout/components/app-bar/app-bar.providers.ts
```diff
@@ -17,5 +19,14 @@ export const TUI_APP_BAR_PROVIDERS: Provider[] = [
...
-    tuiButtonOptionsProvider({appearance: 'action'}),
+    tuiButtonOptionsProvider(() => {
+        const options = inject(TUI_OPTIONS);
+
+        const liquidGlass =
+            options.apis !== 'stable' &&
+            (options.apis.all || options.apis.liquidGlass?.()) &&
+            inject(TUI_PLATFORM) === 'ios';
+
+        return {appearance: liquidGlass ? '' : 'action', size: liquidGlass ? 'm' : 'l'};
```
splincode: what if

```ts
const liquidGlass = inject(TUI_LIQUID_GLASS);

return {
    appearance: liquidGlass() ? '' : 'action',
    size: liquidGlass() ? 'm' : 'l',
};
```

projects/demo/src/components/settings/settings.component.ts
```diff
@@ -99,16 +103,23 @@ export class SettingsComponent {
...
         .pipe(takeUntilDestroyed())
         .subscribe(() => {
-            const {theme, direction, language, platform} = this.form.getRawValue();
+            const {theme, direction, language, platform, liquid} =
+                this.form.getRawValue();
 
             this.switcher.setLanguage(language);
             this.doc.documentElement.setAttribute('dir', direction);
             this.doc.documentElement.setAttribute('data-platform', platform);
+            this.doc.documentElement.classList.toggle('tui-liquid-glass', liquid);
```
splincode: Do we need to toggle `tui-liquid-glass` on `documentElement` manually here? `TuiRoot` already owns this class. Having two writers for the same state can make the demo diverge from real app behavior.

projects/demo/src/components/settings/settings.component.html
```diff
@@ -141,4 +141,25 @@
             </label>
         </tui-segmented>
     </fieldset>
+
+    @if (form.value.platform === 'ios') {
```
splincode: Maybe use a dedicated computed/signal/getter for the selected platform instead of reading `form.value.platform` directly in the template. It would make this UI condition more explicit and consistent with `getRawValue()` usage below.

projects/demo/src/components/settings/settings.component.html
```diff
@@ -141,4 +141,25 @@
...
+        <fieldset>
+            <legend tuiTitle>
+                Apis
```
splincode: APIs

projects/demo/src/components/settings/settings.component.html
```diff
@@ -141,4 +141,25 @@
...
+                <div tuiSubtitle>For experimental apis</div>
+            </legend>
+            <label
+                appearance=""
+                tuiChip
+            >
+                Liquid glass (IOS only)
```
splincode: iOS only

projects/demo/src/pages/components/app-bar/examples/5/component.ts
```diff
@@ -0,0 +1,93 @@
...
+                </div>
+            }
+
+            <div
+                *tuiDynamicHeaderAnchor
+                tuiHeader="h6"
+            >
+                <h6 tuiTitle>Title 3 Long title</h6>
+            </div>
+            @for (_ of '-'.repeat(15); track $index) {
```
splincode: Maybe avoid `'-'.repeat(...)` in demo templates and use readonly arrays from the component instead. It avoids allocations during change detection and prevents users from copying this pattern.

projects/core/components/root/animations.less
```diff
@@ -32,6 +32,12 @@
     }
 }
 
+@keyframes tuiBlur {
```
splincode, line 35: Should liquid-glass transitions/active effects respect reduced motion? `scale(1.25)` + brightness change is a strong visual effect, so it may be worth checking how it behaves when animations are disabled/reduced.

projects/layout/components/app-bar/liquid-glass/app-bar-button.style.less
splincode, line 4: Do we need an explicit fallback for environments where `backdrop-filter` is unsupported or disabled? Liquid glass relies heavily on it, so a simple solid/elevated background fallback could make the component more robust.

```html
@supports not (backdrop-filter: blur(1px)) {
    [tuiAppBarButton] {
        background: var(--tui-background-elevation-1);
    }
}
```

projects/addon-mobile/styles/ios/app-bar.less
```diff
@@ -0,0 +1,15 @@
+.tui-liquid-glass,
```
splincode: Since the app-bar template changed for all platforms, maybe we should have explicit regression coverage for non-liquid-glass app bars too: desktop, android, iOS without liquid glass, and iOS with liquid glass.

Image

projects/demo/src/pages/components/app-bar/examples/7/index.html
```diff
@@ -0,0 +1,42 @@
+<div
```
splincode, line 1: is it expected?

https://github.com/user-attachments/assets/90eba1ca-2282-4bfb-92a4-d8fe68402941

projects/demo/src/pages/components/app-bar/examples/1/index.html
```diff
@@ -1,4 +1,5 @@
 <h3>iOS</h3>
+
```
splincode, line 2: example 2

Image

projects/demo/src/pages/components/app-bar/examples/5/component.ts
```diff
@@ -0,0 +1,93 @@
...
+        TuiCell,
+        TuiDynamicHeader,
+        TuiHeader,
+        TuiTitle,
+    ],
+    template: `
+        <tui-app-bar>
+            <button
+                title="Back"
+                tuiAppBarBack
```
splincode: Should there be a minimum delay? If I accidentally click the button, it can be annoying.

https://github.com/user-attachments/assets/fa288541-21fe-42b8-ab4d-e92331708d44

projects/demo/src/pages/components/app-bar/examples/5/index.html
```diff
@@ -1,81 +1,25 @@
-<tui-segmented [(activeItemIndex)]="activeItemIndex">
-    <button type="button">IOS</button>
-    <button type="button">Android</button>
-</tui-segmented>
+<div class="default">
```
splincode: The blur doesn't disappear if I stop scrolling, wait, and then scroll up.

https://github.com/user-attachments/assets/8a56026e-e1a3-406f-949e-c7078abdbf7e

projects/demo/src/pages/components/app-bar/examples/5/component.ts
vladimirpotekhin: No, this does not imply desktop use.

projects/addon-mobile/styles/ios/app-bar.less
vladimirpotekhin: Don't want to lengthen the internal name unnecessarily.

projects/layout/components/app-bar/app-bar.template.html
vladimirpotekhin: That's what I did initially, but then I started to doubt that additional js was needed here.

projects/layout/components/dynamic-header/dynamic-header-container.directive.ts
vladimirpotekhin: > can keep a stale

No, TuiDynamicHeaderAnchorDirective calls `unregister()`  in onDestroy callback

projects/demo/src/pages/components/app-bar/examples/1/index.html
vladimirpotekhin: This is how platform selection works in the demo version; check it out on the main page right now. For Android, for example, there are two styles. Out of this PR scope

projects/addon-mobile/styles/ios/app-bar.less
vladimirpotekhin: This is how platform selection works in the demo version; check it out on the main page right now. For Android, for example, there are two styles. Out of this PR scope

> explicit regression coverage

by default liquid glass disabled, explicit liquid-glass examples added additionally

projects/layout/components/app-bar/liquid-glass/app-bar-button.style.less
vladimirpotekhin: I don't think a variable for one value makes sense

projects/demo/src/components/settings/settings.component.ts
vladimirpotekhin: Yes, we need. Until the internal roots don't respond to the platform change

projects/demo/src/components/settings/settings.component.html
vladimirpotekhin: Maybe later, when there are more options, but right now I don't see the point in a getter.

projects/core/utils/miscellaneous/provide-taiga.ts
vladimirpotekhin: this is not public interface, public interface TuiOptions contains **`Partial<ExperimentalAPIs>`**

projects/core/components/root/animations.less
vladimirpotekhin: tuiBlur respects reduced motion

projects/demo/src/pages/components/app-bar/examples/7/index.html
vladimirpotekhin: It seems like it is only on demo, due to the fact that the example is not removed from DOM

projects/layout/components/app-bar/liquid-glass/progressive-blur.style.less
vladimirpotekhin: No, that won't work.

vladimirpotekhin: I moved the element up the tree, the z-index is no longer needed

projects/core/utils/miscellaneous/liquid-glass.ts
```diff
@@ -0,0 +1,21 @@
+import {computed, inject, InjectionToken, type Signal} from '@angular/core';
+import {WA_IS_IOS} from '@ng-web-apis/platform';
+
+import {TUI_OPTIONS} from './provide-taiga';
+
+export const TUI_LIQUID_GLASS = new InjectionToken<Signal<boolean>>(
+    ngDevMode ? 'TUI_LIQUID_GLASS' : '',
+    {
+        factory: () => {
+            const options = inject(TUI_OPTIONS);
```
waterplea: `{ apis }` ?

projects/layout/components/dynamic-header/dynamic-header.style.less
```diff
@@ -1,27 +1,53 @@
...
+        position: absolute;
+    }
+
+    [data-platform='ios'].tui-liquid-glass & {
+        .tui-enter {
+            animation-name: tuiFade, tuiSlide, tuiBlur;
 
-.tui-leave {
-    animation-name: tuiFade, tuiSlide;
+            --tui-duration: 500ms;
```
waterplea: This needs to respond to ANIMATION_SPEED, otherwise it's just a hardcode and it also does not obey reduced motion. Maybe add a temp --t-duration: calc(var(--tui-duration) * 5 / 3) on line 29 and then set --tui-duration: var(--t-duration) for .tui-enter/.tui-leave?

projects/layout/components/dynamic-header/dynamic-header.style.less
```diff
@@ -1,27 +1,53 @@
...
+        }
+    }
 }
 
-.tui-enter:not(:first-of-type),
-.tui-leave:not(:first-of-type) {
-    position: absolute;
+[data-platform='ios'].tui-liquid-glass {
+    [tuiHeader] {
+        transition: opacity 300ms linear;
```
waterplea: --tui-duration?

projects/layout/components/app-bar/app-bar.providers.ts
```diff
@@ -17,5 +18,12 @@ export const TUI_APP_BAR_PROVIDERS: Provider[] = [
             subtree: true,
         },
     },
-    tuiButtonOptionsProvider({appearance: 'action'}),
+    tuiButtonOptionsProvider(() => {
+        const liquidGlass = inject(TUI_LIQUID_GLASS);
```
waterplea: Can move inject to arguments to avoid {} and return

projects/layout/components/app-bar/app-bar.providers.ts
```diff
@@ -17,5 +18,12 @@ export const TUI_APP_BAR_PROVIDERS: Provider[] = [
...
+
+        return {
+            appearance: liquidGlass() ? '' : 'action',
```
waterplea: I believe we also need to check if tuiPlatform is iOS.

projects/layout/components/app-bar/app-bar.template.html
```diff
@@ -1,5 +1,11 @@
+@if (liquidGlass()) {
+    <tui-progressive-blur />
```
waterplea, line 2: Let's also add display: none if not under data-platform='ios'

projects/demo/src/pages/components/app-bar/index.html
```diff
@@ -9,9 +9,30 @@
...
-                [content]="$index + 1 | tuiExample"
+                [content]="$index + 1 | tuiExample: 'html,ts,less' : this[$index]"
+                [description]="description"
                 [heading]="example"
-            />
+            >
+                <ng-template #description>
+                    @switch ($index + 1) {
+                        @case (6) {
+                            To enable the liquid-glass mode, specify the option in
```
waterplea: ... on iOS devices ...

so that people know it would only apply on iOS by default

projects/demo/src/pages/components/app-bar/examples/7/index.html
```diff
@@ -0,0 +1,43 @@
...
+    #container
+    tuiDynamicHeaderContainer
+    tuiPlatform="ios"
+    waIntersectionRootMargin="-120px 0px 1000000% 0px"
+    class="container tui-liquid-glass"
+    (scroll)="(0)"
+>
+    <tui-app-bar
+        [attr.tuiTheme]="color() || mode() ? 'dark' : 'light'"
+        [class.tui-liquid-glass_blur]="container.scrollTop > 0"
```
waterplea: Should we put something inside backgrounds? Otherwise the blur effect is not visible

projects/demo/src/pages/components/app-bar/examples/7/index.html
```diff
@@ -0,0 +1,43 @@
+<div
+    #container
+    tuiDynamicHeaderContainer
+    tuiPlatform="ios"
+    waIntersectionRootMargin="-120px 0px 1000000% 0px"
+    class="container tui-liquid-glass"
+    (scroll)="(0)"
```
waterplea: Can we avoid change detection on every scrol event? We only need it for scrollTop > 0? Maybe intersection observer can handle that?

projects/layout/components/app-bar/liquid-glass/app-bar-button.style.less
```diff
@@ -0,0 +1,80 @@
...
+    &.tui-enter {
+        --tui-blur: 1rem;
+
+        animation-name: tuiFade, tuiBlur;
+    }
+}
+
+.tui-liquid-glass [tuiTheme='dark'] [tuiAppBarButton]:where(*&),
+[tuiTheme='dark'] [tuiAppBarButton]:where(*&),
+[tuiTheme='dark'].tui-liquid-glass [tuiAppBarButton]:where(*&) {
```
waterplea: Don't forget [data-platform='ios'] in selectors here

projects/layout/components/dynamic-header/dynamic-header-anchor.directive.ts
```diff
@@ -22,13 +22,25 @@ export class TuiDynamicHeaderAnchorDirective implements AfterViewInit, OnDestroy
     public readonly visible = signal(false);
 
     public ngAfterViewInit(): void {
+        this.container.register(this);
         this.observer.observe(this.view.rootNodes[0], ([e]) => {
-            this.visible.set(!e?.isIntersecting);
+            const isIntersecting = e?.isIntersecting ?? false;
+            const target = e?.target as HTMLElement;
```
waterplea, line 28: No need to use `as`, you can just write e?.target?.classList

projects/layout/components/dynamic-header/dynamic-header-anchor.directive.ts
```diff
@@ -22,13 +22,25 @@ export class TuiDynamicHeaderAnchorDirective implements AfterViewInit, OnDestroy
...
+
+            this.visible.set(!isIntersecting);
+
+            if (isIntersecting) {
+                target.classList.remove('t-header_hidden');
```
waterplea: Is target here the same as elementRef? Can we use host binding with visible instead of manual class management?

vladimirpotekhin: No, it isn't. (It is structural directive) 
But we can manage opacity instead of class. I rewrote it to "opacity"

projects/layout/components/app-bar/app-bar.template.html
vladimirpotekhin: now it is under condition `liquidGlass = inject(TUI_LIQUID_GLASS) && inject(TUI_PLATFORM) === 'ios';`

waterplea: Platform can be dynamically changed, maybe it's better to hide it with CSS?

projects/demo/src/components/settings/settings.component.ts
```diff
@@ -55,7 +55,7 @@ import {TuiForm} from '@taiga-ui/layout';
     providers: [
         tuiSegmentedOptionsProvider({size: 's'}),
         tuiButtonOptionsProvider({appearance: '', size: 's'}),
-        tuiDropdownOptionsProvider({align: 'end'}),
+        tuiDropdownOptionsProvider({align: 'end', maxHeight: 500}),
```
waterplea: No longer needed?

projects/demo/src/pages/components/app-bar/examples/5/index.less
waterplea, line 1: Are these changes relevant?

projects/demo/src/pages/components/app-bar/examples/5/index.ts
```diff
@@ -4,11 +4,12 @@ import {encapsulation} from '@demo/emulate/encapsulation';
 import {TuiPlatform} from '@taiga-ui/cdk';
 import {TuiCell, TuiTitle} from '@taiga-ui/core';
 import {TuiAvatar, TuiSegmented} from '@taiga-ui/kit';
-import {TuiAppBar, TuiDynamicHeader, TuiHeader} from '@taiga-ui/layout';
+import {TuiAppBar, TuiAppBarBack, TuiDynamicHeader, TuiHeader} from '@taiga-ui/layout';
 
 @Component({
     imports: [
         TuiAppBar,
+        TuiAppBarBack,
```
waterplea: Isn't it included in the TuiAppBar already?

projects/demo/src/pages/components/app-bar/examples/7/index.html
```diff
@@ -0,0 +1,82 @@
+<div
+    #container
+    tuiDynamicHeaderContainer
+    tuiPlatform="ios"
+    waIntersectionRootMargin="-120px 0px 1000000% 0px"
+    class="container tui-liquid-glass"
```
waterplea: I wonder if we should mention somehow that you do not need `tui-liquid-glass` class if you enable it globally in your app?

projects/demo/src/pages/components/app-bar/examples/7/index.less
```diff
@@ -0,0 +1,14 @@
...
+    max-block-size: 30rem;
+    inline-size: 20rem;
+    overflow: scroll;
+    overscroll-behavior: none;
+}
+
+tui-app-bar {
+    position: sticky;
+    z-index: 1;
+    inset-block-start: -0.03125rem;
```
waterplea: Why such a weird value here and in other examples?

projects/demo/src/pages/components/app-bar/examples/7/index.less
```diff
@@ -0,0 +1,14 @@
...
+    margin-block-end: 0.75rem;
+    background: var(--tui-background-base);
```
waterplea: Irrelevant

projects/demo/src/pages/components/app-bar/examples/7/index.ts
```diff
@@ -0,0 +1,35 @@
...
+import {WaIntersectionObserver} from '@ng-web-apis/intersection-observer';
+import {TuiPlatform} from '@taiga-ui/cdk';
+import {TUI_LIQUID_GLASS, TuiCell, TuiTitle} from '@taiga-ui/core';
+import {TuiAvatar} from '@taiga-ui/kit';
+import {TuiAppBar, TuiAppBarBack, TuiDynamicHeader, TuiHeader} from '@taiga-ui/layout';
+
+@Component({
+    imports: [
+        TuiAppBar,
+        TuiAppBarBack,
```
waterplea: Unnecessary

projects/demo/src/pages/components/app-bar/examples/7/index.ts
```diff
@@ -0,0 +1,35 @@
...
+        TuiPlatform,
+        TuiTitle,
+        WaIntersectionObserver,
+    ],
+    templateUrl: './index.html',
+    styleUrl: './index.less',
+    encapsulation,
+    changeDetection,
+    providers: [
+        {
```
waterplea: In single line?

projects/layout/components/app-bar/app-bar.template.html
waterplea: Ignore this comment, let's keep detection in the JS, since button params are in JS anyway.

waterplea (approved):

projects/demo/src/pages/components/app-bar/examples/7/index.html
vladimirpotekhin: moved the class up a level so it wouldn't be visible in the example

PR #14213, opened 2026-06-03
fix(kit): `InputColor` hydration

mdlufy:
Fixes #13646


projects/kit/components/input-color/input-color-content.component.ts
```diff
@@ -0,0 +1,57 @@
...
+    styles: `
+        [data-tui-version='${TUI_VERSION}'] {
+            @import './input-color.style.less';
+        }
+    `,
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TuiInputColorContent {
+    protected readonly color = inject(TuiControl) as TuiInputColorComponent;
```
mdlufy: TuiInputColorContent is rendered in a portal, outside the host element's injector hierarchy

inject(TuiInputColorComponent) fails there with NG0201 because the directive token is not explicitly provided — Angular resolves it only through element ancestors. inject(TuiControl) works because tuiAsControl registers the instance under that token, which the portal inherits through its passed injector

splincode (approved):

projects/kit/components/input-color/input-color-content.component.ts
```diff
@@ -0,0 +1,57 @@
...
+        }
+    `,
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {'data-tui-version': TUI_VERSION},
+})
+export class TuiInputColorContent {
+    protected readonly host = inject(TuiInputColorComponent);
+
+    protected readonly filled = computed(() =>
```
splincode, line 30: I think we still have an existing bug in `hexa` mode.

When the value is a valid RGB color without alpha yet, for example `#ffffff`,
`filled()` returns `false` because `hexa` expects 9 chars. As a result:

1. the color preview opacity becomes `0`;
2. changing the opacity slider resets the base color to `#000000`.

I suggest separating "has RGB color" from "has alpha":

```ts
protected readonly hasColor = computed(() => this.host.value().length >= 7);

protected readonly opacity = computed(() => {
    const opacity =
        this.host.format() === 'hexa' && this.host.value().length === 9
            ? Number.parseInt(this.host.value().slice(-2), 16)
            : 255;

    return Number.isNaN(opacity) ? 255 : opacity;
});

protected onOpacity(opacity: number): void {
    const value = this.hasColor() ? this.host.value().slice(0, 7) : '#000000';

    this.host.onChange(`${value}${toHex(opacity)}`);
}
```

And in template:

```html
[style.--t-opacity]="hasColor() ? opacity() / 255 : 0"
```

projects/kit/components/input-color/input-color.component.ts
```diff
@@ -1,60 +1,50 @@
...
-        ngSkipHydration: 'true',
         spellcheck: 'false',
         '[attr.list]': 'null',
         '[disabled]': 'disabled()',
         '[value]': 'value()',
         '(input)': 'onChange($event.target.value)',
     },
 })
+// TODO(v6): rename to TuiInputColorDirective
 export class TuiInputColorComponent extends TuiControl<string> {
```
splincode, line 42: Could we add a small regression test for `hexa` mode?

The risky case is:

1. initial value is `#ffffff`;
2. `format="hexa"`;
3. user changes opacity slider;
4. value should become `#ffffff80`, not `#00000080`.

This would cover the current `filled()` edge case and protect us from the color reset regression.


```ts
it('does not reset base color when opacity changes in hexa mode', () => {
    cy.mount(Test, {
        componentProperties: {
            value: '#ffffff',
            format: 'hexa',
        },
    });

    cy.get('input[tuiInputColor]').should('have.value', '#ffffff');

    cy.get('input[type="range"]').then(($input) => {
        const input = $input[0] as HTMLInputElement;

        input.value = '128';
        input.dispatchEvent(new Event('input', {bubbles: true}));
        input.dispatchEvent(new Event('change', {bubbles: true}));
    });

    cy.get('input[tuiInputColor]').should('have.value', '#ffffff80');
});
```

```ts
@Component({
    standalone: true,
    imports: [FormsModule, TuiInputColor],
    template: `
        <tui-textfield>
            <input
                tuiInputColor
                [(ngModel)]="value"
                [format]="format"
            />
        </tui-textfield>
    `,
})
class Test {
    public value = '#ffffff';
    public format: 'hex' | 'hexa' = 'hexa';
}
```

```ts
it('keeps color preview visible for base color without alpha in hexa mode', () => {
    cy.mount(Test, {
        componentProperties: {
            value: '#ffffff',
            format: 'hexa',
        },
    });

    cy.get('input[type="color"]')
        .should('have.value', '#ffffff')
        .should(($input) => {
            expect($input[0].style.getPropertyValue('--t-opacity')).to.equal('1');
        });
});
```

projects/kit/components/input-color/input-color-content.component.ts
mdlufy: Fixed in https://github.com/taiga-family/taiga-ui/pull/14287

projects/kit/components/input-color/input-color.component.ts
mdlufy: in https://github.com/taiga-family/taiga-ui/pull/14287

vladimirpotekhin (approved):

PR #14214, opened 2026-06-03
chore(demo): change breakpoints path

kenny-not-dead:
Fixes # 

The current button leads to the package on GitHub where these variables were previously contained

image
image


kenny-not-dead:
If there's no package, the link button doesn't display...maybe I should remove the path entirely, similar to the colors, or add package="STYLES"?

image

waterplea:
Yes, I think it's fine to not have a link here at all

kenny-not-dead:
Thanks, then I also removed "path"

splincode (approved):

waterplea (approved):

PR #14219, opened 2026-06-04
fix(kit): `InputPin` remove latest typed symbol

### LINKED ISSUE #13896: 🐞 - The latest typed symbol cannot be removed with a Backspace key press in the InputPin component
### Reproduction url

https://taiga-ui.dev/components/input-pin#basic

### Description

The bug is being stably reproduced in the Docs.

Steps:
1. Write any amount of symbols in the inputs
2. Press backspace to remove the symbols
3. The result is the latest symbol stays in the first cell and cannot be removed, while the expected behavior is that you can delete all the characters

For some reason my media cannot be uploaded so you can find the video https://t.me/taiga_ui/7924/36915

Checked on Safari and Chrome.

### Taiga UI version

5.2.0

### Browser Used

Chrome

### OS Used

macOS

mdlufy:
Fixes #13896


projects/kit/components/input-pin/input-pin.component.ts
```diff
@@ -47,8 +47,12 @@ export class TuiInputPinComponent {
     }
 
     public onSelection(): void {
-        if (this.el.selectionStart === this.el.maxLength) {
-            this.el.setSelectionRange(this.el.maxLength - 1, this.el.maxLength - 1);
+        const {selectionStart, selectionEnd, maxLength, value} = this.el;
+
+        if (selectionStart === maxLength) {
+            this.el.setSelectionRange(maxLength - 1, maxLength - 1);
+        } else if (selectionStart === 0 && selectionEnd === 0 && value) {
```
waterplea: !selectionStart && !selectionEnd, no need to write === 0

waterplea (approved):

projects/kit/components/input-pin/input-pin.component.ts
mdlufy: But !selectionStart conflates 0 and null. Since selectionStart is typed as number | null, the explicit === 0 check is clearer about what we're actually checking here

projects/kit/components/input-pin/test/input-pin.component.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+                <input
+                    maxlength="4"
+                    tuiInputPin
+                    [formControl]="control"
+                />
+            </tui-textfield>
+        `,
+        // eslint-disable-next-line @angular-eslint/prefer-on-push-component-change-detection
+        changeDetection: ChangeDetectionStrategy.Default,
+    })
```
splincode: OnPush?

projects/kit/components/input-pin/test/input-pin.component.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+        inputPin = testComponent.inputPin();
+        el = fixture.nativeElement.querySelector('input[tuiInputPin]');
+    });
+
+    describe('onSelection', () => {
+        it('extends collapsed [0, 0] to [0, 1] when value is non-empty', () => {
+            el.value = '1';
+            el.setSelectionRange(0, 0);
+            inputPin.onSelection();
+            fixture.detectChanges();
```
nsbarsukov: This test feels too artificial. Wouldn't it be better to write a Cypress component test instead?

https://github.com/taiga-family/taiga-ui/tree/pin/projects/demo-cypress

vladimirpotekhin (approved):

projects/kit/components/input-pin/input-pin.component.ts
nsbarsukov: But `selectionStart` is only `null` when the element is an `<input>` whose `type` doesn't support text selection — such as `number`, `email, `date`, `checkbox`, etc. — which isn't our case, right?

projects/kit/components/input-pin/test/input-pin.component.spec.ts
mdlufy: yes, updated

projects/kit/components/input-pin/input-pin.component.ts
mdlufy: hmm yes

projects/demo-cypress/src/tests/input-pin.cy.ts
```diff
@@ -0,0 +1,73 @@
...
+        </tui-textfield>
+    `,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TestInputPin {
+    protected readonly control = new FormControl('', Validators.minLength(4));
+}
+
+describe('InputPin', () => {
+    beforeEach(() => cy.mount(TestInputPin));
```
splincode:
```ts
beforeEach(() => {
    cy.mount(TestInputPin);
    cy.get('#pin-input').as('pin');
});

// ....

it('should allow typing PIN code', () => {
    cy.get('@pin').type('1234');

    cy.get('@pin').should('have.value', '1234');
});
```

mdlufy: Updated

projects/kit/components/input-pin/test/input-pin.component.spec.ts
mdlufy: yes, updated

PR #14223, opened 2026-06-04
fix(core): `Notification` fix font scaling

waterplea:
Fixes #

### DIFF
--- a/projects/styles/components/notification.less
+++ b/projects/styles/components/notification.less
@@ -58,7 +58,7 @@
     &::before {
         position: absolute;
         inset-inline-start: -1rem;
-        inset-block-start: calc(var(--t-offset) + var(--tui-lh));
+        inset-block-start: calc((var(--t-offset) + var(--tui-lh)) * (1 - 0.15 * var(--t-zoom)));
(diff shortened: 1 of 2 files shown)


vladimirpotekhin (approved):

projects/styles/components/notification.less
```diff
@@ -58,7 +58,7 @@
     &::before {
         position: absolute;
         inset-inline-start: -1rem;
-        inset-block-start: calc(var(--t-offset) + var(--tui-lh));
+        inset-block-start: calc((var(--t-offset) + var(--tui-lh)) * (1 - 0.15 * var(--t-zoom)));
```
mdlufy, line 61: magic 0.15 )

mdlufy (approved):

projects/styles/components/notification.less
waterplea: Ususally it would be 0.2 to neutralize 125% zoom effect, but this time it depends on vertical paddng of a notification. We can try switching padding for border, but I guess it's not worth it as this is just a small correction for highest possible scaling.

PR #14227, opened 2026-06-04
fix(core): `Hint` fix positioning and focusing issue on mobile

vladimirpotekhin:
This PR includes a set of fixes related to tooltip/hint positioning and inconsistent behavior on mobile devices.

### Positioning during scale animation

To ensure correct hint positioning, we do not update position during scale animations.

Previously (in 4.x), el.offsetWidth was used instead of getBoundingClientRect().width, which avoided issues where transform: scale() affects layout calculations.
This approach is no longer available, so we prevent position recalculation during the animation to avoid jitter and visual jumps.

### Pointer alignment fix

Fixed incorrect tooltip arrow positioning caused by a coordinate order issue:

x and y values were accidentally swapped ([x, y] → [y, x]) after normalization in [TuiPoint](https://github.com/taiga-family/taiga-ui/pull/12692)

### Mobile / focus / icon click issues

Also fixed several desynchronization issues on mobile:

focus behavior when clicking icons
tooltip open/close inconsistencies

These changes restore the expected behavior and align mobile interactions with previous stable behavior.

### DIFF
--- a/projects/core/components/textfield/textfield.template.html
+++ b/projects/core/components/textfield/textfield.template.html
@@ -5,7 +5,7 @@
 <ng-content select="label" />
 <span
     class="t-content"
-    (pointerdown)="onContentPointerDown($event)"
+    (pointerdown)="input()?.nativeElement?.focus()"
     (waResizeObserver)="$event[0] && onResize($event[0])"
 >
     <ng-content />
--- a/projects/core/components/textfield/textfield.template.html
+++ b/projects/core/components/textfield/textfield.template.html
@@ -5,7 +5,7 @@
 <ng-content select="label" />
 <span
     class="t-content"
-    (pointerdown)="input()?.nativeElement?.focus()"
+    (pointerdown.prevent)="input()?.nativeElement?.focus()"
     (waResizeObserver)="$event[0] && onResize($event[0])"
 >
     <ng-content />
--- a/projects/core/components/textfield/textfield.template.html
+++ b/projects/core/components/textfield/textfield.template.html
@@ -5,7 +5,7 @@
 <ng-content select="label" />
 <span
     class="t-content"
-    (pointerdown.prevent)="input()?.nativeElement?.focus()"
+    (pointerdown)="onContentPointerDown($event)"
     (waResizeObserver)="$event[0] && onResize($event[0])"
 >
     <ng-content />
--- a/projects/core/portals/hint/hint.component.ts
+++ b/projects/core/portals/hint/hint.component.ts
@@ -129,6 +129,7 @@ export class TuiHintComponent {
     private update(left: number, top: number): void {
         if (
             this.isMobile &&
+            this.el.getAnimations?.().length &&
             this.el
                 .getAnimations?.()
                 .every(({effect}) => effect?.getComputedTiming().progress !== null)
--- a/projects/core/portals/hint/hint.component.ts
+++ b/projects/core/portals/hint/hint.component.ts
@@ -142,9 +142,9 @@ export class TuiHintComponent {
             Math.max(GAP, viewport.width + viewport.left - clientWidth - GAP),
         );
 
-        const [beakTop, beakLeft] = this.vvs.correct([
-            rect.top + rect.height / 2 - top,
+        const [beakLeft, beakTop] = this.vvs.correct([
             rect.left + rect.width / 2 - safeLeft,
+            rect.top + rect.height / 2 - top,
         ]);
 
         this.apply(
--- a/projects/core/portals/hint/hint.component.ts
+++ b/projects/core/portals/hint/hint.component.ts
@@ -127,13 +127,12 @@ export class TuiHintComponent {
     }
 
     private update(left: number, top: number): void {
-        const animating = this.el
-            .getAnimations?.()
-            .every(
-                (animation) => animation.effect?.getComputedTiming().progress !== null,
-            );
-
-        if (animating) {
+        if (
+            this.isMobile &&
+            this.el
+                .getAnimations?.()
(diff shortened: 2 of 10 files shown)


projects/core/components/textfield/textfield.template.html
```diff
@@ -5,7 +5,7 @@
 <ng-content select="label" />
 <span
     class="t-content"
-    (pointerdown)="input()?.nativeElement?.focus()"
+    (pointerdown.prevent)="input()?.nativeElement?.focus()"
```
waterplea: It was rolled back to fix some other issue on iOS before:
https://github.com/taiga-family/taiga-ui/pull/12331

projects/core/services/position.service.ts
```diff
@@ -20,7 +20,13 @@ export class TuiPositionService extends Observable<TuiPoint> {
             animationFrame$
                 .pipe(
                     startWith(null),
-                    map(() => this.accessor.getPosition(this.el.getBoundingClientRect())),
+                    map(() => {
+                        const rect = this.el.getBoundingClientRect();
```
waterplea: Why is this necessary?

projects/kit/directives/password/password.directive.ts
```diff
@@ -31,6 +31,7 @@ import {TUI_PASSWORD_OPTIONS} from './password.options';
         '[style.border]':
             'textfield.options.size() === "s" ? "0.25rem solid transparent" : null',
         '(click)': 'toggle()',
+        '(pointerdown.prevent)': '0',
```
waterplea: Why is this needed if you already prevent it in the textfield?

projects/core/components/textfield/textfield.template.html
vladimirpotekhin: yes and its break focusing on clicking on tui-icon, try in main https://taiga-ui.dev/components/input

projects/core/services/position.service.ts
vladimirpotekhin: This was introduced when mobile behavior was added (see screenshot).

Снимок экрана 2026-06-04 в 15 21 40

The scale animation affects getBoundingClientRect() measurements, which causes the positioning to break.

We probably lost the original fix during the 5.x refactoring.

https://github.com/taiga-family/taiga-ui/pull/9547/changes

projects/core/components/textfield/textfield.template.html
waterplea: I believe you, just want to make sure it was just a mistake last time. Maybe there were reasons :)

projects/core/services/position.service.ts
```diff
@@ -20,7 +20,22 @@ export class TuiPositionService extends Observable<TuiPoint> {
...
+                        const width = this.el.clientWidth || rect.width;
+                        const height = this.el.clientHeight || rect.height;
+
+                        return this.accessor.getPosition({
+                            top: rect.top,
```
waterplea: `...rect,` ?

projects/core/services/position.service.ts
waterplea: Oh, I guess it was my fault :( Ok, thanks for fixing!

projects/core/services/position.service.ts
```diff
@@ -20,7 +20,22 @@ export class TuiPositionService extends Observable<TuiPoint> {
             animationFrame$
                 .pipe(
                     startWith(null),
-                    map(() => this.accessor.getPosition(this.el.getBoundingClientRect())),
+                    map(() => {
+                        const rect = this.el.getBoundingClientRect();
+                        const width = this.el.clientWidth || rect.width;
```
waterplea: I think it should be offsetWidth/Height rather than client

projects/core/services/position.service.ts
vladimirpotekhin: https://github.com/taiga-family/taiga-ui/pull/14227#discussion_r3355835815

projects/core/portals/hint/hint.component.ts
```diff
@@ -127,6 +127,16 @@ export class TuiHintComponent {
     }
 
     private update(left: number, top: number): void {
+        const animating = this.el
+            .getAnimations?.()
+            .every(
+                (animation) => animation.effect?.getComputedTiming().progress !== null,
```
waterplea: `({effect})` ?

projects/core/portals/hint/hint.component.ts
```diff
@@ -127,6 +127,16 @@ export class TuiHintComponent {
...
+            );
+
+        if (animating) {
```
waterplea: Maybe just have the condition here, without const?

projects/core/components/textfield/textfield.template.html
```diff
@@ -5,7 +5,7 @@
 <ng-content select="label" />
 <span
     class="t-content"
-    (pointerdown)="input()?.nativeElement?.focus()"
+    (pointerdown)="onContentPointerDown($event)"
```
vladimirpotekhin: revert and remove test

projects/demo-playwright/tests/kit/input-time/input-time.pw.spec.ts
```diff
@@ -407,11 +407,6 @@ test.describe('InputTime', () => {
                     example.locator('tui-textfield:has([tuiInputTime]):first-of-type'),
                 );
             });
-
-            test('is clickable', async () => {
-                await inputTime.clickOnIcon();
-                await expect(inputTime.nativePicker).toBeFocused();
```
vladimirpotekhin, line 413: unfortunately, this is an unreliable test.

waterplea (approved):

splincode (approved):

PR #14230, opened 2026-06-04
feat(core): `Cell` add stretch directive

Yanduz:
# TuiCellStretch Directive

## Problem

When using `tuiCell` inside containers with padding (like `tui-sheet-dialog` with `padding: 0 1rem`), cells have unwanted gaps. Previously, developers had to use CSS hacks:

```css
/* Hack */
.cell {
  margin: 0 -1rem;
  border-radius: 0;
}
```

## Solution

The `tuiCellStretch` directive provides a native way to stretch cells to full container width:

```html
<button tuiCell tuiCellStretch>Full Width Cell</button>
```

image

### In Sheet Dialog

```html
<tui-sheet-dialog>
  <button tuiCell tuiCellStretch>
    <div tuiAvatar="@tui.home"></div>
    <div tuiTitle>
      Home
      <div tuiSubtitle>Main page</div>
    </div>
  </button>
  
  <button tuiCell tuiCellStretch>
    <div tuiAvatar="@tui.user"></div>
    <div tuiTitle>
      Profile
      <div tuiSubtitle>Account settings</div>
    </div>
  </button>
</tui-sheet-dialog>
```

## How It Works

- `inline-size: 100%` — full width
- `margin-inline: -1rem` — compensates parent padding
- `border-radius: 0` (mobile only) — edge-to-edge on mobile devices

Mobile detection uses `WA_IS_MOBILE` (user agent), not screen size.

## Import

```ts
import {TuiCellStretch} from '@taiga-ui/core';
```

### DIFF
--- a/projects/core/components/cell/cell-stretch.directive.ts
+++ b/projects/core/components/cell/cell-stretch.directive.ts
@@ -1,0 +1,10 @@
+import {Directive, inject} from '@angular/core';
+import {WA_IS_MOBILE} from '@ng-web-apis/platform';
+
+@Directive({
+    selector: '[tuiCell][tuiCellStretch]',
+    host: {'[style.border-radius]': 'isMobile ? 0 : null'},
+})
+export class TuiCellStretch {
+    protected readonly isMobile = inject(WA_IS_MOBILE);
+}
--- a/projects/demo/src/pages/components/cell/examples/10/index.html
+++ b/projects/demo/src/pages/components/cell/examples/10/index.html
@@ -1,0 +1,87 @@
+<button
+    tuiButton
+    type="button"
+    (click)="open = true"
+>
+    Open Sheet Dialog
+</button>
+
+<ng-template
+    let-observer
+    [tuiSheetDialogOptions]="options"
+    [(tuiSheetDialog)]="open"
+>
+    <nav>
+        <button
+            tuiCell="l"
+            tuiCellStretch
+            type="button"
+        >
+            <div
+                appearance="primary"
+                tuiAvatar="@tui.home"
+            ></div>
+            <div tuiTitle>
+                Home
+                <div tuiSubtitle>Main page</div>
+            </div>
+        </button>
+
+        <button
+            tuiCell="l"
+            tuiCellStretch
+            type="button"
+        >
+            <div
+                appearance="accent"
+                tuiAvatar="@tui.user"
+            ></div>
+            <div tuiTitle>
+                Profile
+                <div tuiSubtitle>Account settings</div>
+            </div>
+        </button>
+
+        <button
+            tuiCell="l"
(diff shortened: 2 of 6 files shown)


projects/core/components/cell/cell-stretch.directive.ts
```diff
@@ -0,0 +1,35 @@
...
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+class Styles {}
+
+@Directive({
+    selector: '[tuiCell][tuiCellStretch]',
+    host: {
+        'data-tui-version': TUI_VERSION,
+        tuiCellStretch: '',
```
waterplea: These 2 lines are unnecessary, version is already applied by `Cell` and since there's no input names `tuiCellStretch` the attribute is already there as directive selector

projects/core/components/cell/cell-stretch.directive.ts
```diff
@@ -0,0 +1,35 @@
...
+        '[class._mobile]': 'isMobile',
```
waterplea: Let's just do [style.border-radius]="isMobile ? 0 : null"

projects/core/components/cell/cell-stretch.directive.ts
```diff
@@ -0,0 +1,35 @@
...
+    ViewEncapsulation,
+} from '@angular/core';
+import {WA_IS_MOBILE} from '@ng-web-apis/platform';
+import {TUI_VERSION} from '@taiga-ui/cdk/constants';
+import {tuiWithStyles} from '@taiga-ui/cdk/utils/miscellaneous';
+
+@Component({
+    template: '',
+    styles: `
+        [data-tui-version='${TUI_VERSION}'] {
```
waterplea: Maybe we can just put those styles into cell stylesheet? I'm not sure it's worth it to create all this infrastructure just for 2 styles. Maybe just move margin-inline to `host`? Is `inline-size: 100%` required or will it get properly stretched anyway?

waterplea (approved):

nsbarsukov (approved):

projects/demo/src/pages/components/cell/examples/10/index.html
```diff
@@ -0,0 +1,87 @@
...
+<ng-template
+    let-observer
+    [tuiSheetDialogOptions]="options"
+    [(tuiSheetDialog)]="open"
+>
+    <nav>
+        <button
+            tuiCellStretch
+            type="button"
+            [tuiCell]="'l'"
```
vladimirpotekhin: nit: here and below `tuiCell="l"` (without binding)

vladimirpotekhin (approved):

PR #14240, opened 2026-06-05
fix(core): `Hint` prevent global options mutation in nested providers

### LINKED ISSUE #14232: 🐞 - `tuiHintOptionsProvider` mutates parent options
### Reproduction url

https://stackblitz.com/edit/angular-rrcvtvr3

### Description

When you use `tuiHintOptionsProvider` it changes values of the parent options object.

### Taiga UI version

5

### Browser Used

Chrome

### OS Used

macOS

vladimirpotekhin:
Fixes #14232


waterplea (approved):
Ok, so we no longer need the overriding and it was just a relic of the past?

vladimirpotekhin:
> Ok, so we no longer need the overriding and it was just a relic of the past?

Yes, this is no longer relevant

splincode (approved):

PR #14275, opened 2026-06-10
fix(kit): `Textfield[multi]` cannot update viewed items when `updateOn: submit`

### LINKED ISSUE #12361: 🐞 - InputDateMulti and updateOn submit
### Reproduction url

https://taiga-ui.dev/components/input-date-multi/API?sandboxExpanded=true&updateOn=submit

### Description

I can't remove any submitted dates

### Taiga UI version

latest

### Browser Used

Chrome

### OS Used

macOS

nsbarsukov:
Fixes #12361


splincode (approved):

projects/core/components/textfield/textfield-multi/textfield-multi.component.ts
```diff
@@ -100,13 +100,17 @@ export class TuiTextfieldMultiComponent<T> extends TuiTextfieldComponent<T> {
     public readonly item = contentChild(TuiItem, {read: TemplateRef, descendants: true});
     public readonly rows = input(100);
 
+    public get items(): readonly T[] {
+        return (
+            this.cva()?.value() ?? // preferred for updateOn: 'blur' | 'submit'
```
waterplea: I wonder if we actually need control or can just use CVA all the time. In basic textfield we do not require TuiControl for basic inputs, but for multi they are all based on InputChip which IS a TuiControl.

projects/kit/components/input-date-multi/input-date-multi.directive.ts
```diff
@@ -54,10 +54,7 @@ export class TuiInputDateMultiDirective extends TuiInputChipDirective<TuiDay> {
     protected readonly stringify = tuiDirectiveBinding(
         TuiItemsHandlersDirective,
         'stringify',
-        (item) =>
-            this.dateMultiOptions.valueTransformer
-                .fromControlValue([item])[0]
-                ?.toString(this.format().mode, this.format().separator) ?? '',
+        (item) => item?.toString(this.format().mode, this.format().separator) ?? '',
```
waterplea, line 57: Why we no longer need to transform?

waterplea (approved):

projects/kit/components/input-date-multi/input-date-multi.directive.ts
nsbarsukov: After this line was changed
https://github.com/taiga-family/taiga-ui/blob/8288d540427f1ff9d690c19aecd7c5f3ec08415d/projects/core/components/textfield/textfield-multi/textfield-multi.template.html#L14-L19

```diff
- @for (item of control()?.value; track item) {
+ @for (item of items; track item) {
```

```ts
public get items(): readonly T[] {
    return (
        this.cva()?.value() ?? // preferred for updateOn: 'blur' | 'submit'
        this.control()?.value ??
        []
    );
}
```

https://github.com/taiga-family/taiga-ui/blob/8288d540427f1ff9d690c19aecd7c5f3ec08415d/projects/cdk/classes/control.ts#L52

`Textfield[multi]` items uses non transformed internal value as polymorpheus context

```html
[polymorpheusOutletContext]="{$implicit: {item, ... }" // always not-transformed internal value 
```

projects/core/components/textfield/textfield-multi/textfield-multi.component.ts
nsbarsukov: Good point!

Let's drop the excessive fallback — at least for now.

PR #14287, opened 2026-06-11
fix(kit): `InputColor` base color reset when using opacity slider in hexa

mdlufy:
When `format="hexa"` and the user types a 7-character RGB color (e.g. `#ffffff`)
without an alpha part yet, adjusting the opacity slider reset the base color to
`#000000` instead of preserving the entered value.

Root cause: `filled()` conflated two distinct checks:
- "does the value have an RGB part?" (`length >= 7`) — needed in `onOpacity`
- "is the full hexa color entered?" (`length === 9`) — needed to read the alpha byte

The fix removes `filled()` and uses each condition directly where it's needed.
Also adds a `|| 255` fallback in `opacity()` to guard against `NaN` from
`parseInt` on a malformed hex string.

Fixes: opacity slider resets base color to #000000 on partial hexa input
Fixes: color preview invisible (`--t-opacity: 0`) when 7-char hex entered in hexa mode


projects/kit/components/input-color/input-color.component.ts
```diff
@@ -71,14 +71,10 @@ export class TuiInputColorComponent extends TuiControl<string> {
...
-    );
-
     protected readonly opacity = computed(() =>
-        this.filled() && this.format() === 'hexa'
-            ? Number.parseInt(this.value().slice(-2), 16)
-            : 255,
+        this.format() !== 'hexa' || this.value().length !== 9
+            ? 255
+            : Number.parseInt(this.value().slice(-2), 16) || 255,
     );
```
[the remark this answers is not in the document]
mdlufy: fixed

projects/demo-cypress/src/tests/input-color.cy.ts
```diff
@@ -0,0 +1,47 @@
+import {ChangeDetectionStrategy, Component} from '@angular/core';
+import {FormsModule} from '@angular/forms';
+import {TuiTextfield} from '@taiga-ui/core';
+import {TuiInputColor} from '@taiga-ui/kit';
+
+@Component({
```
splincode, line 6: Looks good overall, thanks!

Could you please also add a small regression test for fully transparent HEA value? It would ensure that opacity `0` is preserved and never falls back to `255`.

```ts
@Component({
    imports: [FormsModule, TuiInputColor, TuiTextfield],
    template: `
        <tui-textfield>
            <input
                tuiInputColor
                [(ngModel)]="value"
                [format]="format"
            />
        </tui-textfield>
    `,
})
class TestTransparentInputColor {
    public value = '#ffffff00';
    public format: 'hex' | 'hexa' = 'hexa';
}

it('keeps zero opacity for fully transparent hexa value', () => {
    cy.mount(TestTransparentInputColor);

    cy.get('input[type="color"]').should(($input) => {
        expect($input.get(0).style.getPropertyValue('--t-opacity')).to.equal('0');
    });

    cy.get('input[type="range"]').should('have.value', '0');
});
```

projects/kit/components/input-color/input-color.component.ts
```diff
@@ -71,15 +71,15 @@ export class TuiInputColorComponent extends TuiControl<string> {
         })),
     );
 
-    protected readonly filled = computed(() =>
-        this.format() === 'hex' ? this.value().length === 7 : this.value().length === 9,
-    );
+    protected readonly opacity = computed(() => {
```
splincode: Optional, maybe we can extract `value().length >= 7` into a separate computed property to make the intent clearer?

```ts
protected readonly hasRgb = computed(() => this.value().length >= 7);
```

Then it can be reused in both places:

```ts
protected onOpacity(opacity: number): void {
    const value = this.hasRgb() ? this.value().slice(0, 7) : '#000000';

    this.onChange(`${value}${toHex(opacity)}`);
}
```

```html
[style.--t-opacity]="hasRgb() ? opacity() / 255 : 0"
```

Not a blocker, just a readability suggestion.

mdlufy: updated

projects/demo-cypress/src/tests/input-color.cy.ts
mdlufy: updated

projects/kit/components/input-color/input-color-content.component.ts
```diff
@@ -26,18 +26,17 @@ import {TuiInputColorComponent} from './input-color.component';
...
+    protected readonly hasRgb = computed(() => this.host.value().length >= 7);
 
-    protected readonly filled = computed(() =>
-        this.host.format() === 'hex'
-            ? this.host.value().length === 7
-            : this.host.value().length === 9,
-    );
+    protected readonly opacity = computed(() => {
+        if (this.host.format() !== 'hexa' || this.host.value().length !== 9) {
+            return 255;
```
splincode: I think you can put it in field because it is used in 3 places

```ts
protected readonly maxAlpha = 255;

protected readonly opacityRatio = computed(() =>
    this.hasRgb() ? this.opacity() / this.maxAlpha : 0,
);

protected readonly opacity = computed(() => {
        if (this.host.format() !== 'hexa' || this.host.value().length !== 9) {
            return this.maxAlpha;
        }

        const parsed = Number.parseInt(this.host.value().slice(-2), 16);

        return Number.isNaN(parsed) ? this.maxAlpha : parsed;
    });

```


```html
[style.--t-opacity]="opacityRatio()"
```

mdlufy: updated

splincode (approved):

waterplea (approved):

Issue #14249, opened 2026-06-08
🚀 - Improve font scaling


waterplea:
### Description

When OS implements font scaling, all our components respond to it by increasing font size up until 150%, which can be tracked by `TUI_FONT_OFFSET` signal token that stores by how many pixels default font size is increased (11 corresponds to 150% and it does not go any further). Some of our components still need special treatment. Here are the ones that remain:

- [x] `Cell` needs to be able to stack labels in a column and move accessories to the new line
- [x] `Pin` needs to scale in size according to the fonts
- [x] `Toast` needs to move button to the new line in maximum scaling
- [x] `Badge` needs to scale icons in a unique way to maintain distance from the edges as font size grows~

PR #14299, opened 2026-06-12
feat(kit): `Toast` add font scaling

waterplea:
Partial #14249

### DIFF
--- a/projects/styles/components/toast.less
+++ b/projects/styles/components/toast.less
@@ -39,13 +39,13 @@
         grid-row: 1;
     }
 
-    [tuiButtonX],
-    [tuiIconButton],
+    > [tuiButtonX],
+    > [tuiIconButton],
     > :first-child:not(tui-shrink-wrap, [tuiButton]) {
         grid-row: span 2;
     }
 
-    > tui-icon,
+    > tui-icon:not([tuiBadge]),
     &::after,
     &::before {
         min-inline-size: 1em;
@@ -58,17 +58,17 @@
         color: var(--tui-text-tertiary);
     }
 
-    [tuiAvatar] {
+    > [tuiAvatar] {
         margin-inline-end: 0;
     }
 
-    tui-loader,
-    [tuiButton] {
+    > tui-loader,
+    > [tuiButton] {
         inline-size: fit-content;
         min-inline-size: fit-content;
     }
 
-    &._rearranged [tuiButton] {
+    &._rearranged > [tuiButton] {
         grid-row: 2;
         margin: 0.75rem 0 0 !important;
     }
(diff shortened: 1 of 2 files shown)


vladimirpotekhin (approved):

projects/styles/components/toast.less
```diff
@@ -34,24 +35,41 @@
...
-    [tuiButton] {
+    > tui-loader,
+    > [tuiButton] {
         inline-size: fit-content;
         min-inline-size: fit-content;
     }
+
+    &._rearranged > [tuiButton] {
+        grid-row: 2;
+        margin: 0.75rem 0 0 !important;
```
splincode, line 73: Could we drop `!important` here?

`&._rearranged > [tuiButton]` is more specific than the generic child margin rule and comes later in the file, so the override should work without `!important`.

projects/styles/components/toast.less
```diff
@@ -34,24 +35,41 @@
...
 
-    [tuiAvatar] {
+    > [tuiAvatar] {
         margin-inline-end: 0;
     }
 
-    tui-loader,
-    [tuiButton] {
+    > tui-loader,
+    > [tuiButton] {
```
splincode, line 66: Is narrowing these selectors to direct children intentional?

It looks safer for toast layout because nested buttons, avatars inside the content should not be affected by toast-level layout styles. But this is still a behavior change, so it would be good to make sure existing toast examples with avatar, loader, button are covered by screenshots.

projects/styles/components/toast.less
```diff
@@ -34,24 +35,41 @@
...
         inline-size: fit-content;
         min-inline-size: fit-content;
     }
+
+    &._rearranged > [tuiButton] {
+        grid-row: 2;
+        margin: 0.75rem 0 0 !important;
+    }
```
splincode, line 74:
```less
&._rearranged > [tuiButton]:not([tuiButtonX], [tuiIconButton]) {
    grid-row: 2;
    margin: 0.75rem 0 0;
}
```

Do we need to exclude icon-like buttons here?

If some icon button can also match `[tuiButton]`, this rule would move it to the second row despite the `tuiButtonX`, `tuiIconButton` rules above.

splincode (approved):

nsbarsukov (approved):

projects/styles/components/toast.less
waterplea: All current screenshots pass, I believe it was omitted by accident before, those styles are inteded for direct children only.

projects/styles/components/toast.less
waterplea: Icon buttons do not have tuiButton attribute.

projects/styles/components/toast.less
waterplea: I'll see if we can in my next PR

PR #14303, opened 2026-06-13
fix(layout): `InputSearch` hydration

mdlufy:
Fixes #13646


projects/layout/components/input-search/input-search-content.component.ts
```diff
@@ -0,0 +1,33 @@
...
+    `,
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {'data-tui-version': TUI_VERSION},
+})
+export class TuiInputSearchContent {
+    private readonly tpl = viewChild.required(TemplateRef);
+
+    protected readonly host = inject(TuiInputSearch);
+    protected readonly ef = effect(() => this.host.template.set(this.tpl()));
```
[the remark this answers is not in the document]
mdlufy: no need in angular 19

projects/layout/components/input-search/input-search.component.ts
```diff
@@ -54,6 +55,16 @@ export class TuiInputSearch implements OnChanges {
 
     protected readonly icon = tuiIconStart(inject(TUI_COMMON_ICONS).search, {});
 
+    protected readonly openEffect = effect(() => {
+        const template = this.template();
+
+        if (template && untracked(this.searchOpen)) {
+            this.open();
+        }
+    });
```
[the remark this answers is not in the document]
mdlufy: no need in angular 19

projects/layout/components/input-search/input-search.component.ts
```diff
@@ -87,23 +100,24 @@ export class TuiInputSearch implements OnChanges {
...
         if (
-            container &&
             target !== this.el &&
             tuiIsElement(target) &&
-            !tuiContainsOrAfter(container.nativeElement, target)
+            !tuiContainsOrAfter(this.ref?.rootNodes[0], target)
         ) {
             this.close();
         }
     }
```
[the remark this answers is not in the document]
mdlufy: fixed

projects/layout/components/input-search/input-search.component.ts
```diff
@@ -66,15 +59,19 @@ export class TuiInputSearch implements OnChanges {
...
         this.parent = this.textfield.el.parentElement;
         this.neighbor = this.textfield.el.nextSibling;
-        this.ref = this.service.add(this.template());
-        this.ref.rootNodes[0]?.insertAdjacentElement('afterbegin', this.textfield.el);
+        this.ref = this.service.add(
+            new PolymorpheusComponent(TuiInputSearchContent, this.injector),
+        );
+        this.ref.instance
+            .container()
+            .nativeElement.insertAdjacentElement('afterbegin', this.textfield.el);
```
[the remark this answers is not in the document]
mdlufy: Portal in AddComponent already calls detectChanges() before returning ref, so container() is engaged

projects/layout/components/input-search/index.ts
```diff
@@ -1 +1,2 @@
 export * from './input-search.component';
+export * from './input-search-content.component';
```
splincode: I think we should avoid exporting `TuiInputSearchContent` from the public entry point.

It looks like an internal implementation detail used only to render the popup content dynamically. If we export it here, it becomes part of the public API and will be harder to refactor or remove later.

Maybe we can keep only:

```ts
export * from './input-search.component';
```

Do we usually expose such internal content components as part of the public API?

projects/layout/components/input-search/input-search.component.ts
```diff
@@ -87,23 +84,27 @@ export class TuiInputSearch implements OnChanges {
...
-    protected onArrow(): void {
-        tuiGetClosestFocusable({
-            initial: this.container()?.nativeElement || this.el,
-            root: this.container()?.nativeElement || this.el,
-        })?.focus();
-    }
-
-    protected onFocus({target}: Event): void {
-        const container = this.container();
+    public onFocus({target}: Event): void {
```
splincode: Should we add the same destroyed guard here as in `onArrow`?

`onFocus` can be triggered from a document-level listener, so it feels safer to avoid accessing `this.ref.instance` after the popup has already been destroyed.

For example:

```ts
public onFocus({target}: Event): void {
    if (this.ref?.hostView.destroyed !== false) {
        return;
    }

    const root = this.ref.instance.container().nativeElement;

    if (
        target !== this.el &&
        tuiIsElement(target) &&
        !tuiContainsOrAfter(root, target)
    ) {
        this.close();
    }
}
```

projects/layout/components/input-search/input-search-content.template.html
```diff
@@ -0,0 +1,12 @@
+<div
```
splincode: Maybe we should split this ref into two separate refs: `root` and `content`.

Currently `#container` points to the outer overlay container. Before this change, the focus navigation was based on the content area. Splitting the refs would make the intent clearer:

```html
<div
    #root
    tuiTheme="dark"
    class="t-container"
    (document:focusin)="host.onFocus($event)"
    (keydown.esc)="host.close()"
    (pointerdown.self)="host.close()"
>
    <div
        #content
        class="t-content"
    >
        <ng-container *polymorpheusOutlet="host.tuiInputSearch()" />
    </div>
</div>
```

Then root can be used for outside-focus checks and inserting the textfield, while content can be used for focus navigation.

```ts
public readonly root = viewChild.required<ElementRef<HTMLElement>>('root');
public readonly content = viewChild.required<ElementRef<HTMLElement>>('content');
```

splincode (approved):

projects/layout/components/input-search/index.ts
mdlufy: All other components (fixed in https://github.com/taiga-family/taiga-ui/issues/13646) are already exported

I agree that it is possible to export it excessively, we need to think about how best to do this and in other cases

projects/layout/components/input-search/input-search.component.ts
mdlufy: Good catch. Added the same hostView.destroyed guard for consistency with onArrow - even though the (document:focusin) listener lives inside the content view (so it's removed on destroy), the guard keeps onFocus locally correct instead of relying on that invariant

projects/layout/components/input-search/input-search-content.template.html
mdlufy: One clarification: the focus-navigation behavior here is actually unchanged. The previous version used this.ref.rootNodes[0], which was also the .t-container (the embedded view's root), not .t-content - so the current single ref reproduces the old behavior exactly: tuiGetClosestFocusable({initial: textfield.nextElementSibling /* = .t-content */, root: .t-container})

Also, onFocus has to use .t-container, not .t-content: we move the <input> into .t-container (afterbegin), i.e. outside .t-content. Scoping the outside-focus check to .t-content would treat focusing the input itself as "outside" and close the popup immediately on typing

So a single #container ref is sufficient and correct here. Happy to split into #root/#content if you prefer the explicit intent separation, but it wouldn't change behavior

projects/layout/components/input-search/index.ts
splincode: Makes sense, thanks for the context.

If this is already consistent with the other hydration fixes, I'm okay with keeping it as is in this PR. Maybe we can revisit this separately later and decide whether these internal content components should be exposed from public entry points at all.

projects/layout/components/input-search/input-search.component.ts
splincode: Thanks

projects/layout/components/input-search/input-search-content.template.html
splincode: Thanks for the clarification, that makes sense.

I missed that the previous behavior was also based on the embedded view root and that the input is moved into `.t-container` outside of `.t-content`. In that case, keeping a single `#container` ref is fine with me.

vladimirpotekhin (approved):

projects/layout/components/input-search/input-search.component.ts
```diff
@@ -66,44 +59,62 @@ export class TuiInputSearch implements OnChanges {
...
+            this.searchOpen.set(false);
+
+            return;
+        }
+
         this.el.placeholder = this.placeholder || this.el.placeholder;
         this.parent?.insertBefore(this.textfield.el, this.neighbor);
-        this.ref?.destroy();
+        this.ref.destroy();
+        this.ref = undefined;
```
nsbarsukov: Why do we need to explicitly set the variable to `undefined`, and why isn't the previous approach `if (this.ref?.hostView.destroyed !== false) {...}` sufficient?

projects/layout/components/input-search/input-search.component.ts
```diff
@@ -66,44 +59,62 @@ export class TuiInputSearch implements OnChanges {
...
+
+    protected onArrow(): void {
+        if (this.ref?.hostView.destroyed !== false) {
+            return;
+        }
+
+        const root = this.ref.instance.container().nativeElement;
+        const initial = this.textfield.el.nextElementSibling ?? root;
+
+        tuiGetClosestFocusable({initial, root})?.focus();
```
nsbarsukov: Excessive variable `initial` ?

```ts
const root = this.ref.instance.container().nativeElement;

tuiGetClosestFocusable({
    root,
    initial: this.textfield.el.nextElementSibling ?? root,
})?.focus();
```

projects/layout/components/input-search/input-search.component.ts
```diff
@@ -66,44 +59,62 @@ export class TuiInputSearch implements OnChanges {
...
+            !tuiContainsOrAfter(root, target)
         ) {
             this.close();
         }
     }
+
+    protected onArrow(): void {
+        if (this.ref?.hostView.destroyed !== false) {
+            return;
+        }
```
nsbarsukov: Why this `if`-condition is now required ?
Does it fixes any bug ?

Previous state:
https://github.com/taiga-family/taiga-ui/blob/8d400ed481c0c6bd1e00776a44f911ea2a723c24/projects/layout/components/input-search/input-search.component.ts#L90-L95

projects/layout/components/input-search/input-search.component.ts
mdlufy: Yes, ref undefined was removed, it seems it was redundant

About `if (this.ref?.hostView.destroyed !== false)`
Previously container was a viewChild returning undefined when closed (null-safe by nature), now they touch this.ref, and onArrow's keydown.arrowDown listener lives on the always-present input, so it can fire while the popup is closed - the guard avoids acting on a destroyed/absent ref

projects/layout/components/input-search/input-search.component.ts
mdlufy: Yes, updated

projects/layout/components/input-search/input-search.component.ts
mdlufy: I've tested and looks like `this.ref?....` is enough

projects/layout/components/input-search/input-search.component.ts
mdlufy: I've tested and looks like this.ref?.... is enough

It's for correct reaction, when popup is closed

waterplea (approved):

nsbarsukov (approved):

PR #14306, opened 2026-06-13
fix(kit): `Textarea` hydration

mdlufy:
Fixes #13646


projects/kit/components/textarea/textarea.style.less
```diff
@@ -3,33 +3,96 @@
...
+    }
+}
+
+tui-textfield._with-label[data-size='m'] textarea[tuiTextarea] {
     border-block-start-width: @top-m-label;
     padding-block: 0;
 }
 
-:host-context(tui-textfield._with-label[data-size='l']._with-label[data-size='l']) {
+tui-textfield._with-label[data-size='l'] textarea[tuiTextarea] {
```
splincode: Should be?

```less
tui-textfield:where(*&)._with-label[data-size='l'] textarea[tuiTextarea]:where(*&) {
```

projects/kit/components/textarea/textarea-content.component.ts
```diff
@@ -0,0 +1,34 @@
...
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {'data-tui-version': TUI_VERSION},
+})
+export class TuiTextareaContent {
+    private readonly textRef = viewChild<ElementRef<HTMLSpanElement>>('text');
+
+    protected readonly host = inject(TuiTextareaComponent);
+    protected readonly ef = effect(() => this.host.text.set(this.textRef()));
+}
```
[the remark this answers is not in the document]
mdlufy: in angular 19 it's okay

projects/kit/components/textarea/textarea-content.component.ts
```diff
@@ -0,0 +1,36 @@
...
+    encapsulation: ViewEncapsulation.None,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {'data-tui-version': TUI_VERSION},
+})
+export class TuiTextareaContent {
+    private readonly textRef = viewChild<ElementRef<HTMLSpanElement>>('text');
+
+    protected readonly host = inject(TuiTextareaComponent);
+    protected readonly textfield = inject(TuiTextfieldComponent);
+    protected readonly ef = effect(() => this.host.text.set(this.textRef()));
```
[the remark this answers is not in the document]
mdlufy: no need in angular 19

projects/kit/components/textarea/textarea.style.less
mdlufy: Looks like yes

projects/kit/components/textarea/textarea.component.ts
```diff
@@ -1,56 +1,48 @@
...
     public readonly max = input(this.options.max);
     public readonly content = input(this.options.content);
 
-    public ngAfterViewInit(): void {
-        this.vcr.createEmbeddedView(this.template());
+    constructor() {
+        afterNextRender(() => {
+            this.vcr.createComponent(TuiTextareaContent, {injector: this.injector});
+        });
     }
```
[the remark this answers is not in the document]
mdlufy: no need

projects/kit/components/textarea/textarea.component.ts
```diff
@@ -1,56 +1,50 @@
...
+    public readonly text = signal<ElementRef<HTMLSpanElement> | undefined>(undefined);
     public readonly min = input(this.options.min);
     public readonly max = input(this.options.max);
     public readonly content = input(this.options.content);
 
-    public ngAfterViewInit(): void {
-        this.vcr.createEmbeddedView(this.template());
+    public ngOnInit(): void {
+        this.vcr.createComponent(TuiTextareaContent, {injector: this.injector});
     }
```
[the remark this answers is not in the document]
mdlufy: it's not redudant, we need node injector

projects/kit/components/textarea/textarea-content.component.ts
```diff
@@ -0,0 +1,40 @@
...
+    host: {'data-tui-version': TUI_VERSION},
+})
+export class TuiTextareaContent {
+    private readonly textRef = viewChild<ElementRef<HTMLSpanElement>>('text');
+
+    protected readonly host = inject(TuiTextareaComponent);
+    protected readonly textfield = inject(TuiTextfieldComponent);
+
+    public scrollTo(top: number): void {
+        requestAnimationFrame(() => {
```
vladimirpotekhin: nit: It seems to me that it is better to leave rAF in the `TuiTextareaComponent`, and not in this public method, because it is here due to the peculiarities of the handler (scroll), which is located in the `TuiTextareaComponent`

vladimirpotekhin (approved):

projects/kit/components/textarea/textarea.component.ts
```diff
@@ -1,61 +1,55 @@
...
         // To trigger CD for #text
         '(scroll.once)': 'onScroll()',
         '(scroll.zoneless)': 'onScroll()',
     },
 })
-export class TuiTextareaComponent implements AfterViewInit {
-    private readonly template = viewChild.required(TemplateRef);
-    private readonly options = inject(TUI_TEXTAREA_OPTIONS);
+// TODO(v6): rename to TuiTextareaNative
+export class TuiTextareaComponent implements OnInit {
```
nsbarsukov: Why the `Native` suffix?

This component differs from `TuiInputDateNative`, `TuiNativeSelect`, etc. — it's not a wrapper around a **native browser picker**.

projects/kit/components/textarea/textarea.component.ts
```diff
@@ -1,61 +1,55 @@
...
     public readonly min = input(this.options.min);
     public readonly max = input(this.options.max);
     public readonly content = input(this.options.content);
 
-    public ngAfterViewInit(): void {
-        this.vcr.createEmbeddedView(this.template());
+    public ngOnInit(): void {
+        this.ref = this.vcr.createComponent(TuiTextareaContent, {
+            injector: this.injector,
+        }).instance;
```
nsbarsukov, line 49: Why does `Textarea` use a different approach compared to other similar components?
Is there a reason `TextfieldContent` isn't used here?

projects/kit/components/textarea/textarea-content.component.ts
mdlufy: Hmm, yes, updated

projects/kit/components/textarea/textarea.component.ts
mdlufy: Yes, updated

Suggest to rename current TuiTextareaDirective -> TuiTextareaLimit
and TuiTextareaComponent -> TuiTextareaDirective

projects/kit/components/textarea/textarea.component.ts
mdlufy: Without improvements, tuiAsTextfieldContent will not work, because it fills the inline #vcr inside the tui-textfield (#vcr inside the <span class="t-content"> slot)

For one-liners, this is fine, but for textarea, for example, you need the textarea to be right under the tui-textfield + scrollbar more

projects/kit/components/textarea/textarea.style.less
```diff
@@ -3,33 +3,96 @@
...
+        }
+
+        tui-textfield[data-size='l'] & {
+            margin-block-start: -2.5rem;
+            padding-block-end: 1rem;
+        }
+    }
+}
+
+tui-textfield._with-label[data-size='m']:where(*&) textarea[tuiTextarea]:where(*&) {
```
waterplea: It's enough to have just one :where here and in all next rules

projects/kit/components/textarea/textarea.component.ts
```diff
@@ -1,61 +1,55 @@
...
-    protected readonly isMobile = inject(WA_IS_MOBILE);
 
+    public readonly isMobile = inject(WA_IS_MOBILE);
     public readonly min = input(this.options.min);
     public readonly max = input(this.options.max);
     public readonly content = input(this.options.content);
 
-    public ngAfterViewInit(): void {
-        this.vcr.createEmbeddedView(this.template());
+    public ngOnInit(): void {
```
waterplea, line 46: I think we do not need ngOnInit now, we can just assign ref in definition

projects/kit/components/textarea/textarea.component.ts
```diff
@@ -1,61 +1,55 @@
...
-    private readonly text = viewChild('text', {read: ElementRef});
+    private readonly injector = inject(INJECTOR);
+    private readonly options = inject(TUI_TEXTAREA_OPTIONS);
+    private ref?: TuiTextareaContent;
 
     protected readonly el = tuiInjectElement<HTMLTextAreaElement>();
-    protected readonly textfield = inject(TuiTextfieldComponent<string>);
-    protected readonly isMobile = inject(WA_IS_MOBILE);
 
+    public readonly isMobile = inject(WA_IS_MOBILE);
```
waterplea: If we only need it in content, let's inject it there instead of making it public

projects/kit/components/textarea/textarea.style.less
mdlufy: Fixed

projects/kit/components/textarea/textarea.component.ts
mdlufy: Fixed

projects/kit/components/textarea/textarea.component.ts
mdlufy: Tried this, but moving creation to a field initializer throws NG0200: Circular dependency in DI at runtime. TuiTextareaContent injects TuiTextareaComponent, and the field initializer runs during the host's own construction, so createComponent tries to resolve the host before it's fully constructed and registered in the node injector. Keeping it in ngOnInit, where the host is already available, avoids the cycle

waterplea (approved):

nsbarsukov (approved):

PR #14310, opened 2026-06-14
chore: small refactor

splincode:


projects/demo-playwright/tests/kit/input-date-range/input-date-range.pw.spec.ts
```diff
@@ -264,12 +263,14 @@ test.describe('InputDateRange', () => {
 
             await inputDateRange.textfield.click();
             await calendarSheet.clickOnDay(1);
+            await page.mouse.move(0, 0);
```
vladimirpotekhin, line 266: Why do we need this?

splincode: After `clickOnDay`, the mouse pointer remains over the clicked day, because Playwright performs a real mouse click in the center of the element.

This can leave the calendar in a hover state and make the visual snapshot depend on the last mouse position. The `move(0, 0)` is only needed to normalize the pointer position before taking the screenshot.

It is not about disabled cells specifically. Disabled cells should not have a visible hover state, but the screenshot itself should still not depend on an accidental pointer position after the previous click.

I can also add a short comment here or replace it with a named helper like `moveMouseAway()` to make the intent clearer.

vladimirpotekhin (approved):

nsbarsukov (approved):

PR #14314, opened 2026-06-14
refactor: use display grid in form

splincode:


waterplea:
I believe I have flex-wrap there on purpose because sometimes it's easy to just put "display: flex" on fieldset with a lot of labeled radios so they are just listed in a row + if they do not fit they wrap.

PR #14331, opened 2026-06-15
fix(addon-mobile): keep sheet dropdown open on input blur

### LINKED ISSUE #12975: 🐞 - tuiDropdownMobile gets closed unexpectedly (ios)
### Reproduction url

https://stackblitz.com/edit/angular-q56mwd8k?file=package.json,src%2Fapp%2Fapp.template.html

### Description

1. If there is a textfield inside  tuiDropdownMobile, it gets closed when the field gets unfocused (user taps "done" on a virtual keyboard).
2. If the sheet is short, it gets closed when the textfield gets focused.

The button "long" reproduces the first scenario, "short" the second.

Ios only.

### Taiga UI version

4.66.0

### Browser Used

Safari

### OS Used

iOS

splincode:
Fixes #12975

### Before

https://github.com/user-attachments/assets/2ee9928f-ad24-407a-b30c-ca45ddad4799

### After

https://github.com/user-attachments/assets/88f1e6ac-e43a-4ad5-9589-53d5e880108c

The previous approach changed `TUI_ACTIVE_ELEMENT`, but that is too broad because `focusout` with `relatedTarget === null` is also a valid signal for manual `blur()`. Some components rely on this behavior to detect that focus has left the active zone.

There is no reliable DOM event for the iOS virtual keyboard `Done` button. `enterkeyhint="done"` only changes the virtual keyboard action label and does not prevent the same `blur` / `focusout` flow.

This fix keeps `TUI_ACTIVE_ELEMENT` and `TuiActiveZone` behavior unchanged. Instead, it handles the issue locally for `tuiDropdownMobile="..."` sheet mode: the sheet is no longer closed only because `TuiActiveZoneChange(false)` was emitted after an input blur.

This is safer because mobile sheet dropdowns already have explicit close interactions: backdrop/filler tap, swipe down, `close()` from dropdown context, and option/button actions. Regular dropdowns and non-sheet mobile dropdowns keep the existing active-zone close behavior.

### DIFF
--- a/projects/addon-mobile/directives/dropdown-mobile/dropdown-mobile.directive.ts
+++ b/projects/addon-mobile/directives/dropdown-mobile/dropdown-mobile.directive.ts
@@ -1,17 +1,29 @@
-import {computed, Directive, inject, Input} from '@angular/core';
+import {computed, Directive, inject, Injectable, Input, signal} from '@angular/core';
 import {TUI_IS_MOBILE} from '@taiga-ui/cdk/tokens';
 import {tuiIsHTMLElement} from '@taiga-ui/cdk/utils/dom';
 import {
+    TUI_DROPDOWN_ACTIVE_ZONE_CLOSE_GUARD,
     TUI_DROPDOWN_COMPONENT,
     TuiDropdownDirective,
 } from '@taiga-ui/core/directives/dropdown';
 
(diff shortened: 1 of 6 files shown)


waterplea:
relatedTarget is null when somebody calls manual `blur()` and it is considered leaving the zone, we cannot change this behavior. Is there any other way to detect the "Done" virtual button?

PR #14334, opened 2026-06-15
feat(core): `Cell` allow nested `Title`

waterplea:
Fixes #

### DIFF
--- a/projects/core/components/cell/cell.styles.less
+++ b/projects/core/components/cell/cell.styles.less
@@ -47,6 +47,7 @@
         max-block-size: var(--t-block-size);
         align-items: center;
         align-self: stretch;
+        margin-inline-start: auto;
     }
 
     // Specificity artificially increased for disabled state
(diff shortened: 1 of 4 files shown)


projects/styles/components/toast.less
```diff
@@ -70,6 +70,6 @@
 
     &._rearranged > [tuiButton] {
         grid-row: 2;
-        margin: 0.75rem 0 0 !important;
+        margin: 0.75rem 0 0;
```
mdlufy, line 73: nice without important 👍

mdlufy (approved):

vladimirpotekhin (approved):

projects/core/components/cell/cell.styles.less
```diff
@@ -47,6 +47,7 @@
         max-block-size: var(--t-block-size);
         align-items: center;
         align-self: stretch;
+        margin-inline-start: auto;
```
splincode, line 50: I would avoid setting `margin-inline-start: auto` on the base `[tuiAccessories]` selector.

`[tuiAccessories]` can also be used as a left-side accessory before `[tuiTitle]`, for example with avatar, badged content. Applying auto margin globally pushes such accessories to the inline end and changes existing layouts.

Could we move it to a more specific right-side case instead?

```less
[tuiTitle] ~ [tuiAccessories] {
    margin-inline-start: auto;
}

File: projects/core/components/cell/cell.styles.less [hunk 3]
```diff
@@ -84,27 +85,48 @@
...
 
-        &[tuiFade] {
-            align-items: flex-start;
+            ~ [tuiSubtitle] {
+                justify-content: space-between;
+            }
+        }
+
+        ~ [tuiTitle] {
+            flex-shrink: 3;
```
splincode, line 115: `~ [tuiTitle]` still inherits `flex-grow: 1` from the base `[tuiTitle] { flex: 1 7 auto; }`.

This changes the previous behavior: the right-side title can now grow and split available space with the main title, causing the main title to wrap/fade earlier.

Could we explicitly reset it?

```less
~ [tuiTitle] {
    flex: 0 3 auto;
    margin-inline-start: auto;
    text-align: end;
    align-items: flex-end;
}
```

splincode (approved):

projects/core/components/cell/cell.styles.less
waterplea: Accessories are only on the right and never on the left side.

projects/core/components/cell/cell.styles.less
waterplea: It works fine this way, as we can see on the screenshots

PR #14336, opened 2026-06-15
feat(core): `CellResponsive` add directive

waterplea:
Fixes #14249

### DIFF
--- a/projects/core/components/cell/cell.styles.less
+++ b/projects/core/components/cell/cell.styles.less
@@ -85,7 +85,7 @@
     }
 
     [tuiTitle] {
-        flex: 1 7 auto;
+        flex: 3 7 30%;
         align-items: normal;
         text-align: start;
         gap: 0.25rem 0.5rem;
@@ -96,13 +96,13 @@
 
             ~ [tuiSubtitle] > [tuiFade]:first-child,
             > [tuiFade]:first-child {
-                flex-shrink: 7;
+                flex: 3 7 30%;
                 white-space: nowrap;
             }
 
             ~ [tuiSubtitle] > [tuiFade]:last-child,
             > [tuiFade]:last-child {
-                flex-shrink: 3;
+                flex: 7 3 70%;
                 white-space: nowrap;
             }
 
@@ -112,7 +112,7 @@
         }
 
         ~ [tuiTitle] {
-            flex-shrink: 3;
+            flex: 7 3 70%;
             margin-inline-start: auto;
             text-align: end;
             align-items: flex-end;
--- a/projects/core/components/cell/cell.styles.less
+++ b/projects/core/components/cell/cell.styles.less
@@ -240,6 +240,34 @@
         inline-size: 100%;
         margin-inline: -1rem;
     }
+
+    &[tuiCellResponsive] {
+        display: grid;
+        grid-auto-flow: column;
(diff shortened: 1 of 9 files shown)


projects/core/components/cell/cell.styles.less
```diff
@@ -240,6 +240,34 @@
...
+            grid-template-columns: 1fr;
+        }
+
+        > * {
+            grid-row: 1;
+        }
+
+        &._rearranged {
+            [tuiAccessories] {
+                grid-row: 2;
```
splincode, line 263: Looks good overall, but I think we should fix a couple of layout details before merge.

1. In `_rearranged` state `[tuiAccessories]` should probably span the whole grid:

```less
&._rearranged {
    [tuiAccessories] {
        grid-row: 2;
        grid-column: 1 / -1;
    }
}
```

Without explicit grid-column, CSS Grid auto-placement can put accessories into the first column of the second row, which may expand the avatar column and squeeze the title.

vladimirpotekhin (approved):

projects/demo/src/pages/components/cell/examples/11/index.less
```diff
@@ -0,0 +1,13 @@
...
+    display: flex;
+    flex-direction: column;
+    gap: 2rem;
+}
+
+[tuiCell] {
+    max-inline-size: 25rem;
+
+    &:last-child [tuiAccessories] {
+        grid-row: 1;
```
splincode, line 11: Could we also reset grid-column here?

If `[tuiAccessories]` gets `grid-column: 1 / -1` in the main `_rearranged` styles, this demo-specific override only moves it back to the first row, but it will still span all columns.

```less
&:last-child [tuiAccessories] {
    grid-row: 1;
    grid-column: auto;
}
```

Otherwise this example may accidentally overlap with the avatar, title area.

projects/demo/src/pages/components/cell/index.html
```diff
@@ -8,8 +8,17 @@
             <tui-doc-example
                 [component]="$index + 1 | tuiComponent"
                 [content]="$index + 1 | tuiExample"
+                [description]="description"
                 [heading]="example"
             />
+            <ng-template #description>
+                @switch ($index) {
+                    @case (10) {
+                        Open this example on the phone with OS font size increased by 150% to see cell rearrange for
```
splincode, line 17: What if

```
Increase OS font size to the maximum value to see cell rearrange for better readability.
```

projects/core/components/cell/cell.styles.less
```diff
@@ -240,6 +240,34 @@
         inline-size: 100%;
         margin-inline: -1rem;
     }
+
+    &[tuiCellResponsive] {
+        display: grid;
```
splincode, line 248: Should we keep the same vertical alignment here?

Regular `[tuiCell]` uses align-items: center, but this responsive variant switches the layout to grid. Without explicit align-items, grid defaults to stretch, so avatars, titles and accessories may be aligned differently.

```less
&[tuiCellResponsive] {
    display: grid;
    align-items: center;
    grid-auto-flow: column;
    grid-auto-columns: fit-content(100%);
    grid-template-columns: auto 1fr;
}
```

splincode (approved):

nsbarsukov (approved):

PR #14339, opened 2026-06-16
fix(addon-mobile): close sheet dialog before scroll snap ends

### LINKED ISSUE #12355: 🐞 - SheetDialog close animation is extremely slow on IOS
### Reproduction url

https://taiga-ui.dev/components/sheet-dialog

### Description

I made a quick swipe down gesture, but it takes around 2 seconds for dialog to close, which feels  extremely slow.
Is there a way to fix it?

https://github.com/user-attachments/assets/d5e18bbe-4f6d-4fe7-9938-6451fd014e80

### Taiga UI version

4.59.0

### Browser Used

Safari

### OS Used

iOS

splincode:
Fixes #12355 

### Before

https://github.com/user-attachments/assets/2109171b-22e1-46c0-bb39-f119c4762eb2

### After

https://github.com/user-attachments/assets/60fc2c6a-1ab6-4659-84a4-96defd108943

### DIFF
--- a/projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
+++ b/projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
@@ -117,7 +117,7 @@ export class TuiSheetDialogComponent<I> implements AfterViewInit {
             this.close$.next();
         }
 
-        if (this.el.scrollTop >= this.lastScrollTop) {
+        if (this.el.scrollTop > this.lastScrollTop) {
             this.lastScrollTop = Number.NaN;
         }
     }
--- a/projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
+++ b/projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
@@ -37,15 +37,17 @@ const REQUIRED_ERROR = new Error(ngDevMode ? 'Required dialog was dismissed' : '
         '[class._closeable]': 'context.closable',
         '[style.--tui-offset.px]': 'context.offset',
         '(click.self)': 'close$.next()',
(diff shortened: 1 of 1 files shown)


waterplea:
I think the way to do it is:

1. listen to `touchend` event
2. check the `scrollTop` and if it's higher than initial stop then ignore
3. if it's lower the initial stop start listening to the next `scroll` event
4. if `scrollTop` in the next `scroll` event is even lower then it's being snapped to the top and we can just close the sheet without waiting for it to complete the snap animation

That's a theoretical, needs thorough testing.

projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
```diff
@@ -77,22 +78,64 @@ export class TuiSheetDialogComponent<I> implements AfterViewInit {
...
+
+        if (scrollTop <= 0) {
+            this.close$.next();
+
+            return;
+        }
+
+        if (scrollTop < this.initial) {
+            this.scrollTopAfterTouchEnd = scrollTop;
+        }
```
waterplea, line 105: Let's make this one shorter:
```ts
this.pointers = Math.max(this.pointers - 1, 0);

if (!this.pointers && this.el.scrollTop <= 0) {
    this.close$.next();
} else if (!this.pointers && this.el.scrollTop < this.initial) {
    this.scrollTopAfterTouchEnd = this.el.scrollTop;
}
```
Just reading scrollTop multiple times does not cause multiple reflows:

Image

splincode: fixed

waterplea (approved):

PR #14347, opened 2026-06-16
fix(kit): split edited input chip by separator

### LINKED ISSUE #14242: 🐞 -  The chip is not split when editing
### Reproduction url

https://taiga-ui.dev/v4/components/input-chip/API

### Description

When editing chip, the separator does not split the tag

### Taiga UI version

4.85.0, 5.9.0

### Browser Used

Chrome

### OS Used

macOS

splincode:
Fixes #14242

### DIFF
--- a/projects/kit/components/input-chip/input-chip.component.ts
+++ b/projects/kit/components/input-chip/input-chip.component.ts
@@ -33,6 +33,9 @@ import {TUI_FILE_TEXTS} from '@taiga-ui/kit/tokens';
 import {tuiInjectValue} from '@taiga-ui/kit/utils';
 import {injectContext} from '@taiga-ui/polymorpheus';
 
+import {TuiInputChipDirective} from './input-chip.directive';
+import {tuiParseInputChipValue} from './input-chip.utils';
+
 @Component({
     selector: 'tui-input-chip',
     imports: [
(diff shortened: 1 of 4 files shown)


projects/kit/components/input-chip/input-chip.component.ts
```diff
@@ -113,17 +116,39 @@ export class TuiInputChipComponent<T> {
...
+        }
+
+        if (this.handlers.disabledItemHandler()(item)) {
             return;
         }
 
-        const value = this.value().map((item, index) =>
-            index === this.index ? this.internal() : item,
-        );
+        if (cva instanceof TuiInputChipDirective && tuiIsString(item)) {
```
waterplea: item is always a string here since editing can only happen on string values, we can just use String(item) to satisfy types. And cva is always InputChip so there's no need for else branch

splincode: fixed

projects/kit/components/input-chip/input-chip.component.ts
```diff
@@ -113,17 +116,38 @@ export class TuiInputChipComponent<T> {
...
+            this.handlers,
+        );
+
+        if (!items.length) {
             return;
         }
 
-        const value = this.value().map((item, index) =>
-            index === this.index ? this.internal() : item,
+        cva.setValue(
```
waterplea, line 145: Why not `onChange` like before?

splincode: `setValue` applies `unique()` dedupe and clears textfield value, both matter after splitting by separator, where onChange would leak duplicates

waterplea (approved):

mdlufy (approved):

PR #14349, opened 2026-06-16
refactor(core): `Calendar` add abstract class

waterplea:
Fixes #

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,9 @@
         "Vladimir Potekhin <vladimir.potekh@gmail.com>",
         "Nikita Barsukov <nikita.s.barsukov@gmail.com>"
     ],
-    "workspaces": ["projects/*"],
+    "workspaces": [
+        "projects/*"
+    ],
     "scripts": {
         "build:demo": "nx build demo",
         "build:demo:next": "nx build demo -c next",
@@ -57,25 +59,40 @@
         "typecheck": "tsc --noEmit --strict --incremental false --tsBuildInfoFile null --project tsconfig.spec.json"
     },
     "commitlint": {
-        "extends": ["@taiga-ui/commitlint-config"]
+        "extends": [
+            "@taiga-ui/commitlint-config"
+        ]
     },
     "lint-staged": {
-        "*.less": ["stylelint --fix"],
-        "*.{js,ts,html,svg,yml,md,less,json}": ["prettier --write", "cspell --no-must-find-files"]
+        "*.less": [
+            "stylelint --fix"
+        ],
+        "*.{js,ts,html,svg,yml,md,less,json}": [
+            "prettier --write",
+            "cspell --no-must-find-files"
+        ]
     },
-    "browserslist": ["extends @taiga-ui/browserslist-config"],
+    "browserslist": [
+        "extends @taiga-ui/browserslist-config"
+    ],
     "prettier": "@taiga-ui/prettier-config",
     "stylelint": {
-        "extends": ["@taiga-ui/stylelint-config"],
+        "extends": [
+            "@taiga-ui/stylelint-config"
+        ],
         "overrides": [
             {
-                "files": ["**/demo/**/*.{less,css,scss}"],
+                "files": [
+                    "**/demo/**/*.{less,css,scss}"
+                ],
                 "rules": {
                     "selector-class-pattern": null
                 }
             },
             {
-                "files": ["**/*.scss"],
+                "files": [
+                    "**/*.scss"
+                ],
                 "rules": {
                     "at-rule-allowed-list": null,
                     "at-rule-no-unknown": null,
@@ -156,7 +173,10 @@
         "yarn": "Please use npm instead of yarn to install dependencies"
     },
     "engineStrict": true,
-    "authors": ["Alex Inkin <alexander@inkin.ru>", "Roman Sedov <79601794011@ya.ru>"],
+    "authors": [
+        "Alex Inkin <alexander@inkin.ru>",
+        "Roman Sedov <79601794011@ya.ru>"
+    ],
     "standard-version": {
         "scripts": {
             "postbump": "ts-node ./scripts/postbump.ts",
--- a/package.json
+++ b/package.json
@@ -25,9 +25,7 @@
         "Vladimir Potekhin <vladimir.potekh@gmail.com>",
         "Nikita Barsukov <nikita.s.barsukov@gmail.com>"
     ],
-    "workspaces": [
-        "projects/*"
-    ],
+    "workspaces": ["projects/*"],
     "scripts": {
         "build:demo": "nx build demo",
         "build:demo:next": "nx build demo -c next",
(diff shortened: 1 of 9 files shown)


projects/core/components/calendar/calendar.ts
```diff
@@ -0,0 +1,33 @@
...
+export abstract class AbstractTuiCalendar<
+    T extends TuiDay | TuiDayRange | readonly TuiDay[] =
+        | TuiDay
+        | TuiDayRange
+        | readonly TuiDay[],
+> {
+    public readonly month = model(TuiMonth.currentLocal());
+    public readonly value = model<T | null>(null);
+
+    public readonly min = input(TUI_FIRST_DAY, {
```
splincode, line 19: This refactor changes `TuiCalendarRange` behavior.

Before this PR, `min` and `max` in `TuiCalendarRange` accepted and preserved `null`, and `disabledItemHandler` defaulted to `TUI_FALSE_HANDLER`.

After inheriting from `TuiAbstractCalendar`, `null` min and max values are transformed to `TUI_FIRST_DAY` and `TUI_LAST_DAY`, and the default disabled handler is taken from `TUI_ITEMS_HANDLERS`.

Is this intentional? If this PR is supposed to be a pure refactor, we probably need to preserve the old `TuiCalendarRange` semantics or add explicit tests for the changed behavior.

projects/kit/components/input-date-multi/input-date-multi.directive.ts
```diff
@@ -81,24 +81,22 @@ export class TuiInputDateMultiDirective extends TuiInputChipDirective<TuiDay> {
...
 
-    protected processCalendar(calendar: TuiCalendar): void {
+    protected processCalendar(calendar: AbstractTuiCalendar): void {
         tuiSetSignal(calendar.value, this.value());
         tuiSetSignal(calendar.min, this.min());
         tuiSetSignal(calendar.max, this.max());
-        calendar.month.set(
-            this.value()?.[this.value().length - 1] ?? TuiMonth.currentLocal(),
-        );
+        calendar.month.update((m) => this.value()?.[this.value().length - 1] ?? m);
```
splincode, line 99: This looks like a behavior change for `tuiInputDateMulti`.

Previously, when the value was empty, `processCalendar` reset the calendar month to `TuiMonth.currentLocal()`. Now it keeps the previous calendar month.

Is this intentional? If yes, I would add a test for this behavior. If not, we should keep the previous reset logic.

projects/kit/components/input-date-range/input-date-range.directive.ts
```diff
@@ -66,11 +67,15 @@ export class TuiInputDateRangeDirective extends TuiInputDateBase<TuiDayRange> {
     public readonly minLength = input<TuiDayLike | null>(null);
     public readonly maxLength = input<TuiDayLike | null>(null);
 
-    protected override processCalendar(calendar: TuiCalendarRange): void {
-        super.processCalendar(calendar);
+    protected override processCalendar(
```
splincode, line 70: Since `TuiCalendarRange` already extends `TuiAbstractCalendar`, the union and cast are redundant here.

We can simplify this to:

```ts
protected override processCalendar(calendar: TuiAbstractCalendar): void {
    super.processCalendar(calendar);

    if ('minLength' in calendar && 'maxLength' in calendar) {
        tuiSetSignal(calendar.minLength, this.minLength());
        tuiSetSignal(calendar.maxLength, this.maxLength());
    }
}
```

package.json
```diff
@@ -145,6 +145,7 @@
         "@schematics/angular": "19.2.27",
         "@stackblitz/sdk": "1.11.0",
         "@taiga-ui/configs": "0.540.0",
+        "@taiga-ui/event-plugins": "^5.0.0",
```
splincode: Most dependencies in this block are pinned to exact versions. Could we use `"@taiga-ui/event-plugins": "5.0.0"` instead of `"^5.0.0"` to keep installs reproducible?

projects/core/components/calendar/calendar.ts
```diff
@@ -0,0 +1,33 @@
...
+import {type TuiBooleanHandler} from '@taiga-ui/cdk/types';
+import {TUI_ITEMS_HANDLERS} from '@taiga-ui/core/directives/items-handlers';
+
+@Directive()
+export abstract class AbstractTuiCalendar<
+    T extends TuiDay | TuiDayRange | readonly TuiDay[] =
+        | TuiDay
+        | TuiDayRange
+        | readonly TuiDay[],
+> {
```
splincode: Maybe we can extract this union into a named type to make the generic easier to read:

```ts
type CalendarValue = TuiDay | TuiDayRange | readonly TuiDay[];

@Directive()
export abstract class TuiAbstractCalendar<
    T extends CalendarValue = CalendarValue,
> {
    public readonly value = model<T | null>(null);
}
```

projects/core/components/calendar/calendar.ts
```diff
@@ -0,0 +1,33 @@
+import {Directive, inject, input, model} from '@angular/core';
```
splincode, line 1: `projects/core/components/calendar/calendar.ts` But it does not contain the calendar component. It contains the abstract base class. I would prefer: `abstract-calendar.ts`

projects/kit/components/input-date-multi/input-date-multi.directive.ts
```diff
@@ -81,24 +81,22 @@ export class TuiInputDateMultiDirective extends TuiInputChipDirective<TuiDay> {
         }
     });
 
-    protected readonly calendarOut = effect((onCleanup) => {
-        const subscription = this.calendar()?.dayClick.subscribe((day) => {
-            this.updateValue(day);
-        });
+    protected readonly calendarOut = effect(() => {
+        const value = this.calendar()?.value();
```
splincode, line 85: It works, but semantically it is weaker. `value` is state, not an event. If someone sets calendar value programmatically to `TuiDay`, `input-date-multi` may treat it like a user click and toggle the value.

projects/kit/components/input-date-range/input-date-range.directive.ts
```diff
@@ -66,11 +67,15 @@ export class TuiInputDateRangeDirective extends TuiInputDateBase<TuiDayRange> {
...
+        calendar: AbstractTuiCalendar | TuiCalendarRange,
+    ): void {
+        super.processCalendar(calendar as AbstractTuiCalendar);
 
-        tuiSetSignal(calendar.minLength, this.minLength());
-        tuiSetSignal(calendar.maxLength, this.maxLength());
+        if ('minLength' in calendar && 'maxLength' in calendar) {
```
splincode, line 75: Even better, extract the guard:

```ts
function tuiIsCalendarRange(
    calendar: TuiAbstractCalendar,
): calendar is TuiCalendarRange {
    return 'minLength' in calendar && 'maxLength' in calendar;
}
```

```ts
protected override processCalendar(calendar: TuiAbstractCalendar): void {
    super.processCalendar(calendar);

    if (tuiIsCalendarRange(calendar)) {
        tuiSetSignal(calendar.minLength, this.minLength());
        tuiSetSignal(calendar.maxLength, this.maxLength());
    }
}
```

splincode (approved):

projects/kit/components/input-date-multi/input-date-multi.directive.ts
waterplea: It's intentional

projects/kit/components/input-date-range/input-date-range.directive.ts
waterplea: It's necessary because genertic is narrower

projects/kit/components/input-date-multi/input-date-multi.directive.ts
waterplea: It's only a day if it is clicked, otherwise it's an array

projects/kit/components/input-date-range/input-date-range.directive.ts
waterplea: Only needed once at the moment, no need to extract for now

vladimirpotekhin (approved):

package.json
```diff
@@ -145,6 +145,7 @@
         "@schematics/angular": "19.2.27",
         "@stackblitz/sdk": "1.11.0",
         "@taiga-ui/configs": "0.540.0",
+        "@taiga-ui/event-plugins": "5.0.0",
```
nsbarsukov, line 148: Is it really required to explicitly install it ?

It is already peer dependency and should be installed automatically 
https://github.com/taiga-family/taiga-ui/blob/7c558a1aa53eb00aa75a2e0f603e724108cb039a/projects/cdk/package.json#L56

nsbarsukov (approved):

package.json
waterplea: It is required in the topmost package.json for webtypes of our custom events to work

PR #14376, opened 2026-06-19
fix(addon-doc): correct tab order in `Example` after language switch

### LINKED ISSUE #14241: 🐞 - Names of the tabs in the documentation are confused
### Reproduction url

https://taiga-ui.dev/components/item-group#multiple-choice

### Description

Plays when selecting the Dutch language
https://taiga-ui.dev/components/item-group#multiple-choice

Image

### Taiga UI version

5.9.0

### Browser Used

Chrome

### OS Used

Windows

nullawhale:
Fixes #14241

### DIFF
--- a/projects/addon-doc/components/example/example.component.ts
+++ b/projects/addon-doc/components/example/example.component.ts
@@ -44,7 +44,12 @@ import {BehaviorSubject, map, switchMap} from 'rxjs';
 
 import {TuiDocCode} from '../code';
 import {TUI_DOC_EXAMPLE_OPTIONS} from './example.options';
-import {TuiDocExampleGetTabsPipe} from './example-get-tabs.pipe';
+
+interface TuiDocExampleTab {
+    readonly code: string;
+    readonly key: string;
+    readonly title: PolymorpheusContent;
+}
 
 @Component({
     selector: 'tui-doc-example',
@@ -55,7 +60,6 @@ import {TuiDocExampleGetTabsPipe} from './example-get-tabs.pipe';
         RouterLink,
         TuiButton,
         TuiDocCode,
-        TuiDocExampleGetTabsPipe,
         TuiFullscreen,
         TuiHeader,
         TuiLink,
@@ -107,6 +111,7 @@ export class TuiDocExample implements OnChanges {
         );
 
     protected readonly route = inject(ActivatedRoute);
+    protected readonly previewTab = 'tui-doc-example-preview';
     protected activeItemIndex = 0;
     protected fullscreen = false;
     protected readonly copy = computed(() => this.copyTexts()[0]);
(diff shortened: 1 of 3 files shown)


projects/addon-doc/components/example/example.component.ts
```diff
@@ -124,6 +129,25 @@ export class TuiDocExample implements OnChanges {
...
     );
 
+    protected readonly tabs = computed<readonly TuiDocExampleTab[]>(() => [
+        ...(this.preview()
+            ? [
+                  {
+                      code: '',
+                      key: 'tui-doc-example-preview',
+                      title: this.defaultTab(),
+                  },
```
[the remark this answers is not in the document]
nullawhale: `TUI_DOC_PREVIEW_TEXT` is `InjectionToken<Signal<string>>`, so `this.defaultTab()` is a signal.

splincode (approved):

waterplea (approved):

projects/addon-doc/components/example/example.template.html
```diff
@@ -31,17 +31,16 @@ <h3>
...
-    @if (processor() | tuiDocExampleGetTabs: defaultTab(); as all) {
-        @let tabs = preview() ? all : all.slice(1);
+    @if (tabs(); as tabs) {
         @if (tabs.length) {
             <div class="t-header">
                 <tui-segmented
                     class="t-tabs"
                     [(activeItemIndex)]="activeItemIndex"
                 >
-                    @for (tab of tabs; track tab) {
```
nsbarsukov, line 42: I think this is a one-line fix.

The `tuiDocExampleGetTabs` pipe returns `[defaultTab, ...fileNames]` and the template does track tab — so item 0's identity is a translated label. Languages load asynchronously, so it's 'Preview' on first render and 'Voorbeeld' a moment later. The key changes after the view exists, Angular rebuilds it, and the button lands out of order.

But the list is positional, and `activeItemIndex` already treats it that way. So:
```diff
- @for (tab of tabs; track tab) {
+ @for (tab of tabs; track $index) {
```

in **both** `@for` blocks. Item 0 stays the preview tab whatever the language calls it. 

No interface, no sentinel key, no orphaned `TuiDocExampleGetTabsPipe` (which this PR leaves unused but still exported from `addon-doc`). It also stops a language switch from destroying view 0 and re-instantiating the live example.

nsbarsukov (changes requested):

projects/addon-doc/components/example/example.template.html
nsbarsukov: @nullawhale 
Could you please leave a comment on this issue?
- https://github.com/taiga-family/taiga-ui/issues/14241

This is necessary so I can assign the issue to you.

nsbarsukov (approved):

PR #14386, opened 2026-06-22
ci: auto-merge only clean bot PRs with exactly one commit

splincode:


waterplea (approved):

.github/workflows/auto-merge.yml
```diff
@@ -46,14 +46,27 @@ jobs:
...
       - uses: taiga-family/actions/setup-variables@a11d89eb978fe572206a7b0a026b38ae957acba0 # v2.0.0
+      - id: lint_commit
+        name: Check linting bot commit
+        env:
+          TARGET_SUBJECT: 'chore: apply changes after linting [bot]'
+        run: |
+          echo "found=false" >> "$GITHUB_OUTPUT"
+          git log --format=%s "origin/${{ github.base_ref }}..HEAD" \
+            | grep -Fxq "$TARGET_SUBJECT" \
+            && echo "found=true" >> "$GITHUB_OUTPUT"
```
nsbarsukov: Could we simplify this by replacing the `git log | grep` step (and the `fetch-depth: 0`) with a plain single-commit check? Something like:

```yaml
if: >
  env.IS_TAIGA_FAMILY_BOT_PR_AUTHOR == 'true' &&
  github.event.pull_request.commits == 1
```

The idea: a clean bot PR is expected to be exactly one commit, so if a lint fixup (`chore: apply changes after linting [bot]`) gets pushed, the count goes above 1 and auto-merge is skipped — without having to match a magic subject line or fetch full history.

splincode: fixed

splincode: fixed

nsbarsukov (approved):

Issue #14234, opened 2026-06-05
🗂 - Refactor zoom handling and extend support to hints


vladimirpotekhin:
### Description

Currently, zoom level handling is implemented directly inside the dropdown component. This logic should be moved into a shared service to ensure consistent behavior across components and reduce duplication.

Additionally, zoom handling should be added to hint components so that their positioning and rendering remain correct when browser zoom is applied. To reproduce the issue on the demo, add the zoom / scale property to the body and check the hint positioning. (You can compare it with the dropdowns – they're all fine.)

PR #14397, opened 2026-06-22
feat(cdk): add `tuiGetZoom`

splincode:
Part of https://github.com/taiga-family/taiga-ui/issues/14234

### DIFF
--- a/projects/cdk/utils/dom/get-zoom.ts
+++ b/projects/cdk/utils/dom/get-zoom.ts
@@ -1,0 +1,3 @@
+export function tuiGetZoom(el: Element): number {
+    return el.currentCSSZoom || 1;
+}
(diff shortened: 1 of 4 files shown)


waterplea (approved):

projects/core/portals/hint/hint.component.ts
```diff
@@ -157,9 +157,11 @@ export class TuiHintComponent {
             rect.top + rect.height / 2 - top,
         ]);
 
+        const zoom = tuiGetZoom(this.hint.el);
+
         this.apply(
-            tuiPx(Math.round(top)),
-            tuiPx(Math.round(safeLeft)),
+            tuiPx(Math.round(top / zoom)),
+            tuiPx(Math.round(safeLeft / zoom)),
```
vladimirpotekhin, line 164: Something is wrong here. 

Image
Image

projects/cdk/utils/dom/get-zoom.ts
```diff
@@ -0,0 +1,3 @@
+export function tuiGetZoom(el: Element): number {
```
vladimirpotekhin, line 1: I don't think it's a good idea to add a new API right now. We only have two uses in our codebase.

PR #14401, opened 2026-06-22
feat(core): `Root` apply inert to content under modals

### LINKED ISSUE #12635: 🚀 - Refactor `FocusTrap` to use `inert`
### Description

Due to [incompetence of Webkit devs](https://bugs.webkit.org/show_bug.cgi?id=303022) we might have to refactor FocusTrap to use inert attribute before official bump of browser support to accommodate it and consider it a graceful degradation on older browsers.

waterplea:
Fixes #12635

I'm intentionally leaving out nested modals case because I don't want to overcomplicate the attribute logic – we already have CSS inertness properly configured and attribute is just a fallback for Safari and Firefox, who do not support CSS `interactivity` yet.

### DIFF
--- a/projects/core/portals/modal/modal.service.ts
+++ b/projects/core/portals/modal/modal.service.ts
@@ -21,6 +21,7 @@ export abstract class TuiModalService<T, K = void> extends TuiPortal<T, K> {
         const ref = this.service.add(component);
         const el: HTMLElement = ref.location.nativeElement;
 
+        el.closest('tui-root')?.firstElementChild?.setAttribute('inert', '');
         ref.instance.component.set(new PolymorpheusComponent(this.content));
 
         return () => {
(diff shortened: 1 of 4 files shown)


projects/core/portals/modal/modal.service.ts
```diff
@@ -21,6 +21,7 @@ export abstract class TuiModalService<T, K = void> extends TuiPortal<T, K> {
         const ref = this.service.add(component);
         const el: HTMLElement = ref.location.nativeElement;
 
+        el.closest('tui-root')?.firstElementChild?.setAttribute('inert', '');
```
splincode, line 24: What do you think about the local `getTuiRoot` function, or is it redundant?

```ts
export abstract class TuiModalService<T, K = void> extends TuiPortal<T, K> {
    // ...

    protected override add(
        component: PolymorpheusComponent<TuiModalComponent<T>>,
    ): () => void {
        // ...

        getTuiRoot(el)?.setAttribute('inert', '');
        // ...

        return () => {
            // ...

            Promise.allSettled(getAnimations(el))
                .then(async () => Promise.allSettled(getAnimations(el.firstElementChild)))
                .then(() => {
                    getTuiRoot(el)?.removeAttribute('inert');
                    // ...
                });
        };
    }
}

function getTuiRoot(el: Element | null): Element | null {
 return el.closest('tui-root')?.firstElementChild ?? null;
}

function getAnimations(el: Element | null): ReadonlyArray<Promise<unknown>> {
// ...
```

waterplea: I think we can keep that inline

splincode (approved):

nsbarsukov (approved):

PR #14404, opened 2026-06-23
chore(schematics): migrate tuiGetClosestFocusable

splincode:


projects/cdk/schematics/ng-update/v5/index.ts
```diff
@@ -124,6 +125,10 @@ function main(options: TuiSchema, timings: MigrationStepTiming[]): Rule {
                     name: 'migrateI18nLanguageSignal',
                     step: () => migrateI18nLanguageSignal(tree, options),
                 },
+                {
+                    name: 'migrateGetClosestFocusable',
+                    step: () => migrateGetClosestFocusable(),
```
nsbarsukov: Looks not performance-friendly to create separate iteration only for a argument of single function

Should we re-use/extends logic of `FUNCTION_PARAMETERS_TO_REPLACE` ?

https://github.com/taiga-family/taiga-ui/blob/553d5441c8714724e66ca51bf5e55823e2268c7e/projects/cdk/schematics/ng-update/v5/steps/constants/function-parameters-to-replace.ts#L3

nsbarsukov: Something like

```ts
{
        names: ['tuiGetClosestFocusable'],
        parameters: [{name: 'keyboard', remove: true}],
    },
```

nsbarsukov (approved):

vladimirpotekhin (approved):

PR #14408, opened 2026-06-23
fix(kit): flip chevron on native select interaction

### LINKED ISSUE #12186: 🚀 - If the native select is clicked then the arrow icon should flip
### Description

Image

splincode:
Fixes https://github.com/taiga-family/taiga-ui/issues/12186

### Before

https://github.com/user-attachments/assets/1bfd8947-0611-4a03-85ee-8472620e0c77

### After

https://github.com/user-attachments/assets/6f26a190-8e43-455e-9d92-b6608ce46f7c

### DIFF
--- a/projects/kit/directives/chevron/chevron.directive.ts
+++ b/projects/kit/directives/chevron/chevron.directive.ts
@@ -6,6 +6,7 @@ import {
     inject,
     InjectionToken,
     input,
+    signal,
     ViewEncapsulation,
 } from '@angular/core';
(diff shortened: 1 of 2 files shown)


waterplea:
Let's use this instead and consider it graceful degradation:
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:open

splincode:
@waterplea updated

waterplea (approved):

vladimirpotekhin (approved):

PR #14413, opened 2026-06-23
chore(experimental): `DatePicker` add week indicator

waterplea:
Fixes #


vladimirpotekhin (approved):

projects/experimental/components/calendar/calendar.options.ts
```diff
@@ -0,0 +1,22 @@
+import {type Signal, signal} from '@angular/core';
+import {type TuiDay, TuiDayOfWeek} from '@taiga-ui/cdk/date-time';
+import {type TuiHandler} from '@taiga-ui/cdk/types';
+import {tuiCreateOptions} from '@taiga-ui/cdk/utils/di';
+
+export interface TuiCalendarOptions {
+    readonly dayType: TuiHandler<TuiDay, string>;
+    readonly weekStart: Signal<0 | 1 | 2 | 3 | 4 | 5 | 6>;
```
nsbarsukov: Is just `Signal<number>` better ? 🤔 

```ts
tuiCalendarOptionsProvider({
    weekStart: new Intl.Locale('ar-AF').getWeekInfo().firstDay, // 6
});
```

nsbarsukov (approved):

projects/experimental/components/calendar/calendar.options.ts
nsbarsukov: And maybe rename to `weekFirstDay`?
To be similar  to native naming https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getWeekInfo#firstday

nsbarsukov: Same comment for `dayType` => `weekend`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getWeekInfo#weekend

**UPDATE:** already existed `dayType` is more flexible https://taiga-ui.dev/components/calendar#color-customization ; let's keep it as it

PR #14432, opened 2026-06-25
fix: improve tree shaking of `TuiDay` class for unused `InputDate*` components

nsbarsukov:
## Problem
Create empty angular application.

Create `src/app/provide-proprietary.ts`:
```ts
export function provideProprietary(): Array<EnvironmentProviders | Provider> {
    return [
        // [...]
        tuiInputDateOptionsProvider({icon: '@custom-icon-pack.calendar'}),
        // [...]
    ]
}
```

Add it to `src/app/app.config.ts`:
```ts
import {provideProprietary} from './provide-proprietary';

export const appConfig: ApplicationConfig = {
  providers: [
    // [...]
    provideProprietary(),
  ]
};
```
previous

### New behavior
new

### DIFF
--- a/projects/kit/components/input-date-range/input-date-range.directive.ts
+++ b/projects/kit/components/input-date-range/input-date-range.directive.ts
@@ -37,12 +37,12 @@ import {TUI_INPUT_DATE_RANGE_OPTIONS} from './input-date-range.options';
     hostDirectives: [TuiWithInput, TuiDropdownAuto, MaskitoDirective],
 })
 export class TuiInputDateRangeDirective extends TuiInputDateBase<TuiDayRange> {
-    public override readonly max = input(this.options.max, {
+    public override readonly max = input(this.options.max ?? TUI_LAST_DAY, {
         transform: (max: TuiDay | null): TuiDay =>
             max instanceof TuiDay ? max : TUI_LAST_DAY,
     });
 
-    public override readonly min = input(this.options.min, {
+    public override readonly min = input(this.options.min ?? TUI_FIRST_DAY, {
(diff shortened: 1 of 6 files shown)


projects/kit/components/input-date-range/input-date-range.directive.ts
```diff
@@ -37,12 +37,12 @@ import {TUI_INPUT_DATE_RANGE_OPTIONS} from './input-date-range.options';
     hostDirectives: [TuiWithInput, TuiDropdownAuto, MaskitoDirective],
 })
 export class TuiInputDateRangeDirective extends TuiInputDateBase<TuiDayRange> {
-    public override readonly max = input(this.options.max, {
+    public override readonly max = input(this.options.max ?? TUI_LAST_DAY, {
         transform: (max: TuiDay | null): TuiDay =>
             max instanceof TuiDay ? max : TUI_LAST_DAY,
```
waterplea: Can it be just `max ?? TUI_LAST_DAY` here and everywhere? Is there a meaningful reason for instanceof?

waterplea: Can it be just `max ?? TUI_LAST_DAY` here and everywhere? Is there a meaningful reason for instanceof?

waterplea (approved):

projects/kit/components/input-date-range/input-date-range.directive.ts
nsbarsukov: Good point! Fixed

nsbarsukov: Fixed

splincode (approved):

PR #14437, opened 2026-06-25
chore(experimental): `DatePicker` add cell content customization

waterplea:
Fixes #

### DIFF
--- a/projects/demo/src/pages/components/date-picker/examples/2/index.ts
+++ b/projects/demo/src/pages/components/date-picker/examples/2/index.ts
@@ -1,0 +1,42 @@
+import {Component, signal} from '@angular/core';
+import {FormsModule} from '@angular/forms';
+import {changeDetection} from '@demo/emulate/change-detection';
+import {encapsulation} from '@demo/emulate/encapsulation';
+import {TuiDay} from '@taiga-ui/cdk';
+import {TuiTitle} from '@taiga-ui/core';
+import {tuiCalendarOptionsProvider, TuiDatePicker} from '@taiga-ui/experimental';
+import {TuiInputDate} from '@taiga-ui/kit';
+import {DatePipe} from '@angular/common';
+
+@Component({
+    imports: [FormsModule, TuiDatePicker, TuiInputDate, TuiTitle, DatePipe],
+    templateUrl: './index.html',
+    encapsulation,
+    changeDetection,
+    providers: [tuiCalendarOptionsProvider({showWeek: false})],
+})
+export default class Example {
+    protected readonly value = signal<TuiDay | null>(new TuiDay(2026, 5, 25));
+
+    protected getLabel({day}: TuiDay): string {
+        switch (day) {
+            case 17:
+            case 18:
+            case 19:
+                return 'Sick';
+            case 20:
+                return '••';
+            case 25:
+                return '•••';
+            case 3:
+                return '••';
+            case 30:
+                return 'Q2';
+            case 4:
+            case 5:
+                return '•';
+            default:
+                return '';
+        }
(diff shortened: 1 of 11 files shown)


projects/core/components/carousel/carousel.component.html
```diff
@@ -10,10 +10,12 @@
...
     </span>
 }
-<span class="t-item">
-    <ng-container *ngTemplateOutlet="template(); context: {$implicit: math.min(index() + 1, max())}" />
-    <span
-        class="t-bumper"
-        (waIntersectionObservee)="onIntersection(!!$event[0]?.isIntersecting, 1)"
-    ></span>
-</span>
+@if (max() !== min()) {
```
splincode: This looks reasonable, but it changes the core carousel behavior globally. Is this required for the DatePicker change, or should it be split into a separate PR with test case for the single-item carousel scenario?

projects/demo/src/pages/components/date-picker/examples/2/index.ts
```diff
@@ -0,0 +1,42 @@
...
+
+@Component({
+    imports: [DatePipe, FormsModule, TuiDatePicker, TuiInputDate, TuiTitle],
+    templateUrl: './index.html',
+    encapsulation,
+    changeDetection,
+    providers: [tuiCalendarOptionsProvider({showWeek: false})],
+})
+export default class Example {
+    protected readonly value = signal<TuiDay | null>(new TuiDay(2026, 5, 25));
```
splincode, line 19: Maybe we can keep this example signal-based, but use `model()` and bind it to `ngModel` explicitly:

```ts
protected readonly value = model<TuiDay | null>(new TuiDay(2026, 5, 25));
```

projects/core/components/carousel/carousel.component.html
waterplea: Limits on months can cause only one month to be allowed and carousel did not support this.

projects/demo/src/pages/components/date-picker/examples/2/index.ts
waterplea: model is an input + output, there's no reason to use model instead of regular signal because we do not need an input on Example class.

projects/experimental/components/date-picker/date-picker.component.ts
```diff
@@ -126,9 +128,7 @@ export class TuiDatePicker<
     );
 
     protected readonly disabledMonth = computed(
-        () => (month: number) =>
-            this.month().year * 12 + month < this.min().year * 12 + this.min().month ||
-            this.month().year * 12 + month > this.max().year * 12 + this.max().month,
+        () => (month: TuiMonth) => month < this.min() || month > this.max(),
     );
```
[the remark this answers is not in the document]
waterplea: I'll fix that in the next PR

vladimirpotekhin (approved):

projects/experimental/components/date-picker/date-picker.component.ts
```diff
@@ -160,6 +160,9 @@ export class TuiDatePicker<
 
     public readonly view = model<'day' | 'month' | 'year'>('day');
     public readonly mode = input<T>();
+    public readonly contentDay = input<PolymorpheusContent<TuiContext<TuiDay>>>();
+    public readonly contentMonth = input<PolymorpheusContent<TuiContext<TuiMonth>>>();
+    public readonly contentYear = input<PolymorpheusContent<TuiContext<number>>>();
```
nsbarsukov, line 165: Should this new input-property be mentioned in new example too ?

nsbarsukov (approved):

Issue #13950, opened 2026-04-23
🐞 - Invalid migration on tui-editor repo


splincode:
- [x] TuiDropdownService

I have 

```ts
import {TuiDropdownService} from '@taiga-ui/core';
```

```
✘ [ERROR] TS2724: '"@taiga-ui/core"' has no exported member named 'TuiDropdownService'. Did you mean 'TuiDropdownDriver'? [plugin angular-compiler]

    projects/editor/components/editor/portal/editor-portal.directive.ts:2:23:
      2 │ import {tuiAsViewport, TuiDropdownService, TuiRectAccessor} from '@...
        ╵                        ~~~~~~~~~~~~~~~~~~
```

It should be 

```ts
import {TuiPopupService} from '@taiga-ui/core';
```

- [x] The legacy package does not have addon-mobile in peer deps

```

Application bundle generation complete. [0.625 seconds]

✘ [ERROR] Could not resolve "@taiga-ui/addon-mobile/directives/ripple"
    node_modules/@taiga-ui/legacy/fesm2022/taiga-ui-legacy-components-mobile-dialog.mjs:4:26:
      4 │ import { TuiRipple } from '@taiga-ui/addon-mobile/directives/ripple';
        ╵                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  You can mark the path "@taiga-ui/addon-mobile/directives/ripple" as external to exclude it from the bundle, which will remove this error and leave the unresolved path in the bundle.
/Users/m.ivanov4/WebstormProjects/editor/node_modules/esbuild/lib/main.js:1463
  let error = new Error(text);
              ^
Error: Build failed with 1 error:
node_modules/@taiga-ui/legacy/fesm2022/taiga-ui-legacy-components-mobile-dialog.mjs:4:26: ERROR: Could not resolve "@taiga-ui/addon-mobile/directives/ripple"
    at failureErrorWithLog (/Users/m.ivanov4/WebstormProjects/editor/node_modules/esbuild/lib/main.js:1463:15)
    at /Users/m.ivanov4/WebstormProjects/editor/node_modules/esbuild/lib/main.js:924:25
    at /Users/m.ivanov4/WebstormProjects/editor/node_modules/esbuild/lib/main.js:1341:9
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5) {
  errors: [Getter/Setter],
  warnings: [Getter/Setter]
}
[...]

PR #14452, opened 2026-06-26
chore(schematics): valid migration for tui-accordion

splincode:
Part of #13950


projects/cdk/schematics/ng-update/v5/steps/constants/inputs-to-remove.ts
```diff
@@ -21,4 +21,8 @@ export const INPUTS_TO_REMOVE: RemovableInput[] = [
         inputName: 'tuiDirectionOrder',
         tags: ['table'],
     },
+    {
+        inputName: 'rounded',
+        tags: ['tui-accordion'],
```
nsbarsukov, line 26: Should we also add a comment indicating that the previous `round` behavior can still be achieved using CSS, rather than removing this input prop silently?

And link to this example
https://taiga-ui.dev/components/accordion#custom

mdlufy (approved):

nsbarsukov (approved):

PR #14469, opened 2026-06-30
fix(core): fix `tuiSelectLike` on ios safari

vladimirpotekhin:
On every engine except WebKit, execCommand ignores that cancellation and inserts the text anyway — so it worked. But iOS Safari honors preventDefault() on the beforeinput triggered by execCommand, so the programmatic insert got cancelled and the field stayed empty. 

What's the fix? 🛠️

Skip the preventDefault() guard on mobile (WA_IS_MOBILE):

This is safe because on mobile inputmode="none" already prevents the soft keyboard from appearing — the user can't type manually, so the only source of insertText is our own execCommand. The blocking is still fully in place on desktop, where hardware keyboards make it necessary.

### DIFF
--- a/projects/core/components/textfield/select-like.directive.ts
+++ b/projects/core/components/textfield/select-like.directive.ts
@@ -1,5 +1,5 @@
 import {Directive, inject} from '@angular/core';
-import {WA_IS_ANDROID} from '@ng-web-apis/platform';
+import {WA_IS_ANDROID, WA_IS_MOBILE} from '@ng-web-apis/platform';
 import {tuiInjectElement} from '@taiga-ui/cdk/utils/dom';
 
 import {TUI_TEXTFIELD_OPTIONS} from './textfield.options';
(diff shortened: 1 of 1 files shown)


waterplea (approved):

projects/core/components/textfield/select-like.directive.ts
```diff
@@ -13,7 +13,7 @@ import {TUI_TEXTFIELD_OPTIONS} from './textfield.options';
         tuiSelectLike: '',
         // Click on cleaner icon does not trigger `beforeinput` event --> handle all kind of deletion in input event
         '(beforeinput)':
-            'options.cleaner() && $event.inputType.includes("delete") || $event.preventDefault()',
+            'isMobile || options.cleaner() && $event.inputType.includes("delete") || $event.preventDefault()',
```
nsbarsukov: Virtual keyboard is not the only way to modify value in mobile :(

https://github.com/user-attachments/assets/ab6d62e5-a8c0-43d4-929a-4521af4e96b3

nsbarsukov (approved):

projects/core/components/textfield/select-like.directive.ts
waterplea: Maybe we can consider this an acceptable limitation?

PR #14472, opened 2026-07-01
fix(cdk): `TuiPdfViewer` correct migration

mdlufy:
Fixes #


projects/cdk/schematics/ng-update/v5/steps/constants/identifiers-to-replace.ts
```diff
@@ -169,11 +169,21 @@ export const IDENTIFIERS_TO_REPLACE: ReplacementIdentifierMulti[] = [
     },
     {
         from: {
```
splincode, line 171:
```ts
from: [{
            name: 'TuiPdfViewer',
            moduleSpecifier: '@taiga-ui/kit',
        },
        {
            name: 'TuiPdfViewerComponent',
            moduleSpecifier: '@taiga-ui/kit'
        }
],
```

splincode (approved):

nsbarsukov (approved):

PR #14490, opened 2026-07-02
chore(schematics): preserve control-attr casing in input migrations via `getOriginalAttrText`

mdlufy:
Problem

The v5 template migrations rebuilt moved attributes from parse5's **lowercased**
`attr.name`, guarded by hand-maintained `normalizeAttrName` allowlists. Any
camelCase attribute missing from an allowlist was silently lowercased:

```html
<tui-input-time [formControlName]="c" (ngModelChange)="f($event)">
<!-- ng update -> -->
<input tuiInputTime [formcontrolname]="c" (ngmodelchange)="f($event)" />  <!-- ❌ -->
```

This breaks reactive-form binding and change handlers after `ng update`.

Fix 1 — preserve casing via a shared `getOriginalAttrText`

New helper that slices the attribute straight from the **original template
source** (via parse5's recorded offsets), preserving name casing, quotes and
spacing for *any* attribute — no allowlist to keep in sync. It falls back to
reconstructing `name="value"` only when source location is unavailable, so the
value is never dropped.

Applied across the input-family migrations, removing the per-file
`normalizeAttrName` allowlists and **three** duplicate local copies of the helper:

- **Casing preserved for control bindings:** `input-time`, `-date`,
  `-date-range`, `-date-multi`, `-month`, `-phone`, `-phone-international`,
  `-year`, `-tag`, `-number`.
- **Consolidated onto the shared helper** (behavior-preserving; their allowlists
  already covered the attrs they collected): `combo-box`, `input-date-time`,
  `-month-range`, `-slider`, `multi-select`, `select`.
- **De-duplicated** (dropped local `getOriginalAttrText` copies + dead
  `normalizeAttrName` fallbacks): `input`, `textarea`, `hint-on-legacy-controls`.

Intentional renames are preserved: affix maps in `input-tag`/`input-number`,
`[strictMatcher] -> [matcher]` in `combo-box`, and `slider`'s dedicated
`normalizeSliderAttrName`.

Fix 2 — move control bindings onto `<input>` in combo-box / multi-select / select

While unifying the above, found that these three collected control attributes via
an explicit `CONTROL_ATTRS` allowlist that omitted `[formControlName]` and
`(ngModelChange)`. Those were left on the `<tui-textfield>` wrapper instead of the
generated, control-bearing `<input>`, breaking the binding after migration:

```html
<!-- before (bug) -->
<tui-textfield [formControlName]="c" (ngModelChange)="f($event)">
  <input tuiComboBox />

<!-- after (fix) -->
<tui-textfield>
  <input tuiComboBox [formControlName]="c" (ngModelChange)="f($event)" />
```

Switched their collection to the same `/formcontrol|ngmodel/` regex the other
input migrations already use. This matches the v5 API, where the
`ControlValueAccessor` directive (`tuiComboBox` / `tuiSelect` / `tuiInputChip`)
lives on the `<input>` inside `<tui-textfield>`.


projects/cdk/schematics/ng-update/v5/tests/__snapshots__/schematic-migrate-combo-box.spec.ts.snap
```diff
@@ -326,6 +326,37 @@ exports[`ng-update ComboBox moves [(ngModel)] to generated input when input is a
 }
 `;
 
+exports[`ng-update ComboBox moves [formControlName] and (ngModelChange) to the generated input, preserving camelCase: test.html 1`] = `
+{
+  "0. Before": "
+                <tui-combo-box
+                    [formControlName]="control"
+                    (ngModelChange)="onChange($event)"
```
nsbarsukov, line 334: 😄

nsbarsukov (approved):

vladimirpotekhin (approved):

PR #14494, opened 2026-07-03
refactor: drop `ResizeObserver` directive in favor of event plugin

### LINKED ISSUE #8070: 🚀 - Use `(resize)` event plugin instead of observer
### Description

Then we can drop imports and drop the dependency, possibly.

waterplea:
Fixes #8070

### DIFF
--- a/projects/kit/components/preview/preview.component.ts
+++ b/projects/kit/components/preview/preview.component.ts
@@ -7,7 +7,6 @@ import {
     input,
 } from '@angular/core';
 import {WaMutationObserver} from '@ng-web-apis/mutation-observer';
-import {WaResizeObserver} from '@ng-web-apis/resize-observer';
 import {TUI_FALSE_HANDLER} from '@taiga-ui/cdk/constants';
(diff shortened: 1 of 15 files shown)


projects/kit/components/preview/preview.component.ts
```diff
@@ -101,23 +96,14 @@ export class TuiPreviewComponent {
...
         }
     }
 
-    protected onResize([entry]: readonly ResizeObserverEntry[]): void {
-        if (entry?.contentRect) {
-            this.refresh(entry.contentRect.width, entry.contentRect.height);
-            this.cdr.detectChanges();
-        }
+    protected onResize({clientWidth, clientHeight}: HTMLElement): void {
+        this.refresh(clientWidth, clientHeight);
```
splincode: Is it possible to use `refresh` directly? Or rename the `refresh` method to `onResize`?

splincode (approved):

nsbarsukov (approved):

Issue #13823, opened 2026-04-10
🐞 - `v5` migration issues


hakimio:
### Reproduction url

stackblitz.com

### Description

Some migration issues I have noticed:

- Error parsing `tsconfig.json`: `Unexpected token '/', "/* To lear"... is not valid JSON.` `tsconfig` is `JSONC`. Comments are valid in `JSONC` and the comment was added by the Angular CLI script:
```jsonc
/* To learn more about this file see: https://angular.io/config/tsconfig. */
{
    "compileOnSave": false,
// ...
```

- `WARNING: Formatting of files failed with the following error: spawn ENAMETOOLONG` Not sure if it's because of Windows or because I'm not using prettier.

- `size` attribute doesn't seem to work with the new `tuiAccordion` button
<img width="360" height="167" alt="Image" src="https://github.com/user-attachments/assets/ebe5a00b-96b0-43ee-a12b-9056815aaa56" />

- Dialog `header` option is not migrated to anything. What's the `v5` alternative? Declaration in template?

- `12 | tuiRepeatTimes` is not equivalent to `'-'.repeat(12)`. The pipe produces numbers, the `repeat()` produces strings.

<img width="1127" height="507" alt="Image" src="https://github.com/user-attachments/assets/f7bc4652-4e11-4d53-b798-35833b127d5a" />

- `tuiFilterByInput` migration fails when the item matcher is a signal.

<img width="1893" height="663" alt="Image" src="https://github.com/user-attachments/assets/9413a5ef-8f15-4a18-837a-a1bd31d93dce" />

<img width="2389" height="812" alt="Image" src="https://github.com/user-attachments/assets/1e25a9d4-622c-4ad7-90d4-af4c387b457a" />

<img width="1779" height="271" alt="Image" src="https://github.com/user-attachments/assets/b7964652-7a1f-4571-b5a1-51d79f291124" />

- ~`tuiDirectionOrder` is not migrated~ #13839

- ~`TuiPdfViewerService` is not migrated~ #13840

- ~`TuiIdService` is not migrated~ #13841

- `tui-avatar` migration results in an error when `src` is `SafeResourceUrl`

- ~`TuiOptionNew` is not migrated~ #13842

- Application build **crashes** with `Could not resolve "@taiga-ui/addon-mobile/directives/ripple"` error when `@taiga-ui/legacy` imports are used

### Taiga UI version

5.1.0

### Browser Used

Chrome

### OS Used

Windows

[...]

PR #14497, opened 2026-07-06
fix(cdk): flag removed dialog `header` option during v5 migration

mdlufy:

Part of #13823 (item: *"Dialog `header` option is not migrated to anything"*).

### What is the problem?

In v4 `TuiDialogOptions` had two distinct fields:

- `label: string` — the dialog title;
- `header: PolymorpheusContent` — content rendered **above** the title.

In v5 `header` was removed entirely (it is **not** an alias of `label`). Any
`header` left in dialog options silently produces a type error after `ng update`,
with no guidance.

### What is the solution?

There is no faithful automatic transform: `header` was positioned *above* the
title, so mapping it to `label` would relocate content, and rich
`PolymorpheusContent` cannot be expressed as a `string` label at all. The v5
pattern (per maintainer guidance) is manual: pass an empty `label` and place
`tuiHeader` in the dialog template.

So this migration **flags** each `header` usage with a `TODO` comment pointing to
that pattern and the docs, instead of guessing:

```ts
tuiDialog('content', {
    // TODO: (Taiga UI migration) dialog `header` option (content shown above the title)
    // was removed from `TuiDialogOptions` in v5. To keep content above the title, pass an
    // empty `label` and place `tuiHeader` in the dialog template where you need it. For a
    // plain string title, use `label` instead. See: https://taiga-ui.dev/components/dialog
    header: 'Title',
    size: 'm',
})
```

Detection is scoped to dialog options objects only — `tuiDialog(...)`,
`TuiDialogService.open(...)` and `tuiDialogOptionsProvider(...)` — mirroring
`migrateDialogLegacySizes`, so unrelated `header` keys are never touched.

### DIFF
--- a/projects/cdk/schematics/ng-update/v5/steps/utils/get-dialog-options.ts
+++ b/projects/cdk/schematics/ng-update/v5/steps/utils/get-dialog-options.ts
@@ -17,6 +17,7 @@ export function getDialogOptions(
 ): ObjectLiteralExpression | undefined {
     const isFactory = call.getExpression().getText() === TUI_DIALOG_FACTORY;
     const isProvider = call.getExpression().getText() === TUI_DIALOG_OPTIONS_PROVIDER;
+
     const isMethod = isServiceMethodCall(
         call,
         TUI_DIALOG_SERVICE,
--- a/projects/cdk/schematics/ng-update/v5/steps/utils/get-dialog-options.ts
+++ b/projects/cdk/schematics/ng-update/v5/steps/utils/get-dialog-options.ts
@@ -2,10 +2,13 @@ import {type CallExpression, Node, type ObjectLiteralExpression} from 'ng-morph'
 
 import {isServiceMethodCall} from '../../../../utils/is-service-method-call';
 
+// Exported: reused by migrate-dialog-legacy-sizes to move these imports to legacy.
 export const TUI_DIALOG_FACTORY = 'tuiDialog';
 export const TUI_DIALOG_SERVICE = 'TuiDialogService';
-export const TUI_DIALOG_OPEN_METHOD = 'open';
 export const TUI_DIALOG_OPTIONS_PROVIDER = 'tuiDialogOptionsProvider';
+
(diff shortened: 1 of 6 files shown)


projects/cdk/schematics/ng-update/v5/steps/utils/get-dialog-options.ts
```diff
@@ -0,0 +1,34 @@
+import {type CallExpression, Node, type ObjectLiteralExpression} from 'ng-morph';
+
+import {isServiceMethodCall} from '../../../../utils/is-service-method-call';
+
+export const TUI_DIALOG_FACTORY = 'tuiDialog';
+export const TUI_DIALOG_SERVICE = 'TuiDialogService';
+export const TUI_DIALOG_OPEN_METHOD = 'open';
+export const TUI_DIALOG_OPTIONS_PROVIDER = 'tuiDialogOptionsProvider';
```
splincode, line 8: Should they be exported?

mdlufy: updated

projects/cdk/schematics/ng-update/v5/tests/schematic-migrate-dialog-header.spec.ts
```diff
@@ -95,5 +95,22 @@ describe('ng-update tuiDialog header option', () => {
         expect(result).not.toContain('TODO: (Taiga UI migration)');
     });
 
+    it('adds a TODO for every dialog call with `header` in the same file', async () => {
```
splincode: `schematic-migrate-dialog-header.spec.ts` 

Can you rewrite the file so that there is no expect/toContain, but only a snapshots?

mdlufy: Done

nsbarsukov (approved):

splincode (approved):

PR #14498, opened 2026-07-06
fix(cdk): flag SafeResourceUrl avatar `src` during v5 migration

mdlufy:

Part of #13823 (item: *"`tui-avatar` migration results in an error when `src` is `SafeResourceUrl`"*).

### What is the problem?

`<tui-avatar [src]="x">` migrates to `<span [tuiAvatar]="x">`. In v5 `tuiAvatar`
is proxied to the `iconStart` input, which only accepts a **string** (icon name or
URL). When the bound value is a `SafeResourceUrl` (or any non-string image source)
the migrated template breaks at build time — and the type cannot be inferred from
the template alone.

### What is the solution?

Per maintainer guidance in the issue (drop a comment for bindings, pointing at the
`img`-inside pattern), insert a `TODO` on **raw dynamic** `[src]`/`[(src)]`
bindings:

```html
<!-- TODO: (Taiga UI migration) tuiAvatar accepts only a string (icon name or URL).
     If this value is a SafeResourceUrl or another non-string image source, render it
     via an inner <img> instead: <span tuiAvatar><img [src]="..." alt="" /></span>. -->
<span [tuiAvatar]="avatarUrl"></span>
```

Known-string flows are intentionally skipped, so the comment only appears where the
type is genuinely ambiguous:

- `| tuiIcon` bindings (icon string) — skipped;
- `| tuiFallbackSrc` bindings (already unwrapped into an inner ``) — skipped;
- static string `src="..."` — skipped.

Since the bound type is invisible to the migration, the comment is advisory rather
than a code transform.

### DIFF
--- a/projects/cdk/schematics/ng-update/v5/tests/schematic-migrate-avatar-to-directive.spec.ts
+++ b/projects/cdk/schematics/ng-update/v5/tests/schematic-migrate-avatar-to-directive.spec.ts
@@ -14,8 +14,6 @@ describe('ng-update avatar', () => {
         migrate({template: '<tui-avatar src="tuiIconUser"></tui-avatar>'}),
     );
 
-    // Also guards issue #13823: a raw dynamic [src] binding gets the SafeResourceUrl TODO —
-    // the snapshot's "After" shows the inserted comment above the migrated element.
     it(
         'replaces bound src attribute with bound tuiAvatar',
         migrate({template: '<tui-avatar [src]="avatarUrl" size="m"></tui-avatar>'}),
@@ -91,8 +89,6 @@ describe('ng-update avatar', () => {
         }),
     );
 
-    // Guards Gemini's fix: a bound string literal is not a SafeResourceUrl, so it is migrated
-    // to [tuiAvatar] without a TODO (the snapshot's "After" has no inserted comment).
     it(
         'keeps a bound string literal [src] without a SafeResourceUrl TODO',
         migrate({template: '<tui-avatar [src]="\'assets/avatar.png\'"></tui-avatar>'}),
--- a/projects/cdk/schematics/ng-update/v5/tests/schematic-migrate-avatar-to-directive.spec.ts
+++ b/projects/cdk/schematics/ng-update/v5/tests/schematic-migrate-avatar-to-directive.spec.ts
@@ -2,7 +2,7 @@ import {join} from 'node:path';
 
 import {resetActiveProject} from 'ng-morph';
 
-import {createMigration, runMigration} from '../../../utils/run-migration';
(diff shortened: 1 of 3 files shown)


projects/cdk/schematics/ng-update/v5/tests/schematic-migrate-avatar-to-directive.spec.ts
```diff
@@ -89,5 +89,34 @@ describe('ng-update avatar', () => {
         }),
     );
 
+    it('adds a SafeResourceUrl TODO for a raw dynamic [src] binding (issue #13823)', async () => {
+        const {template} = await runMigration({
+            collection: COLLECTION,
+            template: '<tui-avatar [src]="safeUrl"></tui-avatar>',
+        });
+
+        expect(template).toContain('[tuiAvatar]="safeUrl"');
```
splincode: Why did you choose this approach instead of snapshots?

mdlufy: The behavior is covered by snapshots too — the flagged case is in the updated .snap. I added these explicit assertions on top because a snapshot can't express "this must fail on the bug": toMatchSnapshot just baselines whatever output it sees, so it would happily record the missing/incorrect TODO and pass. The assertions state the actual intent (TODO present for ambiguous bindings, absent for icon/fallback/literal/static), so they fail on regression and read as a spec. Same non-snapshot style as schematic-migrate-dialog-legacy-sizes.spec.ts. Happy to drop to pure snapshots if you'd prefer consistency

mdlufy: Updated

projects/cdk/schematics/ng-update/v5/tests/schematic-migrate-avatar-to-directive.spec.ts
```diff
@@ -89,5 +91,12 @@ describe('ng-update avatar', () => {
         }),
     );
 
+    // Guards Gemini's fix: a bound string literal is not a SafeResourceUrl, so it is migrated
+    // to [tuiAvatar] without a TODO (the snapshot's "After" has no inserted comment).
+    it(
+        'keeps a bound string literal [src] without a SafeResourceUrl TODO',
```
nsbarsukov: Do we really need the such kind of comments?)

mdlufy: Oh no )

splincode (approved):

nsbarsukov (approved):

PR #14499, opened 2026-07-06
fix(cdk): keep `tuiRepeatTimes` loop value numeric in @for blocks

mdlufy:

Part of #13823 (item: *"`12 | tuiRepeatTimes` is not equivalent to `'-'.repeat(12)` — the pipe produces numbers, the `repeat()` produces strings"*).

### What is the problem?

`N | tuiRepeatTimes` iterated a **numeric** index (`0..N-1`). The `*ngFor` and
`*tuiRepeatTimes` migration paths already remap the loop variable to `$index`, but
the **`@for`-block** path only stripped the pipe:

```html
<!-- before -->
@for (n of config.count | tuiRepeatTimes; track n; let isOdd = $odd) { {{n}} }

<!-- after (buggy) -->
@for (n of '-'.repeat(config.count); track n; let isOdd = $odd) { {{n}} }
```

`'-'.repeat(N)` is a **string**, so `n` bound to the `'-'` char: `{{n}}` rendered
`-` and `track n` gave every item the same key.

### What is the solution?

Rewrite the `@for` header the same way the `*tuiRepeatTimes` path does — iterate
with `_`, `track $index`, and alias the original variable back to the numeric
`$index`:

```html
<!-- after (fixed) -->
@for (_ of '-'.repeat(config.count); track $index; let n = $index; let isOdd = $odd) { {{n}} }
```

- Nested `@for` blocks each keep their own `$index` (verified).
- A pipe nested inside the iterable expression (`wrap(b | tuiRepeatTimes)`) or used
  only in the body is left untouched — the rewrite is anchored to the iteration
  segment ending in `| tuiRepeatTimes`.

### DIFF
--- a/projects/cdk/schematics/ng-update/v5/tests/schematic-migrate-repeat-times.spec.ts
+++ b/projects/cdk/schematics/ng-update/v5/tests/schematic-migrate-repeat-times.spec.ts
@@ -2,7 +2,7 @@ import {join} from 'node:path';
 
 import {resetActiveProject} from 'ng-morph';
 
-import {createMigration, runMigration} from '../../../utils/run-migration';
+import {createMigration} from '../../../utils/run-migration';
 
 const COLLECTION = join(__dirname, '../../../migration.json');
 
@@ -196,19 +196,6 @@ describe('ng-update tuiRepeatTimes', () => {
                 '<ng-container *tuiRepeatTimes="let i of 3"><span>{{ i }}</span></ng-container>',
         }),
     );
-
-    it('leaves a @for block with an empty tuiRepeatTimes expression untouched', async () => {
-        const template = '@for (n of   | tuiRepeatTimes; track n) { {{n}} }';
-
-        const {template: result} = await runMigration({
-            collection: COLLECTION,
-            component: REPEAT_TIMES_COMPONENT,
-            template,
-        });
-
-        expect(result).not.toContain("'-'.repeat");
-        expect(result).toContain(template);
-    });
 
     afterEach(() => resetActiveProject());
 });
--- a/projects/cdk/schematics/ng-update/v5/tests/schematic-migrate-repeat-times.spec.ts
+++ b/projects/cdk/schematics/ng-update/v5/tests/schematic-migrate-repeat-times.spec.ts
@@ -2,7 +2,21 @@ import {join} from 'node:path';
 
 import {resetActiveProject} from 'ng-morph';
 
-import {createMigration} from '../../../utils/run-migration';
+import {createMigration, runMigration} from '../../../utils/run-migration';
+
+const COLLECTION = join(__dirname, '../../../migration.json');
+
+const REPEAT_TIMES_COMPONENT = /* TypeScript */ `
+    import {Component} from '@angular/core';
+    import {TuiRepeatTimesPipe} from '@taiga-ui/cdk';
+
+    @Component({
+        standalone: true,
+        templateUrl: './test.html',
+        imports: [TuiRepeatTimesPipe],
+    })
+    export class Test {}
+`;
 
 describe('ng-update tuiRepeatTimes', () => {
     const migrate = createMigration({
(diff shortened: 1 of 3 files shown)


projects/cdk/schematics/ng-update/v5/tests/schematic-migrate-repeat-times.spec.ts
```diff
@@ -193,5 +207,23 @@ describe('ng-update tuiRepeatTimes', () => {
...
+    it('keeps the loop value numeric in @for blocks (issue #13823)', async () => {
+        const {template} = await runMigration({
+            collection: COLLECTION,
+            component: REPEAT_TIMES_COMPONENT,
+            template:
+                '@for (n of config.count | tuiRepeatTimes; track n; let isOdd = $odd) { {{n}} }',
+        });
+
+        // `'-'.repeat(N)` only provides the iteration count
+        expect(template).toContain("'-'.repeat(config.count)");
```
splincode: Why did you choose this approach instead of snapshots?

mdlufy: The two @for-block snapshots in this file already capture the fixed output (I updated them), so snapshot coverage of the behavior exists

I added the explicit-assertion test on top for one specific reason: a snapshot can't express "this must fail on the bug." The old snapshot literally encoded the broken output (track n, {{n}} bound to the '-' string) and stayed green — snapshots record what is produced, not what is correct, so here they masked the regression. A fresh toMatchSnapshot would have just baselined the buggy output and passed

The explicit assertions state the actual invariant — the loop value must stay numeric (track $index + let n = $index, never track n) — so the test genuinely failed before the fix and is robust to unrelated formatting changes. This also matches the existing non-snapshot style in schematic-migrate-dialog-legacy-sizes.spec.ts and schematic-migrate-scroll-service-warning.spec.ts

Happy to convert it to a pure snapshot if you'd prefer strict consistency — but I'd lean on keeping the intent assertion since it documents the correctness property directly

splincode: it's as if it's easier to look at the snapshot anyway

nsbarsukov: I'd still lean toward snapshots here, and I don't think the "old snapshot encoded the broken output" point argues against them.

That snapshot passed on buggy output for one reason: it was regenerated and approved without anyone scrutinizing the diff. That's a review miss, not a property of snapshots — and it's exactly the thing reviewing a snapshot diff is meant to catch. Now that the bug is understood, a regenerated snapshot shows `track $index`, `let n = $index`, and the preserved `let isOdd = $odd` **in full context, all at once** — strictly more information to sign off on than five `toContain` lines.

A couple of concrete concerns with the assertion version:

- **It under-specifies the output.** The five substring checks pass for many different outputs, including subtly wrong ones (corrupted whitespace, a dropped closing `}`, a mangled unrelated attribute, a duplicated block). For a schematic whose output *is* a rewritten file, whole-file comparison is the point; five string peeks are a weaker net.
- **It couples the test to an implementation artifact.** `toContain("'-'.repeat(config.count)")` hard-codes an internal detail we even have to apologize for in a comment. If the schematic later emits a cleaner iteration construct, this test breaks even though the migration got better. `not.toMatch(/track\s+n\b/)` is a fragile negative regex on top of that.

On the one fair point — that a fresh snapshot can't "fail before the fix": that's a one-time authoring concern, not a lifetime one. We can honor it by writing the correct expected snapshot first and running it against the buggy code (mismatch → red), or by rejecting the buggy snapshot in review because the captured output is visibly wrong. After the fix ships, the test's job is regression protection for years, where whole-output coverage wins.

We don't lose the intent either — the invariant already lives in the test title, and we can keep a one-line comment above the snapshot. Proposing we convert this to the `migrate()` snapshot like the rest of the file.

mdlufy: Fair enough — converted it to the migrate() snapshot. The regression-protection argument convinced me: "can't fail before the fix" is a one-time authoring cost, whereas whole-file comparison is the stronger net for a schematic whose output is a rewritten file, and it keeps the assertion coupling ('-'.repeat(...), the track\s+n regex) out of the test.

One thing worth flagging: the exact case was already covered by the keeps @for tail params… snapshot (same component + template), so instead of adding a second identical snapshot I folded the #13823 intent into that test's title + a one-line comment and dropped the standalone assertion test. Its After now reads:


@for (_ of '-'.repeat(config.count); track $index; let n = $index; let isOdd = $odd) { {{n}} }
so track $index / let n = $index / no track n are all visible in one place. Done in df9cea7.

splincode (approved):

projects/cdk/schematics/ng-update/v5/tests/schematic-migrate-repeat-times.spec.ts
```diff
@@ -2,7 +2,21 @@ import {join} from 'node:path';
...
+    import {Component} from '@angular/core';
+    import {TuiRepeatTimesPipe} from '@taiga-ui/cdk';
+
+    @Component({
+        standalone: true,
+        templateUrl: './test.html',
+        imports: [TuiRepeatTimesPipe],
+    })
+    export class Test {}
+`;
```
nsbarsukov, line 19: Should it be re-used in this file if this variable is introduced ?

nsbarsukov (approved):

PR #14501, opened 2026-07-06
fix(core): `TUI_DARK_MODE` persist first theme change on empty storage

mdlufy:
Problem

Saving the theme was lost on the **first** toggle when `localStorage` was empty (it only persisted on the second try, so after a refresh the app fell back to the system theme). Previously masked because the key was already present in storage — surfaced on a fresh origin.

Root cause: persistence ran through an `effect` whose *initial flush* was supposed to consume the `automatic` guard. That relied on the effect flushing before the first user `set()`, which isn't guaranteed — the first change raced the initial flush and got swallowed.

Fix

- Seed the baseline value synchronously (`previous`) so the initial run is skipped without depending on flush timing.
- Mark machine-driven changes (system theme / `reset`) as `automatic` only when the value actually changes, avoiding a lingering guard after a no-op reset.

### DIFF
--- a/projects/core/tokens/dark-mode.ts
+++ b/projects/core/tokens/dark-mode.ts
@@ -10,43 +10,40 @@ export const TUI_DARK_MODE_KEY = new InjectionToken(
     {factory: () => TUI_DARK_MODE_DEFAULT_KEY},
 );
 
-export type TuiDarkMode = WritableSignal<boolean> & {reset(): void};
+export const TUI_DARK_MODE = new InjectionToken<
+    WritableSignal<boolean> & {reset(): void}
+>(ngDevMode ? 'TUI_DARK_MODE' : '', {
+    factory: () => {
+        const storage = inject(WA_LOCAL_STORAGE);
+        const key = inject(TUI_DARK_MODE_KEY);
+        const saved = storage?.getItem(key);
+        const media = inject(WA_WINDOW).matchMedia('(prefers-color-scheme: dark)');
+        const result = signal(saved ? saved === 'true' : media.matches);
+        // Raw setter that does not persist — used for machine-driven changes.
+        const set = result.set.bind(result);
 
-export const TUI_DARK_MODE = new InjectionToken<TuiDarkMode>(
-    ngDevMode ? 'TUI_DARK_MODE' : '',
-    {
-        factory: () => {
-            const storage = inject(WA_LOCAL_STORAGE);
-            const key = inject(TUI_DARK_MODE_KEY);
-            const saved = storage?.getItem(key);
-            const media = inject(WA_WINDOW).matchMedia('(prefers-color-scheme: dark)');
-            const result = signal(saved ? saved === 'true' : media.matches);
-            // Raw setter that does not persist — used for machine-driven changes.
-            const set = result.set.bind(result);
+        // Persist an explicit choice and pin it, even when the value is
+        // unchanged (e.g. picking dark while the system is already dark).
+        const pin = (value: boolean): void => {
+            storage?.setItem(key, String(value));
+            set(value);
+        };
 
-            // Persist an explicit choice and pin it, even when the value is
-            // unchanged (e.g. picking dark while the system is already dark).
-            const pin = (value: boolean): void => {
-                storage?.setItem(key, String(value));
-                set(value);
-            };
+        // Follow the system theme while it has not been pinned (storage empty).
+        fromEvent(media, 'change')
+            .pipe(
+                filter(() => !storage?.getItem(key)),
+                takeUntilDestroyed(),
+            )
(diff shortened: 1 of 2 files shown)


waterplea (approved):

projects/core/tokens/dark-mode.ts
```diff
@@ -17,45 +10,43 @@ export const TUI_DARK_MODE_KEY = new InjectionToken(
...
-    WritableSignal<boolean> & {reset(): void}
->(ngDevMode ? 'TUI_DARK_MODE' : '', {
-    factory: () => {
-        let automatic = true;
-        const storage = inject(WA_LOCAL_STORAGE);
-        const key = inject(TUI_DARK_MODE_KEY);
-        const saved = storage?.getItem(key);
-        const media = inject(WA_WINDOW).matchMedia('(prefers-color-scheme: dark)');
-        const result = signal(Boolean((saved && JSON.parse(saved)) ?? media.matches));
+export type TuiDarkMode = WritableSignal<boolean> & {reset(): void};
```
nsbarsukov: Delete

projects/core/tokens/tests/dark-mode.spec.ts
```diff
@@ -0,0 +1,180 @@
+import {TestBed} from '@angular/core/testing';
```
splincode, line 2:
```ts
import {TestBed} from '@angular/core/testing';
import {WA_LOCAL_STORAGE, WA_WINDOW} from '@ng-web-apis/common';
import {TUI_DARK_MODE, TUI_DARK_MODE_DEFAULT_KEY, type TuiDarkMode} from '@taiga-ui/core';

const KEY = TUI_DARK_MODE_DEFAULT_KEY;

describe('TUI_DARK_MODE', () => {
    let storage: Map<string, string>;
    let systemDark: boolean;
    let listeners: Set<() => void>;

    const getStored = (): string | undefined => storage.get(KEY);

    const changeSystemTheme = (value: boolean): void => {
        systemDark = value;
        listeners.forEach((listener) => listener());
    };

    function setup({
        stored,
        system = false,
    }: {
        stored?: boolean;
        system?: boolean;
    } = {}): TuiDarkMode {
        storage = new Map(stored === undefined ? [] : [[KEY, String(stored)]]);
        systemDark = system;
        listeners = new Set();

        TestBed.configureTestingModule({
            providers: [
                {
                    provide: WA_LOCAL_STORAGE,
                    useValue: {
                        getItem: (key: string) => storage.get(key) ?? null,
                        setItem: (key: string, value: string) => storage.set(key, value),
                        removeItem: (key: string) => storage.delete(key),
                    } as unknown as Storage,
                },
                {
                    provide: WA_WINDOW,
                    useValue: {
                        matchMedia: () => ({
                            get matches() {
                                return systemDark;
                            },
                            addEventListener: (_: string, listener: () => void) =>
                                listeners.add(listener),
                            removeEventListener: (_: string, listener: () => void) =>
                                listeners.delete(listener),
                        }),
                    } as unknown as Window,
                },
            ],
        });

        return TestBed.inject(TUI_DARK_MODE);
    }

    afterEach(() => {
        TestBed.resetTestingModule();
    });

    it('uses system value when storage is empty without persisting it', () => {
        const mode = setup({system: true});

        expect(mode()).toBe(true);
        expect(getStored()).toBeUndefined();
    });

    it('uses stored value over system value', () => {
        const mode = setup({stored: true, system: false});

        expect(mode()).toBe(true);
    });

    it('pins explicit set even when value equals current system value', () => {
        const mode = setup({system: true});

        mode.set(true);
        changeSystemTheme(false);

        expect(getStored()).toBe('true');
        expect(mode()).toBe(true);
    });

    it('persists update', () => {
        const mode = setup();

        mode.update((value) => !value);

        expect(getStored()).toBe('true');
        expect(mode()).toBe(true);
    });

    it('follows system theme while not pinned and resumes after reset', () => {
        const mode = setup();

        changeSystemTheme(true);

        expect(mode()).toBe(true);
        expect(getStored()).toBeUndefined();

        mode.set(false);
        changeSystemTheme(true);

        expect(mode()).toBe(false);
        expect(getStored()).toBe('false');

        mode.reset();
        changeSystemTheme(false);

        expect(mode()).toBe(false);
        expect(getStored()).toBeUndefined();
    });
});
```

nsbarsukov (approved):

projects/core/tokens/dark-mode.ts
mdlufy: updated

projects/core/tokens/tests/dark-mode.spec.ts
mdlufy: Updated

splincode (approved):

PR #14502, opened 2026-07-06
fix(core): `Popup` sync change detection

### LINKED ISSUE #12586: 🐞 - ActionBar doesn't disappear when using regular variables to toggle
### Reproduction url

https://stackblitz.com/edit/angular-jnmhcyjv

### Description

The ActionBar display is not toggled unless something marks the view containing it as dirty. When using regular variables (not signals) to manipulate the ActionBar and its Close button is inside it, it does not close.

You can see this in the reproduction by first calling the ActionBar with a button and then closing it with the Close button. However, if you use a "Close with markForCheck" button, everything works correctly.

I tried to figure out why this was happening. The problem is most likely due to the internal `markViewDirty`, called when processing outputs, marking the wrong view as dirty. I haven't investigated further.

### Taiga UI version

4.61.0

### Browser Used

Chrome

### OS Used

macOS

waterplea:
Fixes #12586

### DIFF
--- a/projects/core/portals/popup/popups.component.ts
+++ b/projects/core/portals/popup/popups.component.ts
@@ -11,7 +11,7 @@ import {tuiProvide} from '@taiga-ui/cdk/utils/di';
 
 import {TuiPopupService} from './popup.service';
 
-export const TUI_CDR = '_tuiCdr';
+export const TUI_CDR = '_tuiCdr' as const;
 
 @Component({
     selector: 'tui-popups',
@@ -27,13 +27,13 @@ export class TuiPopups extends TuiPortals implements AfterViewChecked {
         for (let i = 0; i < this.anchor().vcr.length; i++) {
             const ref = this.anchor().vcr.get(i);
 
-            if (isEmbeddedViewRef<Record<string, any>>(ref)) {
-                ref.context[TUI_CDR]?.markForCheck();
+            if (isEmbeddedViewRef(ref)) {
+                ref.context?.[TUI_CDR]?.markForCheck();
             }
         }
     }
 }
 
-function isEmbeddedViewRef<C>(ref?: ViewRef | null): ref is EmbeddedViewRef<C> {
+function isEmbeddedViewRef(ref?: ViewRef | null): ref is EmbeddedViewRef<any> {
     return !!ref && 'context' in ref;
 }
--- a/projects/core/portals/popup/popups.component.ts
+++ b/projects/core/portals/popup/popups.component.ts
@@ -1,9 +1,17 @@
-import {ChangeDetectionStrategy, Component} from '@angular/core';
+import {
+    type AfterViewChecked,
+    ChangeDetectionStrategy,
+    Component,
+    type EmbeddedViewRef,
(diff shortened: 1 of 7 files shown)


projects/core/portals/popup/popups.component.ts
```diff
@@ -14,4 +22,18 @@ import {TuiPopupService} from './popup.service';
     changeDetection: ChangeDetectionStrategy.Default,
     providers: [tuiProvide(TuiPortalService, TuiPopupService)],
 })
-export class TuiPopups extends TuiPortals {}
+export class TuiPopups extends TuiPortals implements AfterViewChecked {
```
splincode, line 25: Can we avoid `ngAfterViewChecked` here?

The current fix works, but it marks the origin view for check on every popup CD cycle while the popup is opened. It feels a bit too broad and may make OnPush owners checked more often than needed.

Would it be enough to mark origin CDR only on interactions inside `tui-popups` for example `click`, `input`, `change`, `keydown`, maybe with `queueMicrotask`? The reported case is caused by a user action inside the popup, so this would make the bridge event-driven instead of CD-cycle-driven.

If we need to support arbitrary async state changes from popup content, then the current approach is more generic, but it would be nice to document why `AfterViewChecked` is required here.


```ts
const TUI_POPUP_EVENTS = ['click', 'input', 'change', 'keydown', 'submit'] as const;

@Component({
    selector: 'tui-popups',
    imports: [TuiVCR],
    template: '',
    styleUrl: './popups.style.less',
    changeDetection: ChangeDetectionStrategy.Default,
    providers: [tuiProvide(TuiPortalService, TuiPopupService)],
})
export class TuiPopups extends TuiPortals {
    private readonly destroyRef = inject(DestroyRef);
    private readonly elementRef = inject(ElementRef<HTMLElement>);

    constructor() {
        super();

        for (const event of TUI_POPUP_EVENTS) {
            this.elementRef.nativeElement.addEventListener(event, this.markOwnersForCheck);
        }

        this.destroyRef.onDestroy(() => {
            for (const event of TUI_POPUP_EVENTS) {
                this.elementRef.nativeElement.removeEventListener(event, this.markOwnersForCheck);
            }
        });
    }

    private readonly markOwnersForCheck = (): void => {
        queueMicrotask(() => {
            for (let i = 0; i < this.anchor().vcr.length; i++) {
                const ref = this.anchor().vcr.get(i);
    
                if (isEmbeddedViewRef(ref)) {
                    ref.context?.[TUI_CDR]?.markForCheck();
                }
            }
        });
    };

}

function isEmbeddedViewRef(ref?: ViewRef | null): ref is EmbeddedViewRef<Record<string, ChangeDetectorRef>> {
    return !!ref && 'context' in ref;
}
```

splincode (approved):

projects/core/portals/popup/popups.component.ts
waterplea: It's ok, it only does so with opt-in by providing CDR in context, so only for tuiPopup directive, not for every portal content.

nsbarsukov (approved):

PR #14515, opened 2026-07-06
fix(addon-mobile): `SheetDialog` restore `theme-color` on close and make it configurable

### LINKED ISSUE #13690: 🐞 -  Status bar coloring during NotificationMiddle opening
### Reproduction url

-

### Description

When NotificationMiddle opens, the status bar on iPhone Safari (Chrome is OK), Android Chrome is painted gray.
The color change call occurs here:
color | @ | taiga-ui-cdk-services.mjs:135
(anonymous) | @ | taiga-ui-addon-mobile-components-sheet-dialog.mjs:118
ngOnChanges | @ | taiga-ui-cdk-directives-popover.mjs:18
At the time of the call, there are no open dialogs.

### Taiga UI version

4.75.0

### Browser Used

Chrome

### OS Used

Android

loxy:
Three independent failure modes in TuiSheetDialogService / TuiThemeColorService around `<meta name="theme-color">` handling:

- Capture the theme color at the moment the first sheet opens instead of once at construction, so a theme change made after the service was created (e.g. an app-level dark-mode effect) is no longer lost on revert, and nested sheets no longer revert onto the sheet color.
- Make the sheet color configurable via `themeColor` on `TUI_SHEET_DIALOG_OPTIONS` (defaults to `#404040`, `null` opts out) instead of a hardcoded constant that only matched the backdrop in light mode.
- Split TuiThemeColorService: `color` now writes only the `<meta name="theme-color">` tag, while the new `navColor` setter writes the `--tui-theme-color` CSS variable used by navigation chrome. Dimming the address bar no longer flickers the nav.

Adds a jest spec (revert / nesting / opt-out matrix), a Cypress test for the real-browser open/close behavior, and a `themeColor` docs example.

Fixes https://github.com/taiga-family/taiga-ui/issues/13690

### DIFF
--- a/projects/addon-mobile/components/sheet-dialog/sheet-dialog.options.ts
+++ b/projects/addon-mobile/components/sheet-dialog/sheet-dialog.options.ts
@@ -11,6 +11,13 @@ export interface TuiSheetDialogOptions<I = undefined> {
     readonly stops: readonly string[];
     readonly bar: boolean;
     readonly required: boolean;
+    /**
+     * Address-bar (`<meta name="theme-color">`) color applied while the sheet is
+     * open, matching the dimmed backdrop. Pass `null` to leave the address bar
+     * untouched (e.g. apps that manage `theme-color` themselves, or need a value
+     * that tracks dark mode).
+     */
+    readonly themeColor: string | null;
 }
 
 export const TUI_SHEET_DIALOG_DEFAULT_OPTIONS: TuiSheetDialogOptions = {
@@ -23,6 +30,7 @@ export const TUI_SHEET_DIALOG_DEFAULT_OPTIONS: TuiSheetDialogOptions = {
     data: undefined,
     bar: true,
     required: false,
+    themeColor: '#404040',
 };
 
 /**
--- a/projects/addon-mobile/components/sheet-dialog/sheet-dialog.options.ts
+++ b/projects/addon-mobile/components/sheet-dialog/sheet-dialog.options.ts
@@ -11,6 +11,13 @@ export interface TuiSheetDialogOptions<I = undefined> {
     readonly stops: readonly string[];
     readonly bar: boolean;
     readonly required: boolean;
+    /**
+     * Address-bar (`<meta name="theme-color">`) color applied while the sheet is
+     * open, matching the dimmed backdrop. Pass an empty string to leave the
+     * address bar untouched (e.g. apps that manage `theme-color` themselves, or
+     * need a value that tracks dark mode).
+     */
+    readonly themeColor: string;
 }
 
 export const TUI_SHEET_DIALOG_DEFAULT_OPTIONS: TuiSheetDialogOptions = {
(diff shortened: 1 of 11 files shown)


loxy:
Hi @waterplea — this is the v5 PR bundling the three theme-color fixes you greenlit in [discussion #14124](https://github.com/orgs/taiga-family/discussions/14124). Refs #13690.

Marked **draft** on purpose: the fixes are done and covered (jest + a Cypress test), but three API-shape calls are yours to make before I finalize:

1. **Configurable color** — I implemented `themeColor` on `TUI_SHEET_DIALOG_OPTIONS` as a provider/global default (`null` opts out). Do you also want a per-`open()` / directive override, or is the provider default enough?
2. **Split naming & compat** — `color` now writes only `<meta name="theme-color">`, and a new `navColor` writes the `--tui-theme-color` CSS var. OK with that split and a clean v5 break, or would you prefer a back-compat shim on the existing setter?
3. **Revert model** — I capture the initial color at the first open (`0 → 1`) rather than in the constructor. Happy to switch to an initial-color stack if you had a different mechanism in mind for the nested case.

One heads-up from on-device testing: the address-bar (`theme-color`) effect no longer renders on **iOS 26 Safari** (ignores the meta, derives the toolbar from CSS edge backgrounds) or **Chrome dark mode** (overrides it). It still works on Chrome light / iOS ≤ X?, so the fixes are valid — but the `--tui-theme-color` (nav) path is the more durable one now. A CSS-based toolbar approach for iOS 26 might be worth a separate discussion down the line.

waterplea (commented):
This looks good, thank you very much! I think I only have couple of concerns:

1. Maybe let's not change the `TuiThemeColorService` and handle metadata directly in `TuiSheetDialogService` using `Meta`?
2. We can probably not introduce null and just use empty string as falsy value for opt-out? If you prefer more expressive `null` we can still use `!themeColor` in checks since empty string also makes no sense as color
3. What's the purpose of `const {themeColor} = this.options;`? Seems like an extra line, can we just use `this.options.themeColor`?

loxy:
Thank you for the feedback. I'm on vacation currently. Don't know how much time I will have next days. But I will try to address your suggestions and remarks.

loxy:
Thanks, all three make sense — done:

  1. **`TuiThemeColorService` untouched.** The sheet now writes the tag through `Meta`
     directly, so that file and the two navigation examples are out of the diff. Better
     than my split anyway: `color` keeps writing `--tui-theme-color`, so no silent
     breakage for anyone using it to tint nav chrome. And reading the live tag on open
     also picks up colors set without the service.
  2. **Empty string for opt-out.** `themeColor: string`, checked with `!themeColor` —
     which also covers `undefined` from a partial options provider.
  3. **Destructure gone**, `this.options.themeColor` read directly.

  One addition: if the page has no `theme-color` tag at all, revert now calls
  `removeTag` instead of leaving `content=""` behind. Covered by a spec case.

waterplea (approved):
LGTM, thank you!

loxy:
We need a second reviewer, right?

waterplea:
Yes, pinged the team

vladimirpotekhin (approved):

PR #14542, opened 2026-07-09
refactor(kit): `Textarea` refactor scrollable container

waterplea:
Fixes #


splincode (approved):

projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
```diff
@@ -28,7 +29,8 @@ const REQUIRED_ERROR = new Error(ngDevMode ? 'Required dialog was dismissed' : '
     templateUrl: './sheet-dialog.template.html',
     styleUrl: './sheet-dialog.style.less',
     changeDetection: ChangeDetectionStrategy.OnPush,
-    hostDirectives: [TuiAnimated, TuiScrollRef],
+    providers: [tuiProvide(TUI_SCROLL_REF, ElementRef)],
+    hostDirectives: [TuiAnimated],
```
nsbarsukov, line 33: Why not 

```suggestion
    hostDirectives: [TuiAnimated, TuiScrollRef],
```
?

nsbarsukov (approved):

projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
waterplea: TuiScrollRef in future would add animations that control scrollbar positions, not just add provider and we do not want animations on modals, they already have their slideIns

Issue #14543, opened 2026-07-09
🐞 - Textfield multi control never becoming touched in Chrome 150


dzolotarova:
### Reproduction url

https://stackblitz.com/edit/angular-yf8hpqdj?file=src%2Fapp%2Fapp.template.html,src%2Fapp%2Fapp.component.ts

### Description

After updating Chrome to version 150, all textfields[multi] are not becoming `touched`. I’ve looked into this issue a little, and it seems that the following line stopped working after the update:
```ts
'(tuiActiveZoneChange)': '!$event && (el.scrollTo({left: 0}) || cva?.onTouched())'
```
Now `el.scrollTo({left: 0})` return `Promise` instead of `void`

Release notes:
https://developer.chrome.com/release-notes/150#programmatic_scroll_promises

### Taiga UI version

4.90.0

### Browser Used

Chrome

### OS Used

macOS

dzolotarova:
@waterplea hello! I’m sorry to bother you, but this issue has caused our project to malfunction. I’ve created a pull request that fixes this issue. Could you please take a look at it?

waterplea:
Thank you for a quick fix! Coming on Monday

PR #14544, opened 2026-07-09
fix(core): handle scrollTo promise in multi textfield

dzolotarova:
Fixes #[14543](https://github.com/taiga-family/taiga-ui/issues/14543)

### DIFF
--- a/projects/core/components/textfield/textfield-multi/textfield-multi.component.ts
+++ b/projects/core/components/textfield/textfield-multi/textfield-multi.component.ts
@@ -91,8 +91,7 @@ import {WA_WINDOW} from '@ng-web-apis/common';
         '[style.--t-item-height.px]': 'height()',
         '[style.--t-rows]': 'rows',
         '(click.prevent)': 'onClick($event.target)',
-        '(tuiActiveZoneChange)':
-            '!$event && (el.scrollTo({left: 0}) || cva?.onTouched())',
+        '(tuiActiveZoneChange)': 'onActiveZone($event)',
         // TODO: Remove in v5
         '[attr.data-size]': 'options.size()',
         '[class._with-label]': 'hasLabel',
@@ -170,6 +169,13 @@ export class TuiTextfieldMultiComponent<T> extends TuiTextfieldBaseComponent<T> 
 
         if (!selection?.rangeCount || selection.getRangeAt(0)?.collapsed) {
             this.interactiveInput?.nativeElement.focus();
+        }
+    }
+
+    protected onActiveZone(active: boolean): void {
+        if (!active) {
+            this.el.scrollTo({left: 0});
+            this.cva?.onTouched();
         }
     }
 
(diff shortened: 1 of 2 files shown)


splincode (approved):

projects/core/components/textfield/textfield-multi/textfield-multi.component.ts
```diff
@@ -173,6 +172,13 @@ export class TuiTextfieldMultiComponent<T> extends TuiTextfieldBaseComponent<T>
         }
     }
 
+    protected onActiveZone(active: boolean): void {
```
splincode, line 175: Can we add a test so that when the cypress chrome binary is updated to 150, the test should not fail?

```ts
import {ChangeDetectionStrategy, Component} from '@angular/core';
import {FormControl, ReactiveFormsModule} from '@angular/forms';
import {TuiRoot, TuiTextfield} from '@taiga-ui/core';
import {TuiInputChip} from '@taiga-ui/kit';

@Component({
    standalone: true,
    imports: [ReactiveFormsModule, TuiInputChip, TuiRoot, TuiTextfield],
    template: `
        <tui-root>
            <button
                data-testid="outside"
                type="button"
            >
                Outside
            </button>

            <div data-testid="state">
                TOUCHED: {{ control.touched ? 'TRUE' : 'FALSE' }}
            </div>

            <tui-textfield multi>
                <input
                    tuiInputChip
                    [formControl]="control"
                />

                <tui-input-chip *tuiItem />
            </tui-textfield>
        </tui-root>
    `,
})
class TestInputChip {
    protected readonly control = new FormControl<string[]>([]);
}

describe('InputChip | touched state', () => {
    it('marks control as touched when scrollTo returns Promise', () => {
        cy.mount(TestInputChip);

        cy.window().then((win) => {
            cy.stub(win.Element.prototype, 'scrollTo').returns(
                Promise.resolve(true) as never,
            );
        });

        cy.get('[data-testid="state"]').should('contain.text', 'TOUCHED: FALSE');

        cy.get('input[tuiInputChip]').focus();
        cy.get('[data-testid="outside"]').click();

        cy.get('[data-testid="state"]').should('contain.text', 'TOUCHED: TRUE');
    });
});
```

dzolotarova: Yeah, looks like a good idea. I will add it

projects/demo-cypress/src/tests/input-chip-touched.cy.ts
```diff
@@ -0,0 +1,59 @@
+import {ChangeDetectionStrategy, Component} from '@angular/core';
+import {FormControl, ReactiveFormsModule} from '@angular/forms';
+import {TuiRoot, TuiTextfield} from '@taiga-ui/core';
+import {TuiInputChip} from '@taiga-ui/kit';
+
+@Component({
```
splincode, line 6: Could you add also the test in 5.x as well?

dzolotarova: Check please
https://github.com/taiga-family/taiga-ui/pull/14545

nsbarsukov (approved):

PR #14545, opened 2026-07-09
chore: add test for scrollTo promise

dzolotarova:
Fixes #

### DIFF
--- a/projects/demo-cypress/src/tests/input-chip-touched.cy.ts
+++ b/projects/demo-cypress/src/tests/input-chip-touched.cy.ts
@@ -44,9 +44,7 @@ describe('InputChip | touched state', () => {
         cy.mount(TestInputChip);
 
         cy.window().then((win) => {
-            cy.stub(win.Element.prototype, 'scrollTo').returns(
-                Promise.resolve(true) as never,
-            );
+            cy.stub(win.Element.prototype, 'scrollTo').returns(Promise.resolve(true));
         });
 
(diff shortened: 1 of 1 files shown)


projects/demo-cypress/src/tests/input-chip-touched.cy.ts
```diff
@@ -0,0 +1,59 @@
...
+    // `Element.prototype.scrollTo` currently returns `undefined`, but a future Chrome
+    // binary (e.g. bumped to 150) may return a `Promise`. `TuiTextfieldMulti` must mark
+    // the control as touched on blur regardless of what `scrollTo` returns, so we stub it
+    // to return a truthy `Promise` to lock in that behavior independently of the browser.
+    it('marks control as touched on blur even when scrollTo returns a Promise', () => {
+        cy.mount(TestInputChip);
+
+        cy.window().then((win) => {
+            cy.stub(win.Element.prototype, 'scrollTo').returns(
+                Promise.resolve(true) as never,
```
splincode: add eslint ignore

splincode (approved):

projects/demo-cypress/src/tests/input-chip-touched.cy.ts
dzolotarova: Removed the cast

waterplea:
I believe testing for stubbed browser API is excessive. Good that it works in main and thank you for testing, but adding such test is unnecessary.

PR #14546, opened 2026-07-09
feat(core): `ScrollControls` automate use with virtual scroll, deprecate `Scrollable`

waterplea:
Fixes #


splincode (approved):

projects/demo/src/pages/components/select/examples/6/index.html
```diff
@@ -7,10 +7,12 @@
 
     <cdk-virtual-scroll-viewport
         *tuiDropdown
-        tuiScrollable
+        appendOnly
```
nsbarsukov, line 10: Why was it added ?

projects/core/services/position.service.ts
```diff
@@ -30,6 +30,11 @@ export class TuiPositionService extends Observable<TuiPoint> {
                             animations.every(
                                 ({effect}) =>
                                     effect?.getComputedTiming().progress !== null,
+                            ) &&
+                            animations.some(
+                                (a) =>
+                                    'animationName' in a &&
+                                    a.animationName === 'tuiScale',
```
nsbarsukov, line 37: Could you explain why this animation is handled separately?

nsbarsukov (approved):

projects/core/services/position.service.ts
waterplea: Because we only care about this animation since it scales the element and ClientRect is reported incorrectly

projects/demo/src/pages/components/select/examples/6/index.html
waterplea: For some reason scrolling is jerky, you can see it on main. I didn't want to investigate it because it doesn't seem like it's our fault, maybe CDK scrolling. And appendOnly should be fine since we use OnPush and having hundreds of elements is fine, we mostly just don't want them to all be rendered at the same time when dropdown opens since that would lag.

Issue #13869, opened 2026-04-16
🐞 - Fix migrations from 4 to 5


tinkoffspectrum:
### Reproduction url

https://stackblitz.com

### Description

1. Unfortunately, ng-morph is unable to build the tree for my monorepo. It crashed with a Node.js out-of-memory error after 4 hours of execution. I’m not sure what’s wrong, but I used the same method to update to version 4 in the past. It likely pulls in node_modules or other excluded directories; perhaps it should respect the .gitignore file? Or maybe it just can't handle 1200+ libraries.
As a workaround, I put together a .sh script that iterates through apps, libs, and other folders, running them with ROOT_PATH.
2. process.env.ROOT_PATH is ignored during project creation in
    - node_modules/@taiga-ui/cdk/schematics/ng-update/v5/steps/migrate-tui-let.js
    - migrate-doc-i18n-tokens
    - migrate-tokens
3. node_modules/@taiga-ui/cdk/schematics/ng-update/v5/steps/migrate-tokens/update-tsconfig.js doesn't work correctly with Nx monorepos. These projects don't have typeRoots in every tsconfig; typeRoots is typically only in the root tsconfig.base.json, while types in tsconfig.lib.json can exist in every library. As a result, all libs with specified types are skipped. I had to add types manually. Furthermore, the path node_modules/@taiga-ui/tsconfig that the migration tries to insert doesn't even exist. Its purpose is unclear, and the project runs fine without it.
4. Migration for the tuiDropdownMobile directive not exist
5. tuiFallbackSrc for tuiAvatar is the same—no migration available.
6. tuiInputCardOptionsProvider is now tuiInputCardGroupOptionsProvider, but there's no autocomplete in settings and no migration (not even a warning).
7. TuiInputCVC and TuiInputExpire just disappeared. No errors, but no migration to remove them from imports either.
8. TUI_ALERT_POSITION has no migration
9. For tuiLink without [pseudo]="true", it would be helpful to have a migration comment suggesting a CSS fix.
10. tui-avatar-stack used to accept round; now it must be used on the avatar itself. No migration was provided.
11. TUI_CHECKBOX_DEFAULT_OPTIONS has no migration
12. tuiHexToRgb has no migration; it seems to have been deleted.
13. tui-thumbnail-card lost monoHandler. No warnings, no migrations—was it just removed? We used mono icons for payment systems on some cards.
14. Glad TuiCarouselComponent stayed in legacy, as itemsCount was used there
[...]

PR #14594, opened 2026-07-17
chore(schematics): warn on dynamic `[tuiHeader]` with legacy size tokens in v5 migration

mdlufy:
## Which issue does this PR close?

Addresses **#13869 (item 16)** — `<div [tuiHeader]="isMobile() ? 'l' : 'm'">` was not migrated and produced no comment.

## Context

In v5 `tuiHeader` accepts typography tokens instead of size tokens. The v4→v5 map is:

| v4 | v5 |
| --- | --- |
| `xxl` | `h1` |
| `xl` | `h2` |
| `l` | `h3` |
| `m` | `h4` |
| `s` | `h5` |
| `xs` | `h6` |
| `xxs` | `body-l` |

Static (`tuiHeader="l"`) and single-literal bound (`[tuiHeader]="'l'"`) values are already remapped by `ATTR_WITH_VALUES_TO_REPLACE`. But a **dynamic** binding keeps the old tokens inside an arbitrary expression (`isMobile() ? 'l' : 'm'`), which the exact-value replacer can't touch — so it was silently left as-is.

Rewriting quoted literals inside an arbitrary expression is unsafe: single-letter tokens like `'l'` / `'m'` / `'s'` routinely appear in unrelated comparisons (`mode === 'm' ? …`). Following the same reasoning the maintainers gave for `tuiRepeatTimes` in #13823 ("drop a comment"), this adds a TODO instead of a risky auto-rewrite.

## Behavior

A new template step flags a bound `[tuiHeader]` **only** when its value still contains a legacy size-token literal and isn't a single recognized literal:

- `[tuiHeader]="isMobile() ? 'l' : 'm'"` → **TODO added**, binding untouched (manual update).
- `[tuiHeader]="'l'"` → still auto-migrated to `'h3'` by the existing replacer, **no TODO**.
- `[tuiHeader]="headerSize"` / `[tuiHeader]="cond ? 'h3' : 'h4'"` → **no TODO** (no legacy tokens to flag).

```html
<!-- Before -->
<h1 [tuiHeader]="isMobile() ? 'l' : 'm'">Title</h1>

<!-- After -->
<!-- TODO: (Taiga UI migration) `tuiHeader` values changed in v5 from size tokens to
     typography tokens (xxl->h1, xl->h2, l->h3, m->h4, s->h5, xs->h6, xxs->body-l). This
     dynamic binding still contains old size tokens that cannot be migrated automatically —
     update them to the v5 tokens manually. See https://taiga-ui.dev/components/header -->
<h1 [tuiHeader]="isMobile() ? 'l' : 'm'">Title</h1>
```

## Tests

- New `schematic-migrate-header.spec.ts` (dynamic expression with legacy tokens → TODO; single bound literal → still migrated, no TODO).
- Full v5 suite green: **108 suites / 656 tests / 748 snapshots**.


projects/cdk/schematics/ng-update/v5/steps/templates/migrate-header.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const HEADER_ATTR = '[tuiHeader]'.toLowerCase();
+
+// A value that is exactly one recognized size/typography literal is handled by the
+// static value replacer (ATTR_WITH_VALUES_TO_REPLACE) or is already a v5 token.
+const SINGLE_LITERAL =
+    /^'(?:xxl|xl|[lms]|xs|xxs|h1|h2|h3|h4|h5|h6|body-l|body-m|body-s)'$/;
+
+// A v4 size token used as a quoted literal somewhere inside a larger expression.
+const OLD_TOKEN_LITERAL = /'(?:xxl|xl|[lms]|xs|xxs)'/;
```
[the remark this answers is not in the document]
mdlufy: Good catch — fixed in f5347ed. The old-token check now matches both quote styles via a backreference (`/(['"])(?:xxl|xl|[lms]|xs|xxs)\1/`), so `[tuiHeader]='cond ? "l" : "m"'` is flagged too, with a test covering it.

I kept `SINGLE_LITERAL` single-quote-only on purpose: `ATTR_WITH_VALUES_TO_REPLACE` only auto-migrates single-quoted literals, so a lone double-quoted literal like `[tuiHeader]='"l"'` isn't auto-migrated — letting it fall through to the old-token check gives it a TODO instead of being silently skipped.

projects/cdk/schematics/ng-update/v5/steps/templates/migrate-header.ts
```diff
@@ -0,0 +1,63 @@
...
+            .find((attr) => attr.name === HEADER_ATTR)
+            ?.value.trim();
+
+        if (!value || SINGLE_LITERAL.test(value) || !OLD_TOKEN_LITERAL.test(value)) {
+            return;
+        }
+
+        recorder.insertLeft(
+            templateOffset + loc.startOffset,
+            `<!-- ${TODO_MARK} ${DYNAMIC_HEADER_TODO} -->\n`,
```
nsbarsukov: Could we just re-use this approach ?
https://github.com/taiga-family/taiga-ui/blob/a462b8028283e0a4f8fa2e4a549e5c0247c9be9e/projects/cdk/schematics/ng-update/v5/steps/constants/html-comments.ts#L12

For example
https://github.com/taiga-family/taiga-ui/blob/a462b8028283e0a4f8fa2e4a549e5c0247c9be9e/projects/cdk/schematics/ng-update/v5/steps/constants/html-comments.ts#L66-L71

mdlufy: Done in d70b758 — folded the step into a declarative `HTML_COMMENTS` entry (`addHTMLCommentTags`). Same original-template read, same `TODO_MARK` insertion at the element start offset, so the output (and snapshots) are unchanged; the standalone `migrate-header.ts` step, its import and its registration in `migrate-templates` are gone (net -42 lines). The single-vs-double-quote logic is preserved in the entry's `filterFn`.

nsbarsukov (approved):

vladimirpotekhin (approved):

PR #14597, opened 2026-07-18
chore(schematics): auto-remove legacy TuiLegacyDropdownOpenMonitor in v5 migration

mdlufy:
## What's changed?

`TuiLegacyDropdownOpenMonitorDirective` (selector `[tuiDropdownOpenMonitor]`) was a **selector-only compatibility shim** that bridged the legacy and native dropdown open state. In v5 dropdowns handle this natively — and the schematic already maps `tuiDropdownOpen` → `tuiDropdownAuto` — so the marker is redundant.

Unlike the other removed legacy symbols (which map to differently-shaped v5 APIs and can only be warned about), this one has **no replacement to wire up** — the correct migration is simply to remove it. So instead of a warning, this migrates it automatically:

1. **TS** — added to `MODULES_TO_REMOVE`, so `removeModules` drops the `import` and the `imports: [...]` array entry.
2. **Template** — a new step removes the `tuiDropdownOpenMonitor` attribute from elements.

Split out from the warnings batch (#14596), where it was previously a warning entry.

## Before / After

```ts
// TS — before
import {TuiLegacyDropdownOpenMonitorDirective} from '@taiga-ui/legacy';
@Component({imports: [TuiLegacyDropdownOpenMonitorDirective]})
// TS — after
@Component({imports: []})
```

```html
<!-- template — before -->
<button tuiDropdown tuiDropdownOpen tuiDropdownOpenMonitor>Menu</button>
<!-- template — after (tuiDropdownOpen→tuiDropdownAuto by the existing step; monitor removed) -->
<button tuiDropdown tuiDropdownAuto>Menu</button>
```

## Tests

- New `schematic-migrate-dropdown-open-monitor.spec.ts` (TS import + imports-array removal; template attribute removal).
- Full v5 suite green: **108 suites / 656 tests / 748 snapshots**.


projects/cdk/schematics/ng-update/v5/steps/templates/migrate-dropdown-open-monitor.ts
```diff
@@ -0,0 +1,52 @@
...
+
+// parse5 lowercases attribute names, so match against lowercased forms.
+const MONITOR_ATTRS_LOWER = MONITOR_ATTRS.map((name) => name.toLowerCase());
+
+/**
+ * `[tuiDropdownOpenMonitor]` was a selector-only compatibility shim bridging the legacy
+ * and native dropdown open state. v5 dropdowns handle this natively (the schematic maps
+ * `tuiDropdownOpen` -> `tuiDropdownAuto`), so the marker is redundant — drop it.
+ */
+export function migrateDropdownOpenMonitor({
```
vladimirpotekhin: can this be done using [INPUTS_TO_REMOVE](https://github.com/taiga-family/taiga-ui/blob/main/projects/cdk/schematics/ng-update/v5/steps/constants/inputs-to-remove.ts) or [ATTRS_TO_REPLACE](https://github.com/taiga-family/taiga-ui/blob/main/projects/cdk/schematics/ng-update/v5/steps/constants/attrs-to-replace.ts#L118) ?

mdlufy: Done — dropped the bespoke `migrate-dropdown-open-monitor.ts` in favor of the declarative constants: the `tuiDropdownOpenMonitor` marker is now removed via `INPUTS_TO_REMOVE`, and the `TuiLegacyDropdownOpenMonitorDirective` import via `MODULES_TO_REMOVE`. `ATTRS_TO_REPLACE` did not fit here since we drop the attribute rather than rename it. Snapshots updated accordingly (43899a2).

vladimirpotekhin (approved):

nsbarsukov (approved):

PR #14604, opened 2026-07-21
fix(kit): `Textarea` should be resilient to global box-sizing overrides

### LINKED ISSUE #14603: 🐞 - Textarea layout breaks when using global box-sizing: border-box CSS reset
### Reproduction url

https://stackblitz.com/edit/angular-bhuyusfs-eqvj7sxe?file=src%2Fstyles.less

## Description

The textarea layout breaks when a global CSS reset applies `box-sizing: inherit`.

### Reproduction

Apply the following CSS reset:

```css
*, *::before, *::after {
  box-sizing: border-box;
}
```

### Actual behavior

The textarea layout is broken (incorrect sizing/layout).

### Expected behavior

The textarea should render correctly and maintain its expected layout when box-sizing: border-box is applied globally, as this is a common CSS reset used in many projects.

### Taiga UI version

5.15.0

### Browser Used

Chrome

### OS Used

macOS

vladimirpotekhin:
Fixes #14603

### DIFF
--- a/projects/kit/components/textarea/textarea.style.less
+++ b/projects/kit/components/textarea/textarea.style.less
@@ -12,6 +12,7 @@ tui-textarea-content:where(*&) {
     padding-inline-end: calc(var(--t-end) + var(--t-side) + var(--t-space));
     overflow: auto;
     pointer-events: none;
+    box-sizing: content-box;
 
     tui-textfield._with-label[data-size='m'] & {
         margin-block-start: calc(-1.25 * var(--tui-font-offset) - 1.375rem);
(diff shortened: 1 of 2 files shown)


nsbarsukov (approved):

mdlufy (approved):

projects/kit/components/textarea/textarea.style.less
```diff
@@ -12,6 +12,7 @@ tui-textarea-content:where(*&) {
     padding-inline-end: calc(var(--t-end) + var(--t-side) + var(--t-space));
     overflow: auto;
     pointer-events: none;
+    box-sizing: content-box;
```
mdlufy, line 15: But it's a default value?

vladimirpotekhin: Yes, the bug appears under the very common global CSS 

```
*, *::before, *::after {
box-sizing: border-box;
}
```

so we have to specify it explicitly

PR #14624, opened 2026-07-24
fix(core): `DropdownContext` fix click on host element (#11400)

### LINKED ISSUE #11400: 🐞 - Dropdown Context behaviour in click outside menu
### Reproduction url

https://taiga-ui.dev/directives/dropdown-context#context-menu

### Description

The default behavior of the context menu is that when you click outside the menu, it closes. At the moment, if you click inside the block on which DropdownContext is attached, the menu does not close.

Reproduced in "Context menu" example

### Taiga UI version

4.45.0

### Browser Used

Chrome

### OS Used

Linux

Yanduz:
Fixes #11400 
image

# Fix: TuiDropdownContext closes on host element click

## Problem

When using `TuiDropdownContext` directive (context menu triggered by right-click), the dropdown did not close when clicking on the host element itself (e.g., the table row with the directive).

**Behavior before fix:**
- ✅ Click outside the host — dropdown closes
- ✅ Click inside the dropdown menu — dropdown stays open (correct)
- ❌ Click on the host element — dropdown **does not close** (bug)

## Solution

Added an additional check to close the dropdown when the click occurs within the host element bounds.

The fix extends the existing close condition with an `OR` clause:
- Close if click is **outside the active zone** (existing behavior)
- **OR** if click is **inside the host element** (new behavior)

## Result

**Behavior after fix:**
- ✅ Click outside the host — dropdown closes
- ✅ Click inside the dropdown menu — dropdown stays open
- ✅ Click on the host element — dropdown closes (fixed)
- ✅ Click on nested dropdown — dropdown stays open

## Breaking Changes

None. This is a backward-compatible bug fix that only adds the missing close behavior without changing any existing functionality.


mdlufy (approved):

projects/core/directives/dropdown/dropdown-context.directive.ts
```diff
@@ -52,9 +57,13 @@ export class TuiDropdownContext extends TuiRectAccessor {
...
 
     protected closeDropdown(event?: Event): void {
+        const target = event ? tuiGetActualTarget(event) : null;
+
         if (
             !event ||
-            (this.driver.value && !this.activeZone.contains(tuiGetActualTarget(event)))
+            (this.driver.value &&
+                target &&
+                (!this.activeZone.contains(target) || this.el.contains(target)))
```
nsbarsukov, line 65: Could you add Playwright test, please ?

projects/demo-playwright/tests/kit/dropdown-context/dropdown-context.pw.spec.ts
```diff
@@ -52,4 +52,21 @@ test.describe('DropdownContext', () => {
...
     });
+
+    test('closes dropdown when clicking on host element', async ({page}) => {
+        const api = new TuiDocumentationPagePO(page);
+        const example = api.getExample('#context-menu');
+        const tr = example.locator('tr');
+
+        await example.scrollIntoViewIfNeeded();
+        await tr.nth(1).click({button: 'right', position: {x: 1, y: 1}});
+        await api.waitStableState();
```
nsbarsukov, line 64: Add an auto-retrying assertion to ensure that upcoming screenshot tests are actually ready to implement.

```suggestion
        cy.get('tui-dropdown').should('be.visible');
        await api.waitStableState();
```

projects/demo-playwright/tests/kit/dropdown-context/dropdown-context.pw.spec.ts
```diff
@@ -52,4 +52,21 @@ test.describe('DropdownContext', () => {
...
+
+        await expect.soft(example).toHaveScreenshot('06-dropdown-context.png');
+
+        await tr.nth(1).click({position: {x: 1, y: 1}});
+        await api.waitStableState();
```
nsbarsukov, line 70:
```suggestion
        await tr.nth(1).click({position: {x: 1, y: 1}});
        cy.get('tui-dropdown').should('not.exist');
        await api.waitStableState();
```

nsbarsukov (approved):

projects/demo-playwright/tests/kit/dropdown-context/dropdown-context.pw.spec.ts
nsbarsukov: My mistake, it is PW test, not Cypress

I meant
```ts
await expect(page.locator('tui-dropdown')).toBeHidden();
```

projects/demo-playwright/tests/kit/dropdown-context/dropdown-context.pw.spec.ts
nsbarsukov: My mistake, it is PW test, not Cypress

I meant
```ts
await expect(page.locator('tui-dropdown')).toBeAttached();
```

projects/core/directives/dropdown/dropdown-context.directive.ts
```diff
@@ -52,9 +57,13 @@ export class TuiDropdownContext extends TuiRectAccessor {
     }
 
     protected closeDropdown(event?: Event): void {
+        const target = event ? tuiGetActualTarget(event) : null;
+
         if (
             !event ||
```
waterplea: Can be simplified as !target since it is only falsy when there is no event and then we do not need to check && target in the second condition

waterplea (approved):

PR #14625, opened 2026-07-24
chore(schematics): warn on removed TUI_IS_APPLE/TUI_IS_FIREFOX and fix sibling token warnings in v5

mdlufy:
## Which issue does this PR close?

Part of #11917 (v4→v5 ng-update migration backlog).

## What does this PR do?

Adds `ng update` warnings for two removed platform-detection tokens and fixes the three already-shipped siblings that never fired on real code.

### New warnings

| Removed (v4, `@taiga-ui/legacy`) | Guidance |
| --- | --- |
| `TUI_IS_APPLE` | `isApple(inject(WA_NAVIGATOR))` — `isApple` from `@ng-web-apis/platform`, `WA_NAVIGATOR` from `@ng-web-apis/common` |
| `TUI_IS_FIREFOX` | `isFirefox(inject(WA_NAVIGATOR).userAgent)` — `isFirefox` from `@ng-web-apis/platform`, `WA_NAVIGATOR` from `@ng-web-apis/common` |

### Fixes to existing warnings

All five of these tokens were re-exported from `@taiga-ui/legacy` in v4, not `@taiga-ui/cdk`. The three shipped warnings keyed on `@taiga-ui/cdk`, so the exact-match `moduleSpecifier` check in `show-warnings.ts` never matched real imports.

| Token | Before | After |
| --- | --- | --- |
| `TUI_TOUCH_SUPPORTED` | `@taiga-ui/cdk`; suggested non-existent `WA_TOUCH_SUPPORTED` | `@taiga-ui/legacy`; suggests `WA_IS_TOUCH` |
| `TUI_IS_CHROMIUM` | `@taiga-ui/cdk`; suggested non-existent `WA_IS_CHROMIUM` | `@taiga-ui/legacy`; no direct replacement — manual `"chrome" in window` |
| `TUI_IS_STACKBLITZ` | `@taiga-ui/cdk` | `@taiga-ui/legacy` (message unchanged) |

## Notes

`@ng-web-apis/platform` has no `WA_IS_APPLE`, `WA_IS_FIREFOX`, or `WA_IS_CHROMIUM` — the guidance uses the functions/tokens that actually exist. Verified against `origin/v4.x` source and the installed `@ng-web-apis/platform` typings.


vladimirpotekhin (approved):

projects/cdk/schematics/ng-update/v5/steps/constants/migration-warnings.ts
```diff
@@ -185,22 +185,34 @@ export const MIGRATION_WARNINGS: MigrationWarning[] = [
     },
     {
         name: 'TUI_TOUCH_SUPPORTED',
-        moduleSpecifier: '@taiga-ui/cdk',
+        moduleSpecifier: '@taiga-ui/legacy',
         message:
-            'TUI_TOUCH_SUPPORTED has been removed. Use WA_TOUCH_SUPPORTED from @ng-web-apis/platform package instead.',
+            'TUI_TOUCH_SUPPORTED has been removed. Use WA_IS_TOUCH from @ng-web-apis/platform package instead.',
     },
```
nsbarsukov, line 191: Why `projects/cdk/schematics/ng-update/v5/steps/constants/identifiers-to-replace.ts` is not used for this replacement ? 🤔

nsbarsukov (approved):

projects/cdk/schematics/ng-update/v5/steps/constants/migration-warnings.ts
mdlufy: `IDENTIFIERS_TO_REPLACE` only does a 1:1 identifier rename + import swap, so it's safe only when the replacement is API-compatible. Here it isn't: `TUI_TOUCH_SUPPORTED` was an `InjectionToken<boolean>`, while `WA_IS_TOUCH` is an `InjectionToken<Signal<boolean>>`. A bare rename would leave every call site treating a signal as a boolean (e.g. `if (inject(TUI_TOUCH_SUPPORTED))` — the signal object is always truthy), silently breaking runtime behavior, and the schematic can't add the required `()` call at usage sites.

So I kept it as a warning and made the message spell out the difference instead:
> TUI_TOUCH_SUPPORTED has been removed. Use WA_IS_TOUCH from @ng-web-apis/platform instead — note it is a Signal<boolean> (the old token was a plain boolean), so read it with a call: inject(WA_IS_TOUCH)().

The nearby 1:1 renames that *are* API-compatible were moved to `IDENTIFIERS_TO_REPLACE` (see #14643 / #14648).

Issue #14341, opened 2026-06-16
🐞 - Signal Form


Stivin:
### Reproduction url

https://stackblitz.com/edit/stackblitz-starters-6ifbbepn?file=src%2Fapp%2Fapp.html,src%2Fapp%2Fapp.ts

### Description

Hi! There are several issues when using most components with Signal Form. I've converted the TaigaUI form example to use Signal Form, and for convenience, I've also set up a separate repository where I've implemented most of the form components. I had to comment out a lot of code due to errors.
When defining validation, an invalid field immediately shows the error visually. The tui-error component does not work with formField.

[Repository](https://github.com/Stivin/angular-taiga-ui-signal-form)

### Taiga UI version

5.10.0

### Browser Used

Chrome

### OS Used

macOS

Gugge1999:
I also want this improved since Angular 22 has promoted signal form to stable, making it the recommended approach going forward.

riccardoperra:
@waterplea https://github.com/angular/angular/pull/67087 that was something I did for syncing control directive within signals, but still not merged. my workplace design system library followed an approach like your, we then opted for creating our ngcontrol wrapper when formSignals was present in the node injector

Idk if this can be useful, but that's something related of the "FakeControl" provided via formSignals doesn't respect the entire NgControl public interface

waterplea:
We put this issue on the radar for this quarter as a priority.

riccardoperra:
https://github.com/angular/angular/issues/69678 as requested by an angular core team member I've opened an issue in angular repository to understand better this issue

r-arteaga:
Ran into this too on Angular 22 with signal forms, sharing a quick workaround for `tuiSelect` since it took me a bit to figure out what was going on.

The confusing part is the value actually binds with `[formField]`. The closed field shows the right label, so at first glance it looks fine. It's the dropdown that comes up empty, the `tui-data-list` renders with no options at all (just the container comments in the DOM). The exact same template on `[(ngModel)]` or `[formControl]` fills it normally, so the select seems to need something off `NgControl` that the signal forms interop control doesn't expose yet. Matches what's described in angular/angular#69678.

For a quick unblock, I keep the signal form as the source of truth but drive the select through `ngModel` against the model signal instead of `formField`:

```html
<input tuiSelect [ngModel]="model().kind" (ngModelChange)="setKind($event)" />
```
```ts
setKind(v: Kind) {
  this.model.update(m => ({ ...m, kind: v }));
[...]

PR #14629, opened 2026-07-24
feat(core): `Error` support signal forms

nsbarsukov:
Partially solves:

### DIFF
--- a/projects/cdk/classes/form-value-control.ts
+++ b/projects/cdk/classes/form-value-control.ts
@@ -1,15 +1,27 @@
-import {InputSignal, InputSignalWithTransform, ModelSignal, OutputRef} from "@angular/core";
+import {
+    type InputSignal,
+    type InputSignalWithTransform,
+    type ModelSignal,
+    type OutputRef,
+} from '@angular/core';
 
-interface ValidationError {
+/**
+ * @deprecated use import {ValidationError} from '@angular/forms/signals'
+ * Just temporary copy-pasted types until Taiga UI supports Angular <22
+ * https://github.com/angular/angular/blob/main/packages/forms/signals/src/api/rules/validation/validation_errors.ts#L368
+ * TODO: replace all usages `TuiSignalValidationError` by built-in Angular alternative
+ */
+export interface TuiSignalValidationError {
     readonly kind: string;
     readonly message?: string;
 }
 
+// TODO: delete
 interface FormUiControl {
     readonly errors?:
-        | InputSignal<readonly ValidationError[]>
-        | InputSignalWithTransform<readonly ValidationError[], unknown>;
-    
+        | InputSignal<readonly TuiSignalValidationError[]>
+        | InputSignalWithTransform<readonly TuiSignalValidationError[], unknown>;
+
     readonly disabled?: InputSignal<boolean> | InputSignalWithTransform<boolean, unknown>;
     readonly readonly?: InputSignal<boolean> | InputSignalWithTransform<boolean, unknown>;
     readonly invalid?: InputSignal<boolean> | InputSignalWithTransform<boolean, unknown>;
--- a/projects/cdk/classes/form-value-control.ts
+++ b/projects/cdk/classes/form-value-control.ts
@@ -1,0 +1,34 @@
+import {InputSignal, InputSignalWithTransform, ModelSignal, OutputRef} from "@angular/core";
+
+interface ValidationError {
+    readonly kind: string;
+    readonly message?: string;
+}
+
+interface FormUiControl {
+    readonly errors?:
+        | InputSignal<readonly ValidationError[]>
+        | InputSignalWithTransform<readonly ValidationError[], unknown>;
+    
+    readonly disabled?: InputSignal<boolean> | InputSignalWithTransform<boolean, unknown>;
+    readonly readonly?: InputSignal<boolean> | InputSignalWithTransform<boolean, unknown>;
+    readonly invalid?: InputSignal<boolean> | InputSignalWithTransform<boolean, unknown>;
+    readonly pending?: InputSignal<boolean> | InputSignalWithTransform<boolean, unknown>;
+    readonly touched?: InputSignal<boolean> | InputSignalWithTransform<boolean, unknown>;
+    readonly dirty?: InputSignal<boolean> | InputSignalWithTransform<boolean, unknown>;
+
+    readonly touch?: OutputRef<void>;
+
+    focus?(options?: FocusOptions): void;
+    reset?(): void;
+}
+
+/**
+ * @deprecated use import {FormValueControl} from '@angular/forms/signals'
+ * Just temporary copy-pasted types until Taiga UI supports Angular <22
+ * https://github.com/angular/angular/blob/main/packages/forms/signals/src/api/control.ts
+ * TODO: replace all usages `TuiFormValueControl` by built-in Angular alternative
+ */
+export interface TuiFormValueControl<TValue> extends FormUiControl {
+    readonly value: ModelSignal<TValue>;
+}
--- a/projects/cdk/classes/form-value-control.ts
+++ b/projects/cdk/classes/form-value-control.ts
@@ -16,7 +16,7 @@ export interface TuiSignalValidationError {
     readonly message?: string;
 }
 
-// TODO: delete
+// TODO: delete when all usages of `TuiFormValueControl` will be replaced by built-in Angular alternative
 interface FormUiControl {
     readonly errors?:
         | InputSignal<readonly TuiSignalValidationError[]>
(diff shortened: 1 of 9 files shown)


projects/demo-cypress/src/tests/error/signal-forms.cy.ts
```diff
@@ -0,0 +1,208 @@
+/*
+// TODO: Uncomment the whole file when the `@angular/forms/signals` entry point becomes available,
```
nsbarsukov, line 2: All tests are passed are passing in Angular 22

projects/cdk/directives/native-validator/native-validator.directive.ts
```diff
@@ -25,11 +31,24 @@ export class TuiNativeValidator implements Validator {
...
     public readonly tuiNativeValidator = input('Invalid');
     public id = '';
 
     public get control(): AbstractControl | null {
-        return this.control$.value;
+        return (
+            this.control$.value ??
+            // `validate` method is never called in signal forms => this.control$.value is always null
+            // https://github.com/angular/angular/issues/66232#issuecomment-5069162406
+            this.injector.get(NgControl, null, {optional: true, self: true})?.control ??
```
waterplea: Maybe we can use the same approach everywhere and just use ngOnInit with injector and nexting NgControl, rather than relying on validate method? Was the old way used just for cycling dependency avoidance?

vladimirpotekhin (approved):

projects/cdk/directives/native-validator/native-validator.directive.ts
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/14629/commits/ac0ebdb7e74ab0300d8133a14453e1d74c15d223

waterplea: Please wait for me on this PR to take a careful look

Yanduz (approved):

projects/cdk/classes/form-value-control.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Just temporary copy-pasted types until Taiga UI supports Angular <22
+ * https://github.com/angular/angular/blob/main/packages/forms/signals/src/api/rules/validation/validation_errors.ts#L368
+ * TODO: replace all usages `TuiSignalValidationError` by built-in Angular alternative
+ */
+export interface TuiSignalValidationError {
+    readonly kind: string;
+    readonly message?: string;
+}
+
+// TODO: delete
```
mdlufy: nit: we can specify to delete when/after what ?

projects/core/components/error/error.directive.ts
```diff
@@ -104,32 +145,62 @@ export class TuiErrorDirective implements ControlValueAccessor, Validator {
...
-            return getError(context);
+@Directive({
+    selector: 'tui-error[formField]',
+    host: {'(document:tui-validator.zoneless)': 'onValidator($event.detail)'},
+})
+export class TuiErrorField
+    extends AbstractTuiErrorDirective
+    implements TuiFormValueControl<unknown>
+{
+    // TODO: use inject(FORM_FIELD) from `@angular/forms/signals`
```
mdlufy: nit: when we can use FORM_FIELD? after angular 22 ?

projects/demo-cypress/src/tests/error/signal-forms.cy.ts
mdlufy: Are we sure we want to keep it in such a commented form? 

If other components need the same tests, it may be worth creating a separate folder like `signal-form-tests-v22` and storing and adding this folder to the list of exceptions in the cypress config, so that if anything, you can simply remove the entire folder from the exceptions, without having to edit the test file.

projects/demo-cypress/src/tests/native-validator/signal-forms.cy.ts
```diff
@@ -0,0 +1,180 @@
+/*
+// TODO: Uncomment the whole file when the `@angular/forms/signals` entry point becomes available,
```
mdlufy, line 2: Similarly to https://github.com/taiga-family/taiga-ui/pull/14629/changes#r3664730137

mdlufy (approved):

projects/core/components/error/error.directive.ts
```diff
@@ -39,24 +45,66 @@ import {
...
+    protected abstract readonly error: Signal<TuiValidationError | null>;
+    protected readonly el = tuiInjectElement();
+    protected readonly content = inject(TUI_VALIDATION_ERRORS);
+
+    public readonly order = input<readonly string[]>([]);
+
+    protected readonly $ = tuiDirectiveBinding(
+        TuiErrorComponent,
+        'error',
+        computed(() => this.error()),
```
waterplea, line 62: Unnecessary computed wrap, can just be this.error signal

projects/core/components/error/error.directive.ts
```diff
@@ -104,32 +145,62 @@ export class TuiErrorDirective implements ControlValueAccessor, Validator {
...
+export class TuiErrorField
+    extends AbstractTuiErrorDirective
+    implements TuiFormValueControl<unknown>
+{
+    // TODO: use inject(FORM_FIELD) from `@angular/forms/signals`
+    // @ts-expect-error
+    private readonly field = inject(NgControl, {self: true}).field;
+
+    public readonly errors = input<readonly TuiSignalValidationError[]>([]);
+    public readonly disabled = input(false);
```
nsbarsukov: Remove `disable`

projects/core/components/error/error.directive.ts
```diff
@@ -104,32 +145,62 @@ export class TuiErrorDirective implements ControlValueAccessor, Validator {
...
+})
+export class TuiErrorField
+    extends AbstractTuiErrorDirective
+    implements TuiFormValueControl<unknown>
+{
+    // TODO: use inject(FORM_FIELD) from `@angular/forms/signals`
+    // @ts-expect-error
+    private readonly field = inject(NgControl, {self: true}).field;
+
+    public readonly errors = input<readonly TuiSignalValidationError[]>([]);
```
waterplea, line 162: Are these required to be inputs? Would confuse people with autocomplete...

waterplea (approved):

projects/core/components/error/error.directive.ts
nsbarsukov:
```
TS2715: Abstract property error in class AbstractTuiErrorDirective cannot be accessed in the constructor.
```

projects/core/components/error/error.directive.ts
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/14629/commits/1090d6a5c07093d7828965d58127034679b09a97

projects/cdk/classes/form-value-control.ts
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/14629/commits/1090d6a5c07093d7828965d58127034679b09a97

projects/demo-cypress/src/tests/error/signal-forms.cy.ts
nsbarsukov: We cannot store it without commenting because `@angular/forms/signals` entry point is not available in Angular 19.
It will produce Typescript error

projects/demo-cypress/src/tests/native-validator/signal-forms.cy.ts
nsbarsukov: Same comment https://github.com/taiga-family/taiga-ui/pull/14629#discussion_r3713643557

projects/core/components/error/error.directive.ts
nsbarsukov: Unfortunately, it is required to implement `FormValueControl` (signal-based CVA)

```ts
export interface FormValueControl<TValue> extends FormUiControl<TValue> {...}
```

```ts
export interface FormUiControl<TValue> {
  /**
   * An input to receive the errors for the field. If implemented, the `Field` directive will
   * automatically bind errors from the bound field to this input.
   */
  readonly errors?:
    | InputSignal<readonly ValidationError.WithOptionalFieldTree[]>
    | InputSignalWithTransform<readonly ValidationError.WithOptionalFieldTree[], unknown>;
    // No `WritableSignal<readonly ValidationError.WithOptionalFieldTree[]>` option :(

    // [...]
}
```

https://github.com/angular/angular/blob/337053ef1ab6400c3641984511b45c341d306e87/packages/forms/signals/src/api/control.ts#L160

https://github.com/angular/angular/blob/337053ef1ab6400c3641984511b45c341d306e87/packages/forms/signals/src/api/control.ts#L22-L29

PR #14630, opened 2026-07-24
feat(addon-mobile): `TabBar` add liquid glass support

vladimirpotekhin:
Fixes #

### DIFF
--- a/projects/addon-mobile/styles/ios/tab-bar.less
+++ b/projects/addon-mobile/styles/ios/tab-bar.less
@@ -1,0 +1,161 @@
+@t-tab-width: 5.875rem;
+
+&.tui-liquid-glass {
+    nav[tuiTabBar] {
+        justify-content: center;
+        isolation: isolate;
+        inline-size: fit-content;
+        margin-inline: auto;
+        margin-block-end: var(--t-tabbar-offset);
+        padding: 0.25rem;
+        border-radius: 62.5rem;
+        block-size: 3.375rem;
+        backdrop-filter: none;
+
+        --t-tabbar-offset: max(env(safe-area-inset-bottom), 1.25rem);
+        --t-tabbar-bg: rgba(255, 255, 255, 0.7);
+        --t-tabbar-glow: rgba(255, 255, 255, 1);
+        --t-tabbar-active: rgba(255, 255, 255, 0.9);
+        --t-tabbar-bevel: linear-gradient(227deg, rgba(255, 255, 255, 0) 6%, rgba(255, 255, 255, 0.35) 21%, rgba(255, 255, 255, 0.85) 40%, rgba(255, 255, 255, 0.95) 45%, #fff 48%, #fff 50%, #fff 52%, rgba(255, 255, 255, 0.95) 55%, rgba(255, 255, 255, 0.85) 60%, rgba(255, 255, 255, 0.35) 80%, rgba(255, 255, 255, 0) 95%);
+
+        &::before {
+            content: '';
+            position: absolute;
+            inset: 0;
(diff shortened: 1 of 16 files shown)


projects/addon-mobile/styles/ios/tab-bar.less
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        .t-skeleton {
+            flex: none;
+            inline-size: @t-tab-width;
+        }
+
+        &.tui-liquid-glass_fullwidth {
+            inline-size: auto;
+            margin-inline: 1.25rem;
```
nsbarsukov, line 93: I have concerns about including `margin` (this line + the `margin-block` above) as part of the component's internal styles. 

Traditionally, margin is something that Taiga UI consumers manage themselves. 
Should we document these margins as recommended spacing instead of shipping them as built-in styles? 🤔

nsbarsukov: For example, a user might set `inline-size: 100%` expecting the component to stretch to full width, but would unexpectedly run into these built-in margins

https://github.com/user-attachments/assets/7af062b4-25ed-4676-9bdd-45506eb95c72

projects/demo/src/pages/components/tab-bar/examples/5/index.ts
```diff
@@ -0,0 +1,56 @@
...
+
+@Component({
+    selector: 'tui-tab-bar-example-liquid',
+    imports: [TuiTabBar],
+    templateUrl: './index.html',
+    styleUrl: './index.less',
+    encapsulation,
+    changeDetection,
+    providers: [
+        {provide: TUI_LIQUID_GLASS, useValue: true},
```
nsbarsukov, line 23: Should we add comment ?
Something like


```suggestion
        // Not required if `provideTaiga({apis: {liquidGlass: true}})` is already set up
        {provide: TUI_LIQUID_GLASS, useValue: true},
```

projects/addon-mobile/styles/ios/tab-bar.less
```diff
@@ -0,0 +1,161 @@
...
+                );
+            }
+        }
+
+        .t-liquid-glass-overlay {
+            position: absolute;
+            z-index: -1;
+            inset-block: auto calc(-1 * var(--t-tabbar-offset));
+            inset-inline-start: 50%;
+            margin-inline-start: -50vw;
```
nsbarsukov, line 126: Could you explain what this trick does?

nsbarsukov (approved):

projects/addon-mobile/styles/ios/tab-bar.less
vladimirpotekhin: I understand your concerns, but if we leave the indentation on the user side, they'll always have to check the platform and the liquid option on their side (because the old version and the Android version work just fine without extra code). I think this will complicate basic usage.

waterplea (commented):
Please wait for me to take a proper look

projects/demo/src/pages/components/tab-bar/examples/6/index.ts
```diff
@@ -0,0 +1,49 @@
...
+
+@Component({
+    selector: 'tui-tab-bar-example-liquid-full',
+    imports: [TuiTabBar],
+    templateUrl: './index.html',
+    styleUrl: './index.less',
+    encapsulation,
+    changeDetection,
+    providers: [
+        {provide: TUI_LIQUID_GLASS, useValue: true},
```
waterplea, line 17: Should we have the same comment here?

waterplea (approved):

PR #14651, opened 2026-07-28
chore(schematics): migrate remaining v5 leftovers (renames, accordion, CSS vars + font-heading fix)

mdlufy:
## What / Why

Completes the **must-have** part of the v4→v5 public-API audit (ref #11917): leftover symbols/variables that a real user actually imports and that currently get **no** `ng update` guidance. Includes a genuine **bug fix** in the CSS-variable step.

Each entry was individually verified: public in v4 (top-level barrel), gone in v5, correct successor, and — for the auto-renames — a like-for-like symbol kind with no import collision.

## 1. Auto-renames → `IDENTIFIERS_TO_REPLACE`

| From (v4) | To (v5) | Kind |
| --- | --- | --- |
| `TuiDropdownPortal` (`@taiga-ui/core`) | `TuiDropdownContent` (`@taiga-ui/core`) | directive class |
| `TuiWithTextfield` (`@taiga-ui/core`) | `TuiWithInput` (`@taiga-ui/core`) | host-directive class |
| `tuiStringHashToHsl` (`@taiga-ui/core`) | `tuiAutoColor` (`@taiga-ui/kit`) | `(value: string) => string` (identical signature) |
| `tuiSwitchNgDevMode` (`@taiga-ui/testing`) | `tuiSwitchNgDevMode` (`@taiga-ui/jest-config`) | test helper, package move |

## 2. Removed-symbol warnings → `MIGRATION_WARNINGS`

`@taiga-ui/kit` accordion redesign (`<tui-accordion-item>` → restructured `TuiAccordion`):

| Symbol | Guidance |
| --- | --- |
| `TuiAccordionItem` | Declare each item as `<button tuiAccordion>` + `<tui-expand>`; render dynamic lists with a native `@for` block |
| `TuiAccordionItemContent` | Lazy content — wrap it in `<ng-container *tuiItem>` inside the `<tui-expand>` |
| `TuiAccordionItemEagerContent` | Eager content lives directly inside the `<tui-expand>` |

## 3. CSS variables → `migrate-css-variables`

**Bug fix:** the bare `--tui-font-heading` (heading font-family) was silently left un-migrated — the existing regexes only match the numbered `--tui-font-heading-[1-6]`, while the analogous bare `--tui-font-text` token *is* handled. Added the missing bare mapping.

| Variable | → |
| --- | --- |
| `--tui-font-heading` | `--tui-typography-family-display` *(the fix)* |
| `--tui-backdrop` | `--tui-service-backdrop` |
| `--tui-animation-duration-moderate` | `--tui-duration` |
| `--tui-animation-duration-slow` | `--tui-duration-slow` |
| `--tui-animation-curve-expressive-entrance` | `--tui-curve-expressive-entrance` |
| `--tui-slider-track-color` | *removed* → TODO comment (no replacement) |

## Tests

- Snapshot spec `schematic-migrate-remaining-leftovers.spec.ts` — renames / accordion warnings / CSS. The CSS case keeps a numbered `--tui-font-heading-1` as a regression guard so the bug fix can't break the existing numbered mapping.
- Full `ng-update/v5` schematics suite green.

## Audit note

One originally-considered rename was **intentionally dropped**: `TuiInputChipBaseDirective → TuiInputChipDirective`. In v4 both names existed (base class + its concrete subclass), so an auto-rename risks a duplicate import and silently changes what a subclass extends — unsafe for a mechanical remap, and it's a niche base class.


projects/cdk/schematics/ng-update/v5/steps/constants/migration-warnings.ts
```diff
@@ -297,4 +297,22 @@ export const MIGRATION_WARNINGS: MigrationWarning[] = [
...
+        name: 'TuiAccordionItem',
+        moduleSpecifier: '@taiga-ui/kit',
+        message:
+            'TuiAccordionItem (<tui-accordion-item>) has been removed. The accordion was restructured: use TuiAccordion with a tui-expand-based item template from @taiga-ui/kit. See https://taiga-ui.dev/components/accordion',
+    },
+    {
+        name: 'TuiAccordionItemContent',
+        moduleSpecifier: '@taiga-ui/kit',
+        message:
+            'TuiAccordionItemContent ([tuiAccordionItemContent]) has been removed. Lazy accordion content is now projected differently in the restructured TuiAccordion. See https://taiga-ui.dev/components/accordion',
```
nsbarsukov: Let's provide more details (tuiItem) or provide anchor to specific example

mdlufy: Done — expanded the message with concrete steps: declare each item as `<button [tuiAccordion]="open">header</button>` followed by `<tui-expand>content</tui-expand>`, and iterate items with the `*tuiItem` structural directive (plus the accordion doc anchor).

nsbarsukov (approved):

vladimirpotekhin (approved):

PR #14655, opened 2026-07-28
feat(styles): mixin-based theme for scoped/micro frontend usage

vladimirpotekhin:
Part of #10435

## What does this PR do?

Adds `.taiga-ui-theme()` — a mixin entry point for the theme, so it can be applied inside an
arbitrary selector instead of `:root` / `:host`.

Today styles are connected through global entry points:

```less
@import '@taiga-ui/styles/taiga-ui-theme.less';
@import '@taiga-ui/styles/taiga-ui-fonts.less';
@import '@taiga-ui/addon-mobile/styles/taiga-ui-mobile.less'; // optional
```

That is fine for a regular application, but not for micro frontends: `taiga-ui-theme.less` emits
`:root` / `:host` rules, so several micro frontends on one page fight over the same custom
properties and there is no way to isolate them.

With this PR a micro frontend can scope the theme to itself:

```less
@import '@taiga-ui/styles/mixins/theme.less';

[my-app] {
    .taiga-ui-theme();

    // mobile styles have no `:root` / `:host`, so a plain nested import is enough
    @import '@taiga-ui/addon-mobile/styles/taiga-ui-mobile.less';
}
```

Everything the global entry point would put on `:root` / `:host` now lands on `[my-app]`, and
everything it would put on a bare selector is nested under it.

## Notes on `@import` deduplication

`@import` in Less is `once` by default and dedupes by file path, not by insertion point. A few
consequences worth knowing:

**`.taiga-ui-theme()` is not affected — several scopes in one build just work:**

```less
[app-a] { .taiga-ui-theme(); }
[app-b] { .taiga-ui-theme(); }
```

Both get the complete rule set (verified: `accent` = 4 rules in each). The nested `@import` lives
inside the mixin body, so it is resolved once at parse time and the resulting rules are replayed on
every call — dedup never comes into play.

**A raw nested `@import` written twice is affected:**

```less
[app-a] { @import '@taiga-ui/addon-mobile/styles/taiga-ui-mobile.less'; }
[app-b] { @import '@taiga-ui/addon-mobile/styles/taiga-ui-mobile.less'; }
```

`app-a` gets 443 rules, `app-b` gets **0**, silently. If you really need the same file scoped twice
in one compilation, `@import (multiple)` fixes it (449 / 449) — but use it deliberately: `multiple`
propagates down the **whole import subtree**, so files that only define mixins get re-parsed too.
Less then runs *every* same-named definition, and nested calls multiply, so the duplication is
`N^depth`. 

**Do not mix the global and the scoped path in a single Less compilation:**

```less
@import '@taiga-ui/styles/taiga-ui-theme.less';
@import '@taiga-ui/styles/mixins/theme.less';

[my-app] { .taiga-ui-theme(); }
```

Both pull in `theme/appearance.less`, so whichever comes second emits nothing. Separate builds — the
usual micro frontend setup, and also how Angular compiles global styles vs component styles — are
unaffected.

### DIFF
--- a/projects/styles/mixins/theme.less
+++ b/projects/styles/mixins/theme.less
@@ -1,0 +1,67 @@
+@import (once) '@taiga-ui/design-tokens/angular/mixins/animation.less';
+@import (once) '@taiga-ui/design-tokens/angular/mixins/shadow.less';
+@import 'theme/variables.less';
+@import 'theme/palette.less';
+
+// Scoped counterpart of `taiga-ui-theme.less`. Call it inside your own selector to keep
+// the theme out of `:root` / `:host`, e.g. when several micro frontends share a page:
+//
+//     [my-app] {
+//         .taiga-ui-theme();
+//     }
+.taiga-ui-theme() {
+    .tui-animation();
+    .tui-shadow-light();
+    .tui-theme-variables();
+
+    &[dir='rtl'],
+    [dir='rtl'] & {
+        .tui-theme-rtl();
+    }
+
+    @media screen {
+        & {
+            .tui-theme-light();
+        }
+
+        &[tuiTheme='light'],
+        & [tuiTheme='light'] {
+            color-scheme: light;
+
+            .tui-theme-light();
+        }
+
+        &[tuiTheme='dark'],
+        & [tuiTheme='dark'] {
+            color-scheme: dark;
+
+            .tui-theme-dark();
+        }
+    }
+
+    @media print {
+        &,
+        &[tuiTheme],
+        & [tuiTheme] {
+            color-scheme: light;
+
+            .tui-theme-light();
+        }
+    }
+
+    @import '../theme/appearance.less';
+    @import '../theme/appearance/textfield.less';
+
+    // `tuiTheme` can sit on the scoping element itself, which the selectors above do not cover
+    &[tuiTheme='dark'] [tuiAppearance][data-appearance='textfield'] {
+        background-color: var(--tui-background-neutral-1);
+
+        .appearance-hover({
+            background-color: var(--tui-background-neutral-1-hover);
+        });
+
+        .appearance-focus({
+            background-color: transparent !important;
+        });
+    }
+}
--- a/projects/styles/theme/variables.less
+++ b/projects/styles/theme/variables.less
@@ -1,69 +1,13 @@
 @import (inline, once) '@taiga-ui/design-tokens/palette/animation.css';
 @import (inline, once) '@taiga-ui/design-tokens/palette/shadow.css';
 
+@import '../mixins/theme/variables.less';
+
 &:root,
 &:host {
-    // Fonts
-    --tui-typography-family-text: 'Manrope', system-ui, sans-serif;
-    --tui-typography-family-display: 'Manrope', system-ui, sans-serif;
-    --tui-font-offset: calc(var(--t-font-offset, 0) * 1px);
-    --tui-font-scale: calc((22 + var(--t-font-offset, 0)) / 22);
-    // Heading
-    --tui-typography-heading-h1: bold calc(var(--tui-font-offset) + 3.125rem) / calc(56 / 50) var(--tui-typography-family-display);
-    --tui-typography-heading-h2: bold calc(var(--tui-font-offset) + 2.75rem) / calc(48 / 44) var(--tui-typography-family-display);
-    --tui-typography-heading-h3: bold calc(var(--tui-font-offset) + 2.25rem) / calc(40 / 36) var(--tui-typography-family-display);
-    --tui-typography-heading-h4: bold calc(var(--tui-font-offset) + 1.75rem) / calc(32 / 28) var(--tui-typography-family-display);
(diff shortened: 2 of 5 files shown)


projects/addon-mobile/styles/taiga-ui-mobile.less
```diff
@@ -1,61 +1,3 @@
...
-    @import 'common/notification-middle.less';
-    @import 'common/toast.less';
-
-    tui-app-bar {
-        [tuiButton][data-size='l'] {
-            font-weight: normal;
-        }
-    }
-}
+.taiga-ui-mobile();
```
nsbarsukov: What's the difference between:

```less
[my-app] {
    @import '@taiga-ui/styles/taiga-ui-theme.less';
}
```

and:

```less
@import '@taiga-ui/addon-mobile/styles/mixins/mobile.less';

[my-app] {
    .taiga-ui-mobile();
}
```
?

Do they produce the same result? If so, do we really need the new `.taiga-ui-mobile()` mixin?

projects/styles/mixins/fonts.less
```diff
@@ -0,0 +1,6 @@
+// Counterpart of `taiga-ui-fonts.less`. Unlike `.taiga-ui-theme()` it exists only for a uniform
+// API: call it at the root, `@font-face` is global by nature and escapes any scoping selector
+.taiga-ui-fonts() {
+    // stylelint-disable-next-line -- https://github.com/stackblitz/core/issues/2104#issuecomment-1397086512
+    @import (optional) url(https://fonts.googleapis.com/css2?family=Manrope:wght@500;800&display=swap);
+}
```
nsbarsukov: I have doubts that we really need it, even for the sake of API uniformity.

projects/styles/theme/variables.less
```diff
@@ -1,69 +1,13 @@
 @import (inline, once) '@taiga-ui/design-tokens/palette/animation.css';
 @import (inline, once) '@taiga-ui/design-tokens/palette/shadow.css';
```
nsbarsukov, line 2: Should mixins used here too ?

```less
@import '@taiga-ui/design-tokens/angular/mixins/animation.less';
@import '@taiga-ui/design-tokens/angular/mixins/shadow.less';

// [...]

&:root,
&:host {
    .tui-theme-variables();
    .tui-animation();
    .tui-shadow-light();
}
```

projects/styles/theme/appearance/accent.less
```diff
@@ -1,14 +1,3 @@
...
-
-    .appearance-hover({
-        background: var(--tui-background-accent-2-hover);
-    });
-
-    .appearance-active({
-        background: var(--tui-background-accent-2-pressed);
-    });
-}
+.tui-appearance-accent();
```
nsbarsukov: Do all appearances really need to be mixins if there's no `:host` / `:root` inside their files?

Appearances could already be scoped by a selector when needed, even without mixins:

```less
child-app {
    @import 'appearance/accent.less';
}

// Same result — is there any advantage to using the mixin below?
child-app {
    .tui-appearance-accent();
}
```

nsbarsukov (commented):
Do we really need mixin if `@import '@taiga-ui/styles/**/any-file.less'` does not include any `:root` selector ? 🤔

projects/styles/theme/variables.less
vladimirpotekhin: No, this is old theme, `@import (inline, once) '@taiga-ui/design-tokens/palette/shadow.css';` - it contains `:root, :host` already

projects/styles/mixins/theme.less
```diff
@@ -0,0 +1,67 @@
+@import (once) '@taiga-ui/design-tokens/angular/mixins/animation.less';
+@import (once) '@taiga-ui/design-tokens/angular/mixins/shadow.less';
```
nsbarsukov, line 2: They contain `:host`, `:root` . Is it expected ?

projects/styles/mixins/theme.less
```diff
@@ -0,0 +1,67 @@
...
+        &[tuiTheme],
+        & [tuiTheme] {
+            color-scheme: light;
+
+            .tui-theme-light();
+        }
+    }
+
+    @import '../theme/appearance.less';
+    @import '../theme/appearance/textfield.less';
```
nsbarsukov, line 53: Do we need `@import (multiple)` here ?

nsbarsukov: What will happen in the following case ?

```less
[my-app] {
  .taiga-ui-theme();
}

[my-app-2] {
  .taiga-ui-theme(); // without @import (multiple)
  // Still contain style from `theme/appearance.less` ?
}
```

projects/styles/mixins/theme.less
nsbarsukov: My mistake, it is not true

nsbarsukov (approved):

projects/styles/mixins/theme.less
vladimirpotekhin: Yes, in this case everything will be fine. I described it in more detail in the PR body. https://github.com/taiga-family/taiga-ui/pull/14655#issue-5000171163

mdlufy (approved):

PR #14683, opened 2026-08-03
chore(demo): `FloatingContainer` add compact trailing-action example

mdlufy:
## Description

Adds a **Compact** example to the `FloatingContainer` showcase, demonstrating a single action shrunk to its content and pinned to the trailing edge — instead of the default full-width, centered button (think of an iOS-style "Next" keyboard accessory).

By default `[tuiFloatingContainer]` lays its children out in a single-column grid with `justify-items: stretch`, so a button fills the whole width. Overriding `justify-self: end` on the button lets it size to its content and align to the trailing edge:

```html
<footer tuiFloatingContainer="transparent" class="bar">
    <button size="m" tuiButton type="button">Next</button>
</footer>
```

```less
.bar button {
    justify-self: end;
}
```

## Notes

- Pure demo change: new `examples/7/` component + registration on the page (`examples` array and a matching `@case` description). No library code touched.
- Lint clean: ESLint (TS + templates), stylelint, prettier.
- Draft: opened for review of the approach; a rendered screenshot can be added once the demo is served.

### DIFF
--- a/projects/demo/src/pages/components/floating-container/examples/7/index.less
+++ b/projects/demo/src/pages/components/floating-container/examples/7/index.less
@@ -8,11 +8,12 @@
     background: var(--tui-background-elevation-1);
 }
 
-.bar {
-    margin-inline: 1rem;
+.field {
+    margin: 1rem 1rem 0;
+}
 
+.action {
     // shrink the single action to its content and pin it to the trailing edge
-    button {
-        justify-self: end;
-    }
+    justify-self: end;
+    margin-inline: 1rem;
 }
--- a/projects/demo/src/pages/components/floating-container/examples/7/index.less
+++ b/projects/demo/src/pages/components/floating-container/examples/7/index.less
@@ -1,0 +1,18 @@
+.content {
+    position: relative;
+    display: block;
+    inline-size: 18rem;
+    block-size: 30rem;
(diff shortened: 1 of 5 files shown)


projects/demo/src/pages/components/floating-container/examples/7/index.less
```diff
@@ -0,0 +1,18 @@
...
+    position: relative;
+    display: block;
+    inline-size: 18rem;
+    block-size: 30rem;
+    overflow: auto;
+    box-shadow: 0 0.25rem 1.25rem rgba(0, 0, 0, 0.1);
+    background: var(--tui-background-elevation-1);
+}
+
+.bar {
```
waterplea: Unnecessary class, can just add class to button and put both justify-self and margin on button

waterplea (approved):

projects/demo/src/pages/components/floating-container/examples/7/index.less
mdlufy: Done — dropped the `.bar` wrapper, `justify-self` and margin now sit on the button. Also added an input above the pad so the button visibly rises with the keyboard.

vladimirpotekhin (approved):

PR #14688, opened 2026-08-03
feat(kit): confirm model no button appearance

chat-du-cheshire:
## Problem
Designers specify no button appearance. To fix this we need to override styles it makes code brittle.

## Solution
Allow config no button appearance with property


waterplea:
We do not intend to overblow this component with features, it's pretty easy to create your own confirm dialog with the visuals you want, the default one is intended for most common cases.

chat-du-cheshire:
> We do not intend to overblow this component with features, it's pretty easy to create your own confirm dialog with the visuals you want, the default one is intended for most common cases.

Don't you think it's too much to create new component to customize one button?

chat-du-cheshire:
Maybe it's possible to customize this button via provider ?

waterplea:
@chat-du-cheshire how about this?
https://github.com/taiga-family/taiga-ui/pull/14691

chat-du-cheshire:
> @chat-du-cheshire how about this? #14691

lgtm, but maybe it is possible to provide it globally ?

for example we provide confirm config once per app or route?

waterplea:
A different approach is now merged. @chat-du-cheshire if you want this in v4, please cherry pick my changes:
https://github.com/taiga-family/taiga-ui/pull/14691

PR #14689, opened 2026-08-03
feat(kit): v4 confirm modal reject button appearance 

chat-du-cheshire:
same for v4
https://github.com/taiga-family/taiga-ui/pull/14688


waterplea:
A different approach is now merged. @chat-du-cheshire if you want this in v4, please cherry pick my changes:
https://github.com/taiga-family/taiga-ui/pull/14691

PR #14693, opened 2026-08-04
fix(kit): respect animation duration in Pulse

### LINKED ISSUE #14599: 🗂 - Flaky `Pulse` Playwright test
### Description

Image

splincode:
Fixes #14599

### DIFF
--- a/projects/demo-playwright/animations.css
+++ b/projects/demo-playwright/animations.css
@@ -1,3 +1,4 @@
+tui-pulse._playing::before,
 tui-loader > .t-loader .t-icon,
 tui-loader > .t-loader .t-circle {
     animation: none !important;
--- a/projects/demo-playwright/animations.css
+++ b/projects/demo-playwright/animations.css
@@ -1,4 +1,3 @@
-tui-pulse._playing::before,
 tui-loader > .t-loader .t-icon,
 tui-loader > .t-loader .t-circle {
     animation: none !important;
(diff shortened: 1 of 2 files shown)


nsbarsukov (approved):

mdlufy (approved):

waterplea (approved):
I wonder if instead it should follow animation duration 🤔

splincode:
`tuiPulse` is infinite, so waiting for its duration would only move the screenshot to another animation cycle and remain timing-sensitive. That is why I disabled it in the screenshot only stylesheet. If we want to preserve an animated frame visually, I can instead pause it at a deterministic point in the cycle using `animation-play-state` and a negative `animation-delay`. Right?

Like that

```less
tui-pulse._playing::before {
    animation-delay: -500ms !important;
    animation-play-state: paused !important;
}
```

We could pause it at a deterministic point, but it seems like more code for no practical benefit.

waterplea:
I meant that maybe there should be no animation if --tui-duration is 0

splincode:
Do you mean replacing both the hardcoded duration and delay with `calc(var(--tui-duration) * 10 / 3)`, so the current 1s timing is preserved by default and the animation is effectively disabled when --tui-duration is 0?


```css
&::before {
    animation:
        tuiPulse
        calc(var(--tui-duration) * 10 / 3)
        calc(var(--tui-duration) * 10 / 3)
        ease-in-out
        infinite;
}
```

waterplea:
Yes, that's right

splincode:
Updated as suggested: `Pulse` now derives both the animation duration and delay from `--tui-duration` while preserving the default 1s timing. This also disables the animation when reduced motion sets `--tui-duration` to 0, so I removed the Playwright-specific override. Thanks!

PR #14698, opened 2026-08-05
fix(layout): support block content in `tuiList` items

splincode:
## Problem

`tuiList` does not correctly support direct `p` elements inside `li`.

Paragraph margins and whitespace before inline content can affect marker alignment. Positioning all markers outside the content flow also changes the existing spacing of nested ordered lists.

### DIFF
--- a/projects/demo/src/pages/components/list/examples/4/index.html
+++ b/projects/demo/src/pages/components/list/examples/4/index.html
@@ -1,8 +1,13 @@
 <ul tuiList>
     <li>If a list item is long, the bullet stays aligned to the top.</li>
     <li>There are no restrictions on the text length at all. You can write as much text as you want.</li>
-    <!-- prettier-ignore -->
-    <li>This item uses <strong>nested tags</strong> inside which should not break</li>
+    <li>
+        <p [style.margin]="0">
+            This item uses
+            <strong>nested tags</strong>
+            inside which should not break
+        </p>
+    </li>
 </ul>
 
 <ol tuiList>
--- a/projects/demo/src/pages/components/list/examples/4/index.html
+++ b/projects/demo/src/pages/components/list/examples/4/index.html
@@ -2,11 +2,11 @@
     <li>If a list item is long, the bullet stays aligned to the top.</li>
     <li>There are no restrictions on the text length at all. You can write as much text as you want.</li>
     <li>
-        <p [style.margin]="0">
+        <div>
             This item uses
             <strong>nested tags</strong>
             inside which should not break
-        </p>
+        </div>
     </li>
 </ul>
 
--- a/projects/layout/components/list/list.style.less
+++ b/projects/layout/components/list/list.style.less
@@ -10,6 +10,10 @@
         margin-inline-start: 1.25rem;
     }
 
+    ol[tuiList] {
+        margin-inline-start: calc(1.25rem + 1.5ch);
+    }
+
     & > li {
         counter-increment: item;
         margin: 0.5rem 0;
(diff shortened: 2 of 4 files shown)


projects/layout/components/list/list.style.less
```diff
@@ -57,4 +57,14 @@
             margin: 0.625rem 0;
         }
     }
+
+    & > li:has(> :where(p, div)) {
```
vladimirpotekhin: Why are these styles only for (p, div)? Can't we make them default for everything?

splincode: Making it the default would also affect list items with inline text followed by a nested list:

```html
<li>
  First level
  <ol tuiList>
    <li>Second level</li>
  </ol>
</li>
```

With absolute positioning, the marker is removed from the inline flow, which changes the existing spacing before the nested list. For a regular `<li>Text</li>` the visual difference is less noticeable, but its layout behavior also changes. That is why I limited the workaround to items starting with block content.

waterplea: We can instead add float to ::before, but there's something wrong with margins (same without block children and floating, looks like they are initially miscalculated somehow?):

Image

waterplea: Since we use `font-variant-numeric: tabular-nums;` we can try using `ch` units for margins, we can treat 1ch per digit and 0.5ch per dot and get a pretty decent approximation. Seems to be just 3px off when we got 1.2.3.4. in the list:

Image

waterplea: For unordered lists just adding float: inline-start to ::before is enough for both inline multiline text and block children:

Image

splincode: Could you merge [#14702](https://github.com/taiga-family/taiga-ui/pull/14702) so I can test with RTL?

splincode: @waterplea updated

projects/demo/src/pages/components/list/examples/4/index.html
```diff
@@ -1,8 +1,13 @@
 <ul tuiList>
     <li>If a list item is long, the bullet stays aligned to the top.</li>
     <li>There are no restrictions on the text length at all. You can write as much text as you want.</li>
-    <!-- prettier-ignore -->
-    <li>This item uses <strong>nested tags</strong> inside which should not break</li>
+    <li>
+        <p [style.margin]="0">
```
waterplea: Use div instead so there's no need to reset margin

projects/layout/components/list/list.style.less
```diff
@@ -10,6 +10,10 @@
         margin-inline-start: 1.25rem;
     }
 
+    ol[tuiList] {
+        margin-inline-start: calc(1.25rem + 1.5ch);
```
waterplea: When I suggested ch units I thought we would need to make it dynamic based on the number of digits in ::before. No need to use it now.

projects/layout/components/list/list.style.less
```diff
@@ -36,6 +46,14 @@
         margin-inline-end: 0.75rem;
     }
 
+    &:is(ol) > li::before {
+        position: relative;
+        inset-inline-end: 0.5625rem;
```
waterplea: This needs to be 0.75rem:

Image

This also means that list margin needs to be increased:
- Basic list to 1.625rem
- Nested list to 2.5rem (and drop ch units from there per my other comment)

waterplea (commented):
Nice idea with 0 width, a few comments

projects/demo/src/pages/components/list/examples/4/index.html
splincode: @waterplea updated

projects/layout/components/list/list.style.less
splincode: Fixed

projects/layout/components/list/list.style.less
splincode: Fixed

waterplea (approved):

splincode:
@vladimirpotekhin Sorry to bother you. Do you have a chance to take a look at my PR?

vladimirpotekhin (approved):

PR #14701, opened 2026-08-05
fix(core): fix TuiDropdownHover inside ShadowDom (#14699)

### LINKED ISSUE #14699: 🐞 - TuiDropdownHover not working inside ShadowDom
### Reproduction url

https://stackblitz.com/edit/angular-wtbzbjhr-gzefitqr?file=src%2Fapp%2Fapp.component.ts

### Description

`TuiDropdownHover` does not correctly detect pointer movement when the Taiga UI application is rendered inside a Shadow DOM.

When `tui-root` and an element with `tuiDropdownHover` are placed inside a shadow root, hovering over the trigger does not open the dropdown as expected. Depending on the component structure, moving the pointer between the trigger and the dropdown content may also be interpreted as leaving the active hover area, causing the dropdown to remain closed or close unexpectedly.

The same dropdown works correctly when rendered in the regular document DOM, outside Shadow DOM.

### Expected behavior

`tuiDropdownHover` should behave consistently regardless of whether the application is rendered in the regular document DOM or inside a Shadow DOM.

In particular:

* hovering over the trigger should open the dropdown;
* moving the pointer between the trigger and its dropdown content should keep the dropdown open;
* leaving both the trigger and dropdown content should close it according to the configured hide delay;
* nested Shadow DOM boundaries should not break hover detection.

### Actual behavior

Inside Shadow DOM, the directive does not reliably recognize the hovered element as belonging to the dropdown trigger or its active area. As a result, the dropdown does not open or its hover state is lost during normal pointer movement.

### Reproduction

1. Create a component that uses Shadow DOM encapsulation.
2. Place `tui-root` inside that component.
3. Add an element with `tuiDropdownHover` and associated dropdown content.
4. Move the pointer over the trigger.

The dropdown does not behave as it does in the equivalent setup outside Shadow DOM.

### Additional context

The problem is specific to hover-based dropdown interaction. Opening the dropdown through other interaction mechanisms may continue to work.

### Taiga UI version

4.93.0

### Browser Used

Chrome

### OS Used

macOS

MillerSvt:
Fixes #14699

### DIFF
--- a/projects/core/portals/dropdown/dropdown-hover.directive.ts
+++ b/projects/core/portals/dropdown/dropdown-hover.directive.ts
@@ -1,10 +1,11 @@
-import {DOCUMENT} from '@angular/common';
+import {DOCUMENT, isPlatformBrowser} from '@angular/common';
 import {
     contentChild,
     Directive,
     ElementRef,
     inject,
     input,
+    PLATFORM_ID,
     type Signal,
     signal,
     type WritableSignal,
@@ -26,6 +27,8 @@ import {
     fromEvent,
     map,
     merge,
+    NEVER,
+    type Observable,
     of,
     share,
     startWith,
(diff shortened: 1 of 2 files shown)


waterplea:
@MillerSvt could you check why the pipelines are failing?

MillerSvt:
Problem with SSR fixed, but one more test failing. Maybe I can see it in next two-three days...

MillerSvt:
It seems that test just was flaky.

MillerSvt:
Hmm, it works fine on local run right now.

waterplea:
@vladimirpotekhin could you have a look?

vladimirpotekhin:
test is not flaky, the problem is visible here https://taiga-preview-demo--pr14701-fix-tui-dropdown-hover-v5-gy6d41ea.web.app/directives/dropdown-hover#mobile

@MillerSvt check please


https://github.com/user-attachments/assets/b22a4ff4-b38e-4bef-b525-a640809cb8e3

waterplea:
@vladimirpotekhin it is the same on main 🤔

vladimirpotekhin:
Alright, I think we have similar problem here as we had with [tooltips](https://github.com/taiga-family/taiga-ui/blob/main/projects/core/portals/hint/hint-hover.directive.ts#L98): the synthesized mouseover event is not fired again until another element is tapped.

I think we can comment out this test for now and create a separate issue to address this problem.

vladimirpotekhin (approved):

waterplea (approved):

PR #14702, opened 2026-08-06
fix(layout): support RTL for `tuiList` items

splincode:
### Before

image

### After

image

### DIFF
--- a/projects/layout/components/list/list.style.less
+++ b/projects/layout/components/list/list.style.less
@@ -18,6 +18,8 @@
     & > li::before {
         content: counters(item, '.') '.';
         display: inline-flex;
+        unicode-bidi: plaintext;
+        font-variant-numeric: tabular-nums;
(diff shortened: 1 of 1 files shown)


projects/layout/components/list/list.style.less
```diff
@@ -33,6 +35,7 @@
     }
 
     &:is(ol) > li::before {
+        content: '\2066' counters(item, '.') '.\2069';
```
waterplea: I think you can just use `unicode-bidi: plaintext;`. Also why not move those rules to line 18? Is there a reason to isolate them into `:is(ol)`? Doesn't seem like it would hurtn `ul`

splincode: Thanks! Updated it to use `unicode-bidi: plaintext` for both list types.

waterplea (approved):

vladimirpotekhin (approved):

PR #14704, opened 2026-08-06
chore(layout): `Cell`, `Ripple` add ripple effect to Cell (#10976)

### LINKED ISSUE #10976: 🐞 - `Cell` inside `CardLarge` hover/active effect
### Reproduction url

https://taiga-ui.dev/layout/cell#combinations

### Description

We need to have background span whole card, especially on mobile when you tap:

Image

It should support ripple effect on mobile when enabled

### Taiga UI version

4

### Browser Used

Chrome

### OS Used

macOS

Yanduz:
Partial #10976 
image


projects/layout/components/cell/cell.styles.less
```diff
@@ -219,6 +219,11 @@
         margin-inline-end: -1rem;
     }
 
+    &[tuiRipple] {
```
waterplea: Let's move those to interactive cell by default (line 21)

waterplea (approved):

projects/layout/components/cell/cell.styles.less
```diff
@@ -17,6 +17,11 @@
     min-block-size: var(--t-height);
     border-radius: var(--t-radius);
 
+    &:is(button, label):not(:disabled) {
+        position: relative;
```
vladimirpotekhin: It seems like `position: relative` is already on the host. Is it needed here?

vladimirpotekhin (approved):

waterplea:
This approach breaks `tuiConnected` with buttons. Let's drop this and focus on automatic negative margin as we discussed in DMs

PR #14708, opened 2026-08-06
feat(core): `Textfield` moves appearance inputs from `<input>` to `<tui-textfield>`

nsbarsukov:
- `[tuiInput][invalid]` => `tui-textfield[invalid]`
- `[tuiInput][focused]` => `tui-textfield[tuiAppearanceFocus]`
- `[tuiInput][state]` => `tui-textfield[tuiAppearanceState]`

The old inputs of `TuiInputDirective` are deprecated but keep working until v6.

### Why?
`invalid` property is reserved by [Angular signal-based TuiFormValueControl](https://github.com/angular/angular/blob/96b80424c787f7a278b543c19f2c2364041761f4/packages/forms/signals/src/api/control.ts#L52-L56)

https://angular.dev/guide/forms/signals/custom-controls#adding-state-signals

For example:
```html
<input tuiInput [formField]="form.name" [invalid]="true" />
```

throws error

```
✘ [ERROR] NG8022: 
Binding to '[invalid]' is not allowed on nodes using the '[formField]' directive
```

Relates:


projects/core/components/textfield/textfield.component.ts
```diff
@@ -141,6 +145,17 @@ export class TuiTextfieldComponent<T> implements TuiDataListHost<T> {
 
     public readonly content = input<PolymorpheusContent<TuiContext<T>>>();
     public readonly filler = input('');
+    public readonly invalid = input<boolean | null>(null);
+    public readonly readOnly = input(false);
```
nsbarsukov: @taiga-family/core-team should new prop be `readonly` (all letters are lowercased) ? 🤔 

To be similar to Angular naming:
https://github.com/angular/angular/blob/96b80424c787f7a278b543c19f2c2364041761f4/packages/forms/signals/src/api/control.ts#L46

Now it copy-pasted from already existed
https://github.com/taiga-family/taiga-ui/blob/804ceb22166cba8b6ffec9edec74f11bfe279b31/projects/core/components/input/input.directive.ts#L53

projects/core/components/input/input.directive.ts
```diff
@@ -43,31 +45,81 @@ export class TuiInputDirective<T> implements TuiTextfieldAccessor<T> {
...
+    /**
+     * @deprecated use `<tui-textfield [tuiAppearanceState]="..." />` instead
+     * TODO(v6): delete
+     */
     public readonly state = input<TuiInteractiveState | null>(null);
     public readonly value = tuiValue(this.el);
 
     public readonly mode = computed<string | null>(() => {
-        if (this.readOnly()) {
+        const invalid = this.invalid() ?? this.textfield.invalid();
```
waterplea: If Angular's signal forms set this value we will no longer be able to override it with textfield.invalid, shouldn't the order of those be different so textfield.invalid is prioritized?

projects/core/components/textfield/textfield.component.ts
```diff
@@ -141,6 +145,17 @@ export class TuiTextfieldComponent<T> implements TuiDataListHost<T> {
...
+    /**
+     * TODO(v6): expose inputs using host directives API
```
waterplea: We cannot just expose tuiAppearanceFocus because there is no native way to display focus here – it is always controlled, whether by focused computed here which shows focus if it's inside or if the dropdown is open, or by manual user input. So we need some other input name.

projects/core/components/textfield/textfield.component.ts
waterplea: I think we should move readOnly to textfield, instead we should deprecate `readOnly` and add `readonly` input so that with old forms people can manually set it and with signal forms angular will set it automatically via schema

nsbarsukov: The `readOnly` logic has been reverted to its previous state. This PR now focuses on the `invalid`, `tuiAppearanceFocus`, and `tuiAppearanceState` props.

projects/core/components/input/input.directive.ts
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/14708/changes/acbd5ca959aa457f9103c657ffffe74c2c5c7566

projects/core/components/textfield/textfield.component.ts
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/14708/changes/acbd5ca959aa457f9103c657ffffe74c2c5c7566

waterplea (approved):

mdlufy (approved):

PR #14710, opened 2026-08-06
feat(addon-commerce): add configurable currency symbols

### LINKED ISSUE #14551: 🚀 - Configurable DI-token `CURRENCY_SYMBOLS`
### Description

https://github.com/taiga-family/taiga-ui/blob/41d4be71322257e5ed13fb4d76b00f77093d425c/projects/addon-commerce/utils/get-currency-symbol.ts#L7-L8

should become (with backward compatibility) DI-token with map:

```ts
// Just concept, it could be changed!

export const TUI_CURRENCY_SYMBOLS = new InjectionToken<
    Record<NonNullable<TuiCurrencyVariants>, string>
>('', {
    factory: () => {
        const locale = inject(LOCALE_ID);

        const yuan =
            new Intl.NumberFormat(locale, {style: 'currency', currency: 'CNY'})
                .formatToParts(123)
                .find((x) => x.type === 'currency')?.value ?? 'CN¥';

        return {
            [TuiCurrency.ArmenianDram]: '֏',
            [TuiCurrencyCode.ArmenianDram]: '֏',
            // [...]
            [TuiCurrency.YuanRenminbi]: yuan,
            [TuiCurrencyCode.YuanRenminbi]: yuan,
            // [...]
        };
    },
});
```

## Usage example
```ts
// proprietary-providers.ts
{
    provide: TUI_CURRENCY_SYMBOLS,
    useFactory: () => {
        const defaults = inject(TUI_CURRENCY_SYMBOLS, {skipSelf: true});

        return {
            ...defaults,
            [TuiCurrency.YuanRenminbi]: defaults[TuiCurrency.Yen],
            [TuiCurrencyCode.YuanRenminbi]: defaults[TuiCurrency.Yen],
        };
    },
},
```

splincode:
Fixes #14551

- add the configurable `TUI_CURRENCY_SYMBOLS` handler token
- preserve the existing `tuiGetCurrencySymbol` behavior by default
- use the token in `TuiCurrencyPipe` and `TuiAmountPipe`
- add tests for custom currency symbol handlers


projects/addon-commerce/pipes/amount/test/amount.pipe.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+
+describe('TuiAmountPipe', () => {
+    beforeEach(() => {
+        TestBed.configureTestingModule({
+            providers: [
+                TuiAmountPipe,
+                {
+                    provide: TUI_CURRENCY_SYMBOLS,
+                    useValue: (currency: TuiCurrencyVariants): string | null =>
+                        currency === 'USD' ? 'US$' : null,
```
nsbarsukov: Let's also add `inject(TUI_CURRENCY_SYMBOLS, {skipSelf: true})` as a fallback to verify that we can override symbols for specific countries only, while keeping the defaults for the rest. 

\+ 1 more `it(...)`

nsbarsukov: The same comment for `projects/addon-commerce/pipes/currency/test/currency.pipe.spec.ts` below

projects/addon-commerce/utils/format-currency.ts
```diff
@@ -9,8 +10,12 @@ function stringifyCurrency(currency: TuiCurrencyVariants): string {
...
+export function tuiFormatCurrency(
+    currency: TuiCurrencyVariants,
+    currencySymbolHandler: TuiHandler<TuiCurrencyVariants, string | null> =
+        tuiGetCurrencySymbol,
+): string {
     const stringifiedCurrency = stringifyCurrency(currency);
 
-    return tuiGetCurrencySymbol(stringifiedCurrency) || stringifiedCurrency;
+    return currencySymbolHandler(stringifiedCurrency) || stringifiedCurrency;
 }
```
nsbarsukov, line 20: I'm not sure we need both the `stringifyCurrency` and `tuiFormatCurrency` utils. They only exist to pad `number`-type codes with zeros.

Other cases already have fixed formats:

- ISO 4217 alphabetic currency codes (always 3 letters)
- ISO 4217 numeric currency codes (always 3 digits)

Should we deprecate both utils and just extends tuiGetCurrencySymbol with this logic instead?

```ts
export function tuiGetCurrencySymbol(raw: TuiCurrencyVariants): string | null {
    const currency =
        raw &&
        String(raw)
            // Do nothing for already 3-letters alphabetic and numeric currency codes
            .padStart(3, '0');

    switch (currency) {
```

projects/addon-commerce/pipes/amount/test/amount.pipe.spec.ts
splincode: Updated

projects/addon-commerce/utils/format-currency.ts
splincode: updated

nsbarsukov: @splincode 
Could you please review my comment again? 
To clarify, I wasn't suggesting that `tuiStringifyCurrency` should be exposed through the public API.

splincode: @nsbarsukov Thanks for the clarification and for pointing this out! Updated — tuiStringifyCurrency is no longer part of the public API, and numeric currency code normalization now lives directly in tuiGetCurrencySymbol. Is it right?

projects/addon-commerce/utils/format-currency.ts
```diff
@@ -1,16 +1,20 @@
...
+    if (symbol) {
+        return symbol;
+    }
+
+    if (currency === null) {
+        return '';
+    }
 
-    return tuiGetCurrencySymbol(stringifiedCurrency) || stringifiedCurrency;
+    return typeof currency === 'number' ? String(currency).padStart(3, '0') : currency;
```
nsbarsukov, line 19: `tuiGetCurrencySymbol` above already handles `typeof x === 'number'` case.

Do we really need to handle it again ?

nsbarsukov: Is this enough considering the recent changes to `tuiGetCurrencySymbol`'s logic? 

```ts
/**
 * @deprecated Use {@link tuiGetCurrencySymbol} instead.
 */
export function tuiFormatCurrency(currency: TuiCurrencyVariants): string {
    return tuiGetCurrencySymbol(currency) ?? '';
}
```

projects/addon-commerce/pipes/amount/amount.pipe.ts
```diff
@@ -30,7 +31,15 @@ export class TuiAmountPipe implements PipeTransform {
...
     private readonly formatted = computed(() => {
         const format = this.format();
-        const currencySymbol = tuiFormatCurrency(this.currency());
+        const currency = this.currency();
+        let currencySymbol = this.currencySymbolHandler(currency);
+
+        if (!currencySymbol && currency !== null) {
+            currencySymbol =
+                typeof currency === 'number'
+                    ? String(currency).padStart(3, '0')
```
nsbarsukov: `tuiGetCurrencySymbol` above already handles `typeof x === 'number'` case.

Do we really need to handle it again ?

projects/addon-commerce/pipes/currency/currency.pipe.ts
```diff
@@ -1,7 +1,24 @@
...
+            return symbol;
+        }
+
+        if (currency === null) {
+            return '';
+        }
+
+        return typeof currency === 'number'
+            ? String(currency).padStart(3, '0')
+            : currency;
```
nsbarsukov: Same comment as https://github.com/taiga-family/taiga-ui/pull/14710/changes#r3861307549

projects/addon-commerce/pipes/amount/amount.pipe.ts
splincode: Fixed

projects/addon-commerce/pipes/currency/currency.pipe.ts
splincode: Fixed

projects/addon-commerce/utils/format-currency.ts
splincode: Not quite, `tuiFormatCurrency` also preserves unknown or custom currencies.
For example, `tuiFormatCurrency('CSTM')` returns `'CSTM'`, and an unknown numeric code like 1 is formatted as '001'.

`tuiGetCurrencySymbol(currency) ?? ''` would return an empty string in both cases, which would change the existing behavior.

I agree that the numeric handling should not be duplicated though, so I moved that fallback out of the pipes and kept it only in `tuiFormatCurrency`.

splincode: @nsbarsukov friendly ping

nsbarsukov: Okay, let's simplify branching at least 

```ts
export function tuiFormatCurrency(currency: TuiCurrencyVariants): string {
    const fallback =
        typeof currency === 'number' ? String(currency).padStart(3, '0') : currency;

    return tuiGetCurrencySymbol(currency) ?? fallback ?? '';
}
```

projects/addon-commerce/utils/format-currency.ts
```diff
@@ -1,16 +1,20 @@
...
 
-function stringifyCurrency(currency: TuiCurrencyVariants): string {
-    return currency === null || tuiIsString(currency)
-        ? currency || ''
-        : String(currency).padStart(3, '0');
-}
-
+/**
+ * @deprecated Use {@link tuiGetCurrencySymbol} instead.
+ */
```
nsbarsukov, line 8: Add TODO comment to delete in v6

nsbarsukov (approved):

projects/addon-commerce/utils/format-currency.ts
splincode: Fixed

projects/addon-commerce/utils/format-currency.ts
splincode: Fixed

waterplea (approved):

vladimirpotekhin (approved):

PR #14712, opened 2026-08-07
chore(schematics): migrate removed TuiInputRangeModule/TuiInputDateMultiModule to kit in v5 (#11917)

mdlufy:
## Why

Two v4 `@taiga-ui/legacy` input modules had **no `ng update` identifier rename** to their v5 `@taiga-ui/kit` standalone successors, even though the sibling inputs already do (`TuiInputSliderModule`, `TuiInputMonthRangeModule`, `TuiInputDateModule`, `TuiInputDateTimeModule`, `TuiInputDateRangeModule`, `TuiInputPhoneModule`, `TuiInputNumberModule`, `TuiInputTagModule` …). This closes that gap. Part of #11917.

## What

Two entries in `ng-update/v5/steps/constants/identifiers-to-replace.ts`:

| from (`@taiga-ui/legacy`) | → to (`@taiga-ui/kit`) |
| --- | --- |
| `TuiInputRangeModule` | `TuiInputRange` |
| `TuiInputDateMultiModule` | `TuiInputDateMulti` |

Import + `imports: [...]` usage are rewritten (see snapshots).

## Why a plain rename is correct here

- **Established convention**: identical shape to the maintainer-owned `TuiInput*Module@legacy → TuiInput*@kit` renames already in the file.
- **Successors exist & are public on `main`**: `TuiInputRange` (`export class`, selector `tui-input-range` — **same tag as v4**, so templates keep working) and `TuiInputDateMulti` (`export const [...]` standalone array), both importable from `@taiga-ui/kit`.
- **Complements existing template migrations**: `migrate-input-date-multi.ts` already rewrites the `<tui-input-date multiple>` template but the import rename was missing — this fills exactly that half. `input-range` keeps its selector, so the import rename is all that's needed.
- **Not already covered**: neither `from` identifier appears anywhere in `ng-update/v5` on `main`.

## Tests

Added a rename case to each existing spec (`schematic-migrate-input-range`, `schematic-migrate-input-date-multi`); assertions via `migrate()` snapshots only. Full `ng-update/v5` suite green (132 suites / 766 tests).

### DIFF
--- a/projects/cdk/schematics/ng-update/v5/steps/constants/identifiers-to-replace.ts
+++ b/projects/cdk/schematics/ng-update/v5/steps/constants/identifiers-to-replace.ts
@@ -20,6 +20,26 @@ export const IDENTIFIERS_TO_REPLACE: ReplacementIdentifierMulti[] = [
     },
     {
         from: {
+            name: 'TuiInputRangeModule',
+            moduleSpecifier: '@taiga-ui/legacy',
+        },
+        to: {
+            name: 'TuiInputRange',
+            moduleSpecifier: '@taiga-ui/kit',
+        },
+    },
+    {
+        from: {
+            name: 'TuiInputDateMultiModule',
+            moduleSpecifier: '@taiga-ui/legacy',
+        },
+        to: {
+            name: 'TuiInputDateMulti',
+            moduleSpecifier: '@taiga-ui/kit',
+        },
+    },
+    {
+        from: {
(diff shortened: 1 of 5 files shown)


mdlufy:
Superseded by #14706 — these two renames (TuiInputRangeModule/TuiInputDateMultiModule) were consolidated into the single "add missing v5 identifier renames" PR alongside the textfield-directive and doc-documentation renames.

PR #14744, opened 2026-08-12
chore(addon-doc): change sizes for navigation search input

4ooper:


4ooper:
Hi! I think it might make sense to add something like `TUI_DOC_SEARCH_OPTIONS` to DI — what do you think?

waterplea (approved):

nsbarsukov (approved):

PR #14747, opened 2026-08-13
fix(kit): icon appearance for step buttons in tables

### LINKED ISSUE #14612: 🐞 - `InputNumber` step buttons render incorrectly with inherited `table` appearance
### Reproduction url

https://stackblitz.com/edit/angular-uugkzhzv

### Description

`InputNumber` step buttons render incorrectly inside `table[tuiTable]`. `TuiInputNumberStepButtons` applies `appearance = 'table'` from `textfield` and set `min-block-size: calc(100% + 1px)`. But `appearance = 'icon'` for buttons inside `TuiInputNumberStepButtons` can fix that.

### Taiga UI version

5.16.0

### Browser Used

Chrome

### OS Used

macOS

nullawhale:
Fixes #14612


projects/kit/components/input-number/step/input-number-step.component.ts
```diff
@@ -43,6 +44,10 @@ export class TuiInputNumberStepButtons {
     protected readonly appearance = inject(TUI_TEXTFIELD_OPTIONS).appearance;
     protected readonly hold = inject(TuiInputNumberStepService<bigint | number>);
 
+    protected readonly buttonAppearance = computed(() =>
+        this.appearance() === 'table' ? 'icon' : this.appearance(),
```
waterplea: 'icon' has increased icon size due to lack of background, maybe 'flat-grayscale' would look better?

waterplea (changes requested):
Actually, looks like the only change that's necessary is `min-block-size: 0;` added over here:
https://github.com/taiga-family/taiga-ui/blob/main/projects/kit/components/input-number/step/input-number-step.style.less#L18

mdlufy (approved):

nullawhale:
> Actually, looks like the only change that's necessary is `min-block-size: 0;` added over here: https://github.com/taiga-family/taiga-ui/blob/main/projects/kit/components/input-number/step/input-number-step.style.less#L18

Done.

waterplea (approved):

PR #14766, opened 2026-08-14
feat(core): support protected context in tuiDialog

splincode:
## What was done

- Added an explicit `tuiDialog<Result, Data>` overload.
- Allowed dialog components to keep `context` protected.
- Preserved automatic type inference for components with a public context.
- Preserved existing validation for components with multiple contexts.

## Usage

```ts
class DialogComponent {
    protected readonly context =
        injectContext<TuiDialogContext<Result, Data>>();
}

const dialog = tuiDialog<Result, Data>(DialogComponent);
```

## Motivation

`tuiDialog` currently infers data and result types through public component properties. Since protected properties are not included in keyof, components have to expose their dialog context publicly.

The explicit overload provides an opt-in alternative without changing the existing API behavior.

### DIFF
--- a/projects/core/components/dialog/dialog.factory.ts
+++ b/projects/core/components/dialog/dialog.factory.ts
@@ -1,4 +1,10 @@
-import {assertInInjectionContext, inject, INJECTOR, type Injector} from '@angular/core';
+import {
+    assertInInjectionContext,
+    inject,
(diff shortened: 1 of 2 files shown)


waterplea:
Do we need to support this setup? Not sure we want to overcomplicate types for this

PR #14792, opened 2026-08-20
fix(kit): `InputInline` fix for updateOn blur/submit

waterplea:
Fixes #

### DIFF
--- a/projects/demo/src/pages/components/input-inline/examples/2/index.html
+++ b/projects/demo/src/pages/components/input-inline/examples/2/index.html
@@ -7,6 +7,7 @@
             Type a heading
             <input
                 tuiAutoFocus
+                [ngModelOptions]="{updateOn: 'blur'}"
                 [(ngModel)]="heading"
                 (blur)="onBlur()"
                 (keydown.enter.prevent)="toggle()"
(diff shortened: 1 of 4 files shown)


projects/demo/src/pages/components/input-inline/examples/2/index.html
```diff
@@ -7,6 +7,7 @@
             Type a heading
             <input
                 tuiAutoFocus
+                [ngModelOptions]="{updateOn: 'blur'}"
```
vladimirpotekhin: keydown.enter now doesn't save value

vladimirpotekhin (approved):

projects/demo/src/pages/components/input-inline/examples/2/index.html
vladimirpotekhin: (I think it's better to update 1st example)

mdlufy (approved):

projects/demo/src/pages/components/input-inline/examples/2/index.html
waterplea: Worked for me, but I have improved example to handle both save, cancelation and updateOn: 'submit' to better resemble real life situation

PR #14794, opened 2026-08-20
fix(kit): `InputInline` fix for updateOn blur/submit

vladimirpotekhin:
#14792 cherry pick


waterplea (approved):
Let's also change the example to reflect updateOn: 'blur' like I did in the main PR so it can be checked

mdlufy (approved):

Issue #14671, opened 2026-07-31
🐞 - `Textfield[content]` overflows the its container bounds


nsbarsukov:
### Description

https://taiga-ui.dev/v4/components/select/API?content$=2&sandboxWidth=250
<img height="150" alt="Image" src="https://github.com/user-attachments/assets/ba65deb6-ccf3-44ca-a750-6607c3abf804" />

https://taiga-ui.dev/components/select/API?content$=2&sandboxWidth=250
<img height="150" alt="Image" src="https://github.com/user-attachments/assets/f38d446c-bf59-4f5d-8537-c782994eb00f" />

https://taiga-ui.dev/components/combo-box/API?content$=3&sandboxWidth=250

<img height="150" alt="Image" src="https://github.com/user-attachments/assets/57da3bf4-d990-4887-a0bb-ef29a25a9493" />

### Expected behavior
As without `[content]`
https://taiga-ui.dev/components/select/API?sandboxWidth=250

<img height="150" alt="Image" src="https://github.com/user-attachments/assets/abefef08-2634-4b04-a5e9-9b10d4d101f1" />

### Taiga UI version

v4 / 5.17.0

### Browser Used

Chrome

### OS Used

macOS

PR #14799, opened 2026-08-21
fix(kit): textfield content overflow fix v4

gregorycode:
Fixes https://github.com/taiga-family/taiga-ui/issues/14671


projects/core/styles/components/textfield.less
```diff
@@ -270,14 +271,18 @@
...
-        padding-inline-end: calc(var(--t-end, ~'0rem') + var(--t-side) + var(--t-padding));
+        padding-inline-end: var(--t-space-end);
     }
 
     & > .t-template {
         display: flex;
         align-items: center;
         border-radius: inherit;
         color: var(--tui-text-primary);
+        white-space: nowrap;
```
waterplea: This might be breaking change 🤔

gregorycode: I'm not very familiar with this part of the codebase yet, so I may be missing some context. How do you usually handle text wrapping in cases like this? I see that we have white-space: nowrap in v5, but not in v4. Was it a bug all along, or was it done on purpose?

waterplea: I guess we switched default behavior between 4 and 5, not sure we want to do that in a minor release for 4. Most likely we will not break anything by adding it, but I'll wait for the rest of the team's opinion

gregorycode: Yeah, it makes sense.

vladimirpotekhin: I agree, this is a breaking change, and I'd like to avoid it. Maybe just add nowrap to the example?

waterplea: This is pretty minor, I'm not sure it's worth it to create an example. And it's LTS version. Let's just add all of the other CSS and if people will need nowrap – they can add it themselves after they see it doesn't look like they want it to.

gregorycode:
@waterplea , @vladimirpotekhin Updated. Please take a look.

waterplea (approved):

vladimirpotekhin (approved):

projects/demo-playwright/tests/core/textfield/textfield-content.pw.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+        const example = new TuiDocumentationApiPagePO(page).apiPageExample;
+        const textfield = example.locator('tui-textfield');
+
+        await textfield.locator('input').click();
+        await page
+            .locator('button[tuiOption]', {hasText: 'Federated States of Micronesia'})
+            .first()
+            .click();
+
+        await expect.soft(textfield).toHaveScreenshot('textfield-content-ltr.png');
```
nsbarsukov, line 21: Why aren't there any screenshot test failures, given that this PR is supposed to fix a bug reproduced by the new Playwright test?

nsbarsukov: Bug still exists for v4
localhost

waterplea: We just need nowrap in the test, we didn't add it to not cause breaking changes. Might not be worth it for v4, in v5 it's correct.

PR #14802, opened 2026-08-21
feat(core): support labels for buttons

splincode:
## Motivation

`TuiButton` can currently be applied only to `button` and `a` elements.

Using `label` as a button is useful for native controls such as hidden file inputs:

```html
<label
    tuiButton
    for="file"
>
    Upload file
</label>

<input
    id="file"
    type="file"
    hidden
/>
```

This allows opening the native file picker without manually triggering the input with JavaScript.

### DIFF
--- a/projects/core/components/button/test/button.directive.spec.ts
+++ b/projects/core/components/button/test/button.directive.spec.ts
@@ -53,28 +53,6 @@ describe('ButtonDirective', () => {
             >
                 Extra small
             </button>
-
-            <label
-                id="label-button"
-                tuiButton
-                for="file"
-            >
-                Label
-            </label>
-
-            <label
-                id="label-icon-button"
-                tuiIconButton
-                for="file"
-            >
-                Label icon
-            </label>
-
-            <input
-                id="file"
-                type="file"
-                hidden
-            />
         `,
         changeDetection: ChangeDetectionStrategy.OnPush,
     })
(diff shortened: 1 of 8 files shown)


projects/core/components/button/test/button.directive.spec.ts
```diff
@@ -121,4 +143,22 @@ describe('ButtonDirective', () => {
             expect(size).toBe('xs');
         });
     });
+
+    describe('label:', () => {
```
waterplea: This doesn't really test that it works on labels, we need a new example that would be screenshot tested

waterplea (approved):

projects/core/components/button/test/button.directive.spec.ts
splincode: Fixed

projects/demo/src/pages/components/button/examples/9/index.html
```diff
@@ -0,0 +1,12 @@
+<label
```
waterplea: This way input is not keyboard accessible, we need to put it inside the label, absolutely positioned and make it invisible somehow, without hiding keyboard focus 🤔

splincode: fixed

waterplea (approved):

vladimirpotekhin (approved):

Issue #14228, opened 2026-06-04
🚀 - Move components to styles package


waterplea:
### Description

We still have a lot of style based components that are not yet moved to styles package, such as cards, header and other layout components, maybe some more. We need to move all we can to styles.

waterplea:
I think `Cell` and `Card` are the only components left that we can move to the styles package

PR #14820, opened 2026-08-26
refactor(layout): move layout styles to styles package

splincode:
Part of https://github.com/taiga-family/taiga-ui/issues/14228


waterplea (approved):

projects/layout/components/item-group/item-group.directive.ts
waterplea, line 1: Let's keep this one in the layout pack as it requires JS to function (._initialized class)

vladimirpotekhin (approved):

projects/layout/components/item-group/item-group.directive.ts
splincode: done

PR #14824, opened 2026-08-26
fix(cdk): `tuiControlValue` utility supports `NgControl` from signal forms

nsbarsukov:
Relates:

### Problem
```html
<tui-textfield tuiChevron>
  <input
    tuiSelect
    [formField]="f.select"
  />

  <tui-data-list-wrapper
    *tuiDropdown
    [items]="items"
  />
</tui-textfield>
```

throws error on attempt to open dropdown

```
ERROR RuntimeError: 
NG0601: `toSignal()` called with `requireSync` but `Observable` did not emit synchronously.
    at toSignal (rxjs-interop.mjs:143:11)
    at tuiInjectValue (inject-value.ts:10:22)
    at <instance_members_initializer> (select-option.component.ts:44:30
    at new _TuiSelectOption (select-option.component.ts:42:7
    at NodeInjectorFactory.TuiSelectOption_Factory [as factory] (select-option.component.ts:59:22)
```

(same issue for other controls with datalist suggestions: `ComboBox`, `InputChip` etc.)

### Why
Signal forms provide a fake `NgControl` which is not an `AbstractControl` and has no
observable `valueChanges` — its value is exposed as a signal, so reading it inside `computed`
subscribes to it.

### DIFF
--- a/projects/cdk/observables/control-value.ts
+++ b/projects/cdk/observables/control-value.ts
@@ -5,6 +5,7 @@ import {distinctUntilChanged, Observable, startWith} from 'rxjs';
 
 /**
  * Turns form control value changes into ReplaySubject(1)
+ * TODO(v6): refactor it after signal forms are fully supported (Angular >= 22)
  */
 export function tuiControlValue<T>(
     control?: AbstractControl | AbstractControlDirective | null, // TODO: add `InteropNgControl` as possible type after update to Angular 21+
--- a/projects/cdk/observables/control-value.ts
+++ b/projects/cdk/observables/control-value.ts
@@ -1,13 +1,26 @@
+import {computed, type Injector, untracked} from '@angular/core';
+import {toObservable} from '@angular/core/rxjs-interop';
 import {type AbstractControl, type AbstractControlDirective} from '@angular/forms';
-import {Observable, startWith} from 'rxjs';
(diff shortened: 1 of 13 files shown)


projects/cdk/observables/control-value.ts
```diff
@@ -1,13 +1,29 @@
...
+import {toObservable} from '@angular/core/rxjs-interop';
 import {type AbstractControl, type AbstractControlDirective} from '@angular/forms';
-import {Observable, startWith} from 'rxjs';
+import {distinctUntilChanged, Observable, startWith} from 'rxjs';
 
 /**
- * Turns AbstractControl/Abstract-control-directive valueChanges into ReplaySubject(1)
+ * Turns form control value changes into ReplaySubject(1)
  */
 export function tuiControlValue<T>(
```
waterplea, line 10: Let's leave a TODO for v6 to revisit this after signal forms are fully supported

waterplea (approved):

vladimirpotekhin (approved):

projects/cdk/observables/control-value.ts
nsbarsukov: https://github.com/taiga-family/taiga-ui/pull/14824/changes/cdd38d16b906e90aeec272099e28cfe73eaee10e

PR #14829, opened 2026-08-27
fix(kit): toast close button localization

### LINKED ISSUE #14828: Toast close button is not localized: hardcoded "Close" instead of TUI_CLOSE_WORD
### Description

`TuiToastComponent` hardcodes the English word `Close` as the accessible name of its
close button, instead of taking it from `TUI_CLOSE_WORD`. Every other closable surface in
the library already uses the token — `TuiDialogComponent` and `TuiNotification` both do —
so in a non-English app the toast is the only control that announces itself in English.

From `@taiga-ui/kit` 5.18.0, the toast template:

```html
@if (context.closable && !isMobile) {
    <button
        tuiButtonX
        (click.stop.prevent)="context.$implicit.complete()"
    >
        Close
    </button>
}
```

The word is not visible: `tuiButtonX` renders a 24×24 cross and the label is collapsed
with `font-size: 0` and `overflow: hidden` (measured in Chrome via
`getComputedStyle`). So this is purely an accessibility problem — the button looks right
and sounds wrong.

### Impact

With `TUI_LANGUAGE` set to a non-English language, a screen reader announces "Close" for
every toast, while dialogs in the same app announce the localized word. In Russian,
`TUI_RUSSIAN_LANGUAGE.close` is `Закрыть`, and dialogs use it correctly — the
inconsistency is audible only to screen reader users, which is why it can sit unnoticed
for a long time.

### Steps to reproduce

1. Provide a non-English language, e.g.
   `{provide: TUI_LANGUAGE, useValue: signal(TUI_RUSSIAN_LANGUAGE)}`.
2. Open any closable toast: `inject(TuiToastService).open('Hello').subscribe()`.
3. Inspect the close button's accessible name (accessibility pane, or
   `document.querySelector('tui-toast button').textContent`).

**Expected:** the localized close word, the same one a dialog uses in the same app.
**Actual:** `Close`.

### Suggested fix

Read the token the way the dialog already does — inject `TUI_CLOSE_WORD` in
`TuiToastComponent` and interpolate it in place of the literal. I have not opened a PR
because I am not set up to run your test suite; happy to if you would rather have one.

### Environment

- `@taiga-ui/kit` 5.18.0, `@taiga-ui/core` 5.18.0, `@taiga-ui/i18n` 5.18.0
- Angular 21, Chrome 154

Nothing is blocked by this on my side — filing it so it is written down.

---

Written by Claude (Anthropic), working with @DmitrySharabin, who reviewed and tested it.

gregorycode:
Fixes https://github.com/taiga-family/taiga-ui/issues/14828

### DIFF
--- a/projects/kit/components/toast/test/toast.component.spec.ts
+++ b/projects/kit/components/toast/test/toast.component.spec.ts
@@ -1,0 +1,47 @@
+import {ChangeDetectionStrategy, Component} from '@angular/core';
+import {type ComponentFixture, TestBed} from '@angular/core/testing';
+import {WA_IS_MOBILE} from '@ng-web-apis/platform';
+import {provideTaiga, TuiRoot} from '@taiga-ui/core';
+import {TUI_DEFAULT_LANGUAGE, TUI_RUSSIAN_LANGUAGE} from '@taiga-ui/i18n';
+
+import {TuiToastService} from '../toast.service';
+
+describe('Toast', () => {
+    @Component({
+        imports: [TuiRoot],
+        template: `
+            <tui-root />
+        `,
+        changeDetection: ChangeDetectionStrategy.OnPush,
+    })
+    class Test {}
+
+    let fixture: ComponentFixture<Test>;
+    let toasts: TuiToastService;
+
+    function getCloseButton(): HTMLElement | null {
+        return fixture.nativeElement.querySelector('tui-toast button[tuiButtonX]');
+    }
+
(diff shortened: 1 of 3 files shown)


vladimirpotekhin (approved):

waterplea (approved):

gregorycode:
@waterplea, @vladimirpotekhin Hello. The PR can be merged.

PR #14833, opened 2026-08-27
fix(cdk): `AutoFocus` defer iOS focus until entrance animation ends to prevent dialog jump

### LINKED ISSUE #14665: 🐞 - iOS / Confirm dialog animation jumps on open
### Reproduction url

https://taiga-ui.dev/components/confirm

### Description

https://github.com/user-attachments/assets/212b6fba-5fd1-49f7-9eb4-847a7d984c13

### Taiga UI version

5.17.0

### Browser Used

Safari

### OS Used

iOS

mdlufy:
Closes #14665

On iOS the confirm dialog jumped during the opening animation: `TuiConfirm` autofocuses its button, and iOS `focus()` scrolls it into view even with `{preventScroll: true}` when called mid-animation inside a scrollable sheet/dialog.

Now non-text-field elements are focused only after the surrounding `.tui-animated` entrance animation finishes. Autofocus is preserved (it just lands after the dialog settles), with immediate focus as a fallback when there is no animation.

### DIFF
--- a/projects/cdk/directives/auto-focus/handlers/ios.handler.ts
+++ b/projects/cdk/directives/auto-focus/handlers/ios.handler.ts
@@ -1,4 +1,5 @@
 import {type ElementRef, type NgZone, type Renderer2} from '@angular/core';
+import {TUI_ANIMATED} from '@taiga-ui/cdk/directives/animated';
 import {tuiIsPresent, tuiPx} from '@taiga-ui/cdk/utils/miscellaneous';
 
 import {type TuiAutofocusOptions} from '../autofocus.options';
@@ -32,7 +33,13 @@ export class TuiIosAutofocusHandler extends AbstractTuiAutofocusHandler {
         if (this.isTextFieldElement) {
             this.zone.runOutsideAngular(() => this.iosWebkitAutofocus());
         } else {
-            this.element.focus({preventScroll: true});
+            // iOS scrolls to the focused element mid-animation despite preventScroll, jumping the dialog
+            void Promise.allSettled(
+                this.element
+                    .closest(`.${TUI_ANIMATED}`)
+                    ?.getAnimations?.()
+                    .map(async ({finished}) => finished) ?? [],
+            ).then(() => this.element.focus({preventScroll: true}));
         }
     }
 
--- a/projects/cdk/directives/auto-focus/test/auto-focus.directive.spec.ts
+++ b/projects/cdk/directives/auto-focus/test/auto-focus.directive.spec.ts
@@ -8,6 +8,7 @@ import {
 } from '@angular/core';
 import {type ComponentFixture, fakeAsync, TestBed, tick} from '@angular/core/testing';
 import {WA_WINDOW} from '@ng-web-apis/common';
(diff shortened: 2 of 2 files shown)


projects/cdk/directives/auto-focus/handlers/ios.handler.ts
```diff
@@ -32,8 +33,30 @@ export class TuiIosAutofocusHandler extends AbstractTuiAutofocusHandler {
...
+        const animations =
+            this.element.closest(`.${TUI_ANIMATED}`)?.getAnimations?.() ?? [];
+
+        if (!animations.length) {
             this.element.focus({preventScroll: true});
+
+            return;
         }
+
+        const finished: Array<Promise<unknown>> = [];
```
waterplea: Why not Array.from + map?

mdlufy: Fair, switched to `.map`. `getAnimations()` already returns an array, and the callback is `async` to satisfy `promise-function-async`.

projects/cdk/directives/auto-focus/handlers/ios.handler.ts
```diff
@@ -32,8 +33,24 @@ export class TuiIosAutofocusHandler extends AbstractTuiAutofocusHandler {
...
+        const animations =
+            this.element.closest(`.${TUI_ANIMATED}`)?.getAnimations?.() ?? [];
+
+        if (!animations.length) {
             this.element.focus({preventScroll: true});
+
+            return;
         }
+
+        void Promise.allSettled(
```
waterplea: Wouldn't it work for empty array by default? Looks like we do not need if() statement before

projects/cdk/directives/auto-focus/handlers/ios.handler.ts
```diff
@@ -32,8 +33,24 @@ export class TuiIosAutofocusHandler extends AbstractTuiAutofocusHandler {
...
+            animations.map(async (animation) => animation.finished),
```
waterplea: Maybe we can inline it like this?
```js
Promise.allSettled(
  this.element.closest(`.${TUI_ANIMATED}`)?.getAnimations?.().map(async ({finished}) => finished) ?? []
).then(() => this.element.focus({preventScroll: true}));
```
So it's like a one-liner

waterplea (approved):

projects/cdk/directives/auto-focus/handlers/ios.handler.ts
mdlufy: Done, inlined it into `setFocus` as suggested. Dropped the `if` guard too — `allSettled([])` resolves right away and still focuses, so the empty case is covered.

projects/cdk/directives/auto-focus/handlers/ios.handler.ts
mdlufy: Right, removed it — the inlined `?? []` + `allSettled([])` handles the empty case.

nsbarsukov (approved):

PR #10778, opened 2025-04-09
feat(kit): `Textarea` add new component

waterplea:
Fixes #9213
Fixes #8591

PR #14841, opened 2026-08-27
fix(kit): `Textarea` remove excessive end padding

### LINKED ISSUE #14571: 🐞 - `Textarea` has unnecessarily big end padding
### Reproduction url

https://taiga-ui.dev/components/textarea

### Description

There's always 16px of excessive padding when there are no icons/buttons on the right:

Image

### Taiga UI version

5

### Browser Used

Chrome

### OS Used

macOS

splincode:
Fixes #14571

### DIFF
--- a/projects/demo-playwright/tests/kit/textarea/textarea.pw.spec.ts
+++ b/projects/demo-playwright/tests/kit/textarea/textarea.pw.spec.ts
@@ -21,6 +21,24 @@ test.describe('Textarea', () => {
         await expect.soft(textarea).toHaveScreenshot('textarea-line-break-disabled.png');
     });
 
+    test('does not have excessive end padding without content', async ({page}) => {
+        await tuiGoto(page, `${DemoRoute.Textarea}/API`);
+        const {demo} = new TuiDocumentationPagePO(page);
+        const textarea = demo.locator('textarea[tuiTextarea]');
+        const label = demo.locator('label[tuiLabel]');
+        const padding = await textarea.evaluate((element) => {
+            const {paddingInlineEnd, paddingInlineStart} = getComputedStyle(element);
+
+            return {
+                end: Number.parseFloat(paddingInlineEnd),
+                start: Number.parseFloat(paddingInlineStart),
+            };
+        });
+
+        await expect(label).toBeVisible();
+        expect(Math.abs(padding.end - padding.start)).toBeLessThanOrEqual(2);
+    });
+
     test('required textarea is not invalid before touched', async ({page}) => {
         await tuiGoto(page, DemoRoute.Textarea);
         const example = new TuiDocumentationPagePO(page).getExample('#icons');
--- a/projects/styles/components/textfield.less
+++ b/projects/styles/components/textfield.less
@@ -326,7 +326,7 @@ tui-textfield:where(*&) {
     }
 
     textarea ~ .t-content {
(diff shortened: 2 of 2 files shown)


projects/kit/components/textarea/textarea.style.less
```diff
@@ -70,6 +70,14 @@ tui-textarea-content:where(*&) {
     }
 }
 
+textarea[tuiTextarea]:where(*&) ~ [tuiLabel] {
+    min-inline-size: 0.5rem;
+}
+
+textarea[tuiTextarea]:where(*&) ~ .t-content {
+    min-inline-size: 0;
```
waterplea: It's explicitely set as
```css
    textarea ~ .t-content {
        min-inline-size: 0.5rem;
    }
```
Do we know what for?

projects/kit/components/textarea/textarea.style.less
```diff
@@ -70,6 +70,14 @@ tui-textarea-content:where(*&) {
     }
 }
 
+textarea[tuiTextarea]:where(*&) ~ [tuiLabel] {
+    min-inline-size: 0.5rem;
```
waterplea: Why do we need min width for label?

projects/kit/components/textarea/textarea.style.less
splincode: It was introduced together with the new Textarea in #10778. I couldn't find any explanation for the `0.5rem` in the PR history.

For textarea, this empty `.t-content width` is measured into `--t-side`, which then contributes to `padding-inline-end` and `--t-space`, causing the excessive padding from the issue.

I've kept the fix scoped to `textarea[tuiTextarea]` instead of changing the generic textfield rule.

projects/kit/components/textarea/textarea.style.less
splincode: We don't need it here. Removed it and also removed the label-specific assertion from the test.

projects/kit/components/textarea/textarea.style.less
waterplea: There's no reason to do that, there's no other textarea. We need to remove it from textfield styles if we do not need it. Let's remove it, I cannot think of a reason why I added it.

waterplea: Hmm, if I remove it on demo – label disappears

splincode: Hmm, you are right `0.5rem` is needed to keep the label layout intact. The actual issue is that this layout-only minimum gets measured into `--t-side`, so an empty textarea side adds extra end padding. I restored the original `min-inline-size` and moved the fix to `onResize` if the side is only its minimum width, it is treated as 0 for padding calculations. I also saved a regression assertion that the label stays visible.

waterplea: I looks like we just need at least ANY width, maybe we can just put min-inline-size: 1px? 1px is irrelevant error in --t-side calculation.

splincode: Good, `1px` is enough here. I changed `min-inline-size` to `1px` and removed the extra `onResize` logic. I also kept the regression test to make sure the label stays visible and the end padding does not grow again.

waterplea (approved):

nsbarsukov (approved):

PR #14846, opened 2026-08-28
fix(kit): `Stepper` inside `CardLarge`

### LINKED ISSUE #14840: 🐞 - TuiStepper issue inside TuiCardLarge
### Reproduction url

https://stackblitz.com/edit/angular-aafgxz53?file=src%2Fapp%2Fapp.ts

### Description

If you place TuiStepper inside TuiCardLarge, the TuiStepper layout breaks. Also, it’s not possible to override the size of TuiStep

### Taiga UI version

5.21

### Browser Used

Chrome

### OS Used

macOS

splincode:
Fixes https://github.com/taiga-family/taiga-ui/issues/14840

### DIFF
--- a/projects/demo-cypress/src/tests/stepper.cy.ts
+++ b/projects/demo-cypress/src/tests/stepper.cy.ts
@@ -39,6 +39,8 @@ describe('Stepper', () => {
     beforeEach(() => cy.mount(Sandbox));
 
     it('does not inherit CardLarge cell spacing', () => {
+        cy.get('[tuiCardLarge]').compareSnapshot('stepper-card-large');
+
         cy.get('[tuiStep]').each(($step) => {
             cy.wrap($step)
                 .should('have.css', 'margin-left', '0px')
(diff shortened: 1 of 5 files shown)


waterplea (approved):

projects/demo-cypress/src/tests/stepper.cy.ts
```diff
@@ -0,0 +1,54 @@
...
+export class Sandbox {}
+
+describe('Stepper', () => {
+    beforeEach(() => cy.mount(Sandbox));
+
+    it('does not inherit CardLarge cell spacing', () => {
+        cy.get('[tuiStep]').each(($step) => {
+            cy.wrap($step)
+                .should('have.css', 'margin-left', '0px')
+                .and('have.css', 'margin-right', '0px');
```
nsbarsukov, line 47: Would it be better to add a visual snapshot here — either instead of, or in addition to, these assertions?

nsbarsukov (approved):

projects/demo-cypress/src/tests/stepper.cy.ts
splincode: Added `cy.get('[tuiCardLarge]').compareSnapshot('stepper-card-large');`

PR #14856, opened 2026-08-29
refactor(layout): move Card styles to styles package

splincode:
Part of https://github.com/taiga-family/taiga-ui/issues/14228


projects/layout/components/card/collapsed.directive.ts
```diff
@@ -12,7 +12,7 @@ import {tuiWithStyles} from '@taiga-ui/cdk/utils/miscellaneous';
     template: '',
     styles: `
         [data-tui-version='${TUI_VERSION}'] {
-            @import './collapsed.style.less';
```
waterplea, line 15: I think this directive makes no sense without JS so the styles can stay here, not in the styles package

waterplea (approved):

projects/layout/components/card/collapsed.directive.ts
splincode: I moved collapsed.style.less back next to the directive and restored the local import.

mdlufy (approved):

PR #14858, opened 2026-08-29
feat(core): allow content-sized dialogs

### LINKED ISSUE #14836: 🚀 - Being able to define min/max/fixed width/height for Dialogs
### Description

I open dialog with 
```html
<ng-template [(tuiDialog)]="opened"
             [tuiDialogOptions]="{label: 'Lorem Ipsum', size: 'm'}">
```

My content is very long so my dialog is scrollable. I had to wrap my content into
```html
<div class="flex-column gap-4 overflow-auto" style="height: calc(100vh - 12rem)">
```

otherwise, the dialog overflows. I would like to be able to define the height of the dialog to have a more maintainable code. I would like this heigh to take margin into consideration. By than I mean that by default, margins are 2.5rem. If I define the height to 100%, I want the dialog to keep these 2.5rem margins.

Similarly, I would like to be able to define the width.

(Maybe this is already possible with Taiga UI)

splincode:
## Changes

* allow an empty dialog `size`
* let the dialog width be controlled by its content when `size === ''`
* keep existing `s`, `m`, and `l` width presets unchanged
* keep vertical scrolling behavior unchanged
* add regression coverage for the empty size

This uses the existing dialog layout: fixed widths are only applied to the `s`, `m`, and `l` presets, while the base dialog already keeps the viewport margins via `max-inline-size`.

Fixes #14836

### DIFF
--- a/projects/core/portals/dialog/dialog.options.ts
+++ b/projects/core/portals/dialog/dialog.options.ts
@@ -12,7 +12,7 @@ import {type Observable} from 'rxjs';
  * dismissible - close dialog by Esc button or click on overlay (true by default)
  * label - string title for the dialog ('' by default)
  * required - dismissing or closing dialog by X button throws (false by default)
- * size - size of the dialog ('m' by default)
+ * size - size of the dialog ('m' by default, empty string for content width)
  */
 export interface TuiDialogOptions<I> {
     readonly appearance: string;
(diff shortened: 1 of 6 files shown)


projects/core/portals/dialog/dialog.options.ts
```diff
@@ -21,7 +21,7 @@ export interface TuiDialogOptions<I> {
     readonly dismissible: Observable<boolean> | boolean;
     readonly label: string;
     readonly required: boolean;
-    readonly size: TuiSizeL | TuiSizeS;
+    readonly size: '' | TuiSizeL | TuiSizeS;
```
waterplea: Maybe `null` is more explicit than an empty string?

waterplea (commented):
Let's also add an example on demo

projects/core/portals/dialog/dialog.options.ts
splincode: fixed

waterplea (approved):

nsbarsukov (approved):

PR #14861, opened 2026-08-30
fix(kit): reflect unfinished date range picking

### LINKED ISSUE #14827: 🚀 - `InputDateRange` should reflect unfinished picking
### Description

When using InputDateRange and picking the first date, it should reflect the intermediate state

Image

splincode:
Fixes #14827

### DIFF
--- a/projects/cdk/utils/dom/value.ts
+++ b/projects/cdk/utils/dom/value.ts
@@ -79,7 +79,7 @@ export function tuiValue(
             return;
         }
 
-        if (element?.matches(':focus') && 'selectionStart' in element) {
+        if (element?.matches(':focus') && 'selectionStart' in element && element.value) {
             const {selectionStart, selectionEnd} = element;
 
             /**
--- a/projects/kit/components/calendar-range/calendar-range.component.ts
+++ b/projects/kit/components/calendar-range/calendar-range.component.ts
@@ -6,12 +6,14 @@ import {
     linkedSignal,
     model,
     type OnChanges,
+    type OnDestroy,
     type OnInit,
     type SimpleChanges,
     untracked,
 } from '@angular/core';
 import {WA_IS_MOBILE} from '@ng-web-apis/platform';
 import {
+    RANGE_SEPARATOR_CHAR,
     tuiDateClamp,
     TuiDay,
     type TuiDayLike,
@@ -31,7 +33,12 @@ import {
 import {TuiDataList} from '@taiga-ui/core/components/data-list';
 import {TuiIcon} from '@taiga-ui/core/components/icon';
 import {TUI_TEXTFIELD_OPTIONS} from '@taiga-ui/core/components/textfield';
-import {TUI_COMMON_ICONS, tuiAsAuxiliary} from '@taiga-ui/core/tokens';
+import {
+    TUI_COMMON_ICONS,
+    TUI_DATE_FORMAT,
+    TUI_TEXTFIELD_VALUE,
+    tuiAsAuxiliary,
+} from '@taiga-ui/core/tokens';
 import {type TuiSizeL, type TuiSizeS} from '@taiga-ui/core/types';
 import {TUI_OTHER_DATE_TEXT} from '@taiga-ui/kit/tokens';
 
@@ -50,21 +57,19 @@ import {type TuiDayRangePeriod} from './day-range-period';
         tuiProvide(AbstractTuiCalendar<TuiDayRange>, TuiCalendarRange),
         tuiCalendarSheetOptionsProvider({rangeMode: true}),
     ],
-    host: {
-        '[class._mobile]': 'mobile',
-        '(document:keydown.capture)': 'onEsc($event)',
-    },
+    host: {'[class._mobile]': 'mobile'},
 })
 export class TuiCalendarRange
     extends AbstractTuiCalendar<TuiDayRange>
-    implements OnInit, OnChanges
+    implements OnInit, OnChanges, OnDestroy
 {
     /**
      * @deprecated use `item`
      */
     private selectedPeriod: TuiDayRangePeriod | null = null;
+    private readonly format = inject(TUI_DATE_FORMAT);
+    private readonly textfieldValue = inject(TUI_TEXTFIELD_VALUE, {optional: true});
 
-    protected previousValue: TuiDay | TuiDayRange | null = null;
     protected hoveredItem: TuiDay | null = null;
     protected readonly otherDateText = inject(TUI_OTHER_DATE_TEXT);
     protected readonly icons = inject(TUI_COMMON_ICONS);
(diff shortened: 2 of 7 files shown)


projects/kit/components/input-date-range/input-date-range.directive.ts
waterplea, line 1: I don't want this to touch `InputDateRange` directive. Let's instead optionally inject TUI_TEXTFIELD_VALUE (TuiTextfieldComponent.value in v4) and set this signal from within `CalendarRange` component.

projects/kit/components/calendar-range/calendar-range.component.ts
waterplea, line 1: Let's also add `ngOnDestroy` and set single day range value if at the time of destroying `CalendarRange` the range picking is not finished. Otherwise we would need to re-trigger unfinished validator and set first day when dropdown opens with unfinished input. Also our designer said he'd prefer single day range to be selected if you click away/press Esc after single day pick.

waterplea (changes requested):

projects/kit/components/input-date-range/input-date-range.directive.ts
splincode: Done. I removed the changes from `InputDateRange` and now optionally inject `TUI_TEXTFIELD_VALUE` directly into CalendarRange.

projects/kit/components/calendar-range/calendar-range.component.ts
splincode: Done. Added `ngOnDestroy` to commit an unfinished single-day selection as a single-day range.

projects/kit/components/calendar-range/calendar-range.component.ts
```diff
@@ -210,7 +214,12 @@ export class TuiCalendarRange
             this.item.set(this.findItemByDayRange(range));
             this.value.set(range);
         } else {
+            const {mode, separator} = this.format();
+
             this.currentValue.set(day);
+            (this.textfieldValue as WritableSignal<string> | null)?.set(
```
waterplea: I think it is supposed to be writable to begin with, seems like a mistake in token type that should be fixed in definition.

waterplea (approved):
Why did the screenshot fail and selected a disabled date?

projects/kit/components/calendar-range/calendar-range.component.ts
splincode: Fixed

projects/kit/components/calendar-range/calendar-range.component.ts
```diff
@@ -129,13 +134,12 @@ export class TuiCalendarRange
...
-        if (event.key !== 'Escape' || !(this.currentValue() instanceof TuiDay)) {
-            return;
-        }
+    public ngOnDestroy(): void {
+        const value = this.currentValue();
 
-        event.stopPropagation();
-        this.currentValue.set(this.previousValue);
+        if (value instanceof TuiDay) {
+            this.value.set(new TuiDayRange(value, value));
```
waterplea: This does not work, the value is not updated, maybe at destruction it already stops listening to the value changes in the effect?

projects/kit/components/calendar-range/calendar-range.component.ts
```diff
@@ -210,7 +213,12 @@ export class TuiCalendarRange
             this.item.set(this.findItemByDayRange(range));
             this.value.set(range);
         } else {
+            const {mode, separator} = this.format();
+
             this.currentValue.set(day);
+            this.textfieldValue?.set(
```
waterplea, line 219: This keeps caret at the start. We have code that handles caret position inside `tuiValue`, I guess we only need to restore caret position if the previous value was not empty.

splincode: Fixed, thanks! Updated `tuiValue` to restore the previous caret position only when the input value was not empty. For an empty input, the caret now stays at the end after the value is set. Added a regression test for this case as well.

projects/kit/components/calendar-range/calendar-range.component.ts
splincode: Fixed

projects/cdk/utils/dom/value.ts
```diff
@@ -80,18 +80,21 @@ export function tuiValue(
         }
 
         if (element?.matches(':focus') && 'selectionStart' in element) {
+            const restoreCaret = Boolean(element.value);
```
waterplea: You can just move this into the same `if` one line above, because the `else` condition already just tries to set the value without caret manipulation

projects/kit/components/calendar-range/calendar-range.component.ts
```diff
@@ -199,8 +203,18 @@ export class TuiCalendarRange
 
     protected onDayClick(day: TuiDay): void {
         const value = this.currentValue();
+        const min = this.capsMapper(this.min(), value, this.maxLength(), true);
```
waterplea: Why all of this was unnecessary before?

splincode: You're right, this was redundant. I added it only to handle a failing E2E case where the test helper uses `click({force: true})` on a disabled cell, effectively bypassing the normal interaction restrictions. Since TuiCalendar already handles min/max/disabledItemHandler, I removed the duplicated checks from onDayClick.

projects/cdk/utils/dom/value.ts
splincode: Good point

projects/kit/components/input-date/input-date.directive.ts
```diff
@@ -105,14 +105,23 @@ export abstract class TuiInputDateBase<
...
-    protected readonly calendarOut = effect(() => {
-        const value = this.calendar()?.value();
-        const changed = untracked(() => value !== this.toCalendarValue(this.value()));
-        const same = value instanceof TuiDayRange && value.from === value.to;
+    protected readonly calendarOut = effect((onCleanup) => {
+        const calendar = this.calendar();
 
-        if (value && changed && !same) {
-            this.setDate(value);
+        if (!calendar) {
```
waterplea: Do we need it? What if:
```js
const sub = this.calendar()?.value.subscribe(...);

onCleanup(() => sub?.unsubscribe())';
```
Overall feels a bit scary, are we not making any leaks here like keeping a reference to the value signal of destroyed calendar?

waterplea (approved):

projects/kit/components/input-date/input-date.directive.ts
splincode: Updated it to subscribe to the current calendar value and unsubscribe with onCleanup, so we don't keep a reference to the destroyed calendar.

waterplea: Looks like you forgot to push

splincode: Oh, sorry

waterplea (approved):

nsbarsukov (approved):

PR #14866, opened 2026-08-30
fix(kit): close routable dialog after route params change

### LINKED ISSUE #10793: 🐞 - Routable dialog: An error occurred when navigating to the correct route but with different dynamic parameters.
### Reproduction url

Empty

### Description

I am currently using version v4.29.0.
I have the following route configuration:
<pre lang="markdown">
{
  path: `${AdminAppRoute.Details}/:id`,
  children: [
    tuiGenerateDialogableRoute(
      DetailsDialogComponent,
      actionsLogsDetailsDialogOptions,
    ),
  ],
},

const actionsLogsDetailsDialogOptions: RoutableDetailsDialogOptions = {
  dismissible: false,
  size: 'auto',
  data: {
    component: ActionsLogsDetailsComponent,
  },
};</pre>
When I navigate to _"/details/1"_ and click the close button, the app navigates back to _"/list"_ as expected. However, an issue occurs when I navigate to /details/1 and then, from there, click a button that navigates to _"/details/12"_. In this case, clicking the close button keeps the app at _"/details/12"_ instead of navigating back to _"/list"_.

### Taiga UI version

4.29.0

### Browser Used

Chrome

### OS Used

macOS

splincode:
Fixes #10793

### DIFF
--- a/projects/kit/components/routable-dialog/routable-dialog.component.ts
+++ b/projects/kit/components/routable-dialog/routable-dialog.component.ts
@@ -9,7 +9,7 @@ import {takeUntilDestroyed} from '@angular/core/rxjs-interop';
 import {ActivatedRoute, Router} from '@angular/router';
 import {TuiDialogService} from '@taiga-ui/core/portals/dialog';
 import {PolymorpheusComponent} from '@taiga-ui/polymorpheus';
-import {delay, from, of, switchMap} from 'rxjs';
+import {delay, from, of, switchMap, tap} from 'rxjs';
 
 @Component({
     selector: 'tui-routable-dialog',
@@ -20,7 +20,6 @@ export default class TuiRoutableDialog {
     private readonly route = inject(ActivatedRoute);
     private readonly router = inject(Router);
     private readonly injector = inject(INJECTOR);
-    private readonly initialUrl = this.router.url;
     private readonly dialog = inject(TuiDialogService);
 
     constructor() {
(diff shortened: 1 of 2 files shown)


projects/kit/components/routable-dialog/routable-dialog.component.ts
```diff
@@ -39,21 +38,16 @@ export default class TuiRoutableDialog {
...
-            .subscribe({complete: () => this.onDialogClosing()});
+            .subscribe();
     }
 
     private get lazyLoadedBackUrl(): string {
         return (this.route.parent?.snapshot.url || []).map(() => '..').join('/');
     }
 
-    private onDialogClosing(): void {
-        if (this.initialUrl === this.router.url) {
```
waterplea, line 52: Are we sure we no longer need this check? Is the new behavior to close and re-open the dialog if we go from `details/1` to `details/12` URLs as described in original issue?

projects/kit/components/routable-dialog/routable-dialog.component.ts
```diff
@@ -39,21 +38,16 @@ export default class TuiRoutableDialog {
                         this.route.snapshot.data['dialogOptions'],
                     ),
                 ),
+                tap({complete: () => this.navigateToParent()}),
```
waterplea, line 41: Why not in `subscribe`?

waterplea (approved):

projects/kit/components/routable-dialog/routable-dialog.component.ts
splincode: Because `subscribe({complete})` would also be called when `takeUntilDestroyed()` completes the downstream observable on component destruction.

We only want to navigate when the dialog observable itself completes. Keeping `tap({complete})` before `takeUntilDestroyed()` lets us distinguish dialog closing from route/component destruction.

projects/kit/components/routable-dialog/routable-dialog.component.ts
splincode: The check is no longer needed because we now distinguish dialog completion from component destruction instead of comparing URLs.

Changing only route params (`details/1` to `details/12`) reuses the same route/component, so the dialog stays open. When the user closes it, the dialog observable completes and we navigate to the parent.

If navigation goes to another route, the component is destroyed and `takeUntilDestroyed()` unsubscribes from the upstream observable, so `tap({complete})` is not called.

mdlufy (approved):

PR #14870, opened 2026-08-31
feat(layout): `List` add definition list support (#14867)

Yanduz:
Fixes #14867

### DIFF
--- a/projects/layout/directives/list/list.style.less
+++ b/projects/layout/directives/list/list.style.less
@@ -73,3 +73,79 @@
         }
     }
 }
+
+[tuiListV='@{tui-version}']:is(dl) {
+    display: grid;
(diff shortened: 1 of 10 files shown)


projects/layout/directives/list/list.style.less
```diff
@@ -1,6 +1,6 @@
 @import '@taiga-ui/core/styles/taiga-ui-local';
 
-[tuiListV='@{tui-version}'] {
+[tuiListV='@{tui-version}']:is(ul, ol) {
```
waterplea: This bumps specificity, maybe we can avoid this and just add `padding: 0` inside :is(dl)?

waterplea (approved):

mdlufy (approved):

PR #14876, opened 2026-09-01
fix(addon-mobile): mobile dropdown sheet reliably reopens after dismiss

mdlufy:
Mobile `tuiDropdownHover` renders as a sheet dialog. Dismissing the sheet completes the dialog but not the launcher's stream (`tuiIfMap` swallows it), so the dropdown itself wasn't closed — it relied on the mobile hover-reset (`mouseover(backdrop)` → `isHovered=false`), whose event ordering is browser-dependent. In Chrome this wedged after a few open/close cycles and the sheet stopped reopening. Safari was fine.

Fix: close the dropdown when the sheet dialog completes, so `ref` clears and the hover state resets, letting it reopen reliably on the next tap. Scoped to the mobile launcher `TuiDropdownSheetComponent`; where the hover-reset already worked the extra `toggle(false)` is a no-op, so desktop and Safari are unchanged.

Verified on real Chrome and Safari.

### DIFF
--- a/projects/addon-mobile/directives/dropdown-sheet/dropdown-sheet.component.ts
+++ b/projects/addon-mobile/directives/dropdown-sheet/dropdown-sheet.component.ts
@@ -10,6 +10,7 @@ import {TuiSheetDialogService} from '@taiga-ui/addon-mobile/components/sheet-dia
 import {tuiIfMap} from '@taiga-ui/cdk/observables';
 import {TuiDropdownDirective} from '@taiga-ui/core/portals/dropdown';
 import {PolymorpheusOutlet} from '@taiga-ui/polymorpheus';
+import {finalize} from 'rxjs';
 
 import {TuiDropdownSheet} from './dropdown-sheet.directive';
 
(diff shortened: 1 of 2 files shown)


mdlufy:
Closing in favor of #14892.

This PR started from a misdiagnosis ("mobile sheet closes on hover"). Investigation showed two separate things:

- The **device-reproducible bug** — the mobile dropdown sheet not reopening after a few open/close cycles in Chrome — is fixed in #14892 (`finalize` → close the dropdown when the sheet dialog completes). Verified on real Chrome and Safari, with a regression test.
- The **CI E2E failure** of `Closes dropdown on click on overlay` on webkit/firefox is a separate issue: the hover-driven mobile sheet closes on `mouseover` of the overlay (an effect of #14868 where focus isn't in the dropdown zone), so Playwright can't complete the click. The zone-attachment approaches tried here did not fix webkit and regressed Safari, so this direction is abandoned. To be handled separately (either a more robust E2E interaction or a targeted look at #14868 on webkit).

PR #14884, opened 2026-09-02
feat(core): support ShrinkWrap in Hint

### LINKED ISSUE #14881: 🚀 - support ShrinkWrap for balanced tooltip width
## Description

`TuiHint` has a fixed `max-width`, so multiline text can sometimes look visually unbalanced.

For example, when we need to prevent hanging prepositions in Russian typography, using explicit `<br>` produces undesirable results: the hint keeps its maximum width while the text itself occupies a much narrower area.

Without explicit line breaks, browser wrapping can also produce an uneven "Christmas tree" shape.

Example:

```text
Информация о сотруднике доступна только
для пользователей с соответствующими
правами доступа
```

Instead of using `<br>`, non-breaking spaces (`&nbsp;` / `U+00A0`) can be used in places where a line break should not happen.

However, it would be useful if `TuiHint` could also shrink its container to the actual balanced text width.

## Proposed solution

Consider applying `TuiShrinkWrap` to `TuiHint`, similarly to how it is currently used in `TuiToast`.

This could allow hints to:

* keep the configured maximum width;
* shrink to a more natural content width;
* produce more visually balanced multiline text;
* avoid using hardcoded `<br>` for visual formatting.

## Reproduction

Current behavior with `max-width: 20rem`:

1. Text with natural wrapping can produce uneven line lengths.
2. Adding `<br>` improves the line breaks but leaves unnecessary empty horizontal space inside the hint.
3. Using non-breaking spaces prevents undesirable breaks but does not solve the container width itself.

Image

## Expected behavior

The hint should keep its maximum width constraint but shrink to a balanced intrinsic width when the content does not need the full available width.

It would also be useful to add this case to the Hint demo as an example of long multiline text with non-breaking spaces.

splincode:
Fixes #14881


projects/core/portals/hint/hint.component.ts
```diff
@@ -51,18 +56,22 @@ const ARROW_OFFSET = 22;
...
     selector: 'tui-hint',
-    imports: [PolymorpheusOutlet],
+    imports: [PolymorpheusOutlet, TuiShrinkWrap],
     template: `
         <ng-content />
-        <span
-            *polymorpheusOutlet="content() as text; context: hint.context()"
-            [innerHTML]="text"
-        ></span>
+        @if (content()) {
```
waterplea: Why `@if` and what about `<ng-content />`? Shouldn't it wrap the whole thing?

projects/core/portals/hint/hint.component.ts
```diff
@@ -90,6 +99,11 @@ export class TuiHintComponent {
         ?.getAttribute('tuiTheme');
 
     protected readonly appearance = tuiAppearance(this.hint.appearance);
+    protected readonly maxWidth = tuiDirectiveBinding(
+        TuiShrinkWrapDirective,
+        'tuiShrinkWrap',
+        'min(20rem, calc(100% - 1rem))',
```
waterplea: These things are typically static, I wonder if we should use DI for defauilt value/options instead

waterplea (commented):
For some reason the width gets balanced only after the pop-in animation plays out

projects/core/portals/hint/hint.component.ts
splincode: Fixed

projects/core/portals/hint/hint.component.ts
splincode: Fixed

projects/core/portals/hint/hint.component.ts
```diff
@@ -51,17 +53,28 @@ const ARROW_OFFSET = 22;
...
-    imports: [PolymorpheusOutlet],
+    imports: [PolymorpheusOutlet, TuiShrinkWrap],
     template: `
         <ng-content />
-        <span
-            *polymorpheusOutlet="content() as text; context: hint.context()"
-            [innerHTML]="text"
-        ></span>
+
+        @if (text(); as text) {
```
waterplea: Why do we have to differentiate between text and template?

splincode: Because TuiShrinkWrap currently wraps its projected content in a `<span>`. A template can contain block elements, and wrapping such content changed the existing Hint layout, it caused visual regressions in the mobile/custom Hint screenshot tests.

For plain text this is safe and covers the use case from the issue, so I limited ShrinkWrap to string content. Supporting arbitrary template content would probably require making TuiShrinkWrap itself safe for block content, which feels like a separate change.

waterplea: I don't like this, let's roll it back to be always enabled and later I'll think about how to keep it safe for custom templates

splincode: Done, rolled it back

PR #14888, opened 2026-09-02
feat(kit): `LineClamp` improve performance

### LINKED ISSUE #12696: 🚀 - Improving the performance of the tui-line-clamp component
### Description

The current implementation of the `LineClamp` component triggers layout calculations immediately on render. When many instances of `LineClamp` are present on the page (for example, in a table), this causes significant performance issues due to multiple reflows during initial rendering.
See the attached performance screenshot:
Image

### **Proposed Solution**

Instead of performing the overflow calculation during initial render, defer this calculation until a meaningful interaction occurs — when the user hovers over the component.

This would remove unnecessary reflows during initial paint.

### **Notes**

The `overflowChange` output may need reconsideration. A possible optimization: only emit it if there is at least one active subscriber for this output.

### **Alternative Idea**

Extract the overflow calculation into a separate directive so it only runs for components that actually need it (e.g., those with subscribers to `overflowChange`).

vladimirpotekhin:
Fixes #12696

### DIFF
--- a/projects/kit/components/line-clamp/line-clamp.style.less
+++ b/projects/kit/components/line-clamp/line-clamp.style.less
@@ -18,3 +18,15 @@
     overflow: hidden;
     overflow-wrap: anywhere;
 }
+
+@supports (animation-timeline: --name) {
+    @keyframes tuiEmpty {
+        /* comment so that the preprocessor doesn't delete it */
+    }
+
+    .t-wrapper {
+        max-block-size: inherit; // makes the wrapper itself overflow, otherwise detection would depend on `_overflown` it sets
+        animation-name: tuiEmpty, tuiEmpty;
+        animation-timeline: scroll(self block), scroll(self inline);
+    }
+}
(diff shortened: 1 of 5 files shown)


projects/kit/components/line-clamp/line-clamp.style.less
```diff
@@ -18,3 +18,15 @@
     overflow: hidden;
     overflow-wrap: anywhere;
 }
+
+@supports (animation-timeline: --name) {
+    @keyframes tuiEmpty {
```
waterplea, line 23: Maybe we should create global `@tuiDummy` so we can use both here and in tuiPresent?

waterplea (approved):

projects/kit/components/line-clamp/line-clamp.style.less
vladimirpotekhin: ok, I'll do it in a separate pr

nsbarsukov (approved):

Bykiev:
Can you please cherry-pick thees changes to v4 too?

waterplea:
Yes, we will by the next week's release

PR #14889, opened 2026-09-02
fix(kit): `InputFiles` support `accept` / `maxFileSize` validation with signal forms

nsbarsukov:
### Problem 1

```html
<label tuiInputFiles>
  <input
    multiple
    tuiInputFiles
    [accept]="accept()"
    [formField]="f.files"
    [maxFileSize]="maxFileSize()"
    (reject)="onFieldReject($event)"
  />
</label>
```

throws

```
ERROR TypeError: control?.getError is not a function
    at tuiFilesRejected (files.utils.ts:11:37)
    at input-files.directive.ts:54:23
    at map.js:7:37
    at OperatorSubscriber2._this._next (OperatorSubscriber.js:15:21)
    at Subscriber2.next (Subscriber.js:34:18)
    at switchMap.js:14:159
    at OperatorSubscriber2._this._next (OperatorSubscriber.js:15:21)
    at Subscriber2.next (Subscriber.js:34:18)
    at distinctUntilChanged.js:15:28
    at OperatorSubscriber2._this._next (OperatorSubscriber.js:15:21)
```

### Problem 2
After fixing the 1st problem
```ts
it('a file of the wrong format is shown as rejected, not as accepted', () => {
    cy.get('@input').selectFile(NOT_AN_IMAGE, {force: true});

    cy.get('#rejected tui-file').should('have.length', 1);
    // [...]
});
```

```
AssertionError: Timed out retrying after 10000ms: 
Expected to find element: `#rejected tui-file`, but never found it.
```

___

```ts
it('a file over the size limit is shown as rejected, not as accepted', () => {
    cy.get('@input').selectFile(HUGE_IMAGE, {force: true});

    cy.get('#rejected tui-file').should('have.length', 1);
    // [...]
});
```

```
AssertionError: 
Timed out retrying after 10000ms:
Expected to find element: `#rejected tui-file`, but never found it.
```

Relates:

### DIFF
--- a/projects/kit/components/files/input-files/input-files-validator.directive.ts
+++ b/projects/kit/components/files/input-files/input-files-validator.directive.ts
@@ -1,9 +1,10 @@
 import {Directive, inject, type OnChanges, type OnInit} from '@angular/core';
 import {
-    AbstractControl, NG_VALIDATORS,
+    type AbstractControl,
+    NG_VALIDATORS,
     type ValidationErrors,
     type ValidatorFn,
-    Validators
+    Validators,
 } from '@angular/forms';
 import {TuiValidator} from '@taiga-ui/cdk/directives/validator';
 import {tuiProvide} from '@taiga-ui/cdk/utils/di';
@@ -27,8 +28,8 @@ export class TuiInputFilesValidator extends TuiValidator implements OnInit, OnCh
     public accept = this.options.accept;
     public maxFileSize = this.options.maxFileSize;
 
-    public override validate = 
-        (control: AbstractControl): ValidationErrors | null => this.validator(control);
+    public override validate = (control: AbstractControl): ValidationErrors | null =>
+        this.validator(control);
 
     public override ngOnChanges(): void {
         this.update();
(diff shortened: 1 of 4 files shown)


waterplea (approved):

projects/kit/components/files/input-files/input-files-validator.directive.ts
```diff
@@ -30,7 +40,7 @@ export class TuiInputFilesValidator extends TuiValidator implements OnInit, OnCh
     }
 
     private update(): void {
-        this.validate =
```
nsbarsukov, line 33: Earlier I wrote that signal forms aren't compatible with `NG_VALIDATORS`.
That's still the case — I wrote about it here and provided proof with a sandbox
https://github.com/angular/angular/issues/66232#issuecomment-5069162406

But it turned out that `NG_VALIDATORS` doesn't work if there's **no custom** `NG_VALUE_ACCESSOR` above it.
And if there is one, then it works
Here you can check out the logs in the console
https://stackblitz.com/edit/ng-signal-form-template-driven-validator-without-cva?file=src%2Fdumb-cva.ts

**BUT!** You can't overwrite it. Under the hood Angular only binds this method once.
If you overwrite the property at runtime, it ignores the overwrite.

vladimirpotekhin (approved):

PR #14894, opened 2026-09-03
fix(schematics): migrate kit i18n token providers to signal (v5)

mdlufy:
Kit i18n tokens built via `tuiExtractI18n` changed from `Observable<T>` to `Signal<T>` in v5, so `{provide, useValue: of(...)}` providers stop working after `ng update`. Only `TUI_LANGUAGE` and the `TUI_DOC_*` family were covered before.

Adds a migration for the kit i18n token family:

- `useValue: of(x)` → `useValue: signal(x)`
- plain literal `useValue` → wrapped in `signal(...)`
- non-convertible values (streams, `useFactory`) → TODO

`signal` import is added and deduped; values already wrapped in `signal(...)` are left untouched, so re-runs are safe.

### DIFF
--- a/projects/cdk/schematics/ng-update/v5/tests/__snapshots__/schematic-migrate-i18n-signal-tokens.spec.ts.snap
+++ b/projects/cdk/schematics/ng-update/v5/tests/__snapshots__/schematic-migrate-i18n-signal-tokens.spec.ts.snap
@@ -1,0 +1,357 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ng-update migrateI18nSignalTokens adds a TODO when useFactory has a block body: test.ts 1`] = `
+{
+  "0. Before": "
+                import {Component} from '@angular/core';
+                import {TUI_DIGITAL_INFORMATION_UNITS} from '@taiga-ui/kit';
+                import {of} from 'rxjs';
+
+                @Component({
+                    standalone: true,
+                    providers: [
+                        {
+                            provide: TUI_DIGITAL_INFORMATION_UNITS,
+                            useFactory: () => {
+                                return of(['Б']);
+                            },
+                        },
+                    ],
+                })
+                export class TestComponent {}
+            ",
+  "1. After": "
+                import {Component} from '@angular/core';
+                import {TUI_DIGITAL_INFORMATION_UNITS} from '@taiga-ui/kit';
+                import {of} from 'rxjs';
+
+                @Component({
+                    standalone: true,
+                    providers: [
+                        {
+// TODO: (Taiga UI migration) i18n token is a Signal in v5, not an Observable. Provide the value via signal(...), or convert an existing stream with toSignal(...).
+                            provide: TUI_DIGITAL_INFORMATION_UNITS,
+                            useFactory: () => {
+                                return of(['Б']);
+                            },
+                        },
+                    ],
+                })
+                export class TestComponent {}
+            ",
+}
+`;
+
+exports[`ng-update migrateI18nSignalTokens leaves an already signal-wrapped value untouched while migrating siblings: test.ts 1`] = `
+{
+  "0. Before": "
+                import {Component, signal} from '@angular/core';
+                import {
+                    TUI_DIGITAL_INFORMATION_UNITS,
+                    TUI_FILE_TEXTS,
+                } from '@taiga-ui/kit';
+                import {of} from 'rxjs';
+
+                @Component({
+                    standalone: true,
+                    providers: [
+                        {provide: TUI_DIGITAL_INFORMATION_UNITS, useValue: signal(['Б'])},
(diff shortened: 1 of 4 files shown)


projects/cdk/schematics/ng-update/v5/tests/__snapshots__/schematic-migrate-i18n-signal-tokens.spec.ts.snap
```diff
@@ -0,0 +1,277 @@
...
+                import {Component, inject} from '@angular/core';
+                import {TUI_DIGITAL_INFORMATION_UNITS} from '@taiga-ui/kit';
+
+                @Component({
+                    standalone: true,
+                    providers: [
+                        {
+// TODO: (Taiga UI migration) i18n token is a Signal in v5, not an Observable. Provide the value via signal(...), or convert an existing stream with toSignal(...).
+                            provide: TUI_DIGITAL_INFORMATION_UNITS,
+                            useValue: inject(SomeService).units$,
```
nsbarsukov: Could migration automatically wraps any observable inside `toSignal` ?

```ts
{
    provide: TUI_DIGITAL_INFORMATION_UNITS,
    useValue: anyObservable$.pipe(...),
},
```
⬇️ 
```ts
{
    provide: TUI_DIGITAL_INFORMATION_UNITS,
        useValue: toSignal(anyObservable$.pipe(...)),
},
```
___

```ts
{
    provide: TUI_DIGITAL_INFORMATION_UNITS, 
    useFactory: () => of(['Б']),
},
```
⬇️ 
```ts
{
    provide: TUI_DIGITAL_INFORMATION_UNITS,
    useFactory: () => toSignal(of(['Б'])),
},
```
(yep, double conversion is better than manual migration)

mdlufy: Done — streams are auto-wrapped now. One correctness note: `toSignal()` asserts an injection context, and a static `useValue` is not one (it is evaluated at class-definition time when the providers array is built → NG0203), so I wrap via a factory instead:

- `useValue: obs$` → `useFactory: () => toSignal(obs$)`
- `useFactory: () => of(x)` → `useFactory: () => toSignal(of(x))` (params/`deps` preserved)

`of(x)` and plain literals still collapse to `signal(x)` (valid in a static `useValue`). The TODO now stays only for cases we cannot safely transform (block-body factories, `useClass`/`useExisting`).

projects/cdk/schematics/ng-update/v5/tests/__snapshots__/schematic-migrate-i18n-signal-tokens.spec.ts.snap
```diff
@@ -0,0 +1,357 @@
...
+                import {of} from 'rxjs';
+
+                @Component({
+                    standalone: true,
+                    providers: [
+                        {
+// TODO: (Taiga UI migration) i18n token is a Signal in v5, not an Observable. Provide the value via signal(...), or convert an existing stream with toSignal(...).
+                            provide: TUI_DIGITAL_INFORMATION_UNITS,
+                            useFactory: () => {
+                                return of(['Б']);
```
nsbarsukov, line 35: Why not `toSignal` in the same way as this case below ?

```
  "0. Before": "
                import {Component, inject} from '@angular/core';
                import {TUI_DIGITAL_INFORMATION_UNITS} from '@taiga-ui/kit';

                @Component({
                    standalone: true,
                    providers: [
                        {
                            provide: TUI_DIGITAL_INFORMATION_UNITS,
                            useValue: inject(SomeService).units$,
                        },
                    ],
                })
                export class TestComponent {}
            ",
  "1. After": "import { toSignal } from "@angular/core/rxjs-interop";

                import {Component, inject} from '@angular/core';
                import {TUI_DIGITAL_INFORMATION_UNITS} from '@taiga-ui/kit';

                @Component({
                    standalone: true,
                    providers: [
                        {
                            provide: TUI_DIGITAL_INFORMATION_UNITS,
                            useFactory: () => toSignal(inject(SomeService).units$),
                        },
                    ],
                })
                export class TestComponent {}
            ",
}
`;
```

mdlufy: Makes sense — a block body runs inside the factory's injection context exactly like the arrow case, so there's no reason to bail to a TODO there.

Block-body factories now wrap their returned stream too:

```ts
useFactory: () => {
    return of(['Б']);   // → return toSignal(of(['Б']))
}
```

It wraps every return the factory itself owns (so a return after some local `const` setup, or multiple/conditional returns, all work), and deliberately leaves nested-callback returns alone — a `return` inside `.pipe(map(x => { return x; }))` stays untouched. The only thing that still gets a TODO is a block with nothing to return (side-effect-only), where there's no stream to convert.

PR #14896, opened 2026-09-03
chore(demo-playwright): drive mobile dropdown-hover e2e with touch

mdlufy:
The mobile dropdown-hover e2e tests emulate a phone through userAgent only, but the webkit project runs Desktop Safari with hasTouch:false, so they drive a touch-only component (the sheet) with a mouse. Playwright's click() moves the pointer first, that mouseover opens the sheet via tuiDropdownHover, and the sheet backdrop then intercepts the actual click, so the test times out on webkit.

On real touch devices there is no hover, so the component itself works fine. This is the test's input model, not a product bug. The fix stays in the test: enable hasTouch for the mobile block and drive it with tap() instead of click().

### DIFF
--- a/projects/demo-playwright/tests/kit/dropdown-hover/dropdown-hover.pw.spec.ts
+++ b/projects/demo-playwright/tests/kit/dropdown-hover/dropdown-hover.pw.spec.ts
@@ -53,7 +53,7 @@ test.describe('DropdownHover', () => {
             let example!: Locator;
             let po!: TuiDocumentationPagePO;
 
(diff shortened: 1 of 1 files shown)


mdlufy:
Superseded by #14892, which now bundles these test changes (hasTouch + tap) together with the two product fixes they depend on: the core mouse-only-open fix (mobile sheet no longer vanishes on webkit/firefox) and the finalize reopen fix. The mobile e2e only goes green with all three, so keeping them in one PR.

PR #14913, opened 2026-09-06
test: disable CSS transitions in Cypress

splincode:
Disables Taiga UI CSS transitions in Cypress component tests by setting
`--tui-duration: 0ms` on `body`.

This keeps visual regression snapshots deterministic while preserving the
existing `TUI_ANIMATIONS_SPEED = 0` setup for DI-driven animations.

image

### DIFF
--- a/projects/demo-cypress/src/tests/stepper.cy.ts
+++ b/projects/demo-cypress/src/tests/stepper.cy.ts
@@ -1,35 +1,38 @@
 import {ChangeDetectionStrategy, Component} from '@angular/core';
+import {TuiRoot} from '@taiga-ui/core';
 import {TuiStepper} from '@taiga-ui/kit';
 import {TuiCard} from '@taiga-ui/layout';
 
 @Component({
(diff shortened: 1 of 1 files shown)


waterplea (approved):

projects/demo-cypress/src/support/component-index.html
```diff
@@ -20,6 +20,11 @@
         ></div>
 
         <style>
+            body {
+                /* Keep visual regression tests deterministic by disabling CSS transitions */
+                --tui-duration: 0ms;
+            }
```
vladimirpotekhin: Maybe wrap the stepper test in tui-root? That way, we can also verify that the component responds to prefers-reduced-motion

waterplea: Good idea

splincode: fixed

mdlufy (approved):

PR #14915, opened 2026-09-06
fix(addon-mobile): prevent sheet dismissal on focus loss

### LINKED ISSUE #12975: 🐞 - tuiDropdownMobile gets closed unexpectedly (ios)

splincode:
Fixes #12975

### DIFF
--- a/projects/cdk/directives/active-zone/active-zone.directive.ts
+++ b/projects/cdk/directives/active-zone/active-zone.directive.ts
@@ -55,6 +55,12 @@ export class TuiActiveZone implements OnDestroy {
         );
     }
 
+    public containsZone(selector: string): boolean {
+        return this.children.some(
+            (zone) => zone.el.matches(selector) || zone.containsZone(selector),
+        );
+    }
+
     // issue: https://github.com/typescript-eslint/typescript-eslint/issues/11770
     // eslint-disable-next-line @typescript-eslint/no-unused-private-class-members
     private addSubActiveZone(activeZone: TuiActiveZone): void {
--- a/projects/cdk/directives/active-zone/tests/active-zone.directive.spec.ts
+++ b/projects/cdk/directives/active-zone/tests/active-zone.directive.spec.ts
@@ -1,6 +1,7 @@
 import {ChangeDetectionStrategy, Component} from '@angular/core';
 import {type ComponentFixture, TestBed} from '@angular/core/testing';
 import {FormControl, ReactiveFormsModule} from '@angular/forms';
+import {By} from '@angular/platform-browser';
 import {TuiActiveZone} from '@taiga-ui/cdk';
 import {provideTaiga} from '@taiga-ui/core';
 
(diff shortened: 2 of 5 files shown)


projects/core/portals/dropdown/dropdown.providers.ts
```diff
@@ -10,6 +10,11 @@ export const TUI_DROPDOWN_COMPONENT = new InjectionToken<Type<any>>(
     {factory: () => TuiDropdownComponent},
 );
 
+export const TUI_DROPDOWN_CLOSE_ON_ACTIVE_ZONE = new InjectionToken<boolean>(
+    ngDevMode ? 'TUI_DROPDOWN_CLOSE_ON_ACTIVE_ZONE' : '',
+    {factory: () => true},
+);
```
mdlufy: Is there another fix ways to avoid creating new token?

splincode: I couldn't find a cleaner way without introducing some configuration point.

We still need the regular dropdown to close when it leaves the active zone, while `tuiDropdownSheet` should skip this behavior only on mobile. I wouldn't want to put `WA_IS_MOBILE` or addon-mobile-specific logic into the core dropdown.

The token keeps the existing behavior by default and allows `tuiDropdownSheet` to override it locally. If you have another approach in mind, I can try it.

waterplea: I don't like this either, let's keep this open until a better idea comes up

splincode: Reworked it without introducing a new token.

I added TuiActiveZone.containsZone(selector) to query attached child zones and now the dropdown simply checks whether its active zone contains an open tui-modal.

This keeps the existing behavior for regular dropdowns and avoids addon-mobile-specific configuration in core.

mdlufy (approved):

waterplea:
This issue is not zone's business, we leave zone. Sure we would wish to catch this case an not leave it, but this seems impossible thanks to brilliant apple engineers, as usual.

Instead a suggest we do the following:
We already have `tuiDropdownEnabled` which we use to stop dropdown from opening. If we also make dropdown ignore close signals when tuiDropdownEnabled === false – we can then set it to false when DropdownSheet is opened. It is getting closed by the means of SheetDialog tap outside/swipe down anyway, we do not really need zone in that case.

PR #14917, opened 2026-09-06
fix(addon-mobile): close sheet dialog before scroll snap ends

### LINKED ISSUE #12355: 🐞 - SheetDialog close animation is extremely slow on IOS

splincode:
Fixes #12355

### DIFF
--- a/projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
+++ b/projects/addon-mobile/components/sheet-dialog/sheet-dialog.component.ts
@@ -2,6 +2,7 @@ import {
     afterNextRender,
     ChangeDetectionStrategy,
     Component,
+    DestroyRef,
     ElementRef,
     inject,
     viewChildren,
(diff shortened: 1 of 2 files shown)


waterplea:
I think the previous attempt was better. We watched where is the scroll going in that attempt but our problem was we were only watching until next scroll event – instead we should watch for multiple scroll events and check if it keeps going down and not slowing down. Our previous attempt didn't work because when you slightly swiped down, but not enough to close – it would still scroll down a little before slowing down and reversing direction to snap to the previous snap point. It's better to revive that old PR and change it accordingly.

splincode:
@waterplea Thanks! I reworked it based on the previous approach.

Now we track multiple scroll events after touch end/cancel and only close the sheet if it keeps moving down without slowing down. If the scroll slows down or reverses towards the previous snap point, we stop tracking and keep the sheet open.

I also added tests for both cases, including snapping between open stops.

PR #14936, opened 2026-09-08
feat(addon-doc): add "Copy page" action to documentation pages

mdlufy:
## What

Adds a generic **`TUI_DOC_ACTIONS`** content slot to `tui-doc-page` (mirroring `TUI_DOC_LOGO`) — a place for a docs app to pin actions to the trailing edge of the page header. The Taiga docs fill it with a **Copy page** split-button that copies the page as Markdown (heading, description, examples, API) for pasting into an LLM/agent, with a **View as Markdown** item that opens the raw `.md`.

The button, its Markdown fetching, clipboard and the `<link>` discoverability hints all live in the **demo app** — `@taiga-ui/addon-doc` stays free of copy-page/markdown/llms machinery.

## How

- `@taiga-ui/addon-doc` exposes `TUI_DOC_ACTIONS: InjectionToken<PolymorpheusContent>` (default empty). `TuiDocPage` renders it via `*polymorpheusOutlet` in a right-aligned header slot and injects nothing by default — external consumers pay for none of this.
- The demo provides a `CopyPage` component through the slot (`{provide: TUI_DOC_ACTIONS, useValue: new PolymorpheusComponent(CopyPage)}`, mirroring `LOGO_CONTENT` for `TUI_DOC_LOGO`). It resolves the current route to `/<route>.md`, copies via `@angular/cdk` `Clipboard`, and hides itself when no Markdown exists.
- Per-page `.md` generation reuses the existing llms extraction utils to emit one `.md` per page, mirrored to the page URL; it's wired into the `generate-llms` nx target (no standalone npm script).

## Discoverability for agents / crawlers

Provided by the demo's `CopyPage` (not the library), each doc page exposes its Markdown twin so agents and crawlers can find it without executing the app:

- **`<link rel="alternate" type="text/markdown" href="/<route>.md">`** is injected into `<head>` on every doc page. The demo is prerendered, so the link is baked into the static HTML — visible to non-JS crawlers.

> image

- **`<link rel="describedby" href="/llms.txt">`** is injected too, pointing agents at the `llms.txt` overview that covers the docs.
- **`llms.txt`** links straight to each page's Markdown twin, e.g.
  `- [Button](https://taiga-ui.dev/components/button.md)`

## Why a generic slot

Copy-page is docs-site-specific — Markdown twins, clipboard, `llms.txt` — and not the responsibility of a reusable doc-page shell. A generic `TUI_DOC_ACTIONS` slot keeps `addon-doc` free of that machinery while letting any docs app pin its own header actions. For example, Maskito could use the very same slot to pin a StackBlitz sandbox link for the current framework on its framework pages.


projects/addon-doc/components/page/copy-page/copy-page.component.ts
```diff
@@ -0,0 +1,95 @@
...
+    protected readonly icons = {copy: '@tui.copy', chevron: '@tui.chevron-down'} as const;
+
+    protected readonly markdown = toSignal(
+        this.router.events.pipe(
+            filter((event) => event instanceof NavigationEnd),
+            map(() => this.pagePath()),
+            startWith(this.pagePath()),
+            distinctUntilChanged(),
+            switchMap((page) =>
+                this.isBrowser ? from(this.fetchMarkdown(page)) : of(null),
```
vladimirpotekhin, line 52: I’m not sure we should make the button appear based solely on whether the MD file can be fetched. Since this is a public package, it would mean the feature is enabled for all products by default.

Would it make sense to add an option for this, defaulting to false, so that only products that need the “Copy” button (Taiga UI) can explicitly enable it?

projects/addon-doc/components/page/copy-page/copy-page.component.ts
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        await navigator.clipboard.writeText(markdown);
+        this.copied.set(true);
+        setTimeout(() => this.copied.set(false), 2000);
+    }
+
+    protected viewAsMarkdown(): void {
+        this.open.set(false);
+        window.open(`${this.pagePath()}.md`, '_blank');
```
vladimirpotekhin: use `WA_WINDOW` instead of global object

projects/addon-doc/components/page/copy-page/copy-page.template.html
```diff
@@ -0,0 +1,44 @@
...
+            appearance="flat-grayscale"
+            size="s"
+            tuiIconButton
+            type="button"
+            class="t-toggle"
+            [iconStart]="icons.chevron"
+            [tuiDropdown]="menu"
+            [tuiDropdownManual]="open()"
+            (click)="open.set(!open())"
+            (tuiActiveZoneChange)="onActiveZone($event)"
```
vladimirpotekhin: I think that we don't need `tuiActiveZoneChange`, just use `tuiDropdownOpen` https://taiga-ui.dev/directives/dropdown-open#with-custom-host

projects/addon-doc/components/page/copy-page/copy-page.template.html
```diff
@@ -0,0 +1,44 @@
...
+            type="button"
+            class="t-action"
+            [iconStart]="icons.copy"
+            (click)="copy()"
+        >
+            {{ copied() ? 'Copied!' : 'Copy page' }}
+        </button>
+        <span
+            aria-hidden="true"
+            class="t-divider"
```
vladimirpotekhin: maybe use `tuiGroup` directive ?

projects/addon-doc/components/page/copy-page/copy-page.template.html
```diff
@@ -0,0 +1,44 @@
...
+        </button>
+    </div>
+    <ng-template #menu>
+        <tui-data-list>
+            <button
+                tuiOption
+                type="button"
+                (click)="viewAsMarkdown()"
+            >
+                View as Markdown
```
vladimirpotekhin, line 36: It would be better if it were a link with target=_blank

projects/addon-doc/components/page/copy-page/copy-page.component.ts
mdlufy: Done. The button is now gated behind a new `TUI_DOC_COPY_PAGE` token (`InjectionToken<Signal<boolean>>`, default `signal(false)`), so it stays off for library consumers. The demo enables it with `signal(true)` in `app.config.ts`.

projects/addon-doc/components/page/copy-page/copy-page.component.ts
mdlufy: Resolved without `WA_WINDOW` — the menu item is now a plain `<a target="_blank" rel="noreferrer">` link, so `window.open` (global object) is gone entirely.

projects/addon-doc/components/page/copy-page/copy-page.template.html
mdlufy: Done. Dropped `tuiActiveZoneChange` and manual `(click)`/`[tuiDropdownManual]`; now uses `[(tuiDropdownOpen)]="open"` with `#tuiDropdownHost` on the toggle.

projects/addon-doc/components/page/copy-page/copy-page.template.html
mdlufy: Done. The pill is rendered with `tuiGroup` (`[collapsed]="true"`) using `outline-grayscale` buttons for a tight segmented look.

projects/addon-doc/components/page/copy-page/copy-page.template.html
mdlufy: Done. 'View as Markdown' is now an `<a tuiOption [href]="markdownUrl()" target="_blank" rel="noreferrer">`, styled as a plain menu item (`color: inherit; text-decoration: none`).

projects/addon-doc/components/page/copy-page/copy-page.template.html
```diff
@@ -0,0 +1,44 @@
+@if (enabled() && markdown()) {
+    <div
+        tuiGroup
+        class="t-copy-page"
```
waterplea: size="s"

projects/addon-doc/components/page/copy-page/copy-page.template.html
```diff
@@ -0,0 +1,44 @@
...
+        [collapsed]="true"
+        [tuiDropdown]="menu"
+        [(tuiDropdownOpen)]="open"
+    >
+        <button
+            appearance="outline-grayscale"
```
waterplea: Maybe it would look better in secondary-grayscale + non-collapsed group?

projects/addon-doc/components/page/copy-page/copy-page.template.html
```diff
@@ -0,0 +1,44 @@
...
+            size="s"
+            tuiIconButton
+            type="button"
+            class="t-toggle"
+            [iconStart]="icons.chevron"
+        >
+            More
+        </button>
+    </div>
+    <ng-template #menu>
```
waterplea: Let's use *tuiDropdown instead and move it under the tuiGroup

projects/addon-doc/components/page/copy-page/copy-page.template.html
```diff
@@ -0,0 +1,44 @@
...
+        <tui-data-list>
```
waterplea: Menu looks huge, we need size M or S here.

projects/addon-doc/components/page/copy-page/copy-page.style.less
```diff
@@ -0,0 +1,19 @@
...
+    flex: 0 0 auto;
+}
+
+a[tuiOption] {
+    color: inherit;
+    text-decoration: none;
+}
+
+tui-data-list [tuiOption] {
+    padding: 0.625rem 0.75rem;
```
waterplea: Don't override out-of-the-box styles

projects/addon-doc/components/page/copy-page/copy-page.style.less
```diff
@@ -0,0 +1,19 @@
...
+:host {
+    display: inline-block;
+    white-space: nowrap;
+}
+
+.t-copy-page > * {
+    flex: 0 0 auto;
+}
+
+a[tuiOption] {
```
waterplea: This shouldn't be necessary

projects/addon-doc/components/page/copy-page/copy-page.template.html
mdlufy: Done — both buttons and the `tui-data-list` are `size="s"` now.

projects/addon-doc/components/page/copy-page/copy-page.template.html
mdlufy: Done — switched to `secondary-grayscale` and dropped `[collapsed]`, so it is a plain non-collapsed group now.

projects/addon-doc/components/page/copy-page/copy-page.template.html
mdlufy: Done — the toggle button now hosts the dropdown (`tuiDropdown` + `[(tuiDropdownOpen)]`) with `<tui-data-list *tuiDropdown>` under the group; dropped the `#tuiDropdownHost` + `<ng-template #menu>` variant.

projects/addon-doc/components/page/copy-page/copy-page.template.html
mdlufy: Done — `tui-data-list size="s"`, the menu is compact now.

projects/addon-doc/components/page/copy-page/copy-page.style.less
mdlufy: Done — removed the `[tuiOption]` padding override.

projects/addon-doc/components/page/copy-page/copy-page.style.less
mdlufy: Done — removed; `tuiOption` on the `<a>` already renders with no underline and inherited color (verified in the browser), so it was indeed unnecessary.

projects/addon-doc/components/page/copy-page/copy-page.template.html
waterplea: The group itself also needs size="s" – it affects border-radius

projects/addon-doc/components/page/copy-page/copy-page.template.html
```diff
@@ -0,0 +1,44 @@
...
+        </button>
+        <button
+            appearance="secondary-grayscale"
+            size="s"
+            tuiDropdown
+            tuiIconButton
+            type="button"
+            class="t-toggle"
+            [iconStart]="icons.chevron"
+            [(tuiDropdownOpen)]="open"
```
waterplea: Let's align dropdown to the right:

Image

projects/addon-doc/components/page/copy-page/copy-page.template.html
```diff
@@ -0,0 +1,44 @@
+@if (enabled() && markdown()) {
+    <div
+        tuiGroup
```
waterplea: This is unnecessary DOM nesting, `Group` can just be added as a host directive and there will not be a div at all.

projects/addon-doc/components/page/copy-page/copy-page.template.html
mdlufy: Good catch — the outer radius was still `l`. It carries `size="s"` now as well. One wrinkle worth noting: `TuiGroupOptions.size` is typed `TuiSizeL` (only `m`/`l`), so I could not bake it in through `tuiGroupOptionsProvider`. Instead I exposed the group `size` input via the host directive and pass `size="s"` from the doc-page.

projects/addon-doc/components/page/copy-page/copy-page.template.html
mdlufy: Agreed, it reads much better anchored to the button. Added `tuiDropdownAlign="end"`, so the menu now opens flush with the right edge.

projects/addon-doc/components/page/copy-page/copy-page.template.html
mdlufy: Much cleaner this way, thanks. `TuiGroup` is a host directive now, so the wrapper `<div>` is gone and both buttons sit directly on the host element.

projects/addon-doc/components/page/copy-page/copy-page.template.html
waterplea: If group is not typed for S size it does not support S size. Most likely because only L size has increased radius in inputs and buttons. Let's go with M.

waterplea: image

mdlufy: Switched it to `m`. Both `s` and `m` resolve to the same group radius (`--tui-radius-m`, only `l` widens it), so the pill looks identical, just properly typed now. Buttons keep `size="s"`.

waterplea: There's no reason to expose it as an input, let's go with DI options

mdlufy: Updated in https://github.com/taiga-family/taiga-ui/pull/14936/changes/ae26b5f7a1c331798e8476ec38bd70a84e279cc1

projects/addon-doc/tokens/copy-page.ts
```diff
@@ -0,0 +1,6 @@
+import {InjectionToken, type Signal, signal} from '@angular/core';
+
+export const TUI_DOC_COPY_PAGE = new InjectionToken<Signal<boolean>>(
+    ngDevMode ? 'TUI_DOC_COPY_PAGE' : '',
+    {factory: () => signal(false)},
```
waterplea, line 5: Why is it a signal?

mdlufy: Mirroring the existing `TUI_DOC_SEARCH_ENABLED` token, which is a `Signal<boolean>` too — keeps the doc feature-flags consistent and lets a consumer flip Copy page reactively (per route/env). The template reads it straight in `@if` / the `<link>` effect, no extra wiring.

projects/addon-doc/components/page/copy-page/copy-page.component.ts
```diff
@@ -0,0 +1,82 @@
...
+    );
+
+    protected async copy(): Promise<void> {
+        const markdown = this.markdown();
+
+        if (!markdown) {
+            return;
+        }
+
+        await navigator.clipboard.writeText(markdown);
```
nsbarsukov: Why not Clipboard API from `@angular/cdk` ?

projects/addon-doc/components/page/copy-page/copy-page.component.ts
```diff
@@ -0,0 +1,82 @@
...
+        this.copied.set(true);
+        setTimeout(() => this.copied.set(false), 2000);
+    }
+
+    private async fetchMarkdown(url: string): Promise<string | null> {
+        try {
+            const response = await fetch(url);
```
nsbarsukov, line 72: Why fetch instead of angular http client ?

package.json
```diff
@@ -35,6 +35,7 @@
         "cypress:open": "cypress open --browser chrome --project ./projects/demo-cypress",
         "generate:llms": "ts-node ./projects/demo/scripts/llms/llms-generate.ts",
         "generate:llms-full": "ts-node ./projects/demo/scripts/llms/llms-full-generate.ts",
+        "generate:markdown-pages": "ts-node ./projects/demo/scripts/llms/markdown-pages-generate.ts",
```
nsbarsukov: Are all these scripts necessary in the root `package.json`? 

I think we should keep the `scripts` section lean and only include everyday commands.

projects/addon-doc/components/page/copy-page/copy-page.component.ts
mdlufy: `fetch` here keeps addon-doc provider-free — the lib doesn't pull in `HttpClient` anywhere today, so consumers aren't forced to add `provideHttpClient()` just for a Copy button. It also reads the raw `content-type` to skip the GitHub Pages SPA fallback (a missing `.md` comes back as `index.html`, 200 `text/html`), which I'd otherwise have to reconstruct via `observe: 'response'`.

Happy to move it to `HttpClient` if you'd rather keep it idiomatic though — the demo already provides it, so it's a tiny change.

projects/addon-doc/components/page/copy-page/copy-page.component.ts
mdlufy: Switched to CDK `Clipboard` — `copy()` is synchronous now and gets the `execCommand` fallback for free. db576385f15

package.json
mdlufy: Agreed — dropped the three `generate:*` scripts and inlined the `ts-node` calls straight into the `generate-llms` nx target, mirroring the neighbouring `generate-sitemap-file`. 55bf7636989

projects/addon-doc/components/page/page.component.ts
```diff
@@ -56,4 +61,33 @@ export class TuiDocPage {
...
+        // so agents/crawlers discover them without running the app.
+        const doc = inject(DOCUMENT);
+        const pageMarkdown = inject(TuiDocPageMarkdown);
+        const link = doc.createElement('link');
+        const describedBy = doc.createElement('link');
+
+        link.rel = 'alternate';
+        link.type = 'text/markdown';
+        describedBy.rel = 'describedby';
+        describedBy.href = '/llms.txt';
```
vladimirpotekhin: This code doesn’t make sense if the docs using addon-docs don’t have llms.txt pages and the "copy-page" feature is disabled.

mdlufy:
Reworked where this lives after the discussion: copy-page is really a docs-site concern (markdown twins, clipboard, `llms.txt`), not something the generic `tui-doc-page` shell should carry — and even when disabled it was still bundling the button + spinning up the markdown service.

So `addon-doc` now just exposes a generic `TUI_DOC_ACTIONS` content slot (same shape as `TUI_DOC_LOGO`), rendered pinned to the trailing edge of the page header. The whole Copy page implementation — button, `.md` fetch, `@angular/cdk` Clipboard, and the `<link rel="alternate"/"describedby">` hints — moved into the demo app and is provided into the slot via `new PolymorpheusComponent(...)`.

Net: the library carries zero copy-page/markdown/llms code and injects nothing by default (public API actually shrank — the `TUI_DOC_COPY_PAGE` boolean is gone). The same slot is reusable for any docs app — e.g. Maskito could pin a StackBlitz link for the current framework the same way.

projects/addon-doc/components/page/page.component.ts
mdlufy: Updated https://github.com/taiga-family/taiga-ui/pull/14936#issuecomment-5694944686

waterplea (approved):

projects/demo/src/pages/app/copy-page/copy-page.component.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    constructor() {
+        const doc = inject(DOCUMENT);
+        const link = doc.createElement('link');
+        const describedBy = doc.createElement('link');
+
+        link.rel = 'alternate';
+        link.type = 'text/markdown';
+        describedBy.rel = 'describedby';
+        describedBy.href = '/llms.txt';
```
vladimirpotekhin: this link doesn't change, why do we need remove / recreate it in copy-page button ?

vladimirpotekhin (approved):

projects/demo/src/pages/app/copy-page/copy-page.component.ts
mdlufy: No need to recreate, we can attach it statically to index.html

nsbarsukov (approved):

PR #14938, opened 2026-09-08
chore(demo): split AI docs into an "AI tools" section

mdlufy:
## What / why

The docs had a single, long "AI support" page. This splits it into a focused,
top-level **"AI tools"** nav section (sparkles icon, right after Getting Started)
with five short pages, each written as **clean flowing prose** (no boxed example
cards) — in the spirit of the shadcn/ui docs:

| Page | Route |
|---|---|
| Overview (hub) | `/ai/overview` |
| MCP server | `/ai/mcp` |
| Agent skills | `/ai/skills` |
| llms.txt | `/ai/llms` |
| Best practices | `/ai/best-practices` |

This PR also absorbs the previously-pending nav elevation + overview rewrite.

## Changes

- **New** `projects/demo/src/pages/info/ai/*` — five pages + a shared `styles.less`
  and the MCP/skills code snippets.
- **Nav** (`pages.ts`) — new `section: 'AI tools'` block (5 entries in display order)
  right after Getting Started; removed the old "AI support" entry from
  Documentation → More.
- **Order** (`app.config.ts`) — added `'AI tools'` to the `tuiSortPages` exclude set
  so the pages keep their manual order (Overview first), and `'AI tools': '@tui.sparkles'`
  to the section icons.
- **Routes** (`demo-routes.ts`, `app.routes.ts`) — five `/ai/*` routes + a redirect
  `/ai-support` → `/ai/overview` to preserve the old URL.
- **Getting started** — added an **"AI-Ready"** section with an "Explore AI tools" CTA
  linking to the hub.
- **Removed** the old `info/ai-support/` page.

## Note on the "On this page" ToC

These pages are intentionally plain prose (`<h2>` sections), so they don't render the
right-hand ToC — in this doc engine the ToC is generated only from `tui-doc-example`
headings, which also draw the bordered example cards we wanted to avoid here. Keeping
the ToC would require `::ng-deep` to strip the card chrome, which is deprecated.

## Verification

- `prettier --check` and `eslint` green on all new/changed files.
- `nx serve demo`: the "AI tools" section renders 5 pages in order (Overview first);
  every page loads; hub cross-links and the Getting-started CTA resolve to `/ai/*`;
  `/ai-support` redirects to `/ai/overview`. Verified in a real browser.

Demo-only, hence `chore(demo)`.


waterplea (approved):

projects/demo/src/pages/app/getting-started/index.html
```diff
@@ -84,6 +84,29 @@ <h2>Explore Taiga UI</h2>
                 </span>
             </button>
         </article>
+
+        <header tuiHeader="h6">
```
waterplea: This feels like it should have more space above it. Also this section and the one above it are missing from TOC menu.

projects/demo/src/pages/info/ai/overview/index.html
```diff
@@ -0,0 +1,117 @@
...
+                is a query interface over that same data, and skills are the method that consumes it. When they
+                disagree, the live source wins.
+            </div>
+        </tui-doc-example>
+
+        <tui-doc-example
+            heading="Start here"
+            [fullsize]="true"
+            [preview]="false"
+        >
```
nsbarsukov: Let's use the the same style as in https://taiga-ui.dev/getting-started
(group of cards)

Image

projects/demo/src/pages/info/ai/skills/index.html
```diff
@@ -0,0 +1,169 @@
...
+                    tuiLink
+                >
+                    <code>llms-full.txt</code>
+                </a>
+                , because Taiga UI evolves quickly. That keeps the method stable while the facts stay current.
+            </p>
+        </tui-doc-example>
+
+        <tui-doc-example
+            heading="Learn more"
```
nsbarsukov, line 123: The same here

projects/demo/src/pages/app/pages.ts
```diff
@@ -30,6 +30,35 @@ export const pages: DocRoutePages = [
         keywords: 'intro, how to, guide, main, главная, начало, инструкция',
         route: DemoRoute.GettingStarted,
     },
+    // AI tools
```
nsbarsukov: I'm not sure we should place the AI section before all other sections in the documentation. Taiga UI is components-first, so our structure should reflect that priority.

Let's use Angular's official documentation as a reference: https://angular.dev/overview

Their documentation follows this order:
- Introduction
- In-depth Guides
- Build with AI
...

Image

projects/demo/src/pages/app/getting-started/index.html
mdlufy: Both show up in the On this page menu now. I turned the two intro sections into flat `tui-doc-example` blocks, which gives them the anchored heading (and the extra top spacing you flagged) while the code examples above keep their framing.

projects/demo/src/pages/info/ai/overview/index.html
mdlufy: Reworked it into a card group in the getting-started style — the three layers are now `tuiCardLarge` cards linking to their pages.

projects/demo/src/pages/info/ai/skills/index.html
mdlufy: Same treatment here — the Learn more list is a card group now too.

projects/demo/src/pages/app/pages.ts
mdlufy: Agreed, components-first makes sense. Moved the whole AI tools section to the very end of the sidebar (after Tools).

nsbarsukov (approved):

PR #14951, opened 2026-09-09
fix(kit): prevent line clamp emit after destroy

### LINKED ISSUE #14949: 🐞 - `TuiLineClamp` emits `overflownChange` after destruction → NG0953
### Reproduction url

https://stackblitz.com/edit/xf9xzhen

### Description

When a `<tui-line-clamp>` is destroyed, the browser cancels the scroll-driven animation the component uses for overflow detection. The resulting `animationcancel` reaches the component's own listener, which then calls `overflownChange.emit()` on an already destroyed component. Angular reports:

```
NG0953: Unexpected emit for destroyed `OutputRef`. The owning directive/component is destroyed.
```

One warning per destroyed instance, so leaving a page that holds a list of clamped items floods the console — we get 5 at a time when navigating away from a product listing (the clamps sit inside product cards behind `@defer (on viewport)`).

Two things make this more than console noise:

- **Nobody has to subscribe to `overflownChange`.** In `OutputEmitterRef.emit` the `destroyed` check runs *before* the `listeners === null` check, so the warning fires even for an output that is never bound. Our application does not listen to `overflownChange` anywhere, and the reproduction below does not bind it either.
- **It is not dev-only.** `emit` does not guard the `console.warn` behind `ngDevMode` — only the message text is guarded — so the warning survives production builds, just without the descriptive text. Verified in a minified bundle: `destroyed){console.warn(lo(953,!1))`.

## Reproduction

https://stackblitz.com/edit/... ← *(paste link)*

Open the console and press **Destroy** once. Three clamps are destroyed → three NG0953 warnings.

Stack trace:

```
OutputEmitterRef.emit
  at TuiLineClamp.setOverflown
  at TuiLineClamp.onOverflow
  at TuiLineClamp_Template_div_animationcancel_self_0_listener
  at executeListenerWithErrorHandling
```

### Conditions that must be met

The first two are easy to miss, which is why a naive example stays silent:

1. **`@taiga-ui/event-plugins` must be provided** (via `provideTaiga()` or `provideEventPlugins()`). The listener is declared as `(animationcancel.self)`, and without the plugins the `.self` modifier is never wired up — no listener, no bug.
2. **`overflown` must already be `true`** when the component is destroyed. `setOverflown` bails out on `if (this.overflown() !== overflown)`, and `animationcancel` reports `false`. There are two independent ways to get there, and either is enough:
   - the clamp is mounted *after* the first render, so `animationstart` is actually observed — this is what the demo does via `afterNextRender`, and what `@defer` does in a real app. Mounting during bootstrap loses the initial `animationstart` and leaves `overflown === false`, hiding the bug;
   - the user hovers the clamp, and the host directive `TuiLineClampFallback` (`mouseenter` → `apply(measure())`) sets it. Notably this path is active even when `TUI_TIMELINE_SUPPORT` is `true`, so hovering a clamp that was mounted during bootstrap also arms the bug.
3. A browser that supports scroll-driven animations (Chrome), so the `@supports (animation-timeline: --name)` branch is active.

### What is *not* required

Things we ruled out while minimising the reproduction, so they need not be considered:

- **No leave/route animation.** We first suspected the DOM outliving the component because of a fade-out on the page layout. Not needed — a plain `@if` removal reproduces it identically.
- **No router navigation**, no hint interaction, and no binding on `overflownChange`. `@defer` is not required either — in our app it merely happens to be one way of satisfying condition 2.
- **No `<tui-root>` and no Taiga theme stylesheet.** The scroll-driven animation lives in the component's own styles, so the clamp arms itself without either — verified by dropping `@taiga-ui/styles/taiga-ui-theme.less` from the build, which reproduces all the same.
- **Not Angular-version specific.** The reproduction uses Angular 22.1.5, but the `NG0953` check in `OutputEmitterRef.emit` has been there since signal outputs, and `@taiga-ui/kit` 5.23.0 accepts `@angular/core >=19`, so older majors behave the same.

## Root cause

Overflow is detected through scroll-driven animations:

```css
@supports (animation-timeline: --name) {
  @keyframes tuiEmpty {}
  .t-wrapper {
    animation-name: tuiEmpty, tuiEmpty;
    animation-timeline: scroll(self block), scroll(self inline);
  }
}
```

with the animation lifecycle wired up in the template:

```html
<div
    tuiLineClampPosition
    class="t-wrapper"
    (animationcancel.self)="onOverflow(false)"
    (animationstart.self)="onOverflow(true)"
>
```

Tearing the element down cancels those animations, so `animationcancel` fires during destruction and reaches:

```ts
setOverflown(overflown: boolean): void {
    if (this.overflown() !== overflown) {
        this.overflown.set(overflown);
        this.overflownChange.emit(overflown); // ← component already destroyed
    }
}
```

## Expected behaviour

Destroying a `tui-line-clamp` produces no warnings.

## Actual behaviour

One `NG0953` per destroyed instance, in both dev and production builds.

## Suggested fix

Skip the emit once the component is being torn down:

```ts
private readonly destroyed = signal(false);

constructor() {
    inject(DestroyRef).onDestroy(() => this.destroyed.set(true));
}

setOverflown(overflown: boolean): void {
    if (this.destroyed() || this.overflown() === overflown) {
        return;
    }

    this.overflown.set(overflown);
    this.overflownChange.emit(overflown);
}
```

Alternatively, ignore `animationcancel` that is caused by teardown.

## Environment

- `@taiga-ui/kit` / `core` / `cdk` / `i18n` / `styles` 5.23.0, `@taiga-ui/polymorpheus` 5.0.1, `@taiga-ui/event-plugins` 5.0.0
- Angular 22.1.5, standalone components, zoneless change detection
- TypeScript 6.0.3
- Chrome (scroll-driven animations supported)

### Taiga UI version

5.23.0

### Browser Used

Chrome

### OS Used

Windows

splincode:
Fixes #14949

### DIFF
--- a/projects/kit/components/line-clamp/line-clamp.component.ts
+++ b/projects/kit/components/line-clamp/line-clamp.component.ts
@@ -37,11 +37,11 @@ import {TuiLineClampPositionDirective} from './line-clamp-position.directive';
     },
 })
 export class TuiLineClamp {
-    private readonly destroyRef = inject(DestroyRef);
     private readonly offset = inject(TUI_FONT_OFFSET);
     private readonly options = inject(TUI_LINE_CLAMP_OPTIONS);
     private readonly el = tuiInjectElement();
     private readonly overflows = signal(0);
+    private readonly destroyed = signal(false);
 
     protected readonly overflown = signal(false);
 
@@ -77,8 +77,12 @@ export class TuiLineClamp {
         this.showHint() && this.overflown() ? this.content() : '',
     );
 
+    constructor() {
+        inject(DestroyRef).onDestroy(() => this.destroyed.set(true));
+    }
+
     public setOverflown(overflown: boolean): void {
-        if (this.destroyRef.destroyed || this.overflown() === overflown) {
+        if (this.destroyed() || this.overflown() === overflown) {
             return;
         }
 
--- a/projects/kit/components/line-clamp/line-clamp.component.ts
+++ b/projects/kit/components/line-clamp/line-clamp.component.ts
@@ -3,6 +3,7 @@ import {
     ChangeDetectionStrategy,
     Component,
     computed,
+    DestroyRef,
     inject,
     input,
     output,
@@ -36,6 +37,7 @@ import {TuiLineClampPositionDirective} from './line-clamp-position.directive';
     },
 })
 export class TuiLineClamp {
+    private readonly destroyRef = inject(DestroyRef);
     private readonly offset = inject(TUI_FONT_OFFSET);
     private readonly options = inject(TUI_LINE_CLAMP_OPTIONS);
     private readonly el = tuiInjectElement();
@@ -76,10 +78,12 @@ export class TuiLineClamp {
     );
 
     public setOverflown(overflown: boolean): void {
-        if (this.overflown() !== overflown) {
-            this.overflown.set(overflown);
-            this.overflownChange.emit(overflown);
+        if (this.destroyRef.destroyed || this.overflown() === overflown) {
+            return;
         }
+
+        this.overflown.set(overflown);
+        this.overflownChange.emit(overflown);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


waterplea (approved):
Annoying thing :(

vladimirpotekhin (approved):

PR #14957, opened 2026-09-10
fix(addon-charts): support custom pie chart hint appearance

### LINKED ISSUE #6403: 🚀 - ability to use different hint appearances in charts
### Description

 Pie / ring charts provide `onDark` appearance for tooltip. https://github.com/taiga-family/taiga-ui/blob/main/projects/addon-charts/components/pie-chart/pie-chart.component.ts#L42
We need to be able use other appearances

splincode:
Fixes #6403

### DIFF
--- a/projects/addon-charts/components/pie-chart/pie-chart.component.ts
+++ b/projects/addon-charts/components/pie-chart/pie-chart.component.ts
@@ -41,7 +41,6 @@ const TRANSFORM = {
     viewProviders: [
         tuiHintOptionsProvider({
             direction: 'top-end',
-            appearance: 'floating',
             showDelay: 0,
             hideDelay: 0,
         }),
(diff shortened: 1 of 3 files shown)


waterplea (approved):

projects/addon-charts/components/pie-chart/pie-chart.component.ts
```diff
@@ -80,6 +80,10 @@ export class TuiPieChart {
         return this.hintOptions?.content() || '';
     }
 
+    protected get hintAppearance(): string {
+        return this.hintOptions?.appearance() || '';
```
waterplea: This changes default appearance from floating to '', the fallback value needs to be 'floating' and it can be removed from line 44 as it is no longer used.

splincode: fixed

mdlufy (approved):

PR #14958, opened 2026-09-10
fix(core): `SelectLike` should reset form control value on Backspace/Delete

### LINKED ISSUE #14553: 🐞 - Select: Backspace/Delete clears the visible value but not the form model (validation doesn't rerun)
### Reproduction url

https://stackblitz.com/edit/angular-fhj7l32e?file=src%2Fapp%2Fapp.component.ts

### Description

On a [tuiSelect] with the textfield cleaner enabled, pressing Backspace or Delete empties the visible input but leaves the underlying form model unchanged. As a result:
- required (and other validators) keep passing, because the FormControl still holds the previously selected value.
- Validation does not re-run, and the form can be submitted with a stale value that the user can no longer see.
 
The clear (✕) button works correctly, only keyboard deletion is affected.
Seems it was broken in https://github.com/taiga-family/taiga-ui/pull/13874

https://github.com/user-attachments/assets/8f2d8b9d-65eb-4434-837a-9e7da3280323

### Taiga UI version

4.90.0

### Browser Used

Chrome

### OS Used

macOS

nsbarsukov:
Fixes #14553

### DIFF
--- a/projects/core/components/textfield/select-like.directive.ts
+++ b/projects/core/components/textfield/select-like.directive.ts
@@ -1,7 +1,6 @@
 import {Directive, inject} from '@angular/core';
 import {WA_IS_ANDROID, WA_IS_MOBILE} from '@ng-web-apis/platform';
 import {tuiInjectElement} from '@taiga-ui/cdk/utils/dom';
-import {tuiClamp} from '@taiga-ui/cdk/utils/math';
 
 import {TUI_TEXTFIELD_OPTIONS} from './textfield.options';
 
@@ -16,16 +15,18 @@ import {TUI_TEXTFIELD_OPTIONS} from './textfield.options';
         '(beforeinput)':
             '(isMobile && $event.inputType.includes("insertText")) || options.cleaner() && $event.inputType.includes("delete") || $event.preventDefault()',
         '(input.capture)': '$event.inputType?.includes("delete") && clear()',
-        '(keydown.backspace)': 'options.cleaner() && dispatchInputEvent(-1)', // No (input) event if caret is at the beginning
-        '(keydown.delete)': 'options.cleaner() && dispatchInputEvent(1)', // No (input) event if caret is at the end
+        '(keydown.backspace.prevent)':
+            'options.cleaner() && el.value && dispatchInputEvent("deleteContentBackward")',
+        '(keydown.delete.prevent)':
+            'options.cleaner() && el.value && dispatchInputEvent("deleteContentForward")',
         // Hide Android text select handle (bubble marker below transparent caret)
         '(mousedown)': 'prevent($event)',
     },
 })
 export class TuiSelectLike {
-    private readonly el = tuiInjectElement<HTMLInputElement>();
     private readonly isAndroid = inject(WA_IS_ANDROID);
 
+    protected readonly el = tuiInjectElement<HTMLInputElement>();
     protected readonly isMobile = inject(WA_IS_MOBILE);
     protected readonly options = inject(TUI_TEXTFIELD_OPTIONS);
 
@@ -33,20 +34,13 @@ export class TuiSelectLike {
         this.el.value = '';
     }
 
-    protected dispatchInputEvent(direction: -1 | 1): void {
-        const caret = this.el.selectionStart ?? 0;
-
-        if (
-            this.el.value &&
-            tuiClamp(caret + direction, 0, this.el.value.length) === caret
-        ) {
-            this.el.dispatchEvent(
-                new InputEvent('input', {
-                    bubbles: true,
-                    inputType: `deleteContent${direction === 1 ? 'Forward' : 'Backward'}`,
-                }),
-            );
-        }
+    protected dispatchInputEvent(inputType: string): void {
+        this.el.dispatchEvent(
+            new InputEvent('input', {
+                inputType,
+                bubbles: true,
+            }),
+        );
     }
 
     protected prevent(event: MouseEvent): void {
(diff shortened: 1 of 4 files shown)


projects/kit/components/multi-select/multi-select-native/multi-select-native.component.ts
```diff
@@ -26,9 +25,11 @@ import {TuiMultiSelectOption} from '../multi-select-option/multi-select-option.c
     templateUrl: './multi-select-native.template.html',
     changeDetection: ChangeDetectionStrategy.OnPush,
     providers: [tuiAsOptionContent(TuiMultiSelectOption)],
-    hostDirectives: [TuiInputChipDirective, TuiSelectLike],
+    hostDirectives: [TuiInputChipDirective],
     host: {
+        autocomplete: 'off',
         multiple: '',
+        tuiSelectLike: '', // selector for styles
```
waterplea, line 32: Why we no longer can use it as directive?

projects/core/components/textfield/select-like.directive.ts
```diff
@@ -32,6 +33,22 @@ export class TuiSelectLike {
...
+        const caret = this.el.selectionStart ?? 0;
+
+        if (
+            this.el.value &&
+            tuiClamp(caret + direction, 0, this.el.value.length) === caret
+        ) {
+            this.el.dispatchEvent(
+                new InputEvent('input', {
+                    bubbles: true,
+                    inputType: `deleteContent${direction === 1 ? 'Forward' : 'Backward'}`,
```
waterplea: Why does all this matter, isn't it just for the (input.capture) to call clear and for Angular CVA to detect input event?

projects/kit/components/multi-select/multi-select-native/multi-select-native.component.ts
nsbarsukov: **Previous behavior**
https://taiga-ui.dev/components/input-chip#mobile

https://github.com/user-attachments/assets/55549d4f-738e-4ee1-9385-ad5a07feb56c

nsbarsukov: **New behavior**
https://taiga-preview-demo--pr14958-select-like-backspace-form-s85eed6w.web.app/components/input-chip#mobile


https://github.com/user-attachments/assets/81414385-51e2-424c-847d-69e39470fa96

nsbarsukov: We can, but on `<select multiple>` it does nothing useful. And even causes bugs (see videos above)

Everything `SelectLike` does is meant for a text `<input>`:

* `inputmode` / `spellcheck` have no effect on `<select>`
* `<select>` never fires `beforeinput`, and its `input` event has no `inputType`, so `(beforeinput)` / `(input.capture)` do nothing
* the Android `(mousedown)` handler never fires: `select` inside `tui-textfield[multi]` has `pointer-events: none`;
* what's left is keyboard clearing through `el.value = ''`, and on `select[multiple]` that deselects every `<option>` (see video above with bug).

projects/core/components/textfield/select-like.directive.ts
nsbarsukov: It matters for this line

https://github.com/taiga-family/taiga-ui/blob/6518a3da8a7861b6e9a11972f66b2e26910f7f2d/projects/core/components/textfield/select-like.directive.ts#L18

We could simplify this line to this version and everything will still work
```suggestion
                    inputType: 'deleteContentBackward',
```
Just less semantic version

waterplea: I don't think semantics matter here. What is the point of "if", value.length and caret position checks?

nsbarsukov:
```ts
'(keydown.backspace)': '...', // No (input) event if caret is at the beginning
'(keydown.delete)': '...', // No (input) event if caret is at the end
```

https://github.com/taiga-family/taiga-ui/blob/6518a3da8a7861b6e9a11972f66b2e26910f7f2d/projects/core/components/textfield/select-like.directive.ts#L19-L20

This `if` checks if it is this specific case.
If no - do nothing, allow browser implements everything natively.

nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/14958/commits/5679ec396ccd63aeb16b9055518dcacf849b40e2

projects/core/components/textfield/select-like.directive.ts
```diff
@@ -15,23 +15,34 @@ import {TUI_TEXTFIELD_OPTIONS} from './textfield.options';
...
 
+    protected readonly el = tuiInjectElement<HTMLInputElement>();
     protected readonly isMobile = inject(WA_IS_MOBILE);
     protected readonly options = inject(TUI_TEXTFIELD_OPTIONS);
 
     protected clear(): void {
         this.el.value = '';
     }
 
+    protected dispatchInputEvent(inputType: string): void {
```
waterplea: I don't think we need inputType as argument

waterplea (approved):

projects/core/components/textfield/select-like.directive.ts
nsbarsukov: Fixed in https://github.com/taiga-family/taiga-ui/pull/14958/commits/46a0ba90827e881e36d0dc297e25032015d51ba6

mdlufy (approved):

PR #14965, opened 2026-09-11
fix(addon-mobile): `DropdownSheet` reliably reopens after dismiss

### LINKED ISSUE #14964: 🐞 - Mobile `DropdownSheet` does not reopen after dismiss
### Reproduction url

https://taiga-ui.dev/directives/dropdown-hover

### Description

On mobile a dropdown rendered as a sheet (`tuiDropdownHover` + `tuiDropdownSheet`, without `tuiDropdownOpen`) fails to reopen after it has been dismissed (e.g. by tapping the backdrop).

Dismissing the sheet completes the `TuiSheetDialogService.open()` observable, but `tuiIfMap`'s inner `switchMap` swallows that completion, so `TuiDropdownDirective` is never toggled closed and its `ref` is not cleared. The dropdown stays logically "open", so the next tap has nothing to reopen — after a couple of open/dismiss cycles the sheet stops opening entirely.

Sub-problem of #14770; extracted from #14892 so it can be fixed and tested in isolation.

### Taiga UI version

5

### Browser Used

Chrome

### OS Used

macOS

mdlufy:
Closes #14964

### Problem

On mobile a dropdown rendered as a sheet (`tuiDropdownHover` + `tuiDropdownSheet`, without `tuiDropdownOpen`) does not reopen after it has been dismissed via the backdrop. Dismissing completes the `TuiSheetDialogService.open()` observable, but `tuiIfMap`'s inner `switchMap` swallows that completion, so `TuiDropdownDirective` is never toggled closed and its `ref` is not cleared. The dropdown stays logically "open", so the next tap has nothing to reopen — after a couple of open/dismiss cycles the sheet stops opening entirely.

### Fix

Close the dropdown when the sheet dialog completes via `finalize(() => this.dropdown.toggle(false))` on the inner observable, so `ref` clears and the hover state resets through the existing ref-removal branch, letting it reopen reliably on the next tap.

### DIFF
--- a/projects/addon-mobile/directives/dropdown-sheet/test/dropdown-sheet.component.spec.ts
+++ b/projects/addon-mobile/directives/dropdown-sheet/test/dropdown-sheet.component.spec.ts
@@ -1,0 +1,81 @@
+import {ChangeDetectionStrategy, Component} from '@angular/core';
+import {type ComponentFixture, fakeAsync, TestBed, tick} from '@angular/core/testing';
+import {By} from '@angular/platform-browser';
+import {WA_IS_MOBILE} from '@ng-web-apis/platform';
+import {TuiDropdownSheet} from '@taiga-ui/addon-mobile';
+import {
+    provideTaiga,
+    TuiDropdown,
+    TuiDropdownHover,
+    tuiDropdownHoverOptionsProvider,
+    TuiRoot,
+} from '@taiga-ui/core';
+
+describe('TuiDropdownSheet', () => {
+    @Component({
+        imports: [TuiDropdown, TuiDropdownHover, TuiDropdownSheet, TuiRoot],
+        template: `
+            <tui-root>
+                <button
+                    tuiDropdown="Content"
+                    tuiDropdownHover
+                    tuiDropdownSheet
+                    type="button"
+                >
+                    Open
+                </button>
+            </tui-root>
+        `,
+        changeDetection: ChangeDetectionStrategy.OnPush,
+    })
+    class Test {}
+
+    let fixture: ComponentFixture<Test>;
+
+    beforeEach(async () => {
+        TestBed.configureTestingModule({
+            imports: [Test],
+            providers: [
+                provideTaiga(),
+                {provide: WA_IS_MOBILE, useValue: true},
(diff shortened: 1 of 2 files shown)


waterplea (approved):

projects/addon-mobile/directives/dropdown-sheet/test/dropdown-sheet.component.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    it('reopens on the next hover after the sheet is dismissed', fakeAsync(() => {
+        hover();
+        expect(sheet()).not.toBeNull();
+
+        // Dismiss via the backdrop (click.self on the sheet host) without a
+        // hover reset — the case that used to leave the dropdown stuck open.
+        sheet()!.dispatchEvent(new MouseEvent('click', {bubbles: true}));
+        tick();
+        fixture.detectChanges();
+        tick();
```
nsbarsukov, line 73: Why not use a Playwright test instead? 

For current case Jest tests seem too artificial to reliably reproduce the bug and prevent it from being reintroduced.

nsbarsukov (approved):

projects/addon-mobile/directives/dropdown-sheet/test/dropdown-sheet.component.spec.ts
mdlufy: The real user-flow coverage for this is the `Opens mobile version of dropdown on the 2nd time click` e2e in `dropdown-hover.pw.spec.ts` (tap → dismiss on backdrop → tap again). It's entangled with the core hover + `hasTouch`/`.tap()` changes in #14892 — without them the first tap-open and the post-`hideContent` frame flake on webkit/chromium, which is why it was `test.skip` before — so it lives there rather than here.

This jest spec is a deterministic guard for the exact regression this PR fixes: `tuiIfMap`'s `switchMap` swallowing the dialog completion so the dropdown never toggles closed. It fails on the old `.subscribe({complete})` and passes with `finalize`, without relying on touch emulation. Happy to drop it and lean on the e2e in #14892 if you'd prefer this PR test-free.

PR #14971, opened 2026-09-12
feat(addon-doc): version-driven New/Updated navigation badges

mdlufy:
## What

Replaces the free-form per-page `badges` config with a single `version` field and derives everything from it — the navigation `New` / `Updated` badges and an on-page version badge.

`TuiDocRoutePageBase` gains `readonly version?: string` — the library version a page was introduced in:

```ts
// TUI_DOC_PAGES config
{section: 'Components', title: 'Keypad', route: DemoRoute.Keypad, version: '5.19.0'}
```

## Behavior

**Navigation menu** (fully derived, nothing to label by hand):
- **New** (`positive`) on a page whose `version` is within the last **6 minor** releases of the current version (same major).
- A section/group whose **lead (defining) page** is itself recent is marked **New** — the whole thing is new (e.g. `AI tools`), and the lead row isn't badged twice.
- A section/group that merely **gained** a recent page (its lead is older) is marked **Updated** (`info`), and the specific new page keeps its own **New**.
- Badges stay visible while a section/group is collapsed.

**Component page:**
- A neutral **`v{version}+`** badge next to the title. The page looks its `version` up from the page config by title via `TUI_DOC_MAP_PAGES`, so `pages.ts` stays the single source of truth.
- A component page without an explicit `version` falls back to the **current-major baseline** `v{major}.0.0+`. This only applies to pages that carry a `package` (real components/directives) — plain doc pages show nothing — so every component states a since-version without hardcoding `5.0.0` into ~200 entries.

## Current version

- Injected via a new **`TUI_DOC_VERSION`** token — addon-doc no longer imports `TUI_VERSION` from `@taiga-ui/cdk`; the demo provides the token with `TUI_VERSION`. Default is `''` → nothing is badged, a safe no-op for other consumers.
- `New` window = 6 minors. A page annotated with a version from a previous major stops being `New` the instant the current major advances (guarded by a `major === current` check), so a new major never lights up the whole menu. Because the baseline default is display-only and never written into the page data, unversioned components are never flagged `New` either.

## Notes

- Demo carries only genuine, verified 5.x versions: Meter / SearchBar `5.22.0`, ScrollWheel `5.21.0`, Keypad `5.19.0`, File / Pincode `5.9.0`, AI overview `5.23.0`. Everything older relies on the baseline default instead of a placeholder (`Pincode 5.9.0` sits past the `New` window — shows only the on-page version badge, no `New` — a nice both-cases example).
- Navigation badges are precomputed once at construction (inputs are static), so nothing re-scans subpages or re-parses versions on every change detection.

## Tests

- **Navigation:** `New` on a whole new section (lead recent), `Updated` on a section/group that gained a recent page, `New` on the recent non-lead row, and no badge past the window. Plus two guards: a major bump doesn't mark previous-major pages `New`, and a long-standing unversioned page stays quiet a minor into a new major. Versions are computed relative to the current version so they don't rot.
- **Page:** explicit version, current-major baseline, and no-package (no badge) cases for the on-page version badge.

### DIFF
--- a/projects/addon-doc/components/navigation/navigation.component.ts
+++ b/projects/addon-doc/components/navigation/navigation.component.ts
@@ -19,7 +19,11 @@ import {
     TUI_DOC_SEARCH_ENABLED,
     TUI_DOC_SEARCH_TEXT,
 } from '@taiga-ui/addon-doc/tokens';
-import {type TuiDocRoutePage, type TuiDocRoutePages} from '@taiga-ui/addon-doc/types';
+import {
+    type TuiDocRoutePage,
+    type TuiDocRoutePageBadge,
+    type TuiDocRoutePages,
+} from '@taiga-ui/addon-doc/types';
 import {tuiTransliterateKeyboardLayout} from '@taiga-ui/addon-doc/utils';
 import {TuiAutoFocus} from '@taiga-ui/cdk/directives/auto-focus';
 import {tuiControlValue, tuiWatch} from '@taiga-ui/cdk/observables';
@@ -31,6 +35,7 @@ import {TuiLink} from '@taiga-ui/core/components/link';
 import {TuiScrollbar} from '@taiga-ui/core/components/scrollbar';
 import {TUI_COMMON_ICONS} from '@taiga-ui/core/tokens';
 import {TuiAccordion} from '@taiga-ui/kit/components/accordion';
+import {TuiBadge} from '@taiga-ui/kit/components/badge';
 import {TuiDrawer} from '@taiga-ui/kit/components/drawer';
 import {PolymorpheusOutlet} from '@taiga-ui/polymorpheus';
 import {combineLatest, filter, fromEvent, map, of, switchMap, take} from 'rxjs';
@@ -67,6 +72,7 @@ function tuiUniqBy<T extends Record<string, any>>(
         RouterLinkActive,
         TuiAccordion,
         TuiAutoFocus,
+        TuiBadge,
         TuiDataList,
         TuiDocScrollIntoViewLink,
         TuiExpand,
@@ -176,6 +182,10 @@ export class TuiDocNavigation {
 
     protected isActive(route: string): boolean {
         return route === this.active;
+    }
+
+    protected sectionBadge(index: number): TuiDocRoutePageBadge | null {
+        return this.items[index]?.[0]?.badge ?? null;
     }
 
     protected onGroupClick(index: number): void {
--- a/projects/addon-doc/components/navigation/navigation.component.ts
+++ b/projects/addon-doc/components/navigation/navigation.component.ts
@@ -22,6 +22,7 @@ import {
 import {
     type TuiDocRoutePage,
     type TuiDocRoutePageBadge,
+    type TuiDocRoutePageGroup,
     type TuiDocRoutePages,
 } from '@taiga-ui/addon-doc/types';
 import {tuiTransliterateKeyboardLayout} from '@taiga-ui/addon-doc/utils';
@@ -47,6 +48,11 @@ import {
     NAVIGATION_TITLE,
 } from './navigation.providers';
 import {TuiDocScrollIntoViewLink} from './scroll-into-view.directive';
+
+const UPDATED_BADGE = {
+    label: 'Updated',
+    appearance: 'info',
+} satisfies TuiDocRoutePageBadge;
 
 function tuiUniqBy<T extends Record<string, any>>(
     array: readonly T[],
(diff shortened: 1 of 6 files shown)


projects/addon-doc/components/navigation/navigation.component.ts
```diff
@@ -43,6 +49,15 @@ import {
...
 import {TuiDocScrollIntoViewLink} from './scroll-into-view.directive';
 
+interface TuiDocNavigationBadge {
+    readonly label: string;
+    readonly appearance: string;
+}
+
+const NEW_BADGE: TuiDocNavigationBadge = {label: 'New', appearance: 'positive'};
+const UPDATED_BADGE: TuiDocNavigationBadge = {label: 'Updated', appearance: 'info'};
+const TUI_NEW_VERSION_RANGE = 6;
```
waterplea, line 59: Unnecessary TUI_ prefix

projects/addon-doc/components/navigation/navigation.component.ts
```diff
@@ -90,6 +106,13 @@ export class TuiDocNavigation {
 
     private readonly router = inject(Router);
     private readonly doc = inject(DOCUMENT);
+    private readonly currentMajor = Number(TUI_VERSION.split('.')[0]);
```
waterplea, line 109: This is an addon-doc component but now it depends on TUI_VERSION. I think we should add a new token for that and provide it with TUI_VERSION for demo

projects/addon-doc/components/navigation/navigation.component.ts
```diff
@@ -203,6 +248,24 @@ export class TuiDocNavigation {
         }
     }
 
+    private hasRecent(pages: readonly TuiDocRoutePage[] | undefined): boolean {
+        return !!pages?.some((page) => this.isRecent(page.version));
+    }
+
+    private isRecent(version: string | undefined): boolean {
```
waterplea, line 255: version = ''

projects/addon-doc/components/navigation/navigation.component.ts
```diff
@@ -203,6 +248,24 @@ export class TuiDocNavigation {
         }
     }
 
+    private hasRecent(pages: readonly TuiDocRoutePage[] | undefined): boolean {
```
waterplea, line 251: pages: ... = []

projects/addon-doc/components/navigation/navigation.component.ts
```diff
@@ -203,6 +248,24 @@ export class TuiDocNavigation {
...
+    private isRecent(version: string | undefined): boolean {
+        const [major = Number.NaN, minor = Number.NaN] = (version ?? '')
+            .split('.')
+            .map(Number);
+
+        const distance = this.currentMinor - minor;
+
+        return (
+            major === this.currentMajor &&
+            distance >= 0 &&
```
waterplea, line 264: Do we need this check? This case seems invalid, and if it ever happens that we have a component from the future in our docs, "New" badge wouldn't hurt 😄

projects/demo/src/pages/app/pages.ts
```diff
@@ -320,6 +321,7 @@ export const pages: DocRoutePages = [
                 title: 'Dialog',
                 keywords: 'попап, модал, popup, dialog, диалог, modal, окно',
                 route: DemoRoute.Dialog,
+                version: '4.50.0',
```
waterplea, line 324: I think we should not use v4 inside v5 demo, let's clamp all version by 5.0.0, otherwise it's confusing.

waterplea (commented):
I didn't check, but feels like not all components have their version set, let's put it there for all of them.

projects/addon-doc/components/navigation/navigation.component.ts
mdlufy: Updated

projects/addon-doc/components/navigation/navigation.component.ts
mdlufy: Updated

projects/addon-doc/components/navigation/navigation.component.ts
mdlufy: Fixed

projects/demo/src/pages/app/pages.ts
mdlufy: Updated, now v5.0.0 by default. A specific version can be specified in pages.ts

waterplea (approved):

projects/addon-doc/components/navigation/navigation.component.ts
mdlufy: Fixed

projects/addon-doc/components/navigation/navigation.component.ts
mdlufy: I removed this future))

projects/addon-doc/components/navigation/navigation.component.ts
```diff
@@ -203,6 +252,67 @@ export class TuiDocNavigation {
         }
     }
 
+    private mapItemBadges(): ReadonlyMap<
```
vladimirpotekhin, line 251: Maybe we could move this mapper/some methods out of the component and into the utilities?

projects/addon-doc/components/navigation/navigation.component.ts
```diff
@@ -178,6 +215,18 @@ export class TuiDocNavigation {
...
 
+    protected sectionBadge(index: number): TuiDocNavigationBadge | null {
+        return this.sectionBadges[index] ?? null;
+    }
+
+    protected pageBadge(item: TuiDocRoutePage): TuiDocNavigationBadge | null {
+        return this.itemBadges.get(item) ?? null;
+    }
+
+    protected groupBadge(item: TuiDocRoutePageGroup): TuiDocNavigationBadge | null {
```
vladimirpotekhin: `groupBadge` and `pageBadge` methods do the same thing and differ only in their types; perhaps they should be combined?

vladimirpotekhin (approved):

projects/addon-doc/components/navigation/navigation.component.ts
mdlufy: Yes, good idea. Moved in https://github.com/taiga-family/taiga-ui/pull/14971/changes/049963b8e5daf1a9ffe4b84bb7ff8ff579c77acf

projects/addon-doc/components/navigation/navigation.component.ts
mdlufy: Done, combined in https://github.com/taiga-family/taiga-ui/pull/14971/changes/049963b8e5daf1a9ffe4b84bb7ff8ff579c77acf

Issue #14974, opened 2026-09-13
🐞 - MobileCalendar renders blank/truncated months after a fast year jump (virtual-scroll strategy desync)


bjorndmitriev:
### Reproduction url

https://taiga-ui.dev/components/mobile-calendar

### Description

Open the mobile calendar sheet and quickly jump across several years using the horizontal year strip (flicking it fast, or tapping year labels several years apart in quick succession). The vertical month/day list intermittently renders **blank**, or renders only a single truncated row with no month header, or lands on the wrong month (skips the one it should land on) — then self-corrects roughly 300–500ms later, after another scroll tick.

Observed on a real device (iOS Safari) in our production app while paging back through years/months in a `tui-mobile-calendar` date picker.

**Root cause**, traced in `TuiMobileCalendarStrategy.updateRenderedRange`:

https://github.com/taiga-family/taiga-ui/blob/f0012b755e130061fbb43d8385fc3826f3e22a40/projects/addon-mobile/components/mobile-calendar/mobile-calendar.strategy.ts#L128-L158

```ts
private updateRenderedRange(viewport: CdkVirtualScrollViewport): void {
    const offset = viewport.measureScrollOffset();
    const {start, end} = viewport.getRenderedRange();
    ...
    const newRange = {start, end};
    const startBuffer = offset - this.getOffsetForIndex(start);

    if (startBuffer < BUFFER && start !== 0) {
        newRange.start = Math.max(0, this.getIndexForOffset(offset - BUFFER * 2));
        newRange.end = Math.min(dataLength, this.getIndexForOffset(offset + viewportSize + BUFFER));
    } else {
        const endBuffer = this.getOffsetForIndex(end) - offset - viewportSize;

        if (endBuffer < BUFFER && end !== dataLength) {
            newRange.start = Math.max(0, this.getIndexForOffset(offset - BUFFER));
            newRange.end = Math.min(dataLength, this.getIndexForOffset(offset + viewportSize + BUFFER * 2));
        }
    }

    viewport.setRenderedRange(newRange);
    viewport.setRenderedContentOffset(this.getOffsetForIndex(newRange.start));
    this.index$.next(firstVisibleIndex);
}
```

[...]

PR #10023, opened 2024-12-18
fix(addon-mobile): virtual-scroll flickers during scroll in zoneless mode

splincode:
Fixes #9662

### Before

https://github.com/user-attachments/assets/b81ab78d-2aea-40d0-9e50-f90b118e9d40

### After

https://github.com/user-attachments/assets/efc138c5-780e-4b8a-9818-e787dd7b20ef

PR #14983, opened 2026-09-14
fix(addon-mobile): restore variable mobile calendar height

### LINKED ISSUE #14977: 🐞 - MobileCalendar month-height model (weekCount-based) doesn't match the fixed CSS height, causing snap-to-month drift
### Reproduction url

https://taiga-ui.dev/components/mobile-calendar

### Description

`TuiMobileCalendarStrategy` computes scroll offsets assuming each month occupies a **variable** height (label + number of week-rows × week height):

https://github.com/taiga-family/taiga-ui/blob/f0012b755e130061fbb43d8385fc3826f3e22a40/projects/addon-mobile/components/mobile-calendar/mobile-calendar.const.ts#L1-L57

```ts
function getCycle(options) {
    return Array.from({length: options.yearCycle}, (_, i) =>
        Array.from({length: MONTHS_IN_YEAR}, (_, month) =>
            options.label + weekCount({year: i, month, startingYear: options.startingYear}) * options.week));
}
export const ANDROID_LABEL = 64;
export const ANDROID_WEEK = 48;
export const IOS_LABEL = 50;
export const IOS_WEEK = 50;
```

But the actual rendered DOM gives **every** month section a fixed height regardless of its week count:

https://github.com/taiga-family/taiga-ui/blob/f0012b755e130061fbb43d8385fc3826f3e22a40/projects/addon-mobile/components/mobile-calendar/mobile-calendar.style.less#L185-L190

```less
.t-month-wrapper {
    display: flex;
    margin: 0.625rem 0 -0.625rem;
    block-size: 22.125rem; // itemSize: 354px;
    flex-direction: column;
}
```

354px only matches the *6-week* case on iOS (`50 + 6×50 = 350`, close enough). Every other month is shorter in the strategy's own model but is still rendered at a fixed 354px in the DOM. Concretely, for 2026 (iOS constants):

```
2026-01: 5 weeks → strategy height 300px, actual 354px → 54px short
2026-02: 5 weeks → strategy height 300px, actual 354px → 54px short
2026-03: 6 weeks → strategy height 350px, actual 354px → 4px short
2026-04: 5 weeks → strategy height 300px, actual 354px → 54px short
...(most months are 54px short, a few 6-week months are only 4px short; a month never actually gets *taller* than the fixed 354px in this dataset, so the error only ever accumulates in one direction)
```

Since `getOffsetForIndex`/`computeHeight` (used by `scrollToIndex`, i.e. every "snap to start of month" / "jump to year" call) sum up these *undercounted* per-month heights to compute a target scroll offset, while the browser's actual scrollable content is laid out using the *fixed* 354px per month, the computed target increasingly diverges from where that month really starts in the DOM the further you are from the reference point (`STARTING_YEAR`) — typically by multiples of ~54px per month crossed.

**Symptom:** "snap to start of month" after releasing a scroll, and the vertical jump performed by `setYear()`/year-strip navigation, land short of (or past) the actual month boundary, and what should be a single clean snap instead shows the content creeping the rest of the way over several follow-up renders as later `onContentScrolled()`/`scrolledIndexChange` cycles re-measure against the real (fixed-height) layout and correct toward it. On a real device I saw the content settle roughly 60-90px further after letting go of a scroll near a month boundary — consistent with 1-2 months' worth of the ~54px-per-month error above.

A fix needs the constants in `getCycle()`/`weekCount()` to match `.t-month-wrapper`'s actual fixed height (i.e. use `ANDROID_WEEK`/`IOS_WEEK` values whose max (6-week) case equals the real per-month CSS height, or make every month's real height match its computed one instead of a single fixed block-size for all week-counts).

This is a different mechanism from #14974 (buffer heuristic desync on large jumps) and #14976 (unbounded queued `setYear` timers), though all three live in the same `TuiMobileCalendar`/`TuiMobileCalendarStrategy` pair.

### Taiga UI version

5.23.0

### Browser Used

WebView (Capacitor Android), Safari (video reproduction)

### OS Used

Android, iOS

splincode:
Fixes #14977


projects/addon-mobile/components/mobile-calendar/mobile-calendar.style.less
```diff
@@ -185,7 +185,7 @@
 .t-month-wrapper {
     display: flex;
     margin: 0.625rem 0 -0.625rem;
-    block-size: 22.125rem; // itemSize: 354px;
+    block-size: 22.125rem; // MONTH_HEIGHT: 354px;
```
waterplea: The PR says it is restoring dynamic height but the block size for a month is still constant

splincode: Sorry for that. I initially aligned the strategy with the fixed CSS height, but that contradicts the variable-height model and the rollback of #10023.

I've updated the PR to remove the fixed `block-size` and `itemSize`, restore the iOS/Android month-height cycles, and keep the larger render buffer in the custom strategy for #14974.

waterplea: Let's make it 2 sepparate PRs, first let's roll back the fix from [this PR](https://github.com/taiga-family/taiga-ui/pull/10023) that made the height fixed and come up with a different solution to the original issue of blinking in zoneless config. After that works we can get back to fixing buffer issue.

splincode: I'll split it.

splincode: Done

PR #14996, opened 2026-09-15
fix(core): prevent calendar emit after destroy

### LINKED ISSUE #14995: 🐞 - `InputDateRange` - warning NG0953
### Reproduction url

https://stackblitz.com/edit/angular-atntf94t?file=src%2Fapp%2Fapp.ts

### Description

After selecting the date range, we see a warning: _NG0953: Unexpected emit for destroyed OutputRef. The owning directive/component is destroyed._

### Taiga UI version

5.24.0

### Browser Used

Chrome

### OS Used

macOS

splincode:
Fixes #14995


projects/kit/components/calendar-range/calendar-range.component.ts
```diff
@@ -135,11 +143,7 @@ export class TuiCalendarRange
     }
 
     public ngOnDestroy(): void {
```
waterplea, line 137: I don't understand. The warning only fires when we select both dates, not when we select one and click away.

splincode: You're right. Sorry for that. I traced it to the nested `TuiCalendar`. `TuiCalendar.onDayClick` emits `dayClick` first. When the second date is selected, that handler completes the range and closes the dropdown, synchronously destroying the calendar. Execution then returns to `TuiCalendar.onDayClick`, which was calling `value.set(day)` on the already destroyed component, causing `NG0953`.

I've replaced the previous fix with a guard before that trailing `value.set()` and added a regression test where the `dayClick` handler synchronously destroys the fixture.

I also verified the regression test against the old implementation: it fails with the same `NG0953` pointing to `TuiCalendar.onDayClick`. With the fix, the full unit test suite passes.

waterplea: This is annoying. By this logic in every function that has 2 outputs triggered we have to guard against destruction across the whole codebase. Can we just change the order of those in this case?

splincode: Make sense. I changed the order here: `value.set(day)` now runs before `dayClick.emit(day)`, so the model update cannot happen after a `dayClick` handler synchronously destroys the calendar.

waterplea: Some lint issues

splincode: Fixed

waterplea (approved):

vladimirpotekhin (approved):

PR #14998, opened 2026-09-15
fix(styles): restore StackBlitz Less compatibility

### LINKED ISSUE #14994: 📚 - The examples on StackBlitz aren't working.
### What is the affected URL?

_No response_

### Description

When opening any example on StackBlitz, we see an error:
```
Error in turbo_modules/@taiga-ui/styles@5.24.0/mixins/appearance.less (47:31)
Missing closing ')'
```

### Browser Used

Chrome

### OS Used

macOS

splincode:
## Summary

* avoid nested `:not()` inside `:is()` in the invalid appearance selector, which breaks the Less parser used by StackBlitz
* preserve the existing behavior: explicit Taiga UI invalid states still override untouched state, while native `:invalid` ignores `._untouched`
* keep Less and SCSS mixins aligned

Fixes #14994


waterplea (approved):

vladimirpotekhin (approved):
check lint please @splincode

PR #15000, opened 2026-09-15
fix(kit): file validator does not throw when filename is empty

pumano:
cherry-picked from v4
Fixes #


vladimirpotekhin (approved):

waterplea (approved):

projects/kit/components/files/files.validators.ts
```diff
@@ -35,7 +35,7 @@ export function tuiCreateFileFormatValidator(accept: string): ValidatorFn {
 }
 
 function checkFormat({name, type}: TuiFileLike, formats: readonly string[]): boolean {
```
splincode, line 37: Could we also make name optional in TuiFileLike?

readonly name?: string;

The implementation already handles a missing name via optional chaining, so keeping it required in the public type does not reflect the actual supported input. It would also allow us to test this case without as unknown as File.

pumano: It's pure cherry-pick from v4

PR #15003, opened 2026-09-15
 fix(addon-mobile): sync mobile calendar virtual scroll

### LINKED ISSUE #14974: 🐞 - MobileCalendar renders blank/truncated months after a fast year jump (virtual-scroll strategy desync)

splincode:
Fixes #14974


waterplea (approved):
How are we sure it's gonna be enough?

splincode:
@waterplea updated

PR #15008, opened 2026-09-16
fix(core): refresh scrollbar after animation

### LINKED ISSUE #14954: 🐞 - Sheet dialog briefly paints the dialog scrollbar thumb on open (v4)
### Reproduction url

There is no public v4 playground for `tui-sheet-dialog` — taiga-ui.dev serves 5.x, where this is already fixed. Minimal v4 setup we measured (viewport 375×812, `@taiga-ui/core` 4.97.0):

```ts
@Component({
    standalone: true,
    imports: [TuiButton],
    template: `<button tuiButton (click)="open()">Open sheet</button>`,
})
export class AppComponent {
    private readonly sheets = inject(TuiSheetDialogService);

    protected open(): void {
        this.sheets.open('Short content').subscribe();
    }
}
```

### Description

Opening a sheet dialog paints the vertical scrollbar thumb of `.t-dialog` for ~100 ms right after the enter animation ends — a grey bar flickers along the right edge.

Frame-by-frame (t from click, DOM polled on every `requestAnimationFrame`):

| t, ms | `.t-dialog` scrollHeight/clientHeight | `tui-enter` | `.t-bar_vertical` | `.t-thumb` |
|---|---|---|---|---|
| 16 | 1608 / 812 | yes | not yet | — |
| 32–149 | 1589→831 / 812 | yes | `display: none` | inline `height: 52%` |
| **166** | 812 / 812 | no | visible | inline `height: 98%` — **painted** |
| 282 | 812 / 812 | no | visible | inline `height: 100%` → hidden |
| 500 | 812 / 812 | no | removed from DOM | — |

What seems to happen:

1. `.t-dialog` is the scroll container (`overflow: auto`); while entering, the sheet is `translateY(100%)`, so `scrollHeight` is temporarily twice `clientHeight` — `TuiScrollControls` sees a scroll and renders the bar.
2. During the animation the bar is hidden by `.t-dialog > .tui-enter + .t-scrollbars .t-bar_vertical {display: none}`.
3. On `animationend` `tui-enter` is dropped and that guard stops applying — but the thumb still carries the inline height computed while the geometry was still moving (98%), because `TuiScrollbarDirective` recomputes on `throttleTime(100)`.
4. The second guard, `.t-bar_vertical .t-thumb[style*="height: 100%"] {display: none}`, is a string match on exactly `100%`, so `98%` slips through and the thumb paints.
5. The next recompute writes exactly `100%` and it disappears. So the visible window equals one throttle period.

Engine difference (same scenario, same viewport) — the thumb is a 4px bar of `rgb(116,116,116)` at `opacity: .2`, so it only reads against the white sheet, not against the dimmed overlay:

| Engine | `scrollHeight` during the slide | stale thumb height | reaches the sheet | visible |
|---|---|---|---|---|
| Blink | decays 1608 → 812 | 91–98% | yes | yes, 28/255 |
| WebKit | stays 1608 to the end | 51% | no | no, 2/255 |

In 5.23.0 the flash is gone: `TuiScrollControls` has the `_timeline` mode, thumb geometry is driven by CSS `animation-timeline` instead of the throttled JS recompute (both CSS guards are still there verbatim). Filing against 4.x as we are on that line.

### Taiga UI version

4.97.0

### Browser Used

Chrome (also reproduced in WebKit)

### OS Used

macOS

splincode:
## Summary

- refresh `TuiScrollbarDirective` geometry when a CSS animation ends
- keep the existing throttled animation-frame updates unchanged
- add a SheetDialog Playwright regression test that captures the vertical thumb size immediately after the enter animation

Fixes #14954


waterplea:
How is it fixed in v5? Can't we just hide scroll controls with CSS while animation is running?

splincode:
@waterplea In v5 scrollbar geometry is handled by CSS scroll-driven animations (scroll-timeline / animation-timeline), so it stays synchronized without JS measurements.

In v4 we already hide the vertical scroll controls while `.tui-enter / .tui-leave` is present. The flash happens right after animationend: the class is removed before the throttled RAF updates the stale thumb geometry. Handling animationend forces that recalculation at the boundary. Therefore, adding the suggested CSS hiding alone would not fix the remaining frame.

waterplea (approved):

mdlufy (approved):

PR #15009, opened 2026-09-17
fix(core): `Scrollbar` properly react to 'native' option

waterplea:
Fixes #


projects/core/components/textfield/textfield-multi/textfield-multi.style.less
```diff
@@ -2,13 +2,35 @@
 
 tui-textfield[multi][multi]:where(*&) {
     flex-wrap: nowrap;
-    overflow: scroll;
+    overflow: auto;
```
mdlufy, line 5: Is this not a breaking change? in behaviour

mdlufy (approved):

projects/core/components/textfield/textfield-multi/textfield-multi.style.less
waterplea: No

vladimirpotekhin (approved):

PR #15039, opened 2026-09-22
fix(core): safari animation fix for pie/ring charts

### LINKED ISSUE #15037: 🐞 - PieChart / RingChart animation doesn't work in safari
### Reproduction url

https://taiga-ui.dev/charts/pie-chart

### Description

Hover animation in PieChart / RingChart doesn't work in safari
Additionally, hints (labels) doesn't work in `PieChart`  in safari https://taiga-ui.dev/charts/pie-chart#with-labels

https://github.com/user-attachments/assets/4220f2b5-2d20-43f3-9603-33c82c4fa820

### Taiga UI version

5.25.0

### Browser Used

Safari

### OS Used

macOS

gregorycode:
Fixes https://github.com/taiga-family/taiga-ui/issues/15037


projects/core/portals/hint/hint-hover.directive.ts
```diff
@@ -78,6 +77,7 @@ export class TuiHintHover extends TuiDriver {
         alias: 'tuiHintHideDelay',
     });
 
+    public visible = false;
```
gregorycode: Not sure about setting `visible` as public. Maybe it is better to create a getter?

waterplea: What are those changes for? If it fixes tooltips in Safari, why are they broken in Safari only?

gregorycode: There are two independent WebKit differences: a broken resize (hover) animation for chart segments, and hints not showing on hover.

1. Resize (hover animation). The hovered segment was scaled with `[attr.transform]`. Chrome maps that SVG attribute to the CSS transform property, so the stylesheet's transition animates it; WebKit doesn't, leaving the computed transform at `none`, so the segment jumps to full size. Fixed by binding `[style.transform]` instead, which animates identically in both engines.

2. Chart hints (labels). PieChart puts the hint on the `<g>`, so `mouseenter` there starts the show timer (`delay(showDelay)` in `TuiHintHover`, and even with `showDelay: 0` it resolves in a separate task), while the content comes from `pointerenter` on a child `<path>` setting `activeItemIndex`.

So there is a race between two events that Safari orders differently from Chrome. 
- Chrome sends `pointerenter` on the child first, so change detection sets the content before the timer fires and `toggle(true)` finds content. 
- WebKit fires `mouseenter` on the `<g>` before `pointerenter` on the child, so the timer fires while content is still `''` and `TuiHintDirective.toggle()` skips creating the hint. Nothing retries afterwards: the pointer never leaves the host, so no new hover event arrives, and `ngOnChanges` only hides the hint on empty content.                 

So the fix makes `ngOnChanges` show the hint when content arrives after hover has already requested it. Empty content still hides it as before. That's why I made `visible` public. `TuiHintDirective` needs the driver's last decision, but it was private.

waterplea: Hover is just one driver, the fix should work regardless of the particular driver or be implemented inside the problematic driver itself. Looks like changing `(pointerenter)` to `(mouseover)` fixes this discrepancy. Would be great to file a bug against webkit for messing up events order, Firefox and Chrome do it right.

gregorycode: I will take a look. Thanks.

gregorycode: @waterplea You were right, changing `pointerenter` to `mouseover` does the job, and it's a much cleaner solution. I've reverted the hint changes, so the PR now only touches the pie chart. Thank you!

waterplea (approved):

projects/addon-charts/components/pie-chart/pie-chart.template.html
waterplea, line 1: A11y linter brings a good point, but I suppose that's a subject for a separate PR, can we bypass it and create a separate issue, @vladimirpotekhin ?

gregorycode: I was about to ask about that

vladimirpotekhin: ok, I created issue https://github.com/taiga-family/taiga-ui/issues/15055

@gregorycode  let's add 

```
<!-- eslint-disable @angular-eslint/template/mouse-events-have-key-events -->
<!--TODO: https://github.com/taiga-family/taiga-ui/issues/15055  -->
```

before svg tag to bypass linter for now

vladimirpotekhin (approved):

